PR #2, opened 2017-04-22
Blacklist Bandicam's hooking DLL

ifarbod:
It crashes NUI - blacklist it so it doesn't cause any problems. Since Fraps is already blacklisted, this could be blocked too.


blattersturm:
This has been added in internal commit 4cd943aab6b4d88fa0622885a5cbbc9a0f15a914 - the next codebase upload will see this be resolved.

PR #12, opened 2017-05-17
Fix setup / build ability

mcmonkey4eva:
This patch includes a few minor fixes:
- Remove "Adhesive" from component list - that's not in the git repo and as such causes errors trying to run setup with that still there.
- Add `Easy Install` to the requirements list. This is a Python module required to build.
- Add the Windows SDK v10 to requirements list. This or a compatible version is required for building.

On the last one, I had to manually (Text editor + Find-and-replace) retarget every project (too many to do one at a time in VS...) to the Windows SDK latest version.
Might be worthwhile to include a note to that effect.

Also perhaps a general note that this is NOT an easy project to get built!


blattersturm:
The code repository is still not finished in a way that can be built without changes, and the docs are unfinished, and _pull requests can still not be merged_ due to shortcomings in synchronization.

`adhesive` is a real component which is censored from the repository, once everything is done the system will not error out anymore for missing components.

Python itself usually includes `easy_install`, and VS2017 includes the `10.0.15063` SDK.

Also - no you didn't have to manually retarget, there's a `systemversion` directive in `premake5.lua`.

mcmonkey4eva:
`adhesive is a real component which is censored from the repository, once everything is done the system will not error out anymore for missing components.` RIP.
`Python itself usually includes easy_install` Nope! My install did not include it, I had to manually install.
`and VS2017 includes the 10.0.15063 SDK.` Negative! Same as above.
`Also - no you didn't have to manually retarget, there's a systemversion directive in premake5.lua.` Well, it either didn't work or wasn't included properly on github yet :/

PR #15, opened 2017-05-25
Added Exit button to bottom navigation bar

kirkdmo:
I know there is an exit button on the Server page, however this puts an Exit button globally. 
Screenshot


ifarbod:
@kirkdmo Sounds good.

kirkdmo:
So.... I guess this was a no...?

PR #23, opened 2017-06-05
GameCache.cpp Updated

shayandoust:
Resolved an issue where memory traces are still left when Cache error occurs.

### DIFF
--- a/code/client/launcher/GameCache.cpp
+++ b/code/client/launcher/GameCache.cpp
@@ -747,7 +747,9 @@ static void PerformUpdate(const std::vector<GameCacheEntry>& entries)
 	// failed?
 	if (!retval)
 	{
-		FatalError("Fetching game cache failed.");
+		FatalError("Fetching game cache failed. Please retry again.");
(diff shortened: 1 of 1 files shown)


blattersturm:
Hi!

Thanks for contributing to FiveM. However, your pull request is unclear.

1. Please specify what the symptoms of the 'issue' you're trying to resolve are, and how your fix helps fix this.
2. Your code would do nothing, as the application [exists](https://github.com/citizenfx/fivem/blob/67d815e29c71d3b8fd2c8f986e9c194b08c97e41/code/client/common/Error.cpp#L66) after `FatalError` is [called](https://github.com/citizenfx/fivem/blob/67d815e29c71d3b8fd2c8f986e9c194b08c97e41/code/client/common/Error.cpp#L167).
3. You're changing the displayed string in a commit that claims to do only other things.
4. The PR title does not accurately detail the issue you're trying to resolve. Please try making descriptive commit messages and PR titles. For example: **Resolve an issue where memory leaks occur after cache errors**, not **GameCache.cpp Updated**.

We hope you'll take these suggestions in mind, and perhaps contribute again in the future!

PR #27, opened 2017-06-15
NUI: Improvements and fixes

Tungstenium:
Various cursor related improvements and fixes


code/components/nui-core/src/CefInput.cpp
```diff
@@ -55,6 +56,17 @@ namespace nui
 #endif
 
 		g_hasFocus = hasFocus;
+		g_hasCursor = false;
+	}
+
+	void GiveFocus(bool hasFocus, bool hasCursor)
```
blattersturm: Is this code duplication needed? No caller relies on the old call anymore, and to preserve source compatibility the header file could contain `bool hasCursor = false`.

Tungstenium: Then no it isn't needed. I was unaware that you could add default values.

PR #32, opened 2017-06-23
NUI: Mouse cleanup

Tungstenium:
Combines GiveFocus(hasFocus) and GiveFocus(hasFocus, hasCursor)
Improves backwards compatibility with the fake cursor script

### DIFF
--- a/code/components/nui-core/src/CefInput.cpp
+++ b/code/components/nui-core/src/CefInput.cpp
@@ -23,7 +23,7 @@ bool isKeyDown(WPARAM wparam)
 
 namespace nui
 {
-	void GiveFocus(bool hasFocus)
+	void GiveFocus(bool hasFocus, bool hasCursor = false)
 	{
 #if defined(GTA_NY)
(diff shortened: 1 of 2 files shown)


code/components/nui-core/src/CefInput.cpp
```diff
@@ -23,7 +23,7 @@ bool isKeyDown(WPARAM wparam)
 
 namespace nui
 {
-	void GiveFocus(bool hasFocus)
+	void GiveFocus(bool hasFocus, bool hasCursor = false)
```
ejb1123: Error when compiling with vs2017
```
Severity	Code	Description	Project	File	Line	Suppression State
Error	C2572	'nui::GiveFocus': redefinition of default argument: parameter 1	nui-core	C:\FiveM\dev\fivem\code\components\nui-core\src\CefInput.cpp	26	
```
should be 
```C++
void GiveFocus(bool hasFocus, bool hasCursor)
```

PR #50, opened 2017-07-20
Workaround for Native calls not working in threaded server environment

msquerade:
Calling `Function.Call` inside of a thread would normally fail on `ResourceManager::GetCurrent()` (ResourceManager.cpp). Another interesting side effect is that `Function.Call` will get called twice: Once by the main (ScriptHost) thread, and once by the background thread. This can be abused to create a workaround for the (previously) failed assertion by only allowing the main thread to perform this task. The native call will be performed by the 'main' thread, but its value returned to the background thread like in this demo:
```
public class Test : BaseScript
{
	public Test()
	{
		ThreadPool.QueueUserWorkItem( ( state ) => {
			string hostname = Function.Call<string>( Hash.GET_CONVAR, "sv_hostname", "" );
			Debug.WriteLine( "sv_hostname: {0}", hostname );
		} );
	}
}

// prints: sv_hostname: Dev server
```
It's ugly but in my opinion reasonable as a temporary solution for those who wish to use natives in threads.

### DIFF
--- a/code/client/clrcore/InternalManager.cs
+++ b/code/client/clrcore/InternalManager.cs
@@ -18,6 +18,7 @@ namespace CitizenFX.Core
 		private static int ms_instanceId;
 
 		public static IScriptHost ScriptHost { get; private set; }
+		public static int ScriptHostThreadID { get; private set; }
 
 		[SecuritySafeCritical]
 		public InternalManager()
@@ -39,6 +40,7 @@ namespace CitizenFX.Core
 		{
 			ScriptHost = host;
 			ms_instanceId = instanceId;
+			ScriptHostThreadID = Thread.CurrentThread.ManagedThreadId;
 		}
 
 		[SecuritySafeCritical]
--- a/code/client/clrcore/Native.cs
+++ b/code/client/clrcore/Native.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Runtime.InteropServices;
 using System.Security;
+using System.Threading;
 
 namespace CitizenFX.Core.Native
 {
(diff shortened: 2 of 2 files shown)


blattersturm:
Huh, what code path exactly causes the native to get invoked _twice_? As it seems from the code now, any and all natives would end up no-operation: are you sure this thread doesn't execute twice or whatever?

msquerade:
Positive. If we take this example code:

```
public class Test : BaseScript
{
	public Test()
	{
		Debug.WriteLine( "(main) sv_hostname: {0}", Function.Call<string>( Hash.GET_CONVAR, "sv_hostname", "" ) );
		
		ThreadPool.QueueUserWorkItem( ( state ) => {
			Debug.WriteLine( "(thread) sv_hostname: {0}", Function.Call<string>( Hash.GET_CONVAR, "sv_hostname", "" ) );
		} );
	}
}
```

And add this line to the Function.Call method:
```
        public static T Call<T>(Hash hash, params InputArgument[] arguments)
        {
			Debug.WriteLine( "Executing on thread: {0}", Thread.CurrentThread.ManagedThreadId );
			return (T)InvokeInternal(hash, typeof(T), arguments);
        }
```

This is the dumped result:
> Executing on thread: 1
> (main) sv_hostname: Dev server
> Executing on thread: 6
> Executing on thread: 1
> (thread) sv_hostname: Dev server

msquerade:
After more thorough testing I was able to create some cases where the native *doesn't* also get called on the main thread. I don't advise merging this into master.

PR #52, opened 2017-07-27
clrcore:External update for mpgunrunning

mraes:
- Updated Vehicle, Weapon and WeaponComponent hash enums (thanks @thers).
- Added code for correct DisplayName and AttachmentPoint of WeaponComponents.
- Added Enums for (Mk2) Weapon Livery and Livery Color.
- Added Mk2 Tints to Tint enum.


code/client/clrcore/External/VehicleHash.cs
```diff
@@ -1,481 +1,526 @@
-namespace CitizenFX.Core
+﻿namespace CitizenFX.Core
 {
-	public enum VehicleHash : uint
-	{
```
blattersturm: Why is this entire file modified? Are you using `core.autocrlf=true` and did editorconfig get read?

code/client/clrcore/External/Weapon.cs
```diff
@@ -251,6 +333,39 @@ public bool CanUseOnParachute
...
+			}
+		}
+
+		public void SetLivery(WeaponLivery liveryID, WeaponLiveryColor colorID)
+		{
+			if (IsMk2)
+			{
+				WeaponComponent comp = Components.GetMk2CamoComponent((int)liveryID);
+				comp.Active = true;
+				Function.Call((Hash)0x9FE5633880ECD8ED, Game.PlayerPed.Handle, Hash, comp.ComponentHash, (int) colorID);
```
blattersturm: Is this hash not already documented? Where did this hash originate?

mraes: I cannot find it in the natives reference, and as far as I know it has not been given a name yet.
Source of info is http://gtaforums.com/topic/890508-gta-v-gunrunning-weapon-livery-colors/?p=1069683952.

code/client/clrcore/External/VehicleHash.cs
mraes: Should be fixed in https://github.com/mraes/fivem/commit/9a567ffbae1643197d8c044efe957ddd68ec4a4e. You'll also see that some hashes changed because they were just wrong.
Largely copied from https://github.com/crosire/scripthookvdotnet/commit/8e9103a7128b9a9290dddadff10ddca3822fc16d

PR #58, opened 2017-08-08
Slightly better ui

thers:


blattersturm:
@thers why even pr it now :( dunno even if it'll be working fine in game when m61

PR #66, opened 2017-08-22
Avoid extra call to event system for exports

joelwurtz:
It introduce a cache for the reference of the callback function needed to call an export.
It should reduce latency when using exports function.
This cache is reset each time a resource start to take into account updated functions

### DIFF
--- a/data/shared/citizen/scripting/lua/scheduler.lua
+++ b/data/shared/citizen/scripting/lua/scheduler.lua
@@ -185,7 +185,7 @@ if IsDuplicityVersion() then
 
 		return TriggerClientEventInternal(eventName, playerId, payload, payload:len())
 	end
-	
+
 	RegisterServerEvent = RegisterNetEvent
 	RconPrint = Citizen.Trace
 	GetPlayerEP = GetPlayerEndpoint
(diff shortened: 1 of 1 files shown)


data/shared/citizen/scripting/lua/scheduler.lua
```diff
@@ -358,6 +358,9 @@ local function getExportEventName(resource, name)
 	return string.format('__cfx_export_%s_%s', resource, name)
 end
 
+-- callback cache to avoid extra call to serialization / deserizalition process at each time getting an export
```
blattersturm: Typo? 😛 `deserialization`.

data/shared/citizen/scripting/lua/scheduler.lua
```diff
@@ -373,6 +376,9 @@ AddEventHandler(('on%sResourceStart'):format(IsDuplicityVersion() and 'Server' o
 			end)
 		end
 	end
+		
+	-- Clear cache when resource start, so we can get the new exports for a reloaded module
```
blattersturm: It'd be a good idea to also clear on a resource stop, or one may accidentally call an invalid ref in the C++ engine, whereas the original code would error out.

joelwurtz:
should be good

PR #67, opened 2017-08-24
Revert unsequential packet change and reduce how frequently 0x53FFFA3F is transmitted.

bladecoding:
# 1610d7f
I didn't see the 0x53FFFA3F packet anywhere else in the repository. My only guess is that it's used to keep the ping up-to-date on the server end when there is a lull in reliable server->client packets.

# 1472989
I can't repro unsequential packets being dropped due to poor network conditions(out of order, delayed, etc). However I'm able to easily repro unreliable sequential packets being dropped due to network hiccups.

### DIFF
--- a/code/components/citizen-server-impl/src/GameServer.cpp
+++ b/code/components/citizen-server-impl/src/GameServer.cpp
@@ -629,7 +629,7 @@ namespace fx
 						outPacket.Write(packetLength);
 						outPacket.Write(packetData.data(), packetLength);
 
-						targetClient->SendPacket(1, outPacket, (ENetPacketFlag)0);
+						targetClient->SendPacket(1, outPacket, ENET_PACKET_FLAG_UNSEQUENCED);
 
 						client->SetHasRouted();
 					}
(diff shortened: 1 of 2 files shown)


blattersturm:
The 0x53FFFA3F (msgEnd) packet probably was a leftover from the old server, as it actually played a role there.

However, unsequenced packets don't get dropped due to _network conditions_, they get dropped _on the recipient_ due to ENet's handling of them (a sliding window versus directly sending, see [url=https://github.com/freeminer/enet/blob/aefa2bd251d9a08ebdf42ba3253118a4999a48c0]/protocol.c#L469]enet_protocol_handle_send_unsequenced[/url] and compare to [url=https://github.com/freeminer/enet/blob/aefa2bd251d9a08ebdf42ba3253118a4999a48c0/protocol.c#L492]enet_protocol_handle_send_unreliable[/url].

The client change's effects are doubtful (they were only deployed to `production` a few days after the server change was confirmed-working), however the server-side change should _not_ be reverted as it actually _improved_ network reliability from a level where random packets would get dropped 'somewhere' to a level sort-of equivalent to the legacy server, which _also_ sequenced the entire UDP channel.

bladecoding:
Sorry, I managed to miss that part in enet_protocol_handle_send_unsequenced. Although it seems both unsequenced and unreliable have chances to drop delivered packets. I'm curious was the unsequenced change tested separately from the throttle change?

PR #73, opened 2017-09-01
Add possibility to detach current thread

joelwurtz:
This will allow to allow some mods to have "transparent" async.

Like in MySQL, would allow to have async call while using a sync api

```
local dbResult = MySQL.execute(...)
```

The execute method would then call in async the query, then pause the current thread and resume it. It should be detached as i don't want the scheduler to resume it if async call is too slow.

Also having this detached may ensure that if someone use this function it doesn't fucked up the scheduler.

I feel like, there is more safeguard to add here, would just bring this as a first shot will certainly add some more commits.

### DIFF
--- a/data/shared/citizen/scripting/lua/scheduler.lua
+++ b/data/shared/citizen/scripting/lua/scheduler.lua
@@ -9,12 +9,6 @@ function Citizen.CreateThread(threadFunction)
 	})
 end
 
-function Citizen.DetachCurrentThread()
-	table.remove(threads, curThreadIndex)
-
-	return curThread
-end
-
 function Citizen.Wait(msec)
 	curThread.wakeTime = GetGameTimer() + msec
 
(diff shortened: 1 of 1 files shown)


blattersturm:
This doesn't seem to handle CreateThreadNow threads, correct?

izio38:
Correct, but CreatThreadNow is used by itself and by events only, and MySQL Async doesn't use events I guess

blattersturm:
Callers might want to call routines that detach coroutines from events, not supporting this will cause weird undefined behavior even with @joelwurtz's resource.

joelwurtz:
Yes it must handle the CreateThreadNow also, will update this week end

joelwurtz:
Also it's missing a ReattachSystem, as otherwise once a mysql query has been done, Citizen.Wait would be unusable it need to have the following process : 

 1. Query
 2. Detach
 3. Resume from the callback
 4. Reattach
 5. Return result

joelwurtz:
So instead of allowing people to get current thread and let them know all of the logic of the scheduler, i added an await function so that every async function where the last parameter is a callback can be awaitable.

This is a better API imo, as it can be used also with the PerformHttpRequest (and all async functions using a callback) by doing this : 

````
function AwaitablePerformHttpRequest(url, method, data, headers, cb)
    PerformHttpRequest(url, cb, method, data, headers)
end

local Response = Citizen.Await(AwaitablePerformHttpRequest, 'http://test.com', 'GET', "", {})
```

izio38:
Oh this is great! Maybe we could improve it to add possibility to have Cb client -> server and vice-versa?

blattersturm:
@joelwurtz That's some weird syntax, wouldn't it make more sense to invent some sort of minimal promise type than do function calls that aren't actually function calls?

I don't think we'd like to see anything like this in the scheduler anyhow... :/

joelwurtz:
So basically have to add a promise implementation ? And then base the await api on it ?

blattersturm:
@joelwurtz Either that, figure out a better way to handle callbacks, or simply don't have an await API at all and stick with the original plan of exposing the coroutine, as this way of calling is just *nasty*.

joelwurtz:
Would prefer to have a promise implementation, as in the futur would love to build a rpc system between server / client based on this

joelwurtz:
Closed in favor of #75

PR #75, opened 2017-09-02
Add promise / await implementation

joelwurtz:
Replace #73

### DIFF
--- a/code/components/citizen-scripting-lua/src/LuaScriptRuntime.cpp
+++ b/code/components/citizen-scripting-lua/src/LuaScriptRuntime.cpp
@@ -194,7 +194,7 @@ public:
 
 LuaScriptRuntime::~LuaScriptRuntime()
 {
-	
+
 }
 
 static int lua_error_handler(lua_State* L);
@@ -203,7 +203,7 @@ OMPtr<LuaScriptRuntime> LuaScriptRuntime::GetCurrent()
 {
 	OMPtr<IScriptRuntime> runtime;
 	LuaScriptRuntime* luaRuntime;
-	
+
 	assert(FX_SUCCEEDED(fx::GetCurrentScriptRuntime(&runtime)));
 
 #ifdef _DEBUG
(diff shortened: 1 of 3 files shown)


code/components/citizen-scripting-lua/src/LuaScriptRuntime.cpp
```diff
@@ -194,7 +194,7 @@ class LuaScriptRuntime : public OMClass<LuaScriptRuntime, IScriptRuntime, IScrip
 
 LuaScriptRuntime::~LuaScriptRuntime()
 {
-
```
blattersturm: Seriously? Fix your editor. :/

blattersturm:
In general, looks decent. Got any example usage?

joelwurtz:
In mysql it would do the following, async will return promise instead of using callback :

```lua
local promise = MySQL.Async.fetchScalar("SELECT 1)
```

Then people would be able to use the next function

```lua
promise:next(function (result)
     -- Play with data
end)
```

Or there would be able to use Citizen.Await to wait (but in a unblocking way) for the result

```lua
local result = Citizen.Await(promise)
```

On the implementation side, i need to add an error callback / value to my exports, and it would work with something like this

```lua
function Async.fetchScalar(query, params)
    p = promise.new()

    export['mysql-async']:mysql_async_fetch_scalar(query, params, function(result, error)
        if error then
            p:reject(error)
        else
            p:resolve(result)
        end
    end)

    return p
end
```

code/components/citizen-scripting-lua/src/LuaScriptRuntime.cpp
joelwurtz: Don't need fixing, it just try to be consistent, anyway readd them

joelwurtz:
Made some tests with https://github.com/brouznouf/fivem-mysql-async/tree/feature/promise-await

works perfectly

PR #77, opened 2017-09-08
Remove Double Legacy

Metrela:
What I removed is what's throwing the secondary legacy server error message, doesn't seem like it needs to be in the code since it already verifies another way.


blattersturm:
This actually breaks code paths that literally _get used_, so.. how about no?

PR #78, opened 2017-09-10
Tracking slow resource / code

joelwurtz:
Will not say this is perfect but it's a start to know what is slow

### DIFF
--- a/data/shared/citizen/scripting/lua/scheduler.lua
+++ b/data/shared/citizen/scripting/lua/scheduler.lua
@@ -3,9 +3,15 @@ local curThread
 local curThreadIndex
 
 function Citizen.CreateThread(threadFunction)
+	local callerInfo = debug.getinfo(2, 'Sl')
+
 	table.insert(threads, {
 		coroutine = coroutine.create(threadFunction),
-		wakeTime = 0
+		wakeTime = 0,
+		resource = GetCurrentResourceName(),
+		file = callerInfo.short_src,
+		line = callerInfo.currentline,
+		wait = 0
 	})
 end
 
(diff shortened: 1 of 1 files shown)


data/shared/citizen/scripting/lua/scheduler.lua
```diff
@@ -21,17 +27,32 @@ CreateThread = Citizen.CreateThread
...
 
 	-- add new thread and save old thread
 	local oldThread = curThread
 	curThread = t
 
+	local startAt = GetGameTimer()
 	local result, err = coroutine.resume(coro)
+	local time = GetGameTimer() - startAt
+
+	if time > 10 then
```
joelwurtz: Not sure about this limit

joelwurtz:
An example of what we get after this 

slow-example

data/shared/citizen/scripting/lua/scheduler.lua
Hellslicer: `>= 150` for consistency with the underneath hitch warning?

joelwurtz: Not really a hitch warning is for the whole frame where there can be many threads executed here

Hellslicer: IMO something between 50 and 100 would be fine, 10ms is way too low

minipunch:
Why was this not merged? I am currently trying to find the source of some hitch warnings I am getting and this seems like it would be a helpful change.

Jarrrk:
@minipunch because a better resource monitor was implemented - In the F8 console type resmon 1

vecchiotom:
this is serverside though @Jarrrk

PichotM:
probably not pushed since it may impacts client performance
and if you want to use it, you can just edit your sheduler ^_^

minipunch:
@PichotM True.. I might just edit the scheduler manually to use it. What makes you think it might impact client performance?

sirarrowhead:
@minipunch I did it a long time ago, i have not noticed any worse peformance

PR #83, opened 2017-11-20
Remove double / in find resource message

Kngrektor:
https://prnt.sc/hcwsbh
Kinda did this patch on the web, can't imagine it'd break tho.


blattersturm:
breaks category folders, closed.

nice try though.

Kngrektor:
rip 😢

PR #84, opened 2017-11-21
Ability to use specific port for localhost connect

Disquse:
A little update that allow developers use different port. I'm using 33333 port on my local server, so I forced to use Direct connect or console to connect. If "Custom port" is empty, it will use 30120. Also this setting support saving. I don't know can it be useful someone besides me, but still decided to share.
preview

### DIFF
--- a/ext/cfx-ui/src/app/app-nav.component.ts
+++ b/ext/cfx-ui/src/app/app-nav.component.ts
@@ -43,7 +43,6 @@ export class AppNavComponent extends Translation {
 	}
 
 	connectToLocal() {
-		console.log(this.localhostPort);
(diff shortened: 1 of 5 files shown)


michaelcerne:
Maybe we could refactor the "localhost" menu option into just "Quick Connect"

While my primary use would be for a development server (hosted externally), the system could be more widespread for server owners to quickly connect to their own servers.

Just allowing a custom IP along with a port and the renaming of the menu option would be awesome.

Thoughts?

blattersturm:
Perhaps @michaelcerne's feature could be implemented as a separate (2-click? pre-filling the Direct Connect form?) 'reconnect to last' feature.

Disquse:
Good idea, I'll try to implement, thanks

michaelcerne:
Great, let me know if you need any assistance.

PR #85, opened 2017-11-22
Replace legacy LICENSE header on old sources with new website etc.

michaelcerne:
Replaced the old header, which read:
```
/*
 * This file is part of the CitizenFX project - http://citizen.re/
 *
 * See LICENSE and MENTIONS in the root of the source tree for information
 * regarding licensing.
 */
```
with a new, more modern, variant:
```
/*
 * This file is part of FiveM (https://fivem.net), created by CitizenFX (https://github.com/citizenfx)
 *
 * See root directory for information regarding LICENSE and other instruction.
 * 
 */
```

Also, I modified a few LUA scripts using os.is(), which is deprecated. Replaced with os.istarget().

Any reason to keep the old sources, with a 403 domain?


blattersturm:
Giant pull request trying to change every single source file? Yeah, no.

michaelcerne:
Should we just leave the old citizen.re though? Not trying to mess everything up, haha, just wanted to see to the new info. @blattersturm

PR #86, opened 2017-11-23
Save last direct connection to localstorage (quick connect)

michaelcerne:
On direct connect input update, the address is saved to the user's storage.

When the direct connection screen is opened, the input is pre-filled with the last used address.

### DIFF
--- a/ext/cfx-ui/src/app/servers/direct/direct-connect.component.ts
+++ b/ext/cfx-ui/src/app/servers/direct/direct-connect.component.ts
@@ -94,6 +94,7 @@ export class DirectConnectComponent implements OnInit, AfterViewInit {
 
     ngOnInit() {
 		this.addr = localStorage.getItem('directLast');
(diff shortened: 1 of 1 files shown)


Disquse:
Good job, but there is a bug with loading server's data (at least for me). It's doesn't start to ping server that you have in edit, infinite "Loading...". User can fix it by changing any character in edit, but this isn't good anyway. This happens because when you change edit's text by script, this script wouldn't get "ngModelChange" event that trigger "addrChanged".
 You can fix it by adding `this.addrChanged(localStorage.getItem('directLast'));` after `this.addr = localStorage.getItem('directLast'); `

michaelcerne:
Awesome, thanks. I added a commit which should hopefully take care of that problem.

I can't test it as I'm away from my work station, let me know how it goes.

michaelcerne:
Closing as it looks like it was committed by another.

PR #89, opened 2017-12-15
Fixes: Memory leak in CitizenTaskScheduler.Tick

bladecoding:
* Root tasks were added m_inTickTasks and never cleared.
* The memory leak can be seen by simply adding `Debug.WriteLine( $"[TASK] {m_inTickTasks.Count}" );` to the end of CitizenTaskScheduler.Tick and loading up a C# plugin with an OnTick event.
* This is fixed by setting the root m_inTickTasks to null. The locking is also changed a bit to support m_inTickTasks being null. An alternative fix might be to remove from the lastInTickTasks list when a task is completed.

### DIFF
--- a/code/client/clrcore/CitizenTaskScheduler.cs
+++ b/code/client/clrcore/CitizenTaskScheduler.cs
@@ -52,7 +52,8 @@ namespace CitizenFX.Core
 
     class CitizenTaskScheduler : TaskScheduler
     {
-		private List<Task> m_inTickTasks = new List<Task>();
+		private static readonly object m_inTickTasksLock = new object();
+		private List<Task> m_inTickTasks;
 
         private readonly List<Task> m_runningTasks = new List<Task>();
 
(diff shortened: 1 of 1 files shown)


blattersturm:
Is this pull request actually setting `m_inTickTasks` to null anywhere? It does not appear to do so. 😕

RiC0MD:
We've been running this patch for about a week now and good results, memory leak has ceased and no issues reported.

blattersturm:
So the patch does work on the _server_, does it however break the client if applied to the client runtime library?

RiC0MD:
We're on the fly patching the client side with the above PR, so all testing has been client side, but its 100% working great client/server right now. @bladecoding / myself are with FamilyRP, where its presently being used.  Before hand users were reporting 4-12GB of mem usage after 30-45+ mins and FPS loss, after we started patching their games normal GTA mem usage / FPS rates.

blattersturm:
Noted! Just making sure due diligence was applied to the PR, will click that lovely green 'rebase and merge' button then. 😄

RiC0MD:
Perfect, thank you!

PR #90, opened 2017-12-20
wheel camber and wheel track width natives

themonthofjune:
This pull request implements 4 extra vehicle-related natives,
[GS]etVehicleWheelCamber, and [GS]etVehicleWheelTrackWidth

### DIFF
--- a/code/components/extra-natives-five/src/VehicleExtraNatives.cpp
+++ b/code/components/extra-natives-five/src/VehicleExtraNatives.cpp
@@ -151,6 +151,9 @@ const int CanWheelsBreakOffset = 0x893;
 const int BlinkerState = 0x899;
 
 // Wheel class
+const int WheelCamberOffset = 0x008;
+const int WheelCamberInvertedOffset = 0x010;
+const int WheelTrackWidthOffset = 0x030;
 const int WheelTyreRadiusOffset = 0x110;
 const int WheelRimRadiusOffset = 0x114;
 const int WheelTyreWidthOffset = 0x118;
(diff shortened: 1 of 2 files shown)


code/components/extra-natives-five/src/VehicleExtraNatives.cpp
```diff
@@ -151,6 +151,9 @@ const int CanWheelsBreakOffset = 0x893;
 const int BlinkerState = 0x899;
 
 // Wheel class
+const int WheelCamberOffset = 0x008;
```
blattersturm: As seen a few lines below, is equivalent to Xrot. Why have 2 natives doing the same thing?

code/components/extra-natives-five/src/VehicleExtraNatives.cpp
```diff
@@ -352,4 +355,25 @@ static HookFunction initFunction([]()
...
+	{
+		*reinterpret_cast<float *>(wheelAddr + WheelTrackWidthOffset) = context.GetArgument<float>(2);
+	}));
+
+	fx::ScriptEngine::RegisterNativeHandler("GET_VEHICLE_WHEEL_CAMBER", makeWheelFunction([](fx::ScriptContext& context, fwEntity* vehicle, uintptr_t wheelAddr)
+	{
+		context.SetResult<float>(*reinterpret_cast<float *>(wheelAddr + WheelCamberOffset));
+	}));
+
+	fx::ScriptEngine::RegisterNativeHandler("SET_VEHICLE_WHEEL_CAMBER", makeWheelFunction([](fx::ScriptContext& context, fwEntity* vehicle, uintptr_t wheelAddr)
```
blattersturm: Given that XROT also uses 0x008/0x010, has equivalent code now. Is this intended?

ext/natives/codegen_cfx_natives.lua
```diff
@@ -745,12 +745,12 @@ native 'GET_VEHICLE_WHEEL_SPEED'
 	returns 'float'
 	doc [[
 	<summary>
-	Gets speed of a wheel at the tyre. 
+	Gets speed of a wheel at the tyre.
 	Max number of wheels can be retrieved with the native GET_VEHICLE_NUMBER_OF_WHEELS.
 	</summary>
 	<returns>An integer.</returns>
 ]]
-
```
blattersturm: Useful change! Have people still not learned to check diffs to see if their editor has decided to be 'smart' in a third-party codebase?

Please fix and **force-push** to the PR branch. Making a new commit in the same branch, contrary to popular belief, will not fix this kind of screwups.

ext/natives/codegen_cfx_natives.lua
```diff
@@ -1299,6 +1299,40 @@ native 'SET_VEHICLE_WHEEL_XROT'
...
+
+native 'GET_VEHICLE_WHEEL_CAMBER'
+	arguments {
+		Vehicle 'vehicle',
+		int 'wheelIndex'
+	}
+	apiset 'client'
+	returns 'float'
+
+native 'SET_VEHICLE_WHEEL_CAMBER'
```
blattersturm: If this name is more accurate than XROT, rename XROT (and, in case someone else used it, add a compatibility alias - presumably would need a `hash` override and `alias` field in codegen definitions? would need verification by someone familiar with codegen).

PR #91, opened 2017-12-26
Implement Get/SetVehicleWheelXOffset.

Deltanic:
Continuation of https://github.com/citizenfx/fivem/pull/90, fixing commented issues.

Some notes:

`Get/SetVehicleWheelCamber` was removed in favor of `Get/SetVehicleWheelXrot`. An alias can be added in the future. At that point, I suggest to also alias `Get/SetVehicleWheelXrot` to `*WheelXRot` due to capitalization.

`Get/SetVehicleWheelTrackWidth` was renamed to `Get/SetVehicleWheelXOffset`. While `TrackWidth` is a friendly and familiar name, it is technically not correct as it implies the distance between both wheels on a single axle. Instead, track width can be correctly applied like so:
```lua
function SetVehicleFrontTrackWidth(vehicle, width)
  SetVehicleWheelXOffset(vehicle, 0, -width/2)
  SetVehicleWheelXOffset(vehicle, 1, width/2)
end
```

Also added a short doc on the natives, including a note to call the setter every frame as this may not necessarily be clear to everyone. Not sure if examples can contains triple backticks here.  Let me know if that needs to change.

Might work on a more generic `Get/SetVehicleWheelOffset` later, but the Y and Z axis seem to be a bit more funky than X axis so that needs additional work.

### DIFF
--- a/code/components/extra-natives-five/src/VehicleExtraNatives.cpp
+++ b/code/components/extra-natives-five/src/VehicleExtraNatives.cpp
@@ -151,6 +151,7 @@ const int CanWheelsBreakOffset = 0x893;
 const int BlinkerState = 0x899;
 
 // Wheel class
+const int WheelTrackWidthOffset = 0x030;
 const int WheelTyreRadiusOffset = 0x110;
 const int WheelRimRadiusOffset = 0x114;
 const int WheelTyreWidthOffset = 0x118;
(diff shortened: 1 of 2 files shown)


blattersturm:
tl;dm

code/components/extra-natives-five/src/VehicleExtraNatives.cpp
```diff
@@ -151,6 +151,7 @@ const int CanWheelsBreakOffset = 0x893;
 const int BlinkerState = 0x899;
 
 // Wheel class
+const int WheelXOffsetOffset = 0x030;
```
blattersturm: offsetoffset, ok

ext/natives/codegen_cfx_natives.lua
```diff
@@ -1299,6 +1299,42 @@ native 'SET_VEHICLE_WHEEL_XROT'
...
 
+native 'GET_VEHICLE_WHEEL_X_OFFSET'
+	arguments {
+		Vehicle 'vehicle',
+		int 'wheelIndex'
+	}
+	apiset 'client'
+	returns 'float'
+	doc [[
+	<summary>
```
blattersturm: no \<param\> 🥉

code/components/extra-natives-five/src/VehicleExtraNatives.cpp
Deltanic: best

PR #92, opened 2018-01-10
Update SteamComponent.cpp (replacing Source SDK Base 2007 with Grand Theft Auto V)

OfficialSkypo:
In an attempt to replace Source SDK Base 2007 app ID with the actual Grand Theft Auto V app ID, to perhaps show "Playing FiveM" when friends look at what you are playing.

I am not sure whether this even works (and if it does) if it is anything useful.


blattersturm:
This will actually break ticket authentication, and won't be possible without requiring all players to own a Steam copy. It also won't show 'playing FiveM' on the Steam Community web site, showing custom mod names was blocked by Valve a few years ago.

PR #97, opened 2018-01-28
Repair fix

YBa2Cu3O7:
Fixes the repair glitch that occurs when enabling an vehicle extra.

### DIFF
--- a/code/components/extra-natives-five/src/VehicleExtraNatives.cpp
+++ b/code/components/extra-natives-five/src/VehicleExtraNatives.cpp
@@ -11,9 +11,13 @@
 
 #include <Local.h>
 #include <Hooking.h>
+#include <GameInit.h>
 
 #include <limits>
 #include <bitset>
+#include <unordered_set>
+
+static std::unordered_set<fwEntity*> g_skipRepairVehicles{};
 
 static hook::cdecl_stub<fwEntity*(int handle)> getScriptEntity([]()
 {
@@ -363,4 +367,104 @@ static HookFunction initFunction([]()
 	{
 		*reinterpret_cast<float *>(wheelAddr + WheelXOffsetOffset) = context.GetArgument<float>(2);
 	}));
+
+	static struct : jitasm::Frontend
+	{
+		static bool ShouldSkipRepairFunc(fwEntity* VehPointer)
+		{
+			return g_skipRepairVehicles.find(VehPointer) != g_skipRepairVehicles.end();
+		}
+		virtual void InternalMain() override
+		{
+			//restore dl after asm call
+			mov(dl, 1);
+
+			//save parameter registers
+			push(rax);
+			push(rcx);
+			push(rdx);
+			push(r8);
+			
+			sub(rsp, 0x28);
+			mov(rax, (uintptr_t)ShouldSkipRepairFunc);
+			call(rax);
+			add(rsp, 0x28);
+
+			//restore parameter registers
+			pop(r8);
+			pop(rdx);
+			pop(rcx);
+
+			test(al, al);
+			jne("skiprepair");
+			pop(rax);
+			sub(rsp, 0x28);
+			AppendInstr(jitasm::InstrID::I_CALL, 0xFF, 0, jitasm::Imm8(2), qword_ptr[rax + 0x5D0]);
+			add(rsp, 0x28);
+			ret();
+			L("skiprepair");
+			pop(rax);
+			ret();
+		}
+	} asmfunc;
+
+	auto repairFunc = hook::get_pattern("48 8B 03 45 33 C0 B2 01 48 8B CB FF 90 D0 05 00 00 48 8B 4B 20",11);
+	hook::nop(repairFunc, 6);
+	hook::call_reg<2>(repairFunc, asmfunc.GetCode());
+
+	static struct : jitasm::Frontend
+	{
+		static void CleanupVehicle(fwEntity* VehPointer)
+		{
+			g_skipRepairVehicles.erase(VehPointer);
+		}
+		virtual void InternalMain() override
+		{
+			//overwritten assembly code
+			mov(rbx, rcx);
+			mov(qword_ptr[rcx], rax);
+
+			//save registers
+			push(rax);
+			push(rcx);
+			sub(rsp, 0x8);
+
+			//actual cleanup function call
+			mov(rax, (uintptr_t)CleanupVehicle);
+			sub(rsp, 0x20);
+			call(rax);
+			add(rsp, 0x20);
+
+			//restore registers
+			add(rsp, 0x8);
+			pop(rcx);
+			pop(rax);
+			ret();
+		}
+	} vehicleDeconstructorHook;
+	auto vehicleDeconstructor = hook::get_pattern("48 8B 43 20 48 8B 88 B0 00 00 00 75 09 66 FF 89",-0x1D);
+	hook::nop(vehicleDeconstructor, 0x6);
+	hook::call_reg<2>(vehicleDeconstructor, vehicleDeconstructorHook.GetCode());
+
+	OnKillNetworkDone.Connect([]() {
+		g_skipRepairVehicles.clear();
+	});
+
+
+	fx::ScriptEngine::RegisterNativeHandler("SET_VEHICLE_AUTO_REPAIR_DISABLED", [](fx::ScriptContext& context) {
+		auto vehHandle = context.GetArgument<int>(0);
+		auto shouldDisable = context.GetArgument<bool>(1);
+		fwEntity *entity = getScriptEntity(vehHandle);
+		if (shouldDisable) {
+			g_skipRepairVehicles.insert(entity);
+		}
+		else
+		{
+			auto k = g_skipRepairVehicles.find(entity);
+			if (k != g_skipRepairVehicles.end()) {
+				g_skipRepairVehicles.erase(k);
+				return;
+			}
+		}
+	});
 });
--- a/ext/natives/codegen_cfx_natives.lua
+++ b/ext/natives/codegen_cfx_natives.lua
@@ -692,6 +692,18 @@ native 'END_FIND_PICKUP'
 	apiset 'client'
 	returns 'void'
 
+native 'SET_VEHICLE_AUTO_REPAIR_DISABLED'
+	arguments{
+		Vehicle 'vehicle',
+		BOOL 'value'
+	}
+	doc [[
+	<summary>
+	Disables the vehicle from being repaired when a vehicle extra is enabled.
+	</summary>
+	<param name="vehicle">The vehicle to set disable auto vehicle repair.</param>
+	<param name="value">Setting the value to  true prevents the vehicle from being repaired when a extra is enabled. Setting the value to false allows the vehicle from being repaired when a extra is enabled.</param>
+	]]
 native 'GET_VEHICLE_FUEL_LEVEL'
 	arguments {
 		Vehicle 'vehicle'


code/components/extra-natives-five/src/VehicleExtraNatives.cpp
```diff
@@ -6,15 +6,19 @@
 */
 
 #include "StdInc.h"
+#include <unordered_set>
```
blattersturm: isn't this already included by stdinc? even if so, should be lower

code/components/extra-natives-five/src/VehicleExtraNatives.cpp
```diff
@@ -6,15 +6,19 @@
...
 #include <atArray.h>
 
 #include <Local.h>
 #include <Hooking.h>
+#include <GameInit.h>
 
 #include <limits>
 #include <bitset>
 
+static std::unordered_set<fwEntity*> vehicles{};
```
blattersturm: why the `{}`?

code/components/extra-natives-five/src/VehicleExtraNatives.cpp
```diff
@@ -363,4 +367,110 @@ static HookFunction initFunction([]()
 	{
 		*reinterpret_cast<float *>(wheelAddr + WheelXOffsetOffset) = context.GetArgument<float>(2);
 	}));
+
+	static struct : jitasm::Frontend {
```
blattersturm: code conventions: should have { on a newline

code/components/extra-natives-five/src/VehicleExtraNatives.cpp
```diff
@@ -363,4 +367,110 @@ static HookFunction initFunction([]()
...
+		static bool ShouldSkipRepairFunc(fwEntity* VehPointer)
+		{
+
```
blattersturm: random newline?

code/components/extra-natives-five/src/VehicleExtraNatives.cpp
```diff
@@ -363,4 +367,110 @@ static HookFunction initFunction([]()
...
+			for (auto it = vehicles.begin(); it != vehicles.end(); it++) {
```
blattersturm: a) old iterators and not C++11 stuff?
b) std::unordered_set::find, this will be a linear scan and slow

could be replaced with `return g_skipRepairVehicles.find(vehPointer) != g_skipRepairVehicles.end();`

code/components/extra-natives-five/src/VehicleExtraNatives.cpp
```diff
@@ -363,4 +367,110 @@ static HookFunction initFunction([]()
...
+			push(rax);
+			push(rdx);
+
+			sub(rsp, 0x28);
+			mov(rax, (uintptr_t)ShouldSkipRepairFunc);
+			call(rax);
+			add(rsp, 0x28);
+
+			pop(rdx);
+			cmp(al, true); //al is 1 if repair func should be skipped
```
blattersturm: `test(al, al)` to check if 0/nonzero, compilers tend to do this instead; jnz to skip then

code/components/extra-natives-five/src/VehicleExtraNatives.cpp
```diff
@@ -363,4 +367,110 @@ static HookFunction initFunction([]()
...
+			je("skiprepair");
+			pop(rax);
+			sub(rsp, 0x28);
+			AppendInstr(jitasm::InstrID::I_CALL, 0xFF, 0, jitasm::Imm8(2), qword_ptr[rax + 0x5D0]);
```
blattersturm: could this perhaps be an overload added to jitasm.h directly, depending on the return type of qword_ptr[...]?

code/components/extra-natives-five/src/VehicleExtraNatives.cpp
```diff
@@ -363,4 +367,110 @@ static HookFunction initFunction([]()
...
+			sub(rsp, 0x28);
+			mov(rax, (uintptr_t)ShouldSkipRepairFunc);
+			call(rax);
+			add(rsp, 0x28);
+
+			pop(rdx);
+			cmp(al, true); //al is 1 if repair func should be skipped
+			je("skiprepair");
+			pop(rax);
+			sub(rsp, 0x28);
```
blattersturm: technically could unwind stack and jmp, but works as well I'd guess

code/components/extra-natives-five/src/VehicleExtraNatives.cpp
```diff
@@ -363,4 +367,110 @@ static HookFunction initFunction([]()
...
+
+	static auto * repairFunc = hook::pattern("48 8B 03 45 33 C0 B2 01 48 8B CB FF 90 D0 05 00 00 48 8B 4B 20").get_first(11);
+	static auto * vehicleDeconstructor = hook::pattern("41 BE 00 00 80 BF 48 89").get_first(-0x1B);
+	hook::nop(vehicleDeconstructor, 0x6);
+	hook::nop(repairFunc, 6);
+	hook::call_rcx(repairFunc, asmfunc.GetCode());
+
+	static struct : jitasm::Frontend {
+
+		static void LogStub(fwEntity* VehPointer)
```
blattersturm: why is this called LogStub?

code/components/extra-natives-five/src/VehicleExtraNatives.cpp
```diff
@@ -363,4 +367,110 @@ static HookFunction initFunction([]()
...
+			AppendInstr(jitasm::InstrID::I_CALL, 0xFF, 0, jitasm::Imm8(2), qword_ptr[rax + 0x5D0]);
+			add(rsp, 0x28);
+			ret();
+			L("skiprepair");
+			pop(rax);
+			ret();
+		}
+	} asmfunc;
+
+	static auto * repairFunc = hook::pattern("48 8B 03 45 33 C0 B2 01 48 8B CB FF 90 D0 05 00 00 48 8B 4B 20").get_first(11);
```
blattersturm: use `hook::get_pattern`, also no need for it to be `static`.

also, `*` is implied by auto

code/components/extra-natives-five/src/VehicleExtraNatives.cpp
```diff
@@ -363,4 +367,110 @@ static HookFunction initFunction([]()
...
+	static auto * vehicleDeconstructor = hook::pattern("41 BE 00 00 80 BF 48 89").get_first(-0x1B);
+	hook::nop(vehicleDeconstructor, 0x6);
+	hook::nop(repairFunc, 6);
+	hook::call_rcx(repairFunc, asmfunc.GetCode());
+
+	static struct : jitasm::Frontend {
+
```
blattersturm: another random newline

code/components/extra-natives-five/src/VehicleExtraNatives.cpp
```diff
@@ -363,4 +367,110 @@ static HookFunction initFunction([]()
...
+			ret();
+			L("skiprepair");
+			pop(rax);
+			ret();
+		}
+	} asmfunc;
+
+	static auto * repairFunc = hook::pattern("48 8B 03 45 33 C0 B2 01 48 8B CB FF 90 D0 05 00 00 48 8B 4B 20").get_first(11);
+	static auto * vehicleDeconstructor = hook::pattern("41 BE 00 00 80 BF 48 89").get_first(-0x1B);
+	hook::nop(vehicleDeconstructor, 0x6);
```
blattersturm: why is this not below the vehicle destructor code?

code/components/extra-natives-five/src/VehicleExtraNatives.cpp
```diff
@@ -363,4 +367,110 @@ static HookFunction initFunction([]()
...
+			if (count > 0) {
+			}
+		}
+		virtual void InternalMain() override
+		{
+			mov(rcx, rax);
+			mov(r8d, r15d);
+
+			mov(rdi, rcx);
+			push(rax);
```
blattersturm: why save _all_ these registers, aren't some of them callee-saved?

code/components/extra-natives-five/src/VehicleExtraNatives.cpp
```diff
@@ -363,4 +367,110 @@ static HookFunction initFunction([]()
...
+			add(rsp, 0x28);
+			ret();
+			L("skiprepair");
+			pop(rax);
+			ret();
+		}
+	} asmfunc;
+
+	static auto * repairFunc = hook::pattern("48 8B 03 45 33 C0 B2 01 48 8B CB FF 90 D0 05 00 00 48 8B 4B 20").get_first(11);
+	static auto * vehicleDeconstructor = hook::pattern("41 BE 00 00 80 BF 48 89").get_first(-0x1B);
```
blattersturm: pattern isn't sufficiently version-independent

code/components/extra-natives-five/src/VehicleExtraNatives.cpp
```diff
@@ -363,4 +367,110 @@ static HookFunction initFunction([]()
...
+			add(rsp, 0x20);
+
+			pop(r11);
+			pop(r10);
+			pop(r9);
+			pop(r8);
+			pop(rdx);
+			pop(rcx);
+			pop(rax);
+			ret();
```
blattersturm: does this code move the correct value to `r14d` anywhere?

also, in fact, it's the constructor, not the destructor 😕 the destructor starts at `140E89E78` in 1103

code/components/extra-natives-five/src/VehicleExtraNatives.cpp
```diff
@@ -363,4 +367,110 @@ static HookFunction initFunction([]()
...
+	hook::nop(vehicleDeconstructor, 0x6);
+	hook::nop(repairFunc, 6);
+	hook::call_rcx(repairFunc, asmfunc.GetCode());
+
+	static struct : jitasm::Frontend {
+
+		static void LogStub(fwEntity* VehPointer)
+		{
+			auto count = vehicles.erase(VehPointer);
+			if (count > 0) {
```
blattersturm: useless variable and if?

code/components/extra-natives-five/src/VehicleExtraNatives.cpp
```diff
@@ -363,4 +367,110 @@ static HookFunction initFunction([]()
...
+	} vehicleDeconstructorHook;
+
+	hook::call_rcx(vehicleDeconstructor, vehicleDeconstructorHook.GetCode());
+	
+	OnKillNetworkDone.Connect([]() {
+		vehicles.clear();
+	});
+
+
+	fx::ScriptEngine::RegisterNativeHandler("SET_VEHICLE_AUTO_REPAIR", [](fx::ScriptContext context) {
```
blattersturm: `fx::ScriptContext&`, missing &

code/components/extra-natives-five/src/VehicleExtraNatives.cpp
```diff
@@ -363,4 +367,110 @@ static HookFunction initFunction([]()
...
+		auto vehHandle = context.GetArgument<int>(0);
+		auto boolval = context.GetArgument<bool>(1);
+		fwEntity *entity = getScriptEntity(vehHandle);
+		if (!boolval) {
```
blattersturm: `boolval` is a really useful name

ext/natives/codegen_cfx_natives.lua
```diff
@@ -699,6 +699,14 @@ native 'GET_VEHICLE_FUEL_LEVEL'
 	apiset 'client'
 	returns 'float'
 
+native 'SET_VEHICLE_AUTO_REPAIR'
+	arguments {
+		Vehicle 'vehicle',
+		BOOL 'value'
```
blattersturm: at least tag it as `enabled` or so, also add a `doc` comment

code/components/extra-natives-five/src/VehicleExtraNatives.cpp
```diff
@@ -363,4 +367,110 @@ static HookFunction initFunction([]()
...
+			vehicles.insert(entity);
+		}
+		else
+		{
+			auto k = vehicles.find(entity);
+			if (k != vehicles.end()) {
+				vehicles.erase(*k);
```
blattersturm: erase also can take the pointer itself 😕

code/components/extra-natives-five/src/VehicleExtraNatives.cpp
YBa2Cu3O7: https://docs.microsoft.com/en-us/cpp/build/caller-callee-saved-registers

code/components/extra-natives-five/src/VehicleExtraNatives.cpp
YBa2Cu3O7: it is list initialization

blattersturm: not needed, default ctor will be equivalent

code/components/extra-natives-five/src/VehicleExtraNatives.cpp
blattersturm: even then, look at the function to see what is in use and what isn't.

code/components/extra-natives-five/src/VehicleExtraNatives.cpp
ifarbod: Nice variable naming

code/components/extra-natives-five/src/VehicleExtraNatives.cpp
```diff
@@ -363,4 +367,120 @@ static HookFunction initFunction([]()
...
+	static struct : jitasm::Frontend
+	{
+		static void CleanupVehicle(fwEntity* VehPointer)
+		{
+			g_skipRepairVehicles.erase(VehPointer);
+		}
+		virtual void InternalMain() override
+		{
+			//restore rcx after asm call
+			mov(rcx, rdi);
```
blattersturm: `rdi` isn't the place where the vehicle pointer is in all cases! there's other vehicle classes that _don't_ use rdi.

rdx is unused at this point, so you could use `call_reg<2>` or similar

code/components/extra-natives-five/src/VehicleExtraNatives.cpp
```diff
@@ -363,4 +367,120 @@ static HookFunction initFunction([]()
...
+
+			//overwritten assembly code
+			mov(rbx, rcx);
+			mov(qword_ptr[rcx], rax);
+
+			//save registers
+			push(rax);
+			push(rcx);
+			push(rdx);//
```
blattersturm: does `//` here mean 'unused, shouldn't be saved actually'?

code/components/extra-natives-five/src/VehicleExtraNatives.cpp
blattersturm: seems fixed

ext/natives/codegen_cfx_natives.lua
```diff
@@ -692,6 +692,18 @@ native 'END_FIND_PICKUP'
...
+	arguments{
+		Vehicle 'vehicle',
+		BOOL 'value'
+	}
+	doc [[
+	<summary>
+	Disables the vehicle from being repaired when a vehicle extra is enabled.
+	</summary>
+	<param name="vehicle">The vehicle to set disable auto vehicle repair.</param>
+	<param name="value">The bool value to set.</param>
```
blattersturm: does 'the bool value' mean 'enabled'? 'disabled'?

code/components/extra-natives-five/src/VehicleExtraNatives.cpp
```diff
@@ -363,4 +367,120 @@ static HookFunction initFunction([]()
...
+		static bool ShouldSkipRepairFunc(fwEntity* VehPointer)
+		{
+			return g_skipRepairVehicles.find(VehPointer) != g_skipRepairVehicles.end();
+		}
+		virtual void InternalMain() override
+		{
+			//restore rcx after asm call
+			mov(rcx, rbx);
+			//save registers
+			push(rax);
```
blattersturm: you're replacing a `call`, so shouldn't have to save any registers at all (other than `rax` you use yourself), as the compiler already assumes these registers will not be saved.

do make sure to keep stack alignment to 16, so remove registers in pairs of two, or change the stack sub/add alignment.

blattersturm:
Latest changeset doesn't build, please fix whatever is breaking C# NativesFive.cs building (run `prebuild_natives.cmd`) and try building on your own end.

PR #99, opened 2018-02-19
net-tls: fix compound line

jakenotjacob:
Small change to prevent some compilers from yelling.


blattersturm:
you're supposed to use a recent compiler anyway

jakenotjacob:
That or variable initialization not being an expression or referring to itself per C# spec. Compiler works too though.

PR #100, opened 2018-02-27
Disable watermark in main menu

Disquse:
Preview: https://i.imgur.com/Z6qoou4.png
Watermark exists while in-game

### DIFF
--- a/code/components/font-renderer/src/GtaGameInterface.cpp
+++ b/code/components/font-renderer/src/GtaGameInterface.cpp
@@ -9,6 +9,7 @@
 #include "FontRendererImpl.h"
 #include <DrawCommands.h>
 #include <grcTexture.h>
+#include <ICoreGameInit.h>
 
 #include "memdbgon.h"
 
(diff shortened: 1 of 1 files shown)


vecchiotom:
*exists :^)

Cmoen11:
that whole switch statement is a horror to watch..

code/components/font-renderer/src/GtaGameInterface.cpp
```diff
@@ -287,6 +288,17 @@ static InitFunction initFunction([] ()
 	static std::random_device random_core;
 	static std::mt19937 random(random_core());
 
+	static bool shouldDraw = false;
+
+	Instance<ICoreGameInit>::Get()->OnGameRequestLoad.Connect([]()
```
lfuelling: Having to rewrite the HUD because of this crap...

Subtixx: This watermark is just annoying as fuck

blattersturm: @Subtixx
> This watermark is just annoying as fuck

Other than the fact that a line comment on a pull request from 2 years ago isn't a good place to provide feedback, what's so 'annoying as fuck' about it?

Subtixx: A function to let server owners decide on if they want to show it or not. The screenspace is valuable, and having the watermark means 1 less space to put information. I came across this PR when googling how to disable it.

blattersturm: You can still 'put information' in any corner, the watermark doesn't take up some magical 'space' as it's semi-transparent.

'Disabling' it is also not going to happen, it's there for fairly evident recognizability reasons, and other games also tend to have a persistent on-screen overlay as well these days.

lfuelling: "Other software is doing the same" is a pretty garbage standard to develop by imo.

technetium-cfx: Which is why that's not a sole reason to develop by. Locking this thread now, before this derails further.

PR #105, opened 2018-04-10
Updated some natives

GHMatti:
Fixed some native parameter types including some descriptions.

Updated the description of 0xB8A850F20A067EB6 to match my latest research into how that function works.

The values for the other anchors (apart from the known top left) were not obtained through decompiled scripts, but simply by bruteforcing.


GHMatti:
The anchors are behaving more oddly need to research more. The middle one was still anchored to the right.

There might be some more numbers that are not within a byte, but I will not check.

PR #107, opened 2018-04-15
Fix CEF (NUI) on older graphic units

Flatracer:
Reference: https://github.com/daktronics/cef-mixer/issues/10


vecchiotom:
you sir are an hero 😄

blattersturm:
This actually won't fix the issue for old GPUs, for this conditional enabling of GLES2 is still needed in CEF.

Once this is resolved in the public CEF PR, the functionality will be added, but this PR is literally useless at this time.

PR #112, opened 2018-05-28
Fixed Scripthook Bypass

Myrror:
In CheatEngine, the ShAllowed integer can be changed if the right value is selected. This allows bypasses to occur. This patch prevents this.

please let the snail be free,
please merge.

also whitelist @Myrror to #support-plus as i am smart on the FiveM discord


blattersturm:
Very funny. Not really.

PR #114, opened 2018-05-31
Shared Script

freznva:


code/components/citizen-scripting-core/src/ResourceScriptingComponent.cpp
```diff
@@ -64,6 +64,8 @@ ResourceScriptingComponent::ResourceScriptingComponent(Resource* resource)
 #endif
 			);
 
+			auto sharedScripts = metaData->GetEntries("shared_script")
```
blattersturm: Does this even compile? There's a _missing semicolon_.

code/components/citizen-scripting-core/src/ResourceScriptingComponent.cpp
```diff
@@ -78,6 +80,15 @@ ResourceScriptingComponent::ResourceScriptingComponent(Resource* resource)
 					}
 				}
 
+				for (auto& sharedScript : sharedScripts)
```
blattersturm: Code duplication is annoying. It's probably better to make an outer loop that iterates over lists, such as:

```cpp
for (auto& list : { blahScripts, sharedScripts })
{
    for (auto& script : list)
    {
        // ...
    }
}
```

code/components/citizen-scripting-core/src/ResourceScriptingComponent.cpp
```diff
@@ -205,6 +218,19 @@ void ResourceScriptingComponent::CreateEnvironments()
 					}
 				}
 			}
+
+			for (auto& sharedScript : sharedScripts)
```
blattersturm: Same code duplication issue applies here.

code/components/citizen-scripting-core/src/ResourceScriptingComponent.cpp
freznva: Yes, it compiles. But idk how it came out with a semicolon

freznva:
wow, what have i done...

PR #120, opened 2018-06-15
clrcore: Adds Debug.WriteLine overload without formatting arguments

bladecoding:
* This is to match the behavior of System.Diagnostics.Debug.WriteLine which doesn't try to format the string unless you provide additional arguments. This was causing errors if you had unescaped curly brackets in your string and were expecting the System.Diagnostics.Debug.WriteLine behavior. e.g. `Debug.WriteLine("{test}")`

This resolves issues like this https://forum.fivem.net/t/c-debug-writeline-without-parameters-still-tries-to-format-the-string/97667

### DIFF
--- a/code/client/clrcore/Debug.cs
+++ b/code/client/clrcore/Debug.cs
@@ -25,6 +25,11 @@ namespace CitizenFX.Core
         {
             Write("\n");
         }
+				
+        public static void WriteLine(string data)
+        {
+            Write(data + "\n");
+        }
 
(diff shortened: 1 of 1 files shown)


bladecoding:
No since binaries would be referencing the `public static void WriteLine(string, object[])` signature which still exists. You just wouldn't get the fix until you rebuilt your binary.

I did just double check by building a test server resource which references the old citizenfx.core.dll and then using the resource with a server that has the new citizenfx.core.dll.

PR #121, opened 2018-06-17
Fixed drivingStyle`s type in Tasks.DriveTo method

Harry09:


pongo1231:
`DrivingStyle` doesn't cover all the possible driving styles, which is why it's `int`.

Harry09:
So, why `DrivingStyle` still don't have all possible styles?

pongo1231:
Because there's a huge list of possible driving styles?

See https://www.vespura.com/drivingstyle/

Harry09:
Okay, I understood.

PR #98, opened 2018-02-12
Changed russian word "Ошибка" to english "Error".

Eldarien:

PR #124, opened 2018-07-01
Addition of "Error" in brackets next to "Ошибка"

Eric221:
As mentioned in #98, you don't want to remove the "Ошибка". And that's OK. But on the other hand it decreases the usability of this statement. So I thought that a good way will be not touching the word "Ошибка" and leaving it as it is, and adding next to it an English "Error" in brackets.

Case 1: Someone understands the meaning of "Ошибка", so he reads it and moves his eyes below, to the description. Without paying attention to what is in brackets.

Case 2: Someone doesn't understand the meaning of "Ошибка", so he reads it and thinks "What's this?? Idk. Maybe it's a mistake in translation. Anyway, let's continue reading.", then he moves his eyes to the next word, which is in brackets – "Error". He understands it, so he moves his eyes below, to the description.

I think this is a good solution for the problem "nostalgia vs usability".

### DIFF
--- a/ext/cfx-ui/src/assets/locale-en.json
+++ b/ext/cfx-ui/src/assets/locale-en.json
@@ -18,7 +18,7 @@
     "#Servers_Connecting": "Connecting",
     "#Servers_ConnectingTo": "Connecting to {{ serverName }}...",
     "#Servers_ConnectFailed": "Connection failed",
-    "#Servers_Error": "Ошибка",
+    "#Servers_Error": "Ошибка (Error)",
     "#Servers_CloseOverlay": "Close",
     "#Servers_CancelOverlay": "Cancel",
     "#Servers_Message": "{{ message }}",


Eric221:
Some time ago I noticed that you use keywords like "nui", "net", "data", etc. in commit descriptions. I don't know the usage of them is required or not, there's no information about this in `docs` dir. If it is required, please, let me know.

vecchiotom:
seems pretty obvious to me that the reason why this isn't merged yetisnt because of commit names, but because the elements feel like this is unnecessary...

Eric221:
Thank you for accepting. :smiley:

@vecchiotom I am new here, as you can see, and don't know how development works here. What I have done in this PR is just a suggestion. If someone likes what I've done here, he can just merge it, If not, can just close this PR. I think that's how it works.

PR #125, opened 2018-07-12
swapped patterns with non-zero offsets for (mostly shorter ones)

sfinktah:
new patterns all start from the beginning function, and are generally shorter with no increase in wildcards to existing patterns.

patterns appear to match multiple releases, and should be as stable as existing, with less annoying math.

i've only added them as comments, so you can pick and choose.

_n.b., all my work and testing has been performed in the single-player (offline) and is not intended for online play. (Gena disclaimer)_

### DIFF
--- a/code/components/gta-core-five/src/EntityDeletionHacks.cpp
+++ b/code/components/gta-core-five/src/EntityDeletionHacks.cpp
@@ -76,31 +76,37 @@ static T* getTypedEntity(int guid)
 	return static_cast<T*>(entity);
 }
 
+// getScriptEntity_fivem "83 f9 ff 74 31 4c"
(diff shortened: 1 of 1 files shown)


blattersturm:
Comments and therefore the PR not changing any code at all? Perhaps not too suited as a pull request.

blattersturm:
closed since no comment

sfinktah:
Not really prepared to commit actual changes to something I can't build.

Speaking of which - if you could supply the submodule information for whatever you have checked out (and apparently compiling)...   or temporarily commit the submodules as regular folders somewhere, or just make a .zip of a working checkout?

Well, lets just say b1493 wasn't exactly a difficult update.

PR #126, opened 2018-07-12
Suggestion for alternative hook, for easier pool management

sfinktah:
This is the function I hook to increase pool sizes, and it has always worked flawlessly. 

It seems a lot simpler than the multiple patterns you are using at the moment, though I could be missing something.

I've added it as a separate .cpp file, for your perusal, and also included a handy little pointer-match library which I much prefer 

```cpp
    var = mem::pointer(location).rip(4).as<uint32_t>();
```

to 

```cpp
    var = (uint32_t*)(*(int32_t*)location + location + 4);
```

and other things like:

```cpp
    mem::pointer(location).add(3).rip(4).as<uint32_t&> = var;
```

your `get_call()` and `set_call()` functions could be expressed by new methods:

```cpp
    inline pointer pointer::call() const { 
        return offset(1).rip(4); 
    }

    inline pointer pointer::call(const uintptr_t func) {
        as<uint8_t&>() = 0xe8;
        offset(1).as<int&>() = func - value_ - 5;
    }
```

not that there is any reason to switch from your existing `get_call()` etc, I only use them as an example.

_n.b., all my work and testing has been performed in the single-player (offline) and is not intended for online play. (Gena disclaimer)_

### DIFF
--- a/code/components/gta-core-five/src/PoolManagementAlternativeHook.cpp
+++ b/code/components/gta-core-five/src/PoolManagementAlternativeHook.cpp
@@ -1,0 +1,39 @@
+LPVOID allocatePoolMemory_ori;
+
+// #include "memmath.h"
+// Hook(Scan("48 83 ec 28 83 3d ?? ?? ?? ?? 00 75 23")
+//         .add(0x20)
+//         .rip(4)
+//         .as<void*>(), 
+//     allocatePoolMemory, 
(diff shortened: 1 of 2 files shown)


blattersturm:
This is not really suited as a pull request - perhaps a forum issue report, but this isn't actually a working transition to a different hook system :)

Neat idea, however.

PR #127, opened 2018-07-15
An ability to override isPlayerTalking value

Disquse:
This function can be very useful for custom voip systems. It affects on other default voip natives and also plays mp_facial animation without affecting on other anims.
`void OVERRIDE_PLAYER_TALKING(Player player, BOOL state)`

Demonstration: https://gfycat.com/EmbellishedForkedAdmiralbutterfly

### DIFF
--- a/code/components/gta-net-five/src/MumbleVoice.cpp
+++ b/code/components/gta-net-five/src/MumbleVoice.cpp
@@ -331,6 +331,9 @@ static void Mumble_RunFrame()
 	}
 }
 
+static std::bitset<256> g_talkers;
+static std::bitset<256> o_talkers;
+
 static InitFunction initFunction([]()
 {
 	g_mumbleClient = CreateMumbleClient();
@@ -346,6 +349,7 @@ static InitFunction initFunction([]()
 		g_mumbleClient->SetAudioDistance(FLT_MAX);
 
 		Mumble_Disconnect();
+		o_talkers.reset();
 	});
 });
 
@@ -357,8 +361,6 @@ static bool _isAnyoneTalking(void* mgr)
 }
 
 static bool(*g_origIsPlayerTalking)(void*, void*);
-
-static std::bitset<256> g_talkers;
 
 static bool _isPlayerTalking(void* mgr, char* playerData)
 {
@@ -382,7 +384,7 @@ static bool _isPlayerTalking(void* mgr, char* playerData)
 			// actually: netobj owner
 			auto index = netObj[73];
 
-			if (g_talkers.test(index))
+			if (g_talkers.test(index) || o_talkers.test(index))
 			{
 				return true;
 			}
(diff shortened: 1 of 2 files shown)


vecchiotom:
**EmbellishedForkedAdmiralbutterfly** nice link lol

code/components/gta-net-five/src/MumbleVoice.cpp
```diff
@@ -510,6 +512,29 @@ static HookFunction hookFunction([]()
...
 		});
 
+		fx::ScriptEngine::RegisterNativeHandler("OVERRIDE_PLAYER_TALKING", [](fx::ScriptContext& context)
+		{
+			auto isPlayerActive = fx::ScriptEngine::GetNativeHandler(0xB8DFD30D6973E135);
+
+			int playerId = context.GetArgument<int>(0);
+			int state = context.GetArgument<bool>(1);
+
+			if (playerId <= o_talkers.size() && playerId >= 0)
```
blattersturm: `<`, not `<=`?

ext/natives/codegen_cfx_natives.lua
```diff
@@ -1731,6 +1731,14 @@ native 'SET_VEHICLE_WHEEL_X_OFFSET'
 	</summary>
 ]]
 
+native 'OVERRIDE_PLAYER_TALKING'
```
blattersturm: Please add a doc block with a description of the native and the arguments, similarly that it doesn't need to be called every frame.

Of course, if changing the native's name, also change the definition here. 🙂

code/components/gta-net-five/src/MumbleVoice.cpp
```diff
@@ -510,6 +512,29 @@ static HookFunction hookFunction([]()
 			}
 		});
 
+		fx::ScriptEngine::RegisterNativeHandler("OVERRIDE_PLAYER_TALKING", [](fx::ScriptContext& context)
```
blattersturm: Since it's setting a persistent state, `SET_PLAYER_OVERRIDE_TALKING` (or `SET_PLAYER_OVERRIDE_TALK_STATE`, or maybe `SET_PLAYER_TALKING_STATE`, or similar) seems more correct a name.

I'd stick with `SET_PLAYER_TALKING_STATE`, unless R* used `OVERRIDE_` prefixes before.

Disquse: R* uses `OVERRIDE_` prefixes sometimes. So, I will use `SET_PLAYER_OVERRIDE_TALKING`, as you said.

ext/natives/codegen_cfx_natives.lua
```diff
@@ -1731,13 +1731,21 @@ native 'SET_VEHICLE_WHEEL_X_OFFSET'
...
-native 'OVERRIDE_PLAYER_TALKING'
+native 'SET_PLAYER_TALKING_OVERRIDE'
 	arguments {
 		Player 'player',
 		BOOL 'state'
 	}
 	apiset 'client'
 	returns 'void'
+	doc [[
+	<param name="player">The target player.</param>
```
blattersturm: I'm not too sure how much the system will like `<param>` below the summary. :/

Disquse: Native gen is always output summary before params, but I will fix.

PR #129, opened 2018-07-17
unknown -  updated build notes

sfinktah:
has anybody actually tried building this project from scratch lately?

even the one of the CI test builds is broken.

### DIFF
--- a/docs/building.md
+++ b/docs/building.md
@@ -18,4 +18,36 @@ cd code
 premake5 vs2017 --game=five
 ```
 
+## Notes by sfinktah
+```bash
+cd vendor/xz
+cp windows/vs2017/config.h src/common/
+
+cd ../../vendor/fmtlib
+git checkout 4.1.0
+
+cd ../../vendor/protobuf
+git checkout v3.0.2
+
+cd ../../vendor/tbb
+git checkout tbb_2018
+
+# in a developer command window
+cd tbb\build\vs2013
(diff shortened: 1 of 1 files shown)


blattersturm:
The project builds fine as is, your build notes make no sense either.

sfinktah:
"The project builds find as is."

Spoken like someone who hasn't typed

    git submodule update --init --recursive --remote

in a very long time.

blattersturm:
@sfinktah I'm not too sure where the documentation [states to use the --remote flag](https://github.com/citizenfx/fivem/blame/master/docs/building.md#L15), since it does not.

```
--remote
This option is only valid for the update command. Instead of using the superproject’s recorded SHA-1 to update the submodule, use the status of the submodule’s remote-tracking branch. 
```
This seems actually _undesirable_ for building FiveM, and would explain why you had to check out specific submodule versions.

sfinktah:
In retrospect that should have been fairly obvious to me.

Do you perchance have the .pdb files for the pre-compiled v8.dll and v8_platform.dll that come with the binary installation?  

BTW, there are drop-in replacement v8 binaries for 6.8.2 available via nuget, see https://github.com/pmed/v8-nuget  -  though unfortunately they do not have .pdb files either.

sfinktah:
Okay, with a bit of poking and prodding, I have it down to one error.

    code\client\clrcore\Native.cs(9,33,9,37): error CS0246: The type or namespace name 'Hash' could not be found (are you missing a using directive or an assembly reference?)

PR #134, opened 2018-08-23
citizen:scripting:mono: use new profiler API

Hellslicer:
Small PR to fix linux build.

MonoGCEvent was refactored to MonoProfilerGCEvent since mono-5.8.0.103 (https://github.com/mono/mono/commit/ea4e4a9ef6fc42570a23026adbe826cf7248290e)


blattersturm:
Nice, before I wondered 'does this not break old Windows Mono' you already took that into account. 😛 

What's weird is how internally Linux builds still go fine, I'm hoping we didn't end up with odd dependency pinning and using old versions of Mono 😕

PR #149, opened 2018-10-08
v8: fix console failing on undefined values

vecchiotom:
this didnt seem to be fixed with commit 15fd5d3eeaf687544ed5c64e1dc721bab6c04914 which was fixed only null values, leaving undefined values unhandled.


vecchiotom:
i removed `typeof` because it's slower than directly comparing to `undefined`.

PR #150, opened 2018-10-15
Allow OneSync in debug builds

4mmonium:
Allow OneSync in debug builds (requires oneSyncDebugAllowed=1 in the CitizenFX.ini file)


blattersturm:
This'll likely be implemented by checking if `adhesive` exists and setting a compile-time define if not.

Thanks for the effort, however!

PR #152, opened 2018-10-27
Fix for server-client callbacks not working

IllusiveTea:
Should (according to @vecchiotom as he made this and told me to do it for him) fix the issues with callbacks not working correctly


blattersturm:
This will revert to the broken behavior when it comes to reference counting, so won't work at all as a viable 'fix' for remote function references.

PR #154, opened 2018-10-30
clrcore: Change GetDistance method from native call to Vector3 euclidean distance method

MoosheTV:
- Reduces the amount of native calls
- Higher Accuracy:
```C#
var pointA = new Vector3(0f, 0f, 0f);
var pointB = new Vector3(1000f, 1000f, 1000f);


GET_DISTANCE_BETWEEN_COORDS(pointA.X, pointA.Y, pointA.Z, pointB.X, pointB.Y, pointB.Z, 1):
1732.051

Math.Sqrt(pointB.DistanceToSquared(pointA)):
1732.05080756888
```

### DIFF
--- a/code/client/clrcore/External/World.cs
+++ b/code/client/clrcore/External/World.cs
@@ -535,7 +535,7 @@ namespace CitizenFX.Core
 		/// <returns>The distance</returns>
 		public static float GetDistance(Vector3 origin, Vector3 destination)
 		{
-			return Function.Call<float>(Hash.GET_DISTANCE_BETWEEN_COORDS, origin.X, origin.Y, origin.Z, destination.X, destination.Y, destination.Z, 1);
+			return (float) Math.Sqrt(destination.DistanceToSquared(origin));
 		}
 		/// <summary>
 		/// Calculates the travel distance using roads and paths between 2 positions.


neos7:
I think a Native should stay a native, and there is already this
https://github.com/citizenfx/fivem/blob/5bad342ab5dcb8d2f5d10b8823913e8c71802619/code/client/clrcore/Math/Vector3.cs#L615

MoosheTV:
> I think a Native should stay a native, and there is already this
> [fivem/code/client/clrcore/Math/Vector3.cs](https://github.com/citizenfx/fivem/blob/5bad342ab5dcb8d2f5d10b8823913e8c71802619/code/client/clrcore/Math/Vector3.cs#L615)
> 
> Line 615 in [5bad342](/citizenfx/fivem/commit/5bad342ab5dcb8d2f5d10b8823913e8c71802619)
> 
>  public static void Distance(ref Vector3 value1, ref Vector3 value2, out float result)

There is no added benefit for it to remain using this native, and native calls do more to affect performance than calculating the euclidean distance using quick math operations. This is to provide better performance and a more accurate output to any resource making use of this function.

vecchiotom:
there's an even quicker native though, [Vdist2](http://runtime.fivem.net/doc/reference.html#_0xB7A628320EFF8E47), i agree with @neos7, this has no reason to be changed.

d0p3t:
> no reason to be made a non native

Yes there is a reason, namely performance.

A user doesn't care how the result of a method is achieved. The user does however want the fastest operation possible with the same result. There's always an overhead when calling GTA natives, thus creating an unnecessary performance decrease.

This new implementation results in an increase of performance with the same result. Heck, this result is even more precise than the original.

MoosheTV:
> there's an even quicker native than GetDistanceBetweenCoords though, [Vdist2](http://runtime.fivem.net/doc/reference.html#_0xB7A628320EFF8E47), i agree with @neos7, this has no reason to be made a non native. Instead maybe you should use this other native 🤔

No native call can compete against direct arithmetic operations such as multiplication and division. Any native used will cause overhead. The reasons why you would want to use this over a native call would be because it's both faster and more precise.

The only argument I've seen so far for keeping it as a native is for vanity.

PR #156, opened 2018-11-02
Add possibility to clear a timeout in lua

Matspyder51:


derpy54320:
There's a little problem you might have overlooked though. The threads run using a for loop like this: `i = #threads, 1, -1`. So, if you remove a timeout that hasn't run in the current tick, all the threads will move down (because of the `table.remove` call) and result in the current thread that removed the timeout to run again, since it'll be `threads[i]` after a timeout was removed and `i` was decremented. We'd need to insure that if a thread under the current thread is removed, `i` goes down.

An easier solution could be to make the thread in `threads` false (NOT `nil` because that'll break `#threads`) and check `thread` in the tick routine, and remove it from the threads table there (because at that point, removing it will not affect `threads[i]` and the next thread will run properly, and none will be ran twice or skipped).

derpy54320:
Hey, I added thread termination to my current pull request. Check it out, I also used ClearTimeout as an alias so that'll work like you wanted!

PR #171, opened 2019-01-08
citizen:scripting:v8 add generator function to supported tickers

adam-jw:
The preface of this PR is to add support for yielding execution back to the timer without returning from your tick handler.

For example:
```javascript
let modelRequested = false;
setTick(() => {
  if (!modelRequested) {
    RequestModel('...');
    modelRequested = true;
    return;
  } else if (!HasModelLoaded('...')) {
    return;
  }

  ...
});
```

Can be simplified to:
```javascript
setTick(function* () {
  ...

  RequestModel('...')
  while (!HasModelLoaded('...')) {
    yield;
  }

  ...
});
```

### DIFF
--- a/data/shared/citizen/scripting/v8/timer.js
+++ b/data/shared/citizen/scripting/v8/timer.js
@@ -8,6 +8,8 @@
     const tickers = [];
     let animationFrames = [];
 
+    const generatorConstructor = (function*() {}).constructor;
+
     function  setTimer(timer, callback, interval) {
         timers[timer.id] = {
             callback,
@@ -26,7 +28,7 @@
 
     function clearTimer(timer) {
         if (!timer) { return; }
-    
+
         delete timers[timer.id];
     }
 
@@ -91,10 +93,22 @@
             i = tickers.length;
 
             while (i--) {
+                const ticker = tickers[i];
+
                 try {
-                    tickers[i]();
-                } catch(e) {
-                    console.error('Unhandled error: ' + e.toString() + '\n' + e.stack);
+                    if (ticker instanceof generatorConstructor) {
+                        const { next } = ticker();
+                        const { done } = next();
+
+                        // remove ticker if generator is done
+                        if (done === true) {
+                            tickers.splice(i, 1);
+                        }
+                    } else {
+                        ticker();
+                    }
+                } catch (error) {
+                    console.error('Unhandled error: ' + error.toString() + '\n' + error.stack);
                 }
             }
         }
@@ -116,7 +130,7 @@
         }
 
         gameTime = localGameTime;
-				
+
         //Manually fire the callbacks that were enqueued by process.nextTick.
         //Since we override setImmediate/etc, this doesn't happen automatically.
         if (global.process && typeof global.process._tickCallback === "function")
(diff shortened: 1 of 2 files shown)


throwarray:
I'd name it something else and make it a routine instead. A ticker is just while (true) { yield Wait(0) }
```
CreateThread(function* my_tick () {
    yield Wait(1000)
 
    const res = yield myAsyncTask()
 
    console.log(res)
   
    try {
        yield myAsyncTask(true) // catches errors
    } catch(e) {
        console.log(e)
       
        // throw new Error('throws errors')
    }
   
  return 'return results'
}).then(function (res) {
  console.log('Success with res', res)
}).catch(function (err) {
  console.log('Error with err', err)   
})
```
https://pastebin.com/gjsNAcsb

adam-jw:
@throwarray:
I would like to improve the API but would like to see the maintainers thoughts about this change first.

throwarray:
If you don't need to match an internal tick rate this is basically identical to async/await which requires no changes to the runtime.  With generators you could yield anything though and save on allocs now you must yield a promise

thers:
Ah, right, didn't notice `!` :) Anyway, generators aren't easy things and i honestly think that async/await is better, as can be used w/o modifying cfx:v8

adam-jw:
> How's that?

You're invoking a Promise - it won't resolve until the next tick (unless FiveM has a custom implementation here)

data/shared/citizen/scripting/v8/timer.js
```diff
@@ -91,10 +93,22 @@
             i = tickers.length;
 
             while (i--) {
+                const ticker = tickers[i];
+
                 try {
-                    tickers[i]();
-                } catch(e) {
-                    console.error('Unhandled error: ' + e.toString() + '\n' + e.stack);
+                    if (ticker instanceof generatorConstructor) {
```
thers: I'd rather suggest to have different array for generator tickers, i.e. `setTick` would distinguish generators from other functions and put them in different array. Of course that would require another `while` loop down there, but advantage is that we can bail out of generators execution early. We're already doing it like so for animation frames.

blattersturm:
> You're invoking a Promise - it won't resolve until the next tick

Wouldn't this be fixable by both:
1. In the typical setInterval/resolve pattern, doing a check before setting an interval handler.
2. Figuring out where exactly V8 starts executing the job/microtask queue, and triggering this at the end of V8ScriptRuntime::Tick?

thers:
>Wouldn't this be fixable

It would, but i also now see how using generators can be useful, like, you can write down sequence of some initialization things including blocking and non-blocking waits for resources to load and then just finish this sequence, and with current `setTick` you can't easily bail out of it, like, you'd have to return from it if some condition met.

Considering all above, i'd also suggest to not putting generators support into the `setTick` itself but introduce new API, something like `createThread(function *() {})`, even though it is clearly not a "real" thread, it's similar to API from Lua runtime. And yeah, it'd require adding some sort of `wait(milliseconds)`.

thers:
Also, from a discussion in discord: it is also makes sense to add special global constant which you can return from ticker to stop that particular ticker, e.g. `const stopTick = Symbol('stopTick')`. This will bring missing functionality, also will not introduce any new APIs, as we already have a plenty in js runtime.

PR #172, opened 2019-01-09
yojimbo: fix compilation on linux

Hellslicer:


code/components/citizen-server-impl/src/GameServerNet.Yojimbo.cpp
```diff
@@ -1,4 +1,5 @@
 #include <StdInc.h>
+#include <sys/select.h>
```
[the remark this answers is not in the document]
Hellslicer: :heavy_check_mark:

code/components/citizen-server-impl/src/GameServerNet.Yojimbo.cpp
```diff
@@ -1,4 +1,7 @@
 #include <StdInc.h>
+#ifndef _MSC_VER
```
blattersturm: 😕  _WIN32, really.

code/components/net-base/include/NetBase.h
```diff
@@ -21,13 +19,15 @@ typedef SOCKET PlatformSocketType;
 #include <sys/socket.h>
 #include <netinet/in.h>
 
-typedef int PlatformSocketType;
+typedef int SOCKET;
```
blattersturm: Ah. It'd make more sense to replace the `(SOCKET)` with `(PlatformSocketType)` in the yojimbo file then, instead of making this always define `SOCKET`.

code/components/citizen-server-impl/src/GameServerNet.Yojimbo.cpp
Hellslicer: Ok :heavy_check_mark:

code/components/net-base/include/NetBase.h
Hellslicer: :+1:

PR #173, opened 2019-01-11
citizen:scripting:v8 add clearTick and waiting for ticker to resolve

thers:
As a follow up to https://github.com/citizenfx/fivem/pull/171#issuecomment-453018520

This PR introduces:
1. New global function `clearTick` that you can use to halt your ticker execution, accept `tickerId` return from `setTick`.
2. Runtime will now wait for ticker to resolve (if it returns a promise, e.g. is async function) before firing it again.

The latter, *possibly*, can break backwards compatibility, but doubtfully that anyonehas been relying on old behavior, thus I believe we can safely assume previous behavior as a bug and just fix it.

```js
setTick(async () => {
  // Previously this whole ticker would fire on each tick and spawn new promise each time
  // In this PR runtime won't call this ticker again until it got resolved
  await longTakingOp();
});
```

### DIFF
--- a/data/shared/citizen/scripting/v8/timer.js
+++ b/data/shared/citizen/scripting/v8/timer.js
@@ -1,14 +1,21 @@
 // Timers
 
-(function (global) {
+(function(global) {
     let gameTime = Citizen.getTickCount();
 
     const timers = {};
     let timerId = 0;
-    const tickers = [];
+
+    const tickers = {};
+    let tickerId = 0;
+
     let animationFrames = [];
 
-    function  setTimer(timer, callback, interval) {
+    function printError(where, e) {
+        console.error(`Unhandled error in ${where}: ${e.toString()}\n${e.stack}`);
+    }
+
+    function setTimer(timer, callback, interval) {
         timers[timer.id] = {
             callback,
             interval,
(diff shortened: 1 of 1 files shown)


data/shared/citizen/scripting/v8/timer.js
```diff
@@ -72,30 +112,44 @@
...
+            let result;
+            try {
+                result = ticker.callback();
+            } catch (e) {
+                printTickerError(e);
+                continue;
+            }
+
+            // We've got a promise!
+            if (result.then) {
```
adam-jw: Using `typeof result.then === 'function'` is going to be way more reliable at determining whether the result is a Promise or not.

data/shared/citizen/scripting/v8/timer.js
```diff
@@ -1,62 +1,102 @@
...
+        if (!ticker) {
+            return;
+        }
+
+        delete tickers[ticker];
     }
 
-    function setTick(callback) {
-        tickers[tickers.length] = callback;
+    function resolveTicker(ticker) {
```
adam-jw: On L151, you're binding `tickerId` [to the context] - not as an argument...

You probably need to replace `ticker` (`tickers[ticker]`) with `this`

thers: You're right, thanks for pointing this one out!

data/shared/citizen/scripting/v8/timer.js
adam-jw: It probably wouldn't be a bad idea to include this in the `try`/`catch` statement above, or it's own `try`/`catch` statement to avoid dying if you actually haven't got a promise.

PR #174, opened 2019-01-14
citizen:scripting:v8 fix binding of ticker resolver argument

thers:


blattersturm:
thanks, the ridersx, for the broken PR 🌮

PR #178, opened 2019-01-24
vendor: add linux compat to botan

Hellslicer:


blattersturm:
not a good way to do this, also as the botan used for linux (botan_all.cpp) is an entirely different version of botan

these files are meant to be unchanged from a botan configure.py output

PR #183, opened 2019-02-05
Stripped Unused Defines

TYKUHN2:
Stripped unused defines from Launcher.cpp
Seriously... these games aren't supported.


TYKUHN2:
Further note is that I spotted a file (and I bet there are more) that rely on `WIN32` macros. These are not set in any compiler I know of (or at least use) which is something that might want to be looked at.

That being said, I didn't check to see if it was manually defined.

blattersturm:
Not a very useful change. These games might be supported in the future.

PR #192, opened 2019-03-13
scheduler.lua: Add debug info for slow server resource threads

minipunch:
This change to ``scheduler.lua`` helps server owners (or script developers) track down and find the source of hitch frame warnings that are caused by 'slow' code in server sided scripts.

These general idea of these changes were first done a while back by @joelwurtz but it was never put into production. I since modified the scheduler myself to try it out and it has helped me track down which resources and in what events were causing certain hitch frame warnings I was consistently getting but was unsure of where they were coming from exactly. So this change is basically credited to @joelwurtz and I just made some minor changes to his version of the change like adding the event name that triggered the warning.

### DIFF
--- a/data/shared/citizen/scripting/lua/scheduler.lua
+++ b/data/shared/citizen/scripting/lua/scheduler.lua
@@ -37,7 +37,7 @@ function Citizen.CreateThreadNow(threadFunction, eventName)
 	local result, err = coroutine.resume(coro)
 	local runTime = GetGameTimer() - startAt
 
-	if runTime > 110 then
+	if runTime > 150 then
 		print("Executing thread in resource " .. t.resource .. " was slow: " .. runTime .. "ms, please verify what you are doing")
 		if eventName then
 			print("in event: " .. eventName)
@@ -172,7 +172,7 @@ Citizen.SetTickRoutine(function()
 				local result, err = coroutine.resume(thread.coroutine)
 				local runTime = GetGameTimer() - startTime
 
-				if runTime > 110 then
+				if runTime > 150 then
 					print("Executing thread in resource " .. thread.resource .. " was slow: " .. runTime .. "ms, please verify what you are doing")
 				end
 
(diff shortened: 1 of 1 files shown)


minipunch:
> Am getting this issue when running with vRP.
> `Error running system tick function for resource vrp: citizen:/scripting/lua/scheduler.lua:176: attempt to concatenate a nil value (field 'resource') stack traceback: citizen:/scripting/lua/scheduler.lua:176: in function <citizen:/scripting/lua/scheduler.lua:156>`
> 
> Not sure whats causing thread.resource to be nil, seems to report hitches with other resources fine.

@rubbertoe98 It was probably because it wasn't getting the resource name when vRP was using ``SetTimeout()`` because ``GetResourceName()`` wasn't being called for that function previously. I have just made an update to it and it should fix that nil error, let me know if it works better now or not.

PichotM:
Tested and working. Pretty useful I have made some modification to make it better :)
Firstly, I believe it is better to use Citizen.Trace than print.
Secondly, you should add a red color (^1) for each prints 👍

vecchiotom:
remember to add ^7 at the end, so that the color will be reset, otherwise you'll mess up the future output too

PichotM:
What I did ->
`Citizen.Trace(string.format('^1[%s] Executing thread in resource %s was slow: %sms, please verify what you are doing.^7\n', eventName or "Unknown", t.resource, runTime))`

blattersturm:
Closing this PR since it will not be merged. A resource profiler should be implemented as a global API in the host itself, not separately in each HLL.

In addition, this PR does a few slightly.. odd things:

- Changing a lot of blank lines. Why? You should not use editor settings that change parts of files that aren't even touched, especially when making a PR. Please use `git diff` to check before PRing.
- Tagging coroutines with 'event names'. This is wrong on many levels - a normal scope-based profiler would rather work by wrapping the coroutine in a closure and adding a scope around it (with some safety measures for cases where a script induces a Lua error) or something of the likes.
- There's no toggle. Profiling functionality should not require people to run above a threshold to end up spamming the (oftentimes unknowing) user, but should be available on-demand and be added on top of existing warning notices (e.g. if implemented in the host, the usual server-side 'hitch warning' could contain the longest 'exclusive' scopes).

PR #196, opened 2019-03-30
Code cleanup, clearing some warnings and fixing Linux build

t11230:
- Removed some unused code from ComponentLoader.cpp
- Converted some c-style code into more c++-like code
- Constified some functions in ComponentLoader
- Added a couple utility functions to fwPlatformString and marked fwPlatformString::ConvertString as static

### DIFF
--- a/code/client/citicore/ComponentLoader.cpp
+++ b/code/client/citicore/ComponentLoader.cpp
@@ -48,7 +48,6 @@ void ComponentLoader::Initialize()
 	}
 
 	FILE* componentCache = _pfopen(MakeRelativeCitPath(componentsName).c_str(), _P("rb"));
-
 	if (!componentCache)
 	{
 		FatalError("Could not find component cache storage file (components.json).");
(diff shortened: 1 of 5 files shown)


blattersturm:
Hello! Thanks for your contribution.

We appreciate the effort put into this, and will be looking to merge this PR in the coming few days.

Please, as a note, try to discuss any 'refactoring' PRs beforehand in the future - these changes are all relatively harmless, but could very well not have been.

Also, try to not overload PRs for two different purposes in the future - it's seen that you added a second commit in this case that (apparently) does not depend on any new functions you introduced in the other commit; this one could very well have been a separate PR to 'solely' fix whatever compilation error you encountered.

t11230:
Sounds good to me. Where is a good place to discuss future refactorings or code changes?

Also, one thing that I wasn't really sure about when working on this PR was testing a dev build on my computer. I ended up performing the same steps the CI build uses when generating a release archive. This worked and I verified that nothing broke as a result of this PR. Getting those steps figured out took a little digging though. Is there a better way to test out my local changes? If not, I wouldn't mind creating another PR to generate some documentation or scripts to make the process more streamlined.

Thanks!

PR #197, opened 2019-03-31
Fix event handling.

derpy54320:
This little patch aims to fix quite a bit of event handling w/ the scheduler and a tiny bit with CreateThreadNow.

I'm not sure if I should be uploading this somewhere different, but if you want to see a test script for these proposed changes, I put it on pastebin. https://pastebin.com/wtAH0uLq

1. Make event handlers run in order (they didn't before).
2. Made a function to let you change the order they run in, SetEventReversed. They are reversed by default.
3. Made GetCurrentEvent which simply gives back the name of the currently running event.
4. Made RemoveCurrentEventHandler which is also pretty self explanatory.
5. Removing an event handler will not screw up the event handling order (if you removed a handler for the current event before, it would).
6. For neatness sake, event handler tables are now created in one place, getEventHandler.
7. Made runEventHandler simply for the sake of creating an event handler thread w/ less upvalues.
8. Made CreateThreadNow an alias for Citizen.CreateThreadNow for consistency w/ other common functions.
9. Made CreateThreadNow not cause an error, but rather just print the error as to not interrupt whatever created the thread.

### DIFF
--- a/data/shared/citizen/scripting/lua/scheduler.lua
+++ b/data/shared/citizen/scripting/lua/scheduler.lua
@@ -38,15 +38,17 @@ function Citizen.CreateThreadNow(threadFunction)
 	curThread = oldThread
 
 	if err then
-		error('Failed to execute thread: ' .. debug.traceback(coro, err))
-	end
-
-	if resumedThread and coroutine.status(coro) ~= 'dead' then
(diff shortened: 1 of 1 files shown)


derpy54320:
I closed this since I made an update w/ another well-needed feature, thread termination. New pull request instead!

PR #198, opened 2019-03-31
Fix event handling, allow termination of threads.

derpy54320:
This little patch aims to fix quite a bit of event handling w/ the scheduler, a tiny bit with CreateThreadNow, and allows you to terminate threads.

1. Make event handlers run in order (they didn't before).
2. Made a function to let you change the order they run in, SetEventReversed. They are reversed by default.
3. Made GetCurrentEvent which simply gives back the name of the currently running event.
4. Made RemoveCurrentEventHandler which is also pretty self explanatory.
5. Removing an event handler will not screw up the event handling order (if you removed a handler for the current event before, it would).
6. For neatness sake, event handler tables are now created in one place, getEventHandler.
7. Made runEventHandler simply for the sake of creating an event handler thread w/ less upvalues.
8. Made CreateThreadNow an alias for Citizen.CreateThreadNow for consistency w/ other common functions.
9. Made CreateThreadNow not cause an error, but rather just print the error as to not interrupt whatever created the thread.
10. Made TerminateThread (and ClearTimeout), which will mark a thread for removal.

I'm not sure if I should be uploading this test somewhere different, but if you want to see a test script for these proposed changes, I put it on pastebin. https://pastebin.com/wtAH0uLq

### DIFF
--- a/data/shared/citizen/scripting/lua/scheduler.lua
+++ b/data/shared/citizen/scripting/lua/scheduler.lua
@@ -3,10 +3,12 @@ local curThread
 local curThreadIndex
 
 function Citizen.CreateThread(threadFunction)
-	table.insert(threads, {
+	local thread = {
 		coroutine = coroutine.create(threadFunction),
 		wakeTime = 0
-	})
+	}
+	table.insert(threads,thread)
+	return thread
 end
 
 function Citizen.Wait(msec)
@@ -38,15 +40,16 @@ function Citizen.CreateThreadNow(threadFunction)
 	curThread = oldThread
 
 	if err then
-		error('Failed to execute thread: ' .. debug.traceback(coro, err))
-	end
-
-	if resumedThread and coroutine.status(coro) ~= 'dead' then
+		-- only print, rather than cause an error stopping the thread creator.
+		Citizen.Trace('Failed to execute thread: ' .. debug.traceback(coro, err))
+	elseif resumedThread and coroutine.status(coro) ~= 'dead' then
 		table.insert(threads, t)
+		return t
 	end
-
-	return coroutine.status(coro) ~= 'dead'
 end
+
+-- for consistency with CreateThread
+CreateThreadNow = Citizen.CreateThreadNow
 
 local inNext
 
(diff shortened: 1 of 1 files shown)


blattersturm:
We appreciate the effort you've put into these changes, but we'll quite possibly not be accepting this PR.

It adds a lot of API surface to HLL script runtime components that will need to be retained when we end up refactoring the entire event handling logic to be handled in C++, and currently will in fact only provide an advantage in one ScRT, and not all three of them.

The rationale for rewriting event handlers is simple: currently each ScRT is responsible for its own bit of events as they're 'broadcast' across all runtimes, which leads to performance issues (each event/tick/... has HLL code run on even 'uninterested' ScRTs, which in typical use cases - for some reason - means 200+ ScRTs get events they don't care about) and the inability to guarantee ordering _beyond_ a single ScRT (if two resources want to run a specific event, but resource 2 has to run before resource 1, they still can't no matter how much you change the HLL ScRT). We were also intending to completely rework the concept of 'basic' exposed `CreateThread`-style coroutines for FXv2 mode, since as it is now people seem to have quite a number of issues with these, and there's little to no consistency across ScRT implementations when it comes to this.

There have been a lot of refactoring changes being PR'd lately, each and every one of them developed without any prior discussion on whether their PR is even worthwhile or how to _correctly_ implement features like this.

Please, if you're trying to develop a large-scale refactor to an existing project, don't just 'work on it' and PR it, please start a discussion thread on the forums or discuss it in the chat room first - you'll usually get a lot more useful insights and the results end up better for both parties.

derpy54320:
I hate to ask but, where would the proper place to discuss this kind of thing first be? I don't see a relevant area on forums.fivem.net for developing fivem itself. I don't really work on the project often but I love to help when there's something I have the know-how to do!

PichotM:
Discord maybe?

derpy54320:
Will do but, just to leave my ideas on the topic here as well for anyone coming across this to see. Here's how I feel about everything brought up and the reasoning behind all my proposed changes. I understand you're going to be working on a new re-work, but if you're going to be adding order soon anyways, why not start now?

*1. I don't see why having a little more order than not matters.*
I understand different resources will still each handle their own events in a seemingly random order, but I still think it's important that if you put a few event handlers in _the same resource_, you know one thing happens before another. Without this, you invite the possibility of hard-to-reproduce glitches with initialization. I still think at least having order in each resource is important even if not all resources are ordered.

*2. I made sure an old-bug won't resurface.*
If memory serves right, I remember events used to be ordered, but now I see that (before this PR) a global eventKey was used for adding entries to a table. I think this was done so that removal of handlers wouldn't interfere with the handling for loop, but I've taken precautions to insure every handler runs once per event trigger. If a handler is removed while an event is being handled, the index for the handling loop will change accordingly if needed.

*3. Thread termination.*
I don't think you really refuted the thread termination thing, but I did see another PR for it. It was old but I noticed it failed to account for the thread for loop's index changing. I made sure that this won't be the case, as threads are only _marked_ for termination when TerminateThread is called. Meaning, they aren't actually removed until they're found in that loop, so the order will not be affected, no threads run twice, and no threads are skipped (besides the one being terminated obviously).

*4. CreateThreadNow was changes only for consistency's sake.*
All the other commonly used functions don't require you index the Citizen table, so it only makes sense this function shouldn't either. I also made it only print the error if it fails rather than trigger an error, as creating a thread should not stop the calling thread by triggering an error, just like any other thread or even the same thread after the first time it runs.

*5. Anything else I did, didn't affect anything old.*
I made a few local changes, the event handlers get created in one spot instead of like 3 or 4. The other few functions I added are just small utility functions, they don't change the behavior of any already-existing features.

PR #199, opened 2019-04-05
shared: fix fwPlatformString narrowing on linux

Hellslicer:
On linux `fwPlatformString` is a basic `std::string` instead of a `std::wstring`.


blattersturm:
Already resolved by #196.

PR #202, opened 2019-04-14
Added some vehicle window methods and updated RollDownAllWindows

BlackFlash5:
Added following methods:
```csharp
VehicleWindow[] GetAllWindows();
void RollUpAllWindows();
bool HasWindow(VehicleWindowIndex window);
```

Updated following method:
```csharp
void RollDownAllWindows();
```

### DIFF
--- a/code/client/clrcore/External/VehicleWindowCollection.cs
+++ b/code/client/clrcore/External/VehicleWindowCollection.cs
@@ -62,7 +62,18 @@ namespace CitizenFX.Core
 
 		public void RollDownAllWindows()
 		{
-			API.RollDownWindows(_owner.Handle);
+			foreach(VehicleWindow vehicleWindow in this.GetAllWindows())
+			{
+				vehicleWindow.RollDown();
+			}
+		}
+
+		public void RollUpAllWindows()
+		{
+			foreach (VehicleWindow vehicleWindow in this.GetAllWindows())
+			{
+				vehicleWindow.RollUp();
+			}
 		}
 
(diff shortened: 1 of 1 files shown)


code/client/clrcore/External/VehicleWindowCollection.cs
```diff
@@ -48,14 +49,48 @@ internal VehicleWindowCollection(Vehicle owner)
...
+
+		public void RollUpAllWindows()
+		{
+			foreach (VehicleWindow vehicleWindow in this.GetAllWindows())
+			{
+				vehicleWindow.RollUp();
+			}
+		}
+
+		public bool HasWindow(VehicleWindowIndex window)
```
carmineos: You probably want to cover more bones, this is a list of some bone names which might be interested https://hastebin.com/feyaqifaqe.txt you can find some of these in vehicles with a lot of windows like bus and so on..

carmineos: I guess [these ](https://github.com/citizenfx/fivem/blob/30e422b392d835f2c4a499b1f316b59ba4278bc4/code/client/clrcore/External/VehicleWindowCollection.cs#L15) extra windows are among these

BlackFlash5: The vehicle `bus` has more windows like `window_lm1` to `window_lm6` and `window_rm1` to `window_rm5`, but only `window_lf`, `window_lr` and `window_rr` are "working" windows.
If you find a vehicle where the extra windows works, then i can try to find the bones.

PR #204, opened 2019-04-22
cfx-ui: support translation

PichotM:
- Supporting translation for cfx-ui
- Fallback if translation is unknown
- Based on system's language or settings
- Add french translation


PichotM:
:(

PR #207, opened 2019-05-03
handling-loader-five: Fix for fDriveBiasFront get/set

carmineos:
set: Fixed condition that wouldn't have allowed to set 0.9f as value
get: Changed logic to match the one used for the set
fDriveBiasFront and +4 (fDriveBiasRear) are either one 1.0f and the other 0.0f or both are a value among 0 and 1 (whose sum is 1.0f) but stored doubled

### DIFF
--- a/code/components/handling-loader-five/src/HandlingLoader.cpp
+++ b/code/components/handling-loader-five/src/HandlingLoader.cpp
@@ -79,7 +79,21 @@ static float getFloatField(char* handlingChar, uint32_t offset, const char* fiel
 {
 	auto hash = HashRageString(fieldName);
 
-	if (hash == HashRageString("fDriveBiasFront") || hash == HashRageString("fBrakeBiasFront") || hash == HashRageString("fSuspensionBiasFront") ||
+	if (hash == HashRageString("fDriveBiasFront"))
+	{
+		float fDriveBiasFront = *(float*)(handlingChar + offset);
+		float fDriveBiasRear = *(float*)(handlingChar + offset + 4);
+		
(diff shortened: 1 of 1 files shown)


code/components/handling-loader-five/src/HandlingLoader.cpp
```diff
@@ -79,7 +79,20 @@ static float getFloatField(char* handlingChar, uint32_t offset, const char* fiel
 {
 	auto hash = HashRageString(fieldName);
 
-	if (hash == HashRageString("fDriveBiasFront") || hash == HashRageString("fBrakeBiasFront") || hash == HashRageString("fSuspensionBiasFront") ||
+	if (hash == HashRageString("fDriveBiasFront"))
+	{
+		float fDriveBiasFront = *(float*)(handlingChar + offset);
+
+		if (fDriveBiasFront + *(float*)(handlingChar + offset + 4) > 1.0f)
```
carmineos: This could be instead
```c
float fDriveBiasRear = *(float*)(handlingChar + offset + 4);
if ((fDriveBiasFront == 1.0f && fDriveBiasRear == 0.0f) || (fDriveBiasFront == 0.0f && fDriveBiasRear == 1.0f))
{
	return fDriveBiasFront;
}
else
{
	return fDriveBiasFront / 2.0f;
}
```

blattersturm: wow, reviewing your own code changes, cool!

why not force push that part instead

carmineos: kek done

PR #219, opened 2019-05-14
 Add locale for Chinese (zh-cn, zh-tw)

jojllman:


jojllman:
Greetings @blattersturm 

May I have some information about why this PR is closed? I believe this very first Chinese translation is thorough and accomplished, which was also reviewed and tested by my friends.
Could you share some of your thoughts? We'd be very appreciated.

PR #222, opened 2019-05-19
Add support for query strings in fivem:// URIs

dremixam:
Support for query strings in fivem:// URIs would allow to generate unique links with an auth token embedded
`fivem://connect/[IP address]?[Query string]`

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -40,7 +40,7 @@ build_server_windows:
       - out/server.zip
   except:
     - tags
-    #- master
+    - master
     - feature/update-to-1365
     #- feature/update-to-1493
   cache:
@@ -61,7 +61,7 @@ build_proot_linux:
       - fx.tar.xz
   except:
     - tags
-    #- master
+    - master
     - feature/update-to-1365
     - feature/update-to-1604
     - feature/cloning-stuff-releng
--- a/code/client/clrcore/External/Vehicle.cs
+++ b/code/client/clrcore/External/Vehicle.cs
@@ -545,7 +545,7 @@ namespace CitizenFX.Core
 		/// </summary>
 		public float Acceleration
 		{
-			get => API.GetVehicleAcceleration2(Handle);
+			get => API.GetVehicleCurrentAcceleration(Handle);
 		}
 		/// <summary>
 		/// Gets or sets the current RPM of this <see cref="Vehicle"/>.
--- a/code/client/clrcore/InternalManager.cs
+++ b/code/client/clrcore/InternalManager.cs
@@ -290,55 +290,84 @@ namespace CitizenFX.Core
 			Debug.WriteLine("Unhandled exception: {0}", e.ExceptionObject.ToString());
 		}
 
+		private static HashSet<string> ms_assemblySearchPaths = new HashSet<string>();
+
 		public void LoadAssembly(string name)
 		{
 			LoadAssemblyInternal(name.Replace(".dll", ""));
 		}
 
-		static Assembly LoadAssemblyInternal(string name)
+		static Assembly LoadAssemblyInternal(string baseName, bool useSearchPaths = false)
 		{
-			try
+			var attemptPaths = new List<string>();
+			attemptPaths.Add(baseName);
+
+			if (useSearchPaths)
 			{
-				var assemblyStream = new BinaryReader(new FxStreamWrapper(ScriptHost.OpenHostFile(name + ".dll")));
-				var assemblyBytes = assemblyStream.ReadBytes((int)assemblyStream.BaseStream.Length);
+				foreach (var path in ms_assemblySearchPaths)
+				{
+					attemptPaths.Add($"{path.Replace('\\', '/')}/{baseName}");
+				}
+			}
 
-				byte[] symbolBytes = null;
-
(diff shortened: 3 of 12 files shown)


blattersturm:
... what? The commit sequence is a mess, an arbitrary user-provided string shouldn't be an identifier, fivem://connect/ isn't meant to be any sort of API for other applications like 'launchers' to use, and this goes in against literally everything including future plans for user-focused connectivity things.

What issue is this trying to solve, other than weird-ass 'launchers' which don't have any use at all?

dremixam:
Fair enough about the commit sequence, I can come back with something clean.

BUT
this is intended to send a unique auth token generated by a website or a launcher which has more reason to be an identifier than any of the Identifier the game currently uses.

What about future plans ? Because currently there is a LOT we can't do properly without "weird ass launchers". And what fivem://connect/ is meant to be ?

Just two examples, we currently have no reliable identifier and we can't manage wait lists properly.

ASmall317:
What does your fancy pants launcher do that FiveM can't?

smallo92:
So what does your fancy pants launcher do that FiveM can't?

Etuldan:
> So what does your fancy pants launcher do that FiveM can't?

The SteamID seems to be broken (on some configurations), same for the R* (the license hex). 
So we can't rely on SteamID and R* ID. DiscordID is a thing, but as Steam, it could be broken at any time (if Discord servers went down for example).

The goal of this PR seems to be able to log-in a whitelisted server with a secure auth token (provided by a external launcher or a link in a website), to avoid any HTML form inside FiveM with CEF.

PichotM:
From seaborgium: https://forum.fivem.net/t/a-psa-on-steam-identifier-support/517277/48 

Identifier type	Prevalence
xbl	63723 (40.6%)
live	79601 (50.7%)
discord	116414 (74.2%)
steam	142853 (91.1%)
license	156889 (100%)
Total in sample	156889 (100%)

R* id is 100% accurate and discord "The discord identifier should work fine since recent patches to the CnL backend that fixed some incorrect usage of HTTP client instances. With a 74.2% prevalence without any direct push by servers to mandate it, it is a very decent identifier type to use as an account-based fallback."

Etuldan:
Hm, OK for licenseID, my bad.

But, if I need to add someone in my whitelist, and my server rely on license, how can the user is able to get his licenseID (in order I can add him to the whitelist) ?$

Anyway, except for the commit sequence which is messy, is there any negative point (security reason, performance reason, ...) against including a PR (I repeat, if the commit sequence, code is clean, unit tests are OK)

blattersturm:
> how can the user is able to get his licenseID (in order I can add him to the whitelist)

State their identifier in your rejection message, like the official example whitelisting script did.

> Anyway, except for the commit sequence which is messy, is there any negative point (security reason, performance reason, ...) against including a PR (I repeat, if the commit sequence, code is clean, unit tests are OK)

In this case, there's no direct concern, other than the fact this does not work well being an identifier - a session token or other additional data pack should be exposed in a different way (preferably _with_ query string parsing!), using a dedicated kind of native, e.g. (in reference C syntax):

```c
char* GET_PLAYER_JOIN_PARAMETER(char* playerSrc, char* identifierKey);
```

Identifiers are not meant to _change_ for the same user, and systems can and will directly use opaque identifiers as storage mechanism - a transient token or parameter pack should therefore not be an identifier.

Etuldan:
I never said it should be an identifier, only a temporary (for like 5 minute) token.
The goal is:
The user login a  launcher/website (with regular login/password and/or cookie). This website/launcher, if the user is whitelisted and registered, create a token.
The FiveM server will authorize only the temporary token logins and match them with a persistent 'custom userID'

For the rejection message, the user have to manually retype the message on his screen (with possibly misstypes), or read the .log (which is not very user friend, IMO)

blattersturm:
> I never said it should be an identifier, only a temporary (for like 5 minute) token.

The original PR is adding this token as an identifier type, which is incorrect usage.



> For the rejection message, the user have to manually retype the message on his screen (with possibly misstypes), or read the .log (which is not very user friend, IMO)

Selecting text is a thing. Copying text is a thing. If your users can't copy text from a dialog box, what _can_ they reasonably do?

Etuldan:
The original PR is adding "an auth token". Yes, in the code, this is a string, but it's up to server owner to do the magic with it.
Auth token from client, sent to server, server checks it token is valid, if yes, get the "custom internal playerID", and send back this playerID to client (with other players info)


Honestly, I just discovered that the "Connection failed" submessage is selectable. Thanks for the tip ;) 
But unfortunately, his is not usual to have such message box error selectable (for example, the text "Connection failed" isn't selectable, in the Server detail (before we press "Connect", only the IP and Port (but not both at the same time) are selectable.)

So I highly doubt ppl knows that the error message is selectable.

dremixam:
> > 
> 
> In this case, there's no direct concern, other than the fact this does not work well being an identifier - a session token or other additional data pack should be exposed in a different way (preferably _with_ query string parsing!), using a dedicated kind of native, e.g. (in reference C syntax):

OK so if I do it with a native GET_PLAYER_JOIN_PARAMETER it will be ok ? I can do query string parsing and will take care of doing a clean pull request this time.

PR #225, opened 2019-05-22
clrcore: updated ServerWrappers to use API

carmineos:


blattersturm:
Ew, `API.` prefixes everywhere.

PR #227, opened 2019-05-24
Add Bulgarian Language

DoctorAndy:
Add Bulgarian language


d0p3t:
You should push this commit to your `patch-1` branch so your [other PR](https://github.com/citizenfx/fivem/pull/226) gets updated instead. Two PRs is unnecessary

PR #231, opened 2019-05-25
d-typing: declare source as global

izio38:
Just add `declare var source: string;` so it's accessible from server without typescript error.


Disquse:
There's also a lot missed declares, `clearTick` for instance.

izio38:
If anyone have more things to add, let me know here so I can add these at the same time :)
Thanks for the suggestion @Disquse

PR #236, opened 2019-06-06
Add Chinese Language

JimWails:
Add Chinese Language
Chinese-Traditional and Chinese-Simplified
local-cn.json is for Chinese-Simplified
local-zh.json is for Chinese-Traditional


JimWails:
When will it be added to FiveM?

PR #246, opened 2019-06-22
gamer info: add patches to support onesync

balika011:
Let's us to draw gamer tag (https://docs.fivem.net/game-references/gamer-tags/) up to 256 players and 128 peds.


blattersturm:
I take it this is WIP? Not going to accept this without patterns, especially not with a new update coming in the coming weeks and there being _this many_ addresses.

balika011:
glhf making patterns for this

balika011:
Added patterns, but among this much patterns there a surely a few that won't work on the next build...

PR #251, opened 2019-07-02
add SET_PLAYER_NAME native (server)

PichotM:
An useful native for people that do not necessary use steam usernames. It will also replace the name in players.json but will not be synced with the client.
It could also be used with adaptive card to "choose" your username for example.


blattersturm:
No. Not in this way, at least.

PR #253, opened 2019-07-06
Add Spanish Language

aXoZer:
Add Spanish language


aXoZer:
[locale-es.zip](https://github.com/citizenfx/fivem/files/3365049/locale-es.zip)

Added locale-es

PR #257, opened 2019-07-07
Add Spanish language

ZinedineDGN:


ZinedineDGN:
Sorry for my silly question but when will they take the files from my repository?

PR #258, opened 2019-07-11
Implement train track node getter

Disquse:
This native returns train's current track node index.
Example (by TheIndra): https://i.imgur.com/Eo6rwir.png

https://github.com/citizenfx/natives/pull/141

### DIFF
--- a/code/components/extra-natives-five/src/VehicleExtraNatives.cpp
+++ b/code/components/extra-natives-five/src/VehicleExtraNatives.cpp
@@ -155,6 +155,7 @@ const int EnginePowerMultiplierOffset = 0xAC0;
 const int CanWheelsBreakOffset = 0x923; // todo - check?
 const int BlinkerState = 0x929;
 const int WheelieState = 0x14F9;
+const int TrainTrackNodeIndex = 0x14C0;
 
 // Wheel class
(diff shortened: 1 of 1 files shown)


code/components/extra-natives-five/src/VehicleExtraNatives.cpp
```diff
@@ -451,6 +452,8 @@ static HookFunction initFunction([]()
 	fx::ScriptEngine::RegisterNativeHandler("GET_VEHICLE_WHEELIE_STATE", readVehicleMemory<unsigned char, WheelieState>);
 	fx::ScriptEngine::RegisterNativeHandler("SET_VEHICLE_WHEELIE_STATE", writeVehicleMemory<unsigned char, WheelieState>);
 
+	fx::ScriptEngine::RegisterNativeHandler("GET_TRAIN_CURRENT_TRACK_NODE", readVehicleMemory<int, TrainTrackNodeIndex>);
```
blattersturm: CVehicle doesn't go all the way to 0x14C0, please add some sort of vehicle type assertion and return, a sane default if not.

Actually, why don't the things for CBike above do the same?!

Disquse: Fixed.

PR #265, opened 2019-07-28
fix for laft over Russian text

ArMaTeC:
please merge this simple update as its strange to see Russian on an en client


blattersturm:
'Laft over' (sic)? This text is there _intentionally_.

See this PR: https://github.com/citizenfx/fivem/pull/98

PR #267, opened 2019-07-31
onesync: complete CVehicleGameStateDataNode and add new server natives

PichotM:


PichotM:
Thanks @bladecoding for finding some specific types

PR #270, opened 2019-08-05
Increased CNetworkRopeWorldStateData limit

razzie:
I experienced error messages in log and even crashes due to CNetworkRopeWorldStateData limit (20) and other players reported to me too.
This is not my log, but it shows the problem: https://pastebin.com/9bzRENKD

### DIFF
--- a/data/client/citizen/common/data/gameconfig.xml
+++ b/data/client/citizen/common/data/gameconfig.xml
@@ -28,6 +28,10 @@
 							<PoolSize value="256"/>
 						</Item>
 						<Item>
+							<PoolName>CNetworkRopeWorldStateData</PoolName>
+							<PoolSize value="256"/>
+						</Item>
+						<Item>
 							<PoolName>CNetObjPed</PoolName>
 							<PoolSize value="238"/> <!-- 110 + player count -->
 						</Item>


razzie:
I checked the logs and I think the build scripts are broken

blattersturm:
OK, this change actually breaks the entire game, even _with_ this earlier fix.

@razzie Did you even TEST these changes by applying them locally to your own game folder?

razzie:
Are you sure it is my commit that break the game, and not the other one that referenced this pull request?
Because I did test the game on my private server and havent seen anything unusual. Unusual means there were some bugs after a long playtime, like blip issues, but it often happened before my patch.

blattersturm:
Reverting the change to gameconfig.xml instantly resolved crashes for people, both before and after the 'fix' commit, including on a vanilla server. 😕 It's odd how a pool size causes some completely different code to fail.

The 'fix' commit did fix non-OneSync servers, but any OneSync server remained broken.

razzie:
I'm happy to send you a dozen of my crashlogs because of that pool size. Also I repeat, I did test the game with updated xml and was fine

blattersturm:
Heard some reports lower values than 256 (including 128) don't instantly break.. though shouldn't ropes actually be able to be cleaned up after use? If the game leaks ropes even when they're not active anymore that's an unrelated issue :(

razzie:
I suspect the game keeps increasing the rope ID for each new rope even when previous ones are freed properly, but I dont have proof. Also dont know exactly why it crashes the game sometimes, but most often it is just there as an error in the logs

PR #273, opened 2019-08-06
Added Spanish language

TheNikoWTF:
I translated FiveM to Spanish, its a very important language and i dont know why FiveM doesn't have it.
Thanks for your time and hope its get aproved.


sharkiller:
I was about to do this. :+1:

PR #275, opened 2019-08-18
vendor/lua: luaV_gettable modifications

gottfriedleibniz:
See commit messages...

On a similar note, would it be possible to dos2unix, or something along those lines, vendor/lua? There are similar changes I'd like to make in lmathlib, but the line-endings are wildly inconsistent (CRLF vs LF) and the diff's are borderline unreadable.

### DIFF
--- a/code/vendor/lua/src/lbaselib.c
+++ b/code/vendor/lua/src/lbaselib.c
@@ -622,6 +622,10 @@ static int luaB_dot (lua_State *L) {
     lua_checkvector2(L, 1, &x1, &y1);
     lua_checkvector2(L, 2, &x2, &y2);
     lua_pushnumber(L,dot2(x1,y1, x2,y2));
+  } else if (lua_isquat(L, 1)) {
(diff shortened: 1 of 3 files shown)


blattersturm:
> On a similar note, would it be possible to dos2unix, or something along those lines, vendor/lua? 

Odd - might be these were checked in on a system without `core.autocrlf` enabled.

blattersturm:
Anyway, merged this (since I'm assuming these were tested beforehand :) these assumptions often are wrong so I hope I'm right this time).

gottfriedleibniz:
Also, would there be any objections to expanding the vector math API, similar to slimmath in C#? There are a variety of vecmath operations that would benefit from being implemented in pure C. 

Even the simple point in AABB queries can be optimized significantly.

blattersturm:
Nope, as long as it doesn't get too crazy-specific supporting more math without a native invocation boundary would be fine.

PR #277, opened 2019-08-20
 vendor/lua: lmathlib componentwise vector functions

gottfriedleibniz:


blattersturm:
Not sure I agree 100% with re-adding mathlib compatibility routines, but they definitely help with discoverability since a lot of examples/docs are based on older Lua versions.

PR #278, opened 2019-08-20
added ListEvents for lua and js

tabarra:
Since its not the global list of events, perhaps I should rename it to ListResourceEvents?

### DIFF
--- a/data/shared/citizen/scripting/lua/scheduler.lua
+++ b/data/shared/citizen/scripting/lua/scheduler.lua
@@ -374,6 +374,16 @@ function TriggerEvent(eventName, ...)
 	end)
 end
 
+function ListEvents()
+	local events={}
+	for e,_ in pairs(eventHandlers) do
+		if e:sub(1, 5) ~= "__cfx" then
+			table.insert(events, e)
+		end
+	end
+	return events
+end
+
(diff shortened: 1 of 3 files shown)


blattersturm:
Hm, what is the use case for this? Why would a resource not know what events they have registered themselves?

tabarra:
Better control and logging.  
I actually used this to find some dynamic event names from a framework.  
Also, some servers are starting to randomize their event names with UUID, having a list might be useful.

blattersturm:
However, since these methods exist in the global scope, wouldn't it make more sense if you need to 'find event names', to simply prefix a script in the correct ScRT, and monkey-patch AddEventHandler, rather than adding a feature to each HLL ScRT that would only be needed in a very small amount of use cases?

tabarra:
Hmm, how would that monkey-patch work?  
you mean changing `global.addEventListener` or editing all the `addEventListener` in the code to call an intermediary logger function?

tabarra:
Not sure how to do that, for now I'm going to use this patch. 
But feel free to close the PR.

tabarra:
BTW @blattersturm, can you give me an example on how to monkey-patch `addEventListener`?  
I'm curious how this is possible, specially on the Lua env.

PR #280, opened 2019-08-21
onesync: complete CPedHealthDataNode

PichotM:
Tested and working properly.
I did not added `GET_PED_CAUSE_OF_DEATH` because for some unknown reason the returned value in the synctree does not match what should be returned even though the order seems fine.

I get 1891351 instead of 3452007600 for `weapon_fall`


bladecoding:
Looks like unk9 is supposed to be a 13bit short and unk11 shouldn't exist. Here is what the end should look like.
```
auto unk8 = state.buffer.ReadBit();

if (unk8) // unk9 != 0
{
	auto unk9 = state.buffer.Read<short>(13);
}

int causeOfDeath = state.buffer.Read<int>(32);
state.entity->data["causeOfDeath"] = causeOfDeath;

int injuredStatus = state.buffer.Read<int>(2); // Change below 150 HP, injured data?

auto unk13 = state.buffer.ReadBit();

if (unk13)
{
	int unk14 = state.buffer.Read<int>(8);
}
```

PichotM:
nice! @gottfriedleibniz said that something was right before causeOfDeath but I was unable to fix/find it.
I'll test it thanks!

PR #289, opened 2019-09-04
Arabic support

kwmx:
Added locale-ar.json
Update languages.ts along with adding the rtl direction for Arabic and Persian codes.


CorrM:
Nice PR.

PR #292, opened 2019-09-12
Native for overriding ms per game minute value

Disquse:
This native is allowing to set what is game minute in real milliseconds. Default is 2000 (real ms) per 1 in-game minute. For example, changing value to 1000 will speed up in-game clock in 2 times.

## Demonstration:
https://gfycat.com/ColdGregariousBrant

https://github.com/citizenfx/natives/pull/175

### DIFF
--- a/code/components/extra-natives-five/src/TimeExtraNatives.cpp
+++ b/code/components/extra-natives-five/src/TimeExtraNatives.cpp
@@ -1,0 +1,42 @@
+#include "StdInc.h"
+
+#include <Local.h>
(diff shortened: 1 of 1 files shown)


mickeygeecom:
This would be awesome. Currently we're sending server IRL time to all clients pretty often to achieve 'real time' on the server. Would a 1 setting be like normal time then?

Disquse:
@mickeygeecom You will be able to set IRL amount of ms per day. And then, using server, just once sync global time for player.

PR #295, opened 2019-09-23
Label docker container with CI_BUILD_NUMBER

Fireant456:
This used to be a label and was useful for identifying a pulled docker container's version. Can we add this back or explain an alternative way to get the version of a stopped fivem docker container?


Hellslicer:
@Fireant456 I added some fixes to Rocker build and added labels in https://github.com/citizenfx/fivem/pull/296

PR #297, opened 2019-09-25
onesync/natives: GetEntityModel now also returns ped's models

PichotM:
GetEntityModel is now getting ped's models. (untested)


PichotM:
oof let me remove these unwanted commits

PR #301, opened 2019-09-30
Update setInterval

samuelds:
Add args for callback for setInterval

### DIFF
--- a/data/shared/citizen/scripting/v8/timer.js
+++ b/data/shared/citizen/scripting/v8/timer.js
@@ -91,9 +91,7 @@
         setTimer(
             id,
             function() {
-                try {
-                    callback(...argsForCallback);
-                } catch { }
+		callback(...argsForCallback);
             },
             interval
         );
--- a/data/shared/citizen/scripting/v8/timer.js
+++ b/data/shared/citizen/scripting/v8/timer.js
@@ -93,7 +93,7 @@
             function() {
                 try {
                     callback(...argsForCallback);
-                }
+                } catch { }
             },
             interval
         );
--- a/data/shared/citizen/scripting/v8/timer.js
+++ b/data/shared/citizen/scripting/v8/timer.js
@@ -88,7 +88,7 @@
     function setInterval(callback, interval, ...argsForCallback) {
         const id = nextTimerId();
 
-	setTimer(
+        setTimer(
             id,
             function() {
                 try {
--- a/data/shared/citizen/scripting/v8/timer.js
+++ b/data/shared/citizen/scripting/v8/timer.js
@@ -85,10 +85,18 @@
         animationFrames[animationFrames.length] = callback;
     }
 
-    function setInterval(callback, interval) {
+    function setInterval(callback, interval, ...argsForCallback) {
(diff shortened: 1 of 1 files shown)


data/shared/citizen/scripting/v8/timer.js
```diff
@@ -85,10 +85,18 @@
         animationFrames[animationFrames.length] = callback;
     }
 
-    function setInterval(callback, interval) {
+    function setInterval(callback, interval, ...argsForCallback) {
         const id = nextTimerId();
 
-        setTimer(id, callback, interval);
+	setTimer(
```
Hellslicer: Bad indentation
```suggestion
        setTimer(
```

data/shared/citizen/scripting/v8/timer.js
```diff
@@ -85,10 +85,18 @@
...
+            id,
+            function() {
+                try {
+                    callback(...argsForCallback);
+                }
```
Hellslicer: Missing catch or finally after try
```suggestion
                } catch { }
```

data/shared/citizen/scripting/v8/timer.js
```diff
@@ -85,10 +85,18 @@
         animationFrames[animationFrames.length] = callback;
     }
 
-    function setInterval(callback, interval) {
+    function setInterval(callback, interval, ...argsForCallback) {
```
thers: It would be great if you'd also add support for the args passing to the `setTimeout` method as well.

samuelds: That's why I added argsForCallback

samuelds: setTimeout has already:

```js
test(arg) {
   console.log(arg); // Print exemple
}

setTimeout(test, 100, "exemple");
```

thers: Aww, yeah, you're right, it's already there, my bad.

data/shared/citizen/scripting/v8/timer.js
```diff
@@ -85,10 +85,18 @@
...
         const id = nextTimerId();
 
-        setTimer(id, callback, interval);
+        setTimer(
+            id,
+            function() {
+                try {
+                    callback(...argsForCallback);
+                } catch { }
```
thers: This is bad. If something will throw an error inside `callback` we won't ever notice it failing in console. I'd better write `setTimer(id, callback.bind(null, ...argsForCallback), interval);`

samuelds: try and catch is there for that. Just add a print in catch.

thers: And you will do what already is there. See https://github.com/citizenfx/fivem/blob/master/data/shared/citizen/scripting/v8/timer.js#L127-L131

thers: Thing that process every timer callback is already doing that, why doing it here again?

samuelds: I did not pay attention, thx :)

PR #303, opened 2019-10-05
Update to Visual studio 2019 and Boost

GTA5Racer:
Docs were really outdated. If setting up workspace using old docs, it will have alot of errors.


Disquse:
Premake5 binary which is used for CI is not supporting vs2019, consider to update it too (code\tools\ci).

blattersturm:
Updating Boost to 1.70 actually won't work and will produce incompatible builds; however VS16 is required by now, yes.

blattersturm:
Closed for now. Please reopen if wanting to fix this bit.

PR #304, opened 2019-10-06
Fix: cmdline args were not passed when bootstrapping

bladecoding:
This fixes losing cmdline args, such as +connect, when the client updates.


blattersturm:
What about arguments containing double quotes in the original command line?

PR #305, opened 2019-10-10
Ability to control Upperbody in Sequential animations.

mossyblog:
49 will give developers the ability to enable sequential for the upper torso while leaving the lower body controllable. Prime example is a more refined version of "hands up" animations beyond just the normal TaskHandsUp().


Disquse:
49 = 48 (real flag) + 1 (loop), so `UpperBodyControllable` is 48 flag, not 49.

mossyblog:
I've noticed no difference on the results here. 48 and 49 seem to offer no issues?

blattersturm:
32 + 16 + 1, this is a Flags enum so there's no need to actually define flag combos.

PR #307, opened 2019-10-20
Patch weapon components limit

Disquse:
This PR is increasing weapon components pool (CWeaponComponentInfo) limit to 1024 (configurable).

P.S. - Жожа, привет.


Disquse:
Fixed the weapon component struct, that's not really important in this PR, but hey, why not. Thanks to @gottfriedleibniz for sharing his researches.

PR #314, opened 2019-11-17
Add SET_TEXT_FONT_FOR_CURRENT_COMMAND native

summeryukata:
Use before _DRAW_TEXT, param 0 is font id


code/components/config.lua
```diff
@@ -32,11 +32,11 @@ else
 	component 'nui-resources'
 	component 'citizen-game-main'
 	component 'citizen-game-ipc'
-
```
blattersturm: ?

code/components/extra-natives-rdr3/component.lua
```diff
@@ -0,0 +1,3 @@
+includedirs { '../voip-mumble/include' }
```
blattersturm: does it need these currently?

summeryukata:
resolved?

PR #316, opened 2019-11-21
index.d.ts: add missing declarations

PichotM:
- resultAsLong
- resultAsObject
- clearTick
- source

### DIFF
--- a/data/shared/citizen/scripting/v8/index.d.ts
+++ b/data/shared/citizen/scripting/v8/index.d.ts
@@ -10,6 +10,8 @@ interface ResultAsInteger {}
 interface ResultAsFloat {}
 interface ResultAsString {}
 interface ResultAsVector {}
+interface ResultAsLong {}
+interface ResultAsObject {}
 
 type InputArgument =
     string |
@@ -23,7 +25,9 @@ type InputArgument =
     ResultAsInteger |
     ResultAsFloat |
     ResultAsString |
-    ResultAsVector;
+    ResultAsVector |
+    ResultAsLong |
+    ResultAsObject;
 
 interface CitizenInterface {
     trace(...args: string[]): void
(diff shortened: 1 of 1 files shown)


TheDumpap:
Nice, makes things easier :)

Xseba360:
Why is source declared as a string. Isn't it supposed to be a number?

PichotM:
Source is a string

Xseba360:
how come in a `.js` resource the type is `number` ?
```js
addNetEventListener('testEvent', function () {
  // prints 'number' in console
  console.log(typeof source)
  // prints [1], not ['1']
  console.log(JSON.stringify([source]))
});
```

PichotM:
Actually, according to other type definitions, it should be `number|string` but in Lua or in the cpp code 'source' is a string
You can PR this if you want

hojgr:
@PichotM even in Lua its either number or string. `source` in event handlers and commands is `number`, `GetActivePlayers()` has server IDs as string. Leads to insane and hard to debug bugs sometimes...

now I just always `tonumber` any source anywhere. Still wonder why its string sometimes

Xseba360:
This is very confusing.
For example:
https://runtime.fivem.net/doc/natives/?_0x406B4B20
```typescript
// 0x406B4B20 
// GET_PLAYER_NAME
const retval: string = 
	GetPlayerName(
		playerSrc: string
	);
```
The runtime docs specify the `GetPlayerName()` function takes the `source` as a string, but the TypeScript definitions state otherwise (number). Testing this shows passing it as a string fails to return the correct name.
```js
for (const player of GetActivePlayers()) {
  // prints 128
  console.log(player)
  // prints number
  console.log(typeof player)
  // prints Xseba360
  console.log(GetPlayerName(player))
  // prints **Invalid**
  console.log(GetPlayerName(player.toString()))
}
```

Zeemahh:
Correct. On the client, `get_player_name` only takes a number, whereas on the server it takes either a string or number.

On server, iterate from 0-`get_num_player_indices`, calling `get_player_name` passing `get_player_from_index`, returning a string, and you'll see it will return proper name.

PR #321, opened 2019-12-29
Added SET_TEXT_JUSTIFICATION and SET_TEXT_WRAP  native

ThatIsMiyu:
Adds Text Justification and Text Wrap native to draw text right justified / to the right of the screen


blattersturm:
interesting typo, really should somehow manage to fix the CI checks so that this would be caught...

PR #323, opened 2019-12-30
Implement DRAW_LINE and DRAW_POLY for RDR3

Disquse:
This PR is restoring some drawing natives that was unfairly removed from RDR3.

image


DemmyDemon:
These functions are sorely missed for all kinds of script debugging.

PR #360, opened 2020-01-18
Update building.md

Shrimpey:
Added additional section for setting up testing enviroment. All those steps are scattered in different places (Discord, forum) and make it tough to properly build FiveM client for someone new to all of this. I've added all the things that should be sufficient to test modified client. Hopefully, this will encourage more people to try building the project and contribute (it was a big discouragement for me having to look for those additional steps not knowing they even existed in the first place)

### DIFF
--- a/docs/building.md
+++ b/docs/building.md
@@ -1,3 +1,4 @@
+
 # Building FiveM
 To build FiveM's client components, you need the following dependencies:
 * A Windows machine with Visual Studio 2019 (Build Tools/Community is fine) installed with the C++ and UWP workloads.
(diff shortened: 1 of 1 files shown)


nathanctech:
The build.ps1 steps would be very confusing for someone not used to reading powershell as there is a lot of CI related things that you would never used (since it's used for automated builds). Perhaps an abbreviated script would be useful?

Shrimpey:
> The build.ps1 steps would be very confusing for someone not used to reading powershell as there is a lot of CI related things that you would never used (since it's used for automated builds). Perhaps an abbreviated script would be useful?

I mentioned it for additional research, however personally I just copy specific DLL that Im working on to a regular client downloaded from site. Nevertheless, we could create sth that would just copy/paste files without doing all the additional stuff and prepare fully working client, it would probably just be just the case of trimming that build.ps1 script

PR #369, opened 2020-02-10
Added null-terminator to end of event name for latent events

niekschoemaker:
This fixed the bug causing latent events' name to stick around after using a longer name


niekschoemaker:
Tested it on Windows server and windows client with both TriggerLatentServerEvent and TriggerLatentClientEvent

PR #385, opened 2020-02-28
Current ped weapon getter for OneSync

Disquse:
https://github.com/citizenfx/natives/pull/260

### DIFF
--- a/code/components/citizen-server-impl/include/state/SyncTrees_Five.h
+++ b/code/components/citizen-server-impl/include/state/SyncTrees_Five.h
@@ -1131,7 +1131,7 @@ struct CPedGameStateDataNode
 			weapon = state.buffer.Read<int>(32);
 		}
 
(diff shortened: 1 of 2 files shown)


blattersturm:
totally builds
```
C:\projects\fivem\code\components\citizen-server-impl\include\state\SyncTrees_Five.h(1134,1): error C2679: binary '=': no operator found which takes a right-hand operand of type 'uint32_t' (or there is no acceptable conversion) (compiling source file C:\projects\fivem\code\components\citizen-server-impl\src\state\ServerGameState_SyncTrees.cpp) [C:\save\build\server\master\server\windows\citizen-server-impl.vcxproj]
```

blattersturm:
also the getter makes no sense either

and in fact i seem to have slacked off with PR approval too, `GET_VEHICLE_ROOF_LIVERY` etc. are wrong as well

PR #387, opened 2020-02-28
Add simplified Chinese and traditional Chinese support

3125751684:
Add simplified Chinese and traditional Chinese support


3125751684:
Request passed

3125751684:
GitHub branch request goes through

PR #392, opened 2020-03-13
Using json.decode on a null value

imtimon:
It should return null instead of causing an error. This is very useful when fetching data from a database as some JSON columns may have a default value of null, attempting to json.decode one of those null values will cause an error and break the script so a check was necessary.


blattersturm:
I think this is resolved since we use rapidjson by default now?

PR #394, opened 2020-03-18
Add Lithuanian language

R0JEA:
Add Lithuanian language translations to FiveM


Xinerki:
did you use google translate or something? because some translations make no sense

RoJea32:
> did you use google translate or something? because some translations make no sense

which ones? :DD

Xinerki:
"watermark" being translated word for word, ignoring the actual meaning.
"Client script'ai leidžiami" what.
"Tiesioginis ryšys" is also not a very accurate translation

RoJea32:
"vandens ženklas" is not a bad translation, you can call it "logotipas" if you want, it's correct both ways
"client script'ai leidžiami" have a better translation for this? :)
"tiesioginis ryšys" can be corrected to "tiesioginis jungimasis" if you're so picky

Quattro20v:
I think it's a good enough translation. Some words are a bit tricky to translate.

PR #398, opened 2020-03-25
Add MUMBLE_SET_VOLUME_OVERRIDE_MANUAL

Frosty-Ice:
Allows you to reset/change MUMBLE_SET_VOLUME_OVERRIDE if the player is not in your active players for big mode/infinity by manually providing the correct arguments.

https://github.com/citizenfx/natives/pull/268


blattersturm:
Not going to be implemented in this way, rather should have a means of enumerating clients and matching by `[ID]` prefix only, requiring to specify a _matching_ player name is stupid.

If implemented that way it could also be called something like `_BY_SERVER_ID` and not `_MANUAL`.

Rubylium:
Waiting for this ~

Disquse:
Why you closed it?

Frosty-Ice:
I'm not familiar with how player data is stored in `gta:net` so I can't implement it as asked. I get errors when attempting to build the FiveM client to test code, and I refuse to push untested/unverified code.

Disquse:
You may just ask for help here or in Discord guild if you want to finish it. If I understood the context correctly, you can use `NetworkPlayerMgr`. Here's an example:
https://github.com/citizenfx/fivem/blob/9c88f0e7504b9f8f595fdc6e10d1fe78c8dcfe8b/code/components/citizen-resources-gta/src/PlayerScriptFunctions.cpp#L26

Frosty-Ice:
Is the m_session variable here https://github.com/citizenfx/fivem/blob/d53af3474f6eb4b1ce21a881b75a8caac5aabd0f/code/components/voip-mumble/include/MumbleClientState.h#L46-L55
the server id or something else entirely? If not, would you want the server id stored there so you could do searches in `voip-mumble` later with server ids for clients?

Disquse:
Your latest changes looks good tho. Can't test it, but it compiles fine.

Disquse:
You just broke your branch by merging. Try reset "Merge branch" commit and force push without it.

blattersturm:
:(?

PR #403, opened 2020-04-03
fix: return proper rotation

Ascaped:
So I was messing around with entity rotation. I couldn't figure out which order to use, so I messed around with API.GetEntityRotation and this fixed my rotation issues. It returns local rotation of the entity.

### DIFF
--- a/code/client/clrcore/Server/Entity.cs
+++ b/code/client/clrcore/Server/Entity.cs
@@ -38,7 +38,7 @@ namespace CitizenFX.Core
 		{
 			get
 			{
-				return API.GetEntityRotation(this.Handle);
+				return API.GetEntityRotation(this.Handle, 4);
 			}
 			set
 			{


PichotM:
Most scripts use 2

Ascaped:
Sure, yet we have no way of getting any specific order using Entity.Position. So I cannot get the right one which is 4.

blattersturm:
~~All of the rotation orders are the 'right one'. If you want an unambiguous order, use the quaternion.~~

~~Changing the default used here would break compatibility. 😕~~

Wait... this is the server-side wrapper, this doesn't even _have_ a rotation order argument: it just returns the order sent from the network! This doesn't even build at all, and even if it would wouldn't do anything.

What are you even trying to do here? If you want to expose rotation orders for the server wrapper, decompose rotations with the network order into a quaternion and recompose, but that requires some looking at game code.

PR #404, opened 2020-04-04
Add MUMBLE_SET_VOLUME_OVERRIDE_BY_SERVER_ID

Frosty-Ice:
Allows you to reset/change MUMBLE_SET_VOLUME_OVERRIDE if the player is not in your active players for big mode/infinity by providing the server id.

https://github.com/citizenfx/natives/pull/268


code/components/voip-mumble/src/MumbleClientState_User.cpp
```diff
@@ -35,7 +35,10 @@ void MumbleUser::UpdateUser(MumbleProto::UserState& state)
 
 	if (state.has_name())
 	{
-		m_name = ConvertFromUTF8(state.name());
+		std::string name = state.name();
+		m_name = ConvertFromUTF8(name);
+		// Hackish way to pull the serverid from the name
+		m_serverId = atoi(name.substr(1, name.length() - 1).c_str());
```
blattersturm, line 41: Might be worthwhile to add a safeguard to ensure the name is at least 2 characters long to prevent malicious individuals from crashing all connected clients with a std::length_error by joining the Mumble server directly with the name `f` or similar.

code/components/gta-net-five/src/MumbleVoice.cpp
```diff
@@ -670,6 +670,17 @@ static HookFunction hookFunction([]()
 			}
 		});
 
+		fx::ScriptEngine::RegisterNativeHandler("MUMBLE_SET_VOLUME_OVERRIDE_BY_SERVER_ID", [](fx::ScriptContext& context)
+			{
```
blattersturm, line 674: Inconsistent indentation, might be irrelevant.

code/components/voip-mumble/src/MumbleClientState_User.cpp
blattersturm: > (at least, that's what I assume will happen when size() == 1 and substr(1, 0) gets called - maybe it'll be fine too?)

Actually, this seems to be [fine](https://ideone.com/sBqfi3). Hopefully there won't be any odd cases where UTF-8 conversion can break or an empty name is set... might be worth safeguarding anyway.

Frosty-Ice: What would you want the behavior to be if the string isn't long enough? Just don't update the serverId?

blattersturm: Since in that case there won't be a serverId, yeah - just let it default to `0` which is an impossible server ID.

PR #405, opened 2020-04-04
server: added check if ped exists

BlackFlash5:
If you use `player.Character` e.g. shortly after a player connected, then you have a chance of running into:
```
InvokeNative: execution failed: Tried to access invalid entity: 137485
```

This will check if the entity exists, before calling another native.


blattersturm:
While well-intended, I rather think it'd make more sense to ensure that all uses will return a valid handle instead of a handle that (apparently) doesn't exist.

PR #409, opened 2020-04-10
Various fixes of UB and dangling handles

oktonion:


code/client/launcher/CitizenGame.cpp
```diff
@@ -437,7 +437,13 @@ void CitizenGame::Launch(const std::wstring& gamePath, bool isMainGame)
 	}
 
 	// force loading _this_ variant
-	LoadLibrary(MakeRelativeCitPath(L"scripthookv.dll").c_str());
+	HMODULE scripthookvHandle = LoadLibrary(MakeRelativeCitPath(L"scripthookv.dll").c_str());
+
+	if (!scripthookvHandle)
+	{
+		FatalError("Could not load scripthookv.dll.");
```
blattersturm, line 444: Error states ought to be ignored here - this is just to ensure dependent .asi files will load our variant, if present.

The comment could be made more clear instead, since the plugin not being present will fail later during ComponentLoader (and this change will break launching RDR).

code/client/launcher/CitizenGame.cpp
```diff
@@ -571,7 +577,10 @@ void CitizenGame::Launch(const std::wstring& gamePath, bool isMainGame)
 			if (getenv("CitizenFX_ToolMode"))
 			{
 				// pre-load the correct chrome_elf.dll
-				LoadLibraryW(MapRedirectedFilename(L"Social Club/chrome_elf.dll").c_str());
+				HMODULE chromeElfHandle = LoadLibraryW(MapRedirectedFilename(L"Social Club/chrome_elf.dll").c_str());
+
+				if(!chromeElfHandle)
+					return (HMODULE)INVALID_HANDLE_VALUE;
```
blattersturm, line 583: chrome_elf is again a disambiguation, failing to load `libcef.dll` however might lead to confusing errors instead.

code/client/launcher/Console.Logging.cpp
```diff
@@ -172,57 +177,85 @@ struct LoggerInit
...
 	g_logCondVar.notify_all();
 }
 
 void InitLogging()
 {
 	SetThreadName(-1, "MainThrd");
 
-	if (OpenMutex(SYNCHRONIZE, FALSE, L"CitizenFX_LogMutex_Mod") == nullptr)
+	const auto handle_deleter = [](HANDLE handle){ if (handle != INVALID_HANDLE_VALUE) CloseHandle(handle); };
+	using UniqueHandle = std::unique_ptr<std::remove_pointer<HANDLE>::type, decltype(handle_deleter)>;
```
blattersturm, line 193: Is using `decltype` to get the type of a captureless lambda considered fine? Perhaps there's more cases like this straddled throughout code that could benefit from pulling in [WIL](https://github.com/microsoft/wil).

code/client/launcher/Console.Logging.cpp
```diff
@@ -172,57 +177,85 @@ struct LoggerInit
...
-
-		CreateDirectory(MakeRelativeCitPath(L"logs/").c_str(), NULL);
+		mutexHandle.reset(CreateMutex(NULL, TRUE, L"CitizenFX_LogMutex_Mod"));
 
-		SYSTEMTIME curTime;
-		GetSystemTime(&curTime);
+		auto relativeCitPath = MakeRelativeCitPath(L"logs/");
 
-		for (auto& f : std::filesystem::directory_iterator{ MakeRelativeCitPath(L"logs/") })
+		if(0 == CreateDirectory(relativeCitPath.c_str(), NULL))
```
blattersturm, line 207: `if(`? a tad nitpicky here, but could make sense

code/client/launcher/Console.Logging.cpp
```diff
@@ -172,57 +177,85 @@ struct LoggerInit
...
+						SystemTimeToFileTime(&curTime, &curStamp);
+						SystemTimeToFileTime(&fakeTime, &fakeStamp);
+
+						uint64_t curT = ((ULARGE_INTEGER*)&curStamp)->QuadPart;
+						uint64_t fileT = ((ULARGE_INTEGER*)&fakeStamp)->QuadPart;
+
+						if ((curT - fileT) > (7 * 24 * 60 * 60 * uint64_t(10000000)))
+						{
+							std::error_code err_c;
+							std::filesystem::remove(f.path(), err_c); // noexcept overload
```
blattersturm, line 242: Right. This one actually tripped up some actual end users.

code/client/launcher/Console.Logging.cpp
```diff
@@ -172,57 +177,85 @@ struct LoggerInit
...
+				// So we will use modern C++17 to create the full path
+				
+				std::filesystem::create_directories(relativeCitPath.c_str());
+			}
 		}
 
+
 		AsyncTrace(fmt::format("--- BEGIN LOGGING AT {:%c} ---\n", fmt::gmtime(time(NULL))).c_str());
-	}
+	}
```
blattersturm, line 260: yes

code/client/launcher/Main.cpp
```diff
@@ -295,16 +297,26 @@ int WINAPI wWinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, PWSTR pCmdLine
 
 		wcscat_s(systemPath, dll);
 
-		LoadLibrary(systemPath);
+		if(!LoadLibrary(systemPath))
+		{
+			printf("Could not initialize ");
```
blattersturm, line 302: Would it make more sense to use `OutputDebugString` here? CitiLaunch by default does not have a console window, this is too early for `trace`, but it's typically easier to attach a Win32 global debug monitor than it is to change the process subsystem value.

code/client/launcher/Main.cpp
```diff
@@ -295,16 +297,26 @@ int WINAPI wWinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, PWSTR pCmdLine
...
+		{
+			printf("Could not initialize ");
+			wprintf(dll);
+			printf(".\n");
+		}
 	}
 
-	LoadLibrary(MakeRelativeCitPath(L"dinput8.dll").c_str());
-	LoadLibrary(MakeRelativeCitPath(L"steam_api64.dll").c_str());
+	if(!LoadLibrary(MakeRelativeCitPath(L"dinput8.dll").c_str()))
```
blattersturm, line 308: With these extra lines of code it might make more sense to just iterate over an initializer list?

code/client/launcher/Main.cpp
```diff
@@ -386,10 +398,18 @@ int WINAPI wWinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, PWSTR pCmdLine
 
 	if (!toolMode)
 	{
-		if (OpenMutex(SYNCHRONIZE, FALSE, L"CitizenFX_LogMutex") == nullptr)
+		const auto handle_deleter = [](HANDLE handle){ if (handle != INVALID_HANDLE_VALUE) CloseHandle(handle); };
+		using UniqueHandle = std::unique_ptr<std::remove_pointer<HANDLE>::type, decltype(handle_deleter)>;
```
blattersturm, line 402: Code repetition, yes.

code/client/launcher/Main.cpp
```diff
@@ -386,10 +398,18 @@ int WINAPI wWinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, PWSTR pCmdLine
 
 	if (!toolMode)
 	{
-		if (OpenMutex(SYNCHRONIZE, FALSE, L"CitizenFX_LogMutex") == nullptr)
+		const auto handle_deleter = [](HANDLE handle){ if (handle != INVALID_HANDLE_VALUE) CloseHandle(handle); };
+		using UniqueHandle = std::unique_ptr<std::remove_pointer<HANDLE>::type, decltype(handle_deleter)>;
+
+		UniqueHandle mutexHandle(
+			OpenMutex(SYNCHRONIZE, FALSE, L"CitizenFX_LogMutex_Mod"),
```
blattersturm, line 405: _Mod? Also, I'm not sure if it makes sense to close this handle upon scope ending - this logic is entirely unused I believe rather nowadays (we don't have .log.X files to rotate anymore) but if this were still used subprocesses would rotate logs a lot.

code/components/loading-screens-five/src/LoadingScreens.cpp
```diff
@@ -776,7 +776,7 @@ static InitFunction initFunction([] ()
 				}
 			}
 
-			uint64_t loadTimingIdentity = HashString(name) | (16 << 48);
+			uint64_t loadTimingIdentity = HashString(name) | (uint64_t(16) << 48);
```
blattersturm, line 779: Good point. Default MSVC compiler warning verbosity makes it hard to catch some legitimate warnings.

code/components/net-base/src/NetPeerBase.cpp
```diff
@@ -29,7 +29,7 @@ void PeerBase::ProcessPacket(const std::vector<uint8_t>& buffer)
 int PeerBase::ReadCompressedType(Buffer& buffer)
 {
 	uint8_t lead = buffer.Read<uint8_t>();
-	int result;
+	int result = -1;
```
blattersturm, line 32: Not sure if this code was ever used... it might have been some early test of a custom net wrapper. Dead code, yay!

code/components/nui-core/src/NUIInitialize.cpp
```diff
@@ -429,7 +429,13 @@ void HookLibGL(HMODULE libGL)
 {
 	MH_Initialize();
 	MH_CreateHook(GetProcAddress(libGL, "glTexParameterf"), glTexParameterfHook, (void**)&g_origglTexParameterf);
-	MH_CreateHook(GetProcAddress(LoadLibraryW(L"d3d11.dll"), "D3D11CreateDevice"), D3D11CreateDeviceHook, (void**)&g_origD3D11CreateDevice);
+
+	auto d3d11Handle = LoadLibraryW(L"d3d11.dll");
+	if (!d3d11Handle)
+	{
+		FatalError("CitizenFX: Cannot load 'd3d11.dll'. Please (re)install 'Direct X' for your OS.");
```
blattersturm, line 436: Not sure if this needs an error state since CitiLaunch takes in `d3d11.dll` already as well as ViabilityTests->DXGICheck in CitiLaunch doing this as well.

The error phrasing is also a tad odd - `Direct X` -> `DirectX`, and it is typically part of the OS itself nowadays (unlike d3dx*.dll).

code/components/nui-core/src/NUIInitialize.cpp
```diff
@@ -445,13 +451,23 @@ void Component_RunPreInit()
...
-	LoadLibrary(L"bluetoothapis.dll");
+	auto bluetoothapisHandle = LoadLibrary(L"bluetoothapis.dll");
+	(void)(bluetoothapisHandle); // it's ok if there is an error
 
 	// load Chrome dependencies ourselves so that the system won't try loading from other paths
-	LoadLibrary(MakeRelativeCitPath(L"bin/chrome_elf.dll").c_str());
-	LoadLibrary(MakeRelativeCitPath(L"bin/libEGL.dll").c_str());
+	auto chromeElfHandle = LoadLibrary(MakeRelativeCitPath(L"bin/chrome_elf.dll").c_str());
+	if(!chromeElfHandle)
+		MessageBoxW(NULL, L"Could not load bin/chrome_elf.dll.", L"CitizenFX", MB_ICONWARNING | MB_OK);
```
blattersturm, line 460: Users will typically ignore message boxes opened in sequence - this is already a bit of a concern with Ldr errors at this time, but when explicitly making error messages one _should not_ open multiple dialog boxes, since the user will only screenshot the last one.

code/vendor/botan/src/botan_all.cpp
```diff
@@ -33210,9 +33210,9 @@ void GOST_34_11::add_data(const uint8_t input[], size_t length)
          m_position = 0;
          }
       }
-
-   const size_t full_blocks = length / hash_block_size();
-   const size_t remaining   = length % hash_block_size();
+   const auto hash_block_size_v = hash_block_size();
```
blattersturm, line 33213: This is a third-party library; which is also probably somewhat overdue for updating client-side but newer versions have weird defaults that break ROS proxying so we so far only update it server-side as it's more consequential there.

code/components/nui-core/src/NUIInitialize.cpp
oktonion: Totally understandable. But since there is no unified way in the codebase of tracing errors I just took as example code few lines below with messagebox error. Could you give me the cue how to log this please?

oktonion:
@blattersturm Appreciate your comments on the PR. The codebase is so diverse in terms of error handling, tracing logs and mixing WinAPI with C++17 that I was a bit lost there. I'll improve the PR relying on your remarks but I'll need a little bit of help there.

code/client/launcher/Console.Logging.cpp
oktonion: Just following the docs from MS =)

> If the function fails, the return value is zero

Anyway, could we just replace ugly WinAPI with modern C++ filesystem variant a be done with it?

code/client/launcher/Console.Logging.cpp
oktonion: Yes and what for WIL - totally Yes, if it is ok to add another dependency to the project.

blattersturm:
This PR has been ignored for a very long time and is now filled with merge conflicts. Sorry for not getting to it before. :/

PR #412, opened 2020-04-19
murmur: don't send voice to both session and channel targets

PichotM:


code/components/voip-server-mumble/src/client.cpp
```diff
@@ -941,13 +941,15 @@ int Client_voiceMsg(client_t *client, uint8_t *data, int len)
 	} else if ((vt = Voicetarget_get_id(client, target)) != NULL) { /* Targeted whisper */
 		int i;
 		channel_t *ch;
+		std::map<int, bool> targeted_channels;
```
blattersturm, line 944: This is a very hot code path, allocating dynamically multiple times in here could be problematic. Since this is a very small map most likely, using a stack-allocated eastl::fixed_map with, say, a 256-object max (and overflow enabled) might be more suited? You might have to add the vendor component, however.

code/components/voip-server-mumble/src/client.cpp
```diff
@@ -993,7 +995,8 @@ int Client_voiceMsg(client_t *client, uint8_t *data, int len)
 			buffer[0] = (uint8_t) (type | 2);
 			Log_debug("Whisper session %d", vt->sessions[i]);
 			while (Client_iterate(&c) != NULL) {
-				if (c->sessionId == vt->sessions[i]) {
+				channel_t* ch = (channel_t*)client->channel;
+				if (!targeted_channels[ch->id] && c->sessionId == vt->sessions[i]) {
```
blattersturm, line 999: `.find(ch->id) != .end()` is preferred here (non-mutating, your use of `operator[]` will append the current channel with default value - false - on checking), in fact since you're not ever using false _intentionally_ you could also use a `set` - or rather, an `eastl::fixed_set`.

PR #413, opened 2020-04-20
onesync/events: add CWeaponDamageEvent

PichotM:
Works fine.
However, when damageType is 3 I am not sure if it is a mistake but hitGlobalId is zero as f80_1 is false.


blattersturm:
>  However, when damageType is 3 I am not sure if it is a mistake but hitGlobalId is zero as f80_1 is false.

Code makes it appear like 3 is some type of radius damage - which would then be expected that it should apply to all entities in the radius.

PR #418, opened 2020-04-25
Fixed a few parameters which were missed for 1868.

nkjellman:
These values were increased in 1868.


blattersturm:
why merge into a dead branch?!

nkjellman:
> why merge into a dead branch?!

Because I thought the unsupported 1868 build ran from this branch.

PR #420, opened 2020-04-26
onesync: add is_ped_a_player

PichotM:
useful for server-side anticheat


PichotM:
Native https://github.com/citizenfx/natives/pull/281

PR #421, opened 2020-04-27
fix: GET_HASH_KEY crash on server with null value

niekschoemaker:
This fixes the crash when GetHashKey gets called with a nil value on the server side.


niekschoemaker:
Another solution could be to default null to 0 or empty string.
But that depends on expected behavior, think throwing an error would be the appropriate option here. Just not one which crashed the server

PR #422, opened 2020-04-28
onesync: Add GetVehiclePedIsIn

Quattro20v:
My attempt of adding in GetVehiclePedIsIn to the server-side. I think that it would be highly useful to double check what vehicle the player is in on the server-side.

Don't bash me too much tho it's my first time looking at this source code. Plus I am not even sure how to test this unfortunately and run it. And for some reason "citizen-server-impl" didn't show up in the CitizenMP.sln, so I had to manually open the whole folder.

Either way, let me know how does this look, if it makes sense and maybe some tips on how to contribute and test the code in the future.


PichotM:
lastVehicle is only in the struct definition and is never set so it won't work
but the rest should be fine, thanks for this PR

PichotM:
And also use MakeHandle
As I believe it returns the netid

Quattro20v:
Thank you for taking a look! I just checked the lastVehicle, are you sure that lastVehicle is never set? I found some code in ServerGameState.cpp:
line: 1422 `vehicleData->lastVehicle = vehicleData->curVehicle;` 

Also by MakeHandle do you mean MakeScriptHandle? That's what "GET_ALL_VEHICLES" seems to use.

Quattro20v:
Seems like I messed it up, with trying to do the MakeScriptHandle. Maybe someone can shine some light in here.

Also still waiting for some clarification if lastVehicle exists or not.

Quattro20v:
After figuring out how to run the server and test, I got it to finally work. Most of the proper work is in the last 2 commits. I pulled the trigger too soon with the other ones, before getting familiar with the source code and how to test it.

code/components/citizen-server-impl/src/state/ServerGameState_Scripting.cpp
```diff
@@ -802,6 +802,40 @@ static InitFunction initFunction([]()
...
+		if ((lastVehicleArg == true && node->lastVehiclePedWasIn == -1) || (lastVehicleArg == false && node->curVehicle == -1))
+			return (uint32_t)0;
+
+		auto returnEntity = lastVehicleArg == true ? gameState->GetEntity(0, node->lastVehiclePedWasIn) : gameState->GetEntity(0, node->curVehicle);
+
+		if (!returnEntity)
+			return (uint32_t)0;
+
+		// Return current or last vehicle
+		if (lastVehicleArg)
```
jabbink, line 834: 4 lines to return `gameState->MakeScriptHandle(returnEntity);`

Quattro20v: Good catch! Fixed.

blattersturm:
Can you maybe rebase this onto latest master and add a native declaration to ext/native-decls?

Quattro20v:
Yeah for sure, I can do that when I wake up!

Quattro20v:
Native citizenfx/natives#289

blattersturm:
> Native [citizenfx/natives#289](https://github.com/citizenfx/natives/pull/289)

no, natives go in ext/native-decls, not the natives repo

Quattro20v:
Added the native declaration to ext/native-decls

blattersturm:
`.MD`->`.md`, perhaps; filenames are case-sensitive on some systems.

Quattro20v:
Fixed the file name extension. Sorry about that.

blattersturm:
OK, since Travis succeeded (and Appveyor config was broken), this might get merged in soon. :)

Quattro20v:
Sounds good, thank you :)

PR #423, opened 2020-04-28
feature: add 2d convar for mumble voice chat

PichotM:


PichotM:
The voice target fix is also included in this PR

PR #427, opened 2020-05-04
 mumble: fix reconnect 

balika011:
If you don't stop `m_idleTimer` `m_connectTimer` won't be fired.


blattersturm:
Merged via b8cfae48b56b5428309e3cee1cb36e0b4f3514d6.

PR #435, opened 2020-05-07
[OneSync] Move ServerGameState::m_entitiesById to a lock-free implementation

DefCon42:
This should improve performance a bit (especially on Linux) due to the lack of a need for mutexes left and right here.


blattersturm:
> C:\projects\fivem\code\components\citizen-server-impl\include\LRWeakPtr.h(21,1): error C3493: 'newPtr' cannot be implicitly captured because no default capture mode has been specified (compiling source file C:\projects\fivem\code\components\citizen-server-impl\src\state\ServerGameState.cpp) [C:\save\build\server\master\server\windows\citizen-server-impl.vcxproj]

Doesn't seem to build on CI in the current state,

PR #436, opened 2020-05-08
Speed up searches with a .ignore file

sgottelk:


blattersturm:
What tools even read these files? It doesn't seem like, say, VS does.

sgottelk:
Tools that use the `.ignore` file: Visual Studio Code, [ag](https://github.com/ggreer/the_silver_searcher), [ripgrep](https://github.com/BurntSushi/ripgrep)

blattersturm:
... basically, all tools that nobody uses with this repo then, since VSCode C++ support is abysmal, and I've no clue what kind of edgy person would use those other tools at all (and scanning vendor could be useful FWIW).

sgottelk:
well Visual Studio project-wide search is pretty abysmal, so VS Code is nice when i'm searching for code, but fair enough

blattersturm:
it improved in 16.5 a lot anyway

PR #439, opened 2020-05-09
Add server sided DeleteEntity native declaration

Jazzuh:


ext/native-decls/DeleteEntity.md
```diff
@@ -0,0 +1,14 @@
...
+ns: CFX
+apiset: server
+---
+## DELETE_ENTITY
+
+```c
+void DELETE_ENTITY(Entity entity);
+```
+
+Deletes the specified entity. This is server sided version of this function, which can be used with OneSync enabled
```
blattersturm, line 11: Not sure why this needs a specific sentence (with grammar mistakes) since the API set field already says such.

PR #440, opened 2020-05-10
Create GetPlayerIdentifiers.md

hassony18:


Zeemahh:
This function only exists in Lua, afaik.

hassony18:
Indeed. I've just checked the other languages and it doesn't work there. However, I am not sure how I can edit a native for a specific language, as you can see: 
image

There are 4 languages but when I click on "Improve this document" it takes me to a single file where I can only edit C language. 
image


Could you please lead me to the proper way to edit it for LUA for example?

blattersturm:
It's not a native. It doesn't belong in natives.

It's already documented in the Lua-specific reference: https://docs.fivem.net/docs/scripting-reference/runtimes/lua/functions/GetPlayerIdentifiers/

hassony18:
Alright thanks for the clarification. I will close this PR just after you answer my question above: how do I edit a native for a specific language (like LUA only)

d0p3t:
> Alright thanks for the clarification. I will close this PR just after you answer my question above: how do I edit a native for a specific language (like LUA only)

You can't edit anything specific for a language tab. For examples just use markdown. See https://runtime.fivem.net/doc/natives/?_0x06843DA7060A026B

PR #441, opened 2020-05-10
Update GetPlayerIdentifier.md

hassony18:


ext/native-decls/GetPlayerIdentifier.md
```diff
@@ -10,7 +10,17 @@ char* GET_PLAYER_IDENTIFIER(char* playerSrc, int identifier);
 
 
 ## Parameters
-* **playerSrc**: 
-* **identifier**: 
+* **playerSrc**: the player you want to get his identifier
+* **identifier**: the identifier id:
+```
+0 = steam
```
blattersturm, line 16: Nope, that's entirely wrong, both in formatting as well as factually. The identifier list is an array without explicit indexes or ordering, if an identifier isn't present the index will be something else.

PR #442, opened 2020-05-11
schedular: only deserialize event structure if there is an event handler

rubbertoe98:


blattersturm:
Why? The event function doesn't even get invoked if there isn't a handler registered with C++.

rubbertoe98:
Ah I didn't know the function doesn't get invoked if there's no handler nw.

PR #447, opened 2020-05-17
Additional blip color

Local9:
Some additional blip colors in the Library referenced from the documents, if doc updates are required also let me know and I'll add a pull request for that also.


d0p3t:
What's with the weird diff? Different line endings?

TomGrobbe:
Looks like he probably formatted the file and it added indents for every line. If you hide whitespace diff it shows the actual code changes.

Local9:
Sorry, I made a quick edit to it in Visual Studio Code which has affected the file format. Is there a set CRLF you want me to set locally to not have this happen in the future?

Kngrektor:
There's a https://github.com/citizenfx/fivem/blob/master/code/.editorconfig (dunno how up to date it is). If VSC doesn't support it by default https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig probably does the job.

Local9:
Close: Will fix my line endings first

PR #450, opened 2020-05-20
Increase TxdStore limit

Mick-King:
This increases TxdStore from 65500 to 480000 so more textures can be streamed.


blattersturm:
That's not enough, this is hardcoded to a 16-bit ID so you're going to have to patch a lot of game code to not break things. Did you even test this?

PR #451, opened 2020-05-20
Basic resmon version for RedM

Ktos93:


blattersturm:
Copying exact files that should be shared across games is bad form.

PR #454, opened 2020-05-24
FiveM Server crasher exploit [HOTFİX]

umutsantur00:
Hey, in your fivem application system have joke code. "fivem\code\components\citizen-server-impl\serc\InitConnectMethod.cpp" this area have traceback joke code. If any guy send php-get exploit from any server:  server will be don. FxServer gets crash. Please hold on and fix it!
`// verify
	if (ticketExpiry < utcTime)
	{
		trace("Connecting player: ticket expired\n");
		return false; <-- Please change this area to true, cause this area calling back this verify system. This is very dangerous for all fivem servers. Change this fork.
	}`


technetium-cfx:
Hi, this does not appear as a valid PR. What exactly do you mean by joke code? If you want us to review a PR, please explain what it is that you solve, and make sure to not 1989 old commits with 7705 changed files.

PR #455, opened 2020-05-25
Update of czech locale

rex2630:


rex2630:
I dont know why u guys decided to go with Weblate translation when there is done PR some time now, new commit is just horrrible, since it adds string twice and doesnt add some strings that are missing.

rex2630:
Updated on Weblate

PR #460, opened 2020-05-28
Optimize ValidateEntity function

SloPro:
When the first set of ifs are true, checking g_entityCreationList
can be skipped as the result will not be changed.

Second if changed to direct return instantly breaks out of the loop.


blattersturm:
Random returns are bad form generally, `break;` and checking the boolean before further iteration seems better?

PR #469, opened 2020-06-05
Citcore: Add cmd's check on adding to avoid duplicated and a cmd proposal to remove all ace for an object

izio38:


izio38:
Is it normal that server build fail on linux ?

code/client/citicore/se/Security.cpp
```diff
@@ -74,6 +74,11 @@ std::string Context::SaveSnapshot()
 
 void Context::AddPrincipalInheritance(const Principal& child, const Principal& parent)
 {
+	if (DoesPrincipalInheritanceExists(child, parent))
+	{
+		console::Printf("security", "A principal inheritance already exists with this parent and child.\n");
```
blattersturm, line 79: Warning on this is a little odd. :(

izio38: Should we completely remove this ?

izio38:
Still waiting for a review :)

PR #471, opened 2020-06-06
webadmin: added console write ability

jcfoxtrot54:
This is my second pull request, I apologise for its mess but I've done my best to clean it up.

Simply adds the ability to write to console from the webadmin panel provided webadmin as a resource has the correct permissions.


Korioz:
I don't understand C family well but i think these commits should be done in the txAdmin repository cause txAdmin is included as a submodule in FiveM if i'm not wrong.

Jamelele:
> I don't understand C family well but i think these commits should be done in the txAdmin repository cause txAdmin is included as a submodule in FiveM if i'm not wrong.

This is not for txAdmin, it's for the other panel at the /webadmin endpoint.

jcfoxtrot54:
> 
> 
> I don't understand C family well but i think these commits should be done in the txAdmin repository cause txAdmin is included as a submodule in FiveM if i'm not wrong.

Sorry I've been running around like a headless chicken the past couple of days. This is inteded for the webadmin resource, not for txAdmin, as @Jamelele explained

Korioz:
I believed that old webadmin had been removed but it's not the case, don't understand why cause txAdmin has a lot of updates / commits made every day and i think that most part of server owners prefer & use txAdmin rather than webadmin 🙄

blattersturm:
This PR seems to contain a number of unrelated changes. Can you make a separate branch with only the relevant changes?

In addition to that, a compatibility API for webadmin plugins is planned for txAdmin at some point.

jcfoxtrot54:
My PR's a mess, for some reason I pushed a couple of test functions. I'll close this and open a new PR on a separate branch

PR #475, opened 2020-06-21
Add pushing individual version pushing to docker

Fireant456:
Allows for version switching on docker installations


blattersturm:
This Docker Hub image seems dead at this time, so this PR is not needed anymore - it'll need to be done when/if we ever revive this image.

PR #487, opened 2020-07-06
Mumble Voip for RedM

Ktos93:
I know,I need add some stuff like Camera Matrix and Input/Output device change but this is good base to start mumble on RedM. Now servers can use TokoVoip or SaltyChat but players  don't like to use additional applications


code/components/gta-net-rdr3/MumbleVoice.cpp
```diff
@@ -0,0 +1,694 @@
...
+}
+
+static HookFunction hookFunction([]()
+{
+
+
+    rage::scrEngine::OnScriptInit.Connect([]()
+    {
+
+        g_preference = hook::get_pattern<char>("D0 3E 7B 45 01 00 00 00 E0 4D 27 45", 0x2B4);
```
Disquse, line 378: Are you sure this pattern is viable enough? Can't find it on 1207.58 and 1232.62. It seems you hardcoded vtable's address to your pattern, which is not good.

Ktos93: I dont have any issues with this pattern i test this code with 4 friends and works fine. But if you want how i can improve this ?

Disquse: That pattern may work fine with the current version of the game which is used by RedM (1207.80 iirc). But it will be broken for newer game patches (which is gonna be ported to RedM I hope). Patterns are used to avoid such breakage between different versions of the game. And your pattern is literally just hardcoded address (0x1457B3ED0?) which will be different for other game versions. In short, you have to find a more viable pattern for that address. Feel free to ping me in CFX's discord and I will try to help you.

Ktos93: Memory pattern changed this should be better.

Ktos93: The script now uses a pointer to get the address of the voice settings structure

minipunch: Don't you need to mask the memory pattern? Would you need to try different versions of the game and see which bytes stay static and which changes between versions?

Ktos93: Memory pattern is valid on both version but 
Disquse say this memory region isn't safe for this

PourrezJ:
Any news?

PR #488, opened 2020-07-08
fix: new CEF download link

PichotM:
Introduced by https://github.com/citizenfx/fivem/commit/207c47fe785fed2f91fc08bc9b73990e188239f0


blattersturm:
Not added for a reason as it's temporary.

PR #491, opened 2020-07-14
onesync: add CClearPedTasksEvent server event

SloPro:
Seems to work fine when tested with 2 players.
Example usage:
```lua
AddEventHandler("clearPedTasksEvent", function(sender, data)
    print(sender, GetPlayerName(sender), json.encode(data))
    -- 128, SloPro, { "pedId": 1, "unk1": true }
end)
```
Can be cancelled to prevent syncing.

### DIFF
--- a/code/components/citizen-server-impl/src/state/ServerGameState.cpp
+++ b/code/components/citizen-server-impl/src/state/ServerGameState.cpp
@@ -3249,7 +3249,7 @@ struct CClearPedTasksEvent
 	void Parse(rl::MessageBuffer& buffer)
 	{
 		pedId = buffer.Read<uint16_t>(13);
-		unk1 = buffer.Read<uint8_t>(1);
+		immediately = buffer.Read<uint8_t>(1);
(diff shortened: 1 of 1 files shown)


code/components/citizen-server-impl/src/state/ServerGameState.cpp
```diff
@@ -3244,6 +3244,24 @@ struct CVehicleComponentControlEvent
...
+		unk1 = buffer.Read<uint8_t>(1);
+	}
+
+	inline std::string GetName()
+	{
+		return "clearPedTasksEvent";
+	}
+
+	int pedId;
+	bool unk1;
```
gottfriedleibniz, line 3261: Should double check this in your testing, but I believe unk1 corresponds to: ``Clear Tasks Immediately``

SloPro: It does appear so, I'll change the name of it then.

PR #492, opened 2020-07-16
Create locale-tr.json (Turkish)

alperb:
Most of the phrases are correct however some can be mistranslated.


d0p3t:
Translations are not done anymore via user PRs. See https://translations.cfx.re/projects/citizenfx/cfx-ui/

PR #495, opened 2020-07-26
Increase "ClothStore" poolsize to 600

rjmxtechnology:
Disregard see comments below


rjmxtechnology:
Tbh I just picked a random number in the file that I saw, guess I misinterpreted what the "ClothStore" is used for but makes sense now after bubble mentioned "*ld" files.  600 is a safer bet

PR #496, opened 2020-07-27
Increase "ClothStore" poolsize from 60 to 600

rjmxtechnology:
A few months back I hit this limit with a bunch of models with cloth physics (yld files) seems I am back at that roadblock again (this time with only 4 models that have cloth physics). The current value of 60 does seem a bit low so increasing this doesn't seem to do much harm at 600. Probably can be tweaked in the future if necessary.

(redid this PR because the last one was absolutely horridly confusing)

### DIFF
--- a/data/client/citizen/common/data/gameconfig.xml
+++ b/data/client/citizen/common/data/gameconfig.xml
@@ -154,7 +154,7 @@
 						</Item>
 						<Item>
 							<PoolName>ClothStore</PoolName>
-							<PoolSize value="60"/>
+							<PoolSize value="600"/>
 						</Item>
 						<Item>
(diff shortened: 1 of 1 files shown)


rjmxtechnology:
Not directly correlational but checking the [commit](https://github.com/citizenfx/fivem/blob/639746e9e2ffb9e774c317006c29ba72866846e7/data/client_rdr/citizen/common/data/gameconfig_pc.xml#L421) from Disquse regarding RedM's gameconfig.xml the "ClothStore" poolsize is set to 750 there . 

While this might not be the **_ideal_** comparison as it's 2 different games (one being newer than the other), it's probably worth noting. 

When testing on my client I set values of 300k+  and didn't seem to have any issues, so theoretically 600 should not be an issue in the slightest

PR #498, opened 2020-08-04
feat(mumble): support external voice server

PichotM:
This has been tested with the mumble server component removed and using two external grumble servers

### DIFF
--- a/code/components/gta-net-five/src/MumbleVoice.cpp
+++ b/code/components/gta-net-five/src/MumbleVoice.cpp
@@ -165,8 +165,7 @@ static struct
 	volatile int nextConnectDelay;
 	volatile uint64_t nextConnectAt;
 
-	std::string overrideAddress;
-	volatile int overridePort;
+	boost::optional<net::PeerAddress> overridePeer;
 
 	concurrency::concurrent_queue<std::function<void()>> mainFrameExecQueue;
 } g_mumble;
@@ -178,7 +177,7 @@ static void Mumble_Connect()
 
 	_initVoiceChatConfig();
 
-	g_mumbleClient->ConnectAsync(!g_mumble.overrideAddress.empty() ? net::PeerAddress::FromString(fmt::sprintf("%s:%d", g_mumble.overrideAddress, g_mumble.overridePort), g_mumble.overridePort, net::PeerAddress::LookupType::NoResolution).get() : g_netLibrary->GetCurrentPeer(), fmt::sprintf("[%d] %s", g_netLibrary->GetServerNetID(), g_netLibrary->GetPlayerName())).then([](concurrency::task<MumbleConnectionInfo*> task)
+	g_mumbleClient->ConnectAsync(g_mumble.overridePeer ? *g_mumble.overridePeer : g_netLibrary->GetCurrentPeer(), fmt::sprintf("[%d] %s", g_netLibrary->GetServerNetID(), g_netLibrary->GetPlayerName())).then([](concurrency::task<MumbleConnectionInfo*> task)
 	{
 		try
 		{
@@ -505,8 +504,7 @@ static HookFunction initFunction([]()
 	OnKillNetworkDone.Connect([]()
 	{
 		g_mumbleClient->SetAudioDistance(0.0f);
-		g_mumble.overrideAddress.empty();
-		g_mumble.overridePort = NULL;
+		g_mumble.overridePeer = {};
 
 		Mumble_Disconnect();
 		o_talkers.reset();
@@ -852,10 +850,18 @@ static HookFunction hookFunction([]()
 			auto address = context.GetArgument<const char*>(0);
 			int port = context.GetArgument<int>(1);
 
-			g_mumble.overrideAddress = address;
-			g_mumble.overridePort = port;
+			boost::optional<net::PeerAddress> overridePeer = net::PeerAddress::FromString(fmt::sprintf("%s:%d", address, port), port);
 
-			Mumble_Disconnect(true);
+			if (overridePeer)
+			{
+				g_mumble.overridePeer = overridePeer;
+
+				Mumble_Disconnect(true);
+			}
+			else
+			{
+				trace("Exception: Couldn't resolve Mumble server address.\n");
+			}
 		});
 
 		scrBindGlobal("GET_AUDIOCONTEXT_FOR_CLIENT", getAudioContext);
--- a/code/components/gta-net-five/src/MumbleVoice.cpp
+++ b/code/components/gta-net-five/src/MumbleVoice.cpp
@@ -860,7 +860,7 @@ static HookFunction hookFunction([]()
 			}
 			else
 			{
-				trace("Exception: Couldn't resolve Mumble server address.\n");
+				throw std::exception("Couldn't resolve Mumble server address.");
 			}
 		});
 
--- a/code/components/gta-net-five/src/MumbleVoice.cpp
+++ b/code/components/gta-net-five/src/MumbleVoice.cpp
@@ -165,6 +165,9 @@ static struct
 	volatile int nextConnectDelay;
 	volatile uint64_t nextConnectAt;
 
+	std::string overrideAddress;
+	volatile int overridePort;
+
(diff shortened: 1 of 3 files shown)


code/components/gta-net-five/src/MumbleVoice.cpp
```diff
@@ -165,6 +165,9 @@ static struct
 	volatile int nextConnectDelay;
 	volatile uint64_t nextConnectAt;
 
+	std::string overrideAddress;
```
blattersturm, line 168: Please store this as a net::PeerAddress directly so resolution cost/failures can be handled at native call time.

code/components/gta-net-five/src/MumbleVoice.cpp
```diff
@@ -832,6 +841,23 @@ static HookFunction hookFunction([]()
...
+			context.SetResult<bool>(g_mumble.connected ? true : false);
+		});
+		
+		
+		fx::ScriptEngine::RegisterNativeHandler("MUMBLE_SET_SERVER_ADDRESS", [](fx::ScriptContext& context)
+		{
+			auto address = context.GetArgument<const char*>(0);
+			int port = context.GetArgument<int>(1);
+
+			g_mumble.overrideAddress = address;
```
blattersturm, line 855: Do net::PeerAddress resolution here _and check if the std::optional is valid_ before trying to dereference/use .get(). If it's not valid, `throw std::exception("Couldn't resolve Mumble server address.")` or so to propagate to script.

This thing you're doing now will lead to a nasty _deferred_ crash if the address is invalid.

code/components/gta-net-five/src/MumbleVoice.cpp
```diff
@@ -498,6 +505,8 @@ static HookFunction initFunction([]()
 	OnKillNetworkDone.Connect([]()
 	{
 		g_mumbleClient->SetAudioDistance(0.0f);
+		g_mumble.overrideAddress.empty();
```
blattersturm, line 508: You're.. checking if it's empty, and discarding the result of that check. :P

`.clear()` or the universal (which will also work if it's a `net::PeerAddress`) `var = {};` should be preferred.

ext/native-decls/MumbleIsConnected.md
```diff
@@ -0,0 +1,12 @@
+---
+ns: CFX
+apiset: client
+---
+## MUMBLE_IS_CONNECTED
+
+```c
+BOOL MUMBLE_IS_CONNECTED();
+```
+
```
blattersturm, line 10: A description? :(

ext/native-decls/MumbleSetServerAddress.md
```diff
@@ -0,0 +1,13 @@
+---
+ns: CFX
+apiset: client
+---
+## MUMBLE_SET_SERVER_ADDRESS
+
+```c
+void MUMBLE_SET_SERVER_ADDRESS(char* address, int port);
+```
+
```
blattersturm, line 10: Another description missing, such as for instance 'Changes the Mumble server address to connect to, and reconnects to the new address.'

code/components/gta-net-five/src/MumbleVoice.cpp
```diff
@@ -175,7 +178,7 @@ static void Mumble_Connect()
 
 	_initVoiceChatConfig();
 
-	g_mumbleClient->ConnectAsync(g_netLibrary->GetCurrentPeer(), fmt::sprintf("[%d] %s", g_netLibrary->GetServerNetID(), g_netLibrary->GetPlayerName())).then([](concurrency::task<MumbleConnectionInfo*> task)
+	g_mumbleClient->ConnectAsync(!g_mumble.overrideAddress.empty() ? net::PeerAddress::FromString(fmt::sprintf("%s:%d", g_mumble.overrideAddress, g_mumble.overridePort), g_mumble.overridePort, net::PeerAddress::LookupType::NoResolution).get() : g_netLibrary->GetCurrentPeer(), fmt::sprintf("[%d] %s", g_netLibrary->GetServerNetID(), g_netLibrary->GetPlayerName())).then([](concurrency::task<MumbleConnectionInfo*> task)
```
blattersturm, line 181: Actually, you can even store an optional<net::PeerAddress> directly, so you can do something like

```cpp
g_mumble.overridePeer ? *g_mumble.overridePeer : g_netLibrary->GetCurrentPeer()
```

PichotM:
It should be better now

code/components/gta-net-five/src/MumbleVoice.cpp
```diff
@@ -832,6 +839,31 @@ static HookFunction hookFunction([]()
...
+
+			if (overridePeer)
+			{
+				g_mumble.overridePeer = overridePeer;
+
+				Mumble_Disconnect(true);
+			}
+			else
+			{
+				trace("Exception: Couldn't resolve Mumble server address.\n");
```
blattersturm, line 863: Ideally, one'd _actually_ throw to propagate error state to script correctly.

ravicant:
> two external grumble servers

I mean i tried to use it grumble ALC permission are fucked how u tested it pls if u can help me

ViolentMinds:
Has this been confirmed to work yet? We could really use this to get around the 64 player limit

blattersturm:
... what "64 player limit"?

ViolentMinds:
The people who run a FiveM server I play on said in the Mumble documentation there is a "64" player limit

When we play and the server has more than 64 players, clients would get a slight lock up then "rp-radio is using 384MB of memory on the server" pop up (and sometimes FiveM would crash because of it) or something like that anyways

Then we would also have consistent audio cut outs across the entire server. Especially with many people nearby.

They said because Mumble is not optimized for more than 64 on the FiveM server

blattersturm:
That's complete nonsense and sounds more like an issue with this "rp-radio". Various servers are using the built in umurmur server with 600+ players, are you sure you're not being confused with the "mumble-voip" resource that has zero relation to built-in functionality or limits?

Also, what "Mumble documentation"? There are no such limits at all, so no idea where these "people" read this.

ViolentMinds:
I literally have no idea all I know is what they told me on the FiveM "FivePDRP" server so I'm just here trying to figure out if there is a fix as they are using Teamspeak TOKO right now which I cant stand

blattersturm:
Seems more like incapability of said server owners to work with built in functionality than any actual limitations.

ViolentMinds:
Very true. What would you suggest in order to get them back to using Mumble on the FiveM server? Externally hosted mumble like this mentions, better than internally?

blattersturm:
No, the built in server is capable of running fine as long as a script is used that splits by channels.
There's various of those, including the "mumble-voip" aforementioned but there's a few others people released as well.

ViolentMinds:
True perhaps mumble-voip is what is not good over 64p as they say. Its frustrating as hell when audio cuts out across entire server or sometimes would bug out and you would hear someone in global. They never figured out a fix

blattersturm:
One I've heard of is "pma-voice", no idea if it works well but it's apparently less flawed than the other one.

ViolentMinds:
They probably wanted rp-radio api for mumble-voip because of the animation of pulling out the police walkie talkie and all that but  lol RIP.  Hmm I will look into that one ty

AvarianKnight:
pma-voice has its own version of [rp-radio](https://github.com/AvarianKnight/rp-radio) most of the mumble-voip -> pma-voice exports and everything are the same

ViolentMinds:
> pma-voice has its own version of [rp-radio](https://github.com/AvarianKnight/rp-radio) which just converts the mumble-voip exports to pma-voice (they have basically the same interface)

Oh ok I think maybe I will suggest pma-voice instead. It seems they dont know much about scripts and all that as they frequently have issues. But neither do I. I've never worked with Mumble or FiveM but have some coding knowledge and experience with game servers. 

Do you know if it works well with 100+ players?

AvarianKnight:
We run it with 150 players plus I think a few other servers use it.

ViolentMinds:
Sounds good thank you I will suggest the owners of FivePDRP use it!

PR #501, opened 2020-08-06
fix(ui-build): stop focus jumping between iframes on tab presses

dwjft:
https://forum.cfx.re/t/nui-iframe-management-tabbing-between-frames/1491744/12

### DIFF
--- a/ext/ui-build/data/root.html
+++ b/ext/ui-build/data/root.html
@@ -30,6 +30,7 @@ window.focusStack = [];
 
 let handoverBlob = {};
 let serverAddress = '';
+let frameEscapeFunctions = {};
 
 registerPushFunction(function(type, ...args) {
 	switch (type) {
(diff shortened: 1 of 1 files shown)


blattersturm:
Very tiny nitpick, but most the file uses semicolons and some of your diff doesn't. Can you fix this and squash+force-push such?

JulienGelmar:
This change unfortunately breaks iFrames that are opened inside a NUI iframe. Example code:
```<html>
<head>
<title>reproduction steps</title>
</head>
<body>
<iframe src="http://html.net/tutorials/php/lesson11.php"></iframe>
</body>
</html>
```

If you now attempt to for example focus the search input up top, you will notice any focus events are now being ignored

blattersturm:
> If you now attempt to for example focus the search input up top, you will notice any focus events are now being ignored

Hm - so a nested iframe will blur the patent frame if focused? Would it be possible for you or @dwjft to PR a tweak to this that would e.g. check the hierarchy of a newly focused frame?

Personally I'm only able to look at this in a little while. 😕

JulienGelmar:
To clarify, the search input I am speaking of is inside the secondary iframe. Every input I am clicking inside this html.net website will instantly lose focus.

After some additional testing I can also say that it attempts to focus the input, but instantly loses it.

dwjft:
Looking.

dwjft:
https://github.com/citizenfx/fivem/pull/505

PR #502, opened 2020-08-07
1-8 keybind inputs for RedM

jostster:
Added the 1-8 keybinds for weapon quickselect into the settings so that they can be accessed in the `IsControlJustPressed()`


Disquse:
It seems R* added those hotkeys. Check if https://github.com/citizenfx/fivem/pull/503 has the same changes like yours?

jostster:
Changes done in #503

PR #505, opened 2020-08-10
fix(nui): only refocus iframe if we are in the parent body

dwjft:
This will only re-focus if we tabbed to the body element, and not a child iframe.


blattersturm:
> This branch has conflicts that must be resolved

Uh, did you accidentally apply to an old tree?

dwjft:
If you want to help test this locally please change your `root.html` in `/FiveM Application Data/citizen/ui.zip`.

My initial testing seemed to work fine but more eyes couldn't hurt.

blattersturm:
Threw it in anyway, if it's worse in a way we'll notice soon enough.

PR #507, opened 2020-08-10
Add additional in-tree vendor components.

DefCon42:
This PR adds two in-tree vendor components:

**1. citizen_util:** A general set of useful utilities and containers. Currently contains object_pool, a fast pool-based allocator, and shared_reference, a shared/weak pointer implementation with the control block detached from the object itself, allowing for destruction and deallocation of the value without breaking weak pointers.

**2. citizen_enet:** An enet fork. Currently, the only change is using the aforementioned object pools for fixed-sized data, though this may change later down the line; as an example, using STL containers or other stuff for data.

These will be used in a later PR for various modifications to the server implementation.


blattersturm:
Should this be considered WIP or can this be safely merged while assuming the build won't break (in case CI goes and has a hiccup again)?

DefCon42:
Though I've build the server component, I'll check other components to make sure CI isn't completely broken.

DefCon42:
Other components don't have dependencies on the vendor files for now, but they seem to be parsed fine, making this OK to merge.

PR #511, opened 2020-08-16
2020 Redesign

aiden-z-source:
This matches the default loading screen to the new 2020 redesign.


FlawwsX:
You have a couple of images you can share?

aiden-z-source:
mDROG28BvU

blattersturm:
What? This is even _less_ functional than what existed before.

PR #512, opened 2020-08-16
onesync/server: implement GET_ALL_OBJECTS

SloPro:
Returns an array of all object handles (objects + doors)


blattersturm:
> Because that's what it's called in client-side C#:

That's only because `object` would conflict with System.Object.

Please keep it as 'object'.

SloPro:
Do you think a GET_ALL_ENTITIES general native would also be beneficial to add? So people could do GetEntityType() by themselves if needed

PR #513, opened 2020-08-16
zh-cn translation

hugoshao:
Add Simplified Chinese translation file


blattersturm:
Translations don't go in direct PRs, sorry.

PR #515, opened 2020-08-16
feat(cfx-ui): streamer mode

sharkykh:
This allows streamers to hide sensitive information on the UI, such as server IP addresses.
I based this on the `devMode` setting, hopefully this is good enough :)

### Interface settings:
Setting defaults to **disabled**.

settings

### Linked profiles / accounts:
settings_account

### Profile Avatar:
avatar

### When connecting:
Added new string for `Connecting to server`.

connecting

### On connection failure:
connection-failed

### Server list filter / direct connect:
The input field is now a password field when streamer mode is enabled.

server-filter
server-filter-selected


sharkykh:
I can't replicate this state I found on Discord (probably because I am only using a local server),
but I assume it is being used after all, so I'm looking into hiding the IP from that as well.

Disquse:
You can also hide these identities I think.
image

sharkykh:
Implemented and updated the screenshots.

jaymo1011:
Would it be possible to add an indication where the text is hidden conveying "hey! you're in streamer mode!" incase a user enables it, forgets, and then wonders *why IP invisible* and *looks like red rectangle* when selected?

Sounds like something that will come up a lot in support questions and giving an immediate indication as to *why* the text is hidden seems like an easy way to prevent that.

sharkykh:
@jaymo1011 
I am actually reworking the server filter this very moment to be a password field.
At first I couldn't get it the way I wanted to, now I figured it out.
It looks much better IMO:

server-filter
server-filter-selected

But regardless, having a "Streamer Mode is enabled" notification seems like a good idea.
I don't really have a design idea in mind, though.

jaymo1011:
> But regardless, having a "Streamer Mode is enabled" ~notification~ warning(?) seems like a good idea.
> **I don't really have a design idea in mind, though.**

With it being a password field, it seems somewhat self explanatory now, a quick "Why is the server address hidden when I type it in?" in the client manual FAQ should *hopefully* clear up most questions.

If not, another section below this with some message conveying *streamer mode enabled therefore input is hidden* should take care of all but the most unobservant of people 😛 
image

blattersturm:
Hi, we'll review and hopefully merge this soon. :)

Perhaps it might be better to pass this to a persistent convar so native code can read it too (and it follows usual setting persistence), though I don't know if `seta` is exposed to UI currently.

sharkykh:
Hi, thank you!

That's a neat idea.
My C++ knowledge is... non-existent, but I will try to do it myself if I can :)

blattersturm:
If `seta` _is_ exposed to a invokeNative call and not just `set`, that's all you'd have to do as everything would be exposed in gameService already - if not, one should follow the current convar setting logic and make one for seta, presumably.

Reference as to seta in the [docs](https://docs.fivem.net/docs/client-manual/console-commands/#seta) except archiving is implemented now and that page is outdated!

sharkykh:
Alright, should be good. I can't build FiveM on my machine, so I tested everything but invoking the native (which is kind of ironic).
I used `seta ui_streamerMode true|false` on the NUI console and it updated in real time, *should* be good.
I'll try the EXE from the AppVeyor build artifacts if that works...

JimmisAtha:
Is it possible to get FiveM on PlayStation???

technetium-cfx:
No, FiveM and RedM are for PC only. Please don't ask such questions on GitHub in the future, we have a wonderful community on the [forums](https://forum.cfx.re/) and [Discord](https://discord.gg/fivem) 🙂

PR #517, opened 2020-08-18
Fix for bigger servers

dutchplayers:
This allowes servers to use more vehicle resources.

Some servers are for high end or friends only.

### DIFF
--- a/data/client/citizen/common/data/gameconfig.xml
+++ b/data/client/citizen/common/data/gameconfig.xml
@@ -474,7 +474,7 @@
 						</Item>
         					<Item>
 							<PoolName>HandlingData</PoolName>
-							<PoolSize value="6000"/>
+							<PoolSize value="8000"/>
 						</Item>
 						<Item>
 							<!-- Size of the CWaypointRecording streaming module. -->


smallo92:
You honestly don't need more vehicles then what the game already allows. You are only going to cause more and more issues for your players by doing so. For example the "city bug" or "texture loss" that everyone complains about is directly caused by your monstrous amount of custom vehicles.

dutchplayers:
We want to make our server how we like it. The solution is already there and our users do edit their own client sided files while you can do it for everybody. If your server causes city bug for players, it's your problem. So if we like to allow high end  gaming only its up  to us.

It's just a kind request since it does not cause any crashes or whatever. 

Yes we have a lot resources, 430 now. It's working fine for us and we want to keep expanding since we still hasve 60fps.

DamonOnYT:
This isn't a "fix". This is just you wanting players with less expensive PC's to lag more by changing 1 number.

izio38:
Thanks for the big fix !

... `uhm` `uhm`

dutchplayers:
You are toxic. 

It is a fix for bigger servers.

dutchplayers:
We decide how we make our server you dont decide that. Go play on your own server and leave us alone.

We want to ask de dev to support bigger servers. Thank you.

dutchplayers:
> 
> 
> This isn't a "fix". This is just you wanting players with less expensive PC's to lag more by changing 1 number. Just because you have a good PC doesn't mean everyone else who plays FiveM does.

Dear if your server uses 2000 handling data it wont give lag to players if limit is 8000. :')

dutchplayers:
Or just work harder and buy better pc

technetium-cfx:
> You are toxic.
No one is being toxic. All responses so far have been constructive and descriptive. Calling people out because you are not getting the response you wanted is however, by definition, toxic behavior.

> We decide how we make our server you dont decide that. Go play on your own server and leave us alone.
You are countering your own argument here. You cannot expect us to merge a change that affects _every_ server if only you are having issues. There are good reasons not to increase the limit, and getting a better PC is not one of them. We're open to improvements, but not open to changes that only suit one server.

PR #519, opened 2020-08-20
add: onesync_clearPeds

arthur-rl:
Took onesync_clearArea and changed it for peds

### DIFF
--- a/code/components/citizen-server-impl/src/state/ServerGameState.cpp
+++ b/code/components/citizen-server-impl/src/state/ServerGameState.cpp
@@ -432,7 +432,7 @@ struct SyncCommandState
 		maybeFlushBuffer = {};
 		frameIndex = 0;
 		client = {};
-		hadTime = false;
+		hadTime = false
 	}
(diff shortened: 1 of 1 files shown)


jaymo1011:
Is this something that **needs** to be handled by a built-in command? 

This seems like more of a job for scripting as [`GET_ALL_PEDS`](https://github.com/citizenfx/fivem/blob/637237e69bd73c154a8106a4f26af106748832d1/code/components/citizen-server-impl/src/state/ServerGameState_Scripting.cpp#L584) exists on the server now.

Korioz:
I personally think that's completely useless and can be scripted by people's themselves 🤷

Local9:
Can't `ClearAreaOfPeds` already do this? Or is this something else?

https://runtime.fivem.net/doc/natives/?_0xBE31FD6CE464AC59

blattersturm:
Not entirely sure what this tries to accomplish - this can be trivially handled by script, hopefully.

PR #521, opened 2020-08-27
Create Statebags abstraction for JS

Matspyder51:
A whole abstraction for using State Bags in Javascript:

Server:
```js
GlobalState.set('test', 'Global Test');
GlobalState.set('test2', 'Global Test 2', true);

onNet('someRandomEvent', () => {
    const entity = new Citizen.Entity(GetVehiclePedIsIn(GetPlayerPed(source)));
    // Player extend from Entity so all methods are available, only the constructor is different
    const player = new Citizen.Player(source);

    entity.setMeta('meta', 'Server Side Meta'); // will set a meta available serverside through all resources
    entity.setReplicatedMeta('meta', 'Replicated Meta'); // will set a replicated state bag

    console.log(entity.getMeta('meta')); // Server Side Meta
    console.log(entity.getReplicatedMeta('meta')); // Replicated Meta
}
```

Client:
```js
console.log(GlobalState.get('test')); // null
console.log(GlobalState.get('test2')); // Global Test 2

const me = new Citizen.Player();
const meToo = new Citizen.Player(-1);
const theOtherOne = new Citizen.Player(YourGetClosestPlayerFunction());
const veh = new Citizen.Entity(GetVehiclePedIsIn(PlayerPedId()));

veh.setMeta('meta', 'client value'); // will set a meta available clientside through all resources

console.log(veh.getMeta('meta')); // client value
console.log(veh.getReplicatedMeta('meta')); // Replicated Meta (Setted before by the server)
```

Meta & Replicated Meta can have the same key, their automaticly prefixed to prevent overriding

I use the "beta" of private properties of javascript, that's why you can see `#bag` in Entity class, that prevent get/set from any other class than Entity, but since nothing exist for 'protected' in JS, I use a small trick by using instanceof inside the Entity class to automaticly set the good prefix (entity/player) to the bag name

I gave access to Entity & Player trough Citizen namespace to prevent any conflicts with a lot of scripts which are already using Entity/Player in their code

Added definitions to index.d.ts, since an interface can't declare a class, I changed to export Citizen as a namespace and not a global var, it will not change anything for the runtime, shouldn't too for Typescript auto completion

### DIFF
--- a/data/shared/citizen/scripting/v8/index.d.ts
+++ b/data/shared/citizen/scripting/v8/index.d.ts
@@ -1,4 +1,4 @@
-/// <reference path="./natives_universal.d.ts"/>
+/// <reference path="./natives_server.d.ts"/>
 
 interface IntPtrInitialized {}
 interface FloatPtrInitialized {}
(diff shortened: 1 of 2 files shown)


blattersturm:
Sorry for taking so long to review but - this is a tad inconsistent with other APIs. 😕 

Ideally, this would use a proxy object so the API is similar to the Lua one.

lfshr:
@blattersturm I believe some of this has been done now with #569 covering the second half. Could probably close this one off.

PR #526, opened 2020-09-03
feat(onesync): custom distance culling radius for entities

PichotM:


PichotM:
it does basically the same thing as https://media.giphy.com/media/loLPei2rdrMqotZ4Jd/giphy.gif
but instead of setting a custom radius for a player, we set it for a specific entity
It is relatively useful for a player that needs to control an entity outside its scope

zirconiium:
big vouch for this PR. It's an absolute must-have rn for 1sync/infinity, i personnally need it for my black-hole event but whatever, i hope this will be pushed. nice pr.

ShadowOfVoid:
+1 really liked the idea and i think its really needed !

code/components/citizen-server-impl/include/state/ServerGameState.h
```diff
@@ -407,6 +408,12 @@ struct SyncEntityState
 
 	virtual ~SyncEntityState();
 
+	inline float GetDistanceCullingRadius()
+	{
+		// #TODO1S: figure out a good value for this
+		return overrideCullingRadius != NULL ? (overrideCullingRadius * overrideCullingRadius) : (424.0f * 424.0f);
```
blattersturm, line 414: `!= 0.0f`, maybe pre-multiply on set?

code/components/citizen-server-impl/include/state/ServerGameState.h
```diff
@@ -383,6 +383,7 @@ struct SyncEntityState
 	uint64_t lastFrameIndex;
 	uint16_t uniqifier;
 	uint32_t creationToken;
+	float overrideCullingRadius;
```
blattersturm, line 386: `= 0.0f;`. Without any default set either here or in the constructor it'll be uninitialized memory, which could be _anything_!

code/components/citizen-server-impl/src/state/ServerGameState_Scripting.cpp
```diff
@@ -1093,4 +1093,23 @@ static InitFunction initFunction([]()
...
+		{
+			float radius = context.GetArgument<float>(1);
+
+			if (radius > 0.0f)
+			{
+				entity->overrideCullingRadius = radius;
+			}
+			else
+			{
+				entity->overrideCullingRadius = NULL;
```
blattersturm, line 1109: again, NULL is for a pointer (and won't exist on Linux), `0.0f` is more likely and by then will actually not need an extra `if` (unless you're trying to keep the 'negative float' part).

blattersturm (changes requested):
Memory model stuff.

PichotM:
Fixed

blattersturm (approved):

PR #527, opened 2020-09-06
fix(net/five): update 2060 CAlterWantedLevelEvent pool check patterns

PichotM:


PichotM:
i think these patterns are fine, but I would recommend making sure it is

PR #528, opened 2020-09-08
fix(native-decls): fix typo in SetEntityDistanceCullingRadius

PichotM:
dustance?


blattersturm:
are you sure the code doesn't need to be changed to name the native cullung everywhere 😛

PR #531, opened 2020-09-11
Patch fly through windscreen in netgames + extra natives

Disquse:
Preview: https://gfycat.com/tautgenerousfiddlercrab
Changed params: https://gfycat.com/lametemptingiberiannase
Can be disabled using `SET_PED_CONFIG_FLAG` and 32 flag set to false.

Natives:
```cpp
BOOL SET_FLY_THROUGH_WINDSCREEN_PARAMS(float vehMinSpeed, float unkMinSpeed, float unkModifier, float minDamage);
void RESET_FLY_THROUGH_WINDSCREEN_PARAMS();
```


JamesV455455:
This would be an amazing addon, +1.

nikez:
With the risk of annoying bubble; this works and would be a solid addition! (tested on 3017 w/ custom client)

PR #532, opened 2020-09-12
onesync: add CPlayerCreationDataNode

Korioz:
Maybe it's faster to grab model from `CPlayerCreationDataNode` rather `CPlayerAppearanceDataNode`.
I'm still a real beginner so tell me if i've done this pr wrong.

There's a lot of other's data in `CPlayerCreationDataNode` i'm actually researching on this to improve the node parser.


csx47:
ok

PR #535, opened 2020-09-20
Added new useful Ped, Vector, and Color Methods

manups4e:
The latest PR was a disaster, this should work

### DIFF
--- a/code/client/clrcore/External/Ped.cs
+++ b/code/client/clrcore/External/Ped.cs
@@ -1599,5 +1599,20 @@ namespace CitizenFX.Core
 		{
 			return !ReferenceEquals(ped, null) && ped.Exists();
 		}
+
+		/// <summary>
+		/// Get the first entity in front of the relative Ped 
+		/// </summary>
+		/// <param name="ped">The <see cref="Ped"/> the Raycast will start from</param>
+		/// <param name="distance">Max distance of the Raycast</param>
+		/// <returns>Returns the first Entity encountered in a distance specified</returns>
+		public static Entity GetEntityInFrontOfPed(this Ped ped, float distance = 5f)
+		{
+			RaycastResult raycast = World.Raycast(ped.Position, ped.GetOffsetPosition(new Vector3(0f, distance, 0f)), IntersectOptions.Everything);
+			if (raycast.DitHitEntity)
+				return (Entity)raycast.HitEntity;
+			return null;
+		}
+
 	}
 }
--- a/code/client/clrcore/External/Ped.cs
+++ b/code/client/clrcore/External/Ped.cs
@@ -1603,12 +1603,11 @@ namespace CitizenFX.Core
 		/// <summary>
 		/// Get the first entity in front of the relative Ped 
 		/// </summary>
-		/// <param name="ped">The <see cref="Ped"/> the Raycast will start from</param>
 		/// <param name="distance">Max distance of the Raycast</param>
 		/// <returns>Returns the first Entity encountered in a distance specified</returns>
-		public static Entity GetEntityInFrontOfPed(this Ped ped, float distance = 5f)
+		public Entity GetEntityInFront(float distance = 5f)
 		{
-			RaycastResult raycast = World.Raycast(ped.Position, ped.GetOffsetPosition(new Vector3(0f, distance, 0f)), IntersectOptions.Everything);
+			RaycastResult raycast = World.Raycast(Position, GetOffsetPosition(new Vector3(0f, distance, 0f)), IntersectOptions.Everything);
 			if (raycast.DitHitEntity)
 				return (Entity)raycast.HitEntity;
 			return null;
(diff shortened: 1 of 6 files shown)


code/client/clrcore/Math/Vector2.cs
```diff
@@ -1514,5 +1514,26 @@ public override bool Equals(object value)
...
+			{
+				return new float[] { X, Y };
+			}
+			catch (Exception ex)
+			{
+				Debug.WriteLine($"ToArray exception: {ex.Data}");
+			}
+			return null;
+		}
+
```
carmineos: what's the point of using the try-catch block?

code/client/clrcore/Math/Vector3.cs
```diff
@@ -1802,5 +1802,26 @@ public float DistanceToSquared(Vector3 other)
...
+			try
+			{
+				return new float[] { X, Y, Z };
+			}
+			catch (Exception ex)
+			{
+				Debug.WriteLine($"ToArray exception: {ex.Data}");
+			}
+			return null;
+		}
```
carmineos: what's the point of using the try-catch block?

code/client/clrcore/Math/Vector4.cs
```diff
@@ -1442,5 +1442,26 @@ public override bool Equals(object value)
...
+			{
+				return new float[] { X, Y, Z, W };
+			}
+			catch (Exception ex)
+			{
+				Debug.WriteLine($"ToArray exception: {ex.Data}");
+			}
+			return null;
+		}
+
```
carmineos: what's the point of using the try-catch block?
Here would make sense because of the `static` method, but Vector4 is a struct and can't be null.

code/client/clrcore/External/Ped.cs
```diff
@@ -1599,5 +1599,109 @@ public static bool Exists(Ped ped)
...
+				Vector3 vehcoords = v.Position;
+				float distance = World.GetDistance(vehcoords, Coords);
+				if (closestDistance == -1 || closestDistance > distance)
+				{
+					closestVehicle = v;
+					closestDistance = distance;
+				}
+			}
+			return new Tuple<float, Vehicle>(closestDistance, closestVehicle);
+		}
```
carmineos, line 1682:
```cs
public Tuple<float, Vehicle> GetClosestVehicle(Vector3 Coords)
{
    Vehicle[] vehs = World.GetAllVehicles();
    float closestDistance = Single.MaxValue;
    
    Vehicle closestVehicle = null;
    foreach (Vehicle v in vehs)
    {
        float distanceSquared = Vector3.DistanceSquared(v.Position, Coords);
        
        if (distanceSquared < closestDistance)
        {
            closestVehicle = v;
            closestDistance = distanceSquared;
        }
    }
    return new Tuple<float, Vehicle>((float)Math.Sqrt(closestDistance), closestVehicle);
}
```
Avoid equality comparisons on floats like that, anyway you can compare distance using squared ones which should be faster.

carmineos: Once added this, you can avoid writing duplicated code for the version which uses the Ped.Position property instead and just use
```cs
public Tuple<float, Vehicle> GetClosestVehicle()
{
    return GetClosestVehicle(Position);
}
```

carmineos: And then do the same for the `GetClosestPed` methods.

carmineos (changes requested):

code/client/clrcore/Math/Vector4.cs
manups4e: i've removed the ToArray natives as they're actually already in the API.. i didn't see it 😅

code/client/clrcore/External/Ped.cs
manups4e: i see 🤔 i'll do it now

code/client/clrcore/Math/Color.cs
```diff
@@ -397,5 +398,20 @@ private static void CheckARGBValues (int alpha,int red,int green,int blue)
 		}
 
         public static Color Transparent => Color.FromArgb(0);
+
+		public Color ToColor(string color)
+		{
+			try
+			{
+				return Color.FromArgb(int.Parse(color.Replace("#", ""),
+							 System.Globalization.NumberStyles.AllowHexSpecifier));
```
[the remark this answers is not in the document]
manups4e: i see.. i've never tried ColorTranslator.FromHtml(String).. so i made this up

code/client/clrcore/External/Ped.cs
carmineos: If you check the [`World`](https://github.com/citizenfx/fivem/blob/master/code/client/clrcore/External/World.cs) class, you'll notice similar API are already there, maybe you could add a method to the Entity class instead which will just invoke those ones.

manups4e: yeah that's actually true.. i didn't see it because it was commented out.. i'll edit that!!

PR #538, opened 2020-09-21
Added new enum for Props and new Methods for Entity class to find closest Vehicles, Peds, Props

manups4e:


manups4e:
i don't get why is the FXServer = 1 build not building correctly 😅 i actually only edited Client API

Local9:
This will be very handy, will be much nicer to have a central version instead of the horrors I've implemented.

manups4e:
i'd like to get a review about this :D

code/client/clrcore/External/PropHash.cs
```diff
@@ -0,0 +1,11456 @@
...
+		ss1_rd1_19 = -1907152414,
+		ss1_rd1_20 = 817229101,
+		ss1_rd1_21 = 520636882,
+		ss1_rd1_22 = 319369684,
+		ss1_rd1_24 = 2041413403,
+		ss1_rd1_25 = 1743903652,
+		ss1_rd1_26 = -546026801,
+		ss1_rd1_27 = -248844740,
+		ss1_rd1_28 = 2077754264,
+		ss1_rd1_29 = -19
```
Disquse, line 9188: Are you sure those are *prop* model hashes? There's lots of ped/vehicle model hashes.

manups4e: yes as most vehicles can be spawn as object (not interactable)

blattersturm:
Useful as reference for future C# API improvements.

PR #539, opened 2020-09-22
feat(statebags) "State bag" support for JS

xIAlexanderIx:
Based on lua:
```javascript
    // server
    const serverId = GetPlayers()[0];
    const ped = GetPlayerPed(serverId);

    const ply = Player(serverId);
    const veh = Entity(GetVehiclePedIsIn(ped));

    ply.state.test = { ok : 'yup'}
    veh.state.test = { ok : 'yes' }

    GlobalState.mode = 'open'
```

```javascript
    // Client
    const serverId = GetPlayerServerId(PlayerId());
    const ped = PlayerPedId();

    const ply = Player(serverId);
    const veh = Entity(GetVehiclePedIsIn(ped));

    console.log(ply.state.test); // should print { "ok": "yup" }
    console.log(veh.state.test); // should print { "ok": "yes" }
    console.log(GlobalState.mode); // should print "open"

    ply.state.set('test', { ok: 'set from client' }, true)
    veh.state.set('test', { ok: 'set from client' }, true)
```

### DIFF
--- a/data/shared/citizen/scripting/v8/main.js
+++ b/data/shared/citizen/scripting/v8/main.js
@@ -565,7 +565,7 @@ const EXT_LOCALFUNCREF = 11;
 		__ext: EXT_ENTITY,
 
 		__pack: () => {
-			return toString(NetworkGetNetworkIdFromEntity(this.__data));
+			return String(NetworkGetNetworkIdFromEntity(this.__data));
 		},
 
 		__unpack: (data, t) => {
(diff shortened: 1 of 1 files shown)


nikez:
Tested this - worked fine for me!

xIAlexanderIx:
There is no extend method for msgpack in js, so not sure if we want to keep the pack stuff
```javascript
__pack: () => {
    return String(this.__data);
},

__unpack: (data, t) => {
    const ref = Number(data);
    return new Proxy({ __data: ref }, playerTM);
}
```

PR #543, opened 2020-09-24
Simple fix for the mass-connection method

MrSentex:
With this fix mass connections from an user will be ratelimited making the method non possible with a single user.
(If someone could do the attack even with this fix probably he is stealling cfxTickets from clients of his server).
I do not have experience with C++ so if the code is wrong let me know <3.

### DIFF
--- a/code/components/citizen-server-impl/src/InitConnectMethod.cpp
+++ b/code/components/citizen-server-impl/src/InitConnectMethod.cpp
@@ -357,6 +357,8 @@ static InitFunction initFunction([]()
 			cb(json(nullptr));
 		});
 
+		instance->SetComponent(new fx::TokenRateLimiter(1.0f, 3.0f));
+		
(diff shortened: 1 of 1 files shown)


blattersturm:
Isn't the server already doing anti-replay for existing tickets?

https://github.com/citizenfx/fivem/blob/b9670a862e2b3067de3d0e792c0e49873c29da06/code/components/citizen-server-impl/src/InitConnectMethod.cpp#L113-L124

MrSentex:
Yes but still crash the server at the moment.

blattersturm:
> Yes but the method still crashing the server or making it unplayable.

Any profile/dumps for that, though? 😕 

Are you sure this change doesn't, then, even?

MrSentex:
Maybe the check is too early and can make the server over use RAM if the ticket is random each request.
I'm only trying to help if the code you specify do the same work and does it fine I'm happy with that.

PR #544, opened 2020-09-24
Prevent packet sniffing

MrSentex:
Taking advantage of the fact that the server now supports https this is a good idea for making more difficult to sniff the initConnect creds.
Also this help people to build Nginx reverse proxies to make customs ratelimits and prevent Layer 7 attacks.

### DIFF
--- a/code/components/net/src/NetLibrary.cpp
+++ b/code/components/net/src/NetLibrary.cpp
@@ -828,7 +828,7 @@ static concurrency::task<std::optional<std::string>> ResolveUrl(const std::strin
 	{
 		// same as above, we need a record
 		skyr::url_record record;
-		record.scheme = "http";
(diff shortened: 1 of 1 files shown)


blattersturm:
We might consider this albeit not likely to be by merging the PR (single-line change w/ a long comment?). Also, I believe server list joins *already* use TLS where applicable, it's just direct connection that doesn't, as you astutely noticed.

MrSentex:
Even with the reverse proxy provided when you connect through the server list  is only the client who uses TLS. The server recieve the connection in HTTP.

And yes sry about that. I'm new doing this things.

Jerem584:
Most useless things I’ve ever seen, everything that is on client side can be sniffed just by knowing how the game source code work. 

And as another guy said every server already use tls

PR #546, opened 2020-10-02
fix(scripting): don't override pre-existing native handler

PichotM:


blattersturm:
... why? What does this accomplish?

Isigar:
Just asking but if you cannot override native executors will not work, their override natives as GetResourceNum and etc so it can be detected if their inject code.

PR #547, opened 2020-10-04
Fix interior extra natives for b2060

Disquse:
TLDR: R* added some weird flag field into CInteriorProxy class struct which is moved interior inst pointer to +8 bytes (flag + 4 bytes of padding), they also removed 8 bytes of padding after inst pointer field, so other interior proxy fields wasn't affected. 

Also contains various small fixes such as:
- Fixed an issue when missing proxy->inst->mloModel caused native execution failures.
- Better structs.
- Fixed typos.

### DIFF
--- a/code/components/extra-natives-five/src/InteriorExtraNatives.cpp
+++ b/code/components/extra-natives-five/src/InteriorExtraNatives.cpp
@@ -7,6 +7,17 @@
 #include <nutsnbolts.h>
 #include <atPool.h>
 #include <DirectXMath.h>
+#include <CrossBuildRuntime.h>
+
+#define DECLARE_ACCESSOR(x) \
+	decltype(impl.m2060.x)& x()        \
+	{                       \
+		return (Is2060() ? impl.m2060.x : impl.m1604.x);   \
+	} \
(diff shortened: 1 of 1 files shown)


balika011:
Using templates would be a lot nicer solution

blattersturm:
> Using templates would be a lot nicer solution

How would you see templates used here? I'm mainly thinking of the following, but there's probably a better solution here?

```cpp
struct Blank {};

template<bool Enable, typename T>
using TypeIf = std::conditional_t<Enable, T, Blank>;

// ...
TypeIf<(Build >= 1868), uint32_t> m_field;
```

Disquse:
> 
> 
> > Using templates would be a lot nicer solution
> 
> How would you see templates used here? I'm mainly thinking of the following, but there's probably a better solution.
> 

I was thinking about it, but since there's 3 fields that needs to be changed between 2 versions of the game, I thought that in this case I will be forced to use struct for those 3 fields, so I decided to use the current way. Anyway if you want I can refactor it.

blattersturm:
Merging as is for now.

PR #549, opened 2020-10-11
Resmon and Mumble voip for RedM

Ktos93:
Mumble voip  require add change input and output device + matrix

resmon


code/components/devtools-five/src/ResourceTimeWarnings.cpp
```diff
@@ -321,6 +325,7 @@ static InitFunction initFunction([]()
 
 					ImGui::Text("%s", resource->GetName().c_str());
 					ImGui::NextColumn();
+
```
Disquse, line 328: Empty line?

code/components/devtools-five/src/ResourceTimeWarnings.cpp
```diff
@@ -369,7 +374,7 @@ static InitFunction initFunction([]()
 						}
 
 						ImGui::NextColumn();
-
+						#ifdef GTA_FIVE
```
Disquse, line 377: I guess it's better to add preprocessor stuff without tabulation.

code/components/devtools-five/src/StreamingDebug.cpp
```diff
@@ -1,3 +1,4 @@
+
```
Disquse, line 1: Why empty line in here?

code/components/devtools-five/src/StreamingDebug.cpp
```diff
@@ -655,3 +656,4 @@ static InitFunction initFunction([]()
 		ImGui::End();
 	});
 });
+
```
Disquse, line 659: And here too?

code/components/devtools-five/src/StreamingProgress.cpp
```diff
@@ -1,9 +1,9 @@
+
```
Disquse, line 1: Same for this file.

code/components/devtools-five/src/WorldEditor.cpp
```diff
@@ -5,9 +5,9 @@
 * regarding licensing.
 */
 
+
```
Disquse, line 8: And this..

code/components/gta-net-rdr3/src/MumbleVoice.cpp
```diff
@@ -0,0 +1,770 @@
+#include "StdInc.h"
```
Disquse, line 1: I guess it will be better to share codebase between RDR3 and GTA5 for such stuff. You're re-using too much stuff anyway.

Disquse (changes requested):

code/components/gta-net-rdr3/src/MumbleVoice.cpp
Ktos93: But using ifdef for every native, pattern and registered native?
I think it will make a terrible chaos in this file

blattersturm: Block along the lines of:

```cpp
#ifdef IS_RDR3
constexpr const auto patternBlah = "AA BB CC ? DD";
constexpr const auto nativeBlah = 0xDEADC0DEDEADC0DE;
#elif defined(GTA_FIVE)
constexpr const auto patternBlah = "AA BB AC ? DD";
constexpr const auto nativeBlah = 0xC0C0FEED;
#endif
```

would also work?

Disquse: You can also see `LoadStreamingFile.cpp`, it shares lots of code between games.

Ktos93: gta-five-net MumbleVoice.cpp updated and symlink created

code/components/gta-net-five/src/MumbleVoice.cpp
```diff
@@ -898,34 +984,43 @@ static HookFunction hookFunction([]()
 				g_mumbleClient->SetChannel("Root");
 			}
 		});
-
+#endif
```
Disquse, line 987: #elif ?

Ktos93: orginal native isnt discovered. Create new custom native?

Ktos93:
anything else to do? not to mention looking for other patterns for device and matrix

Disquse (changes requested):
`MumbleVoice_BindNetLibrary` is the most important part to make Mumble work, but for some reason you ifdef'ed it do work only for V. So Mumble won't work without it, not even talking about crashes due to `g_netLibrary` is being nullptr. Please add `MumbleVoice_BindNetLibrary` and call this function after `g_netLibrary = NetLibrary::Create();` in `gta-net-rdr3`

Disquse:
Also an advice for future PRs: please don't merge 2 absolutely not related to each other things in 1 PR.

PR #553, opened 2020-10-18
Update russian language

Vinipux322:


blattersturm:
https://translations.cfx.re/

Even though it's unmonitored lately (and perhaps broken - could @technetium-cfx maybe take a look at managing this? nah just kidding he never does anything he's supposed to do) that's not an excuse to make language PRs here.

PR #554, opened 2020-10-18
Add extra row to resmon with cumulative summary

SloPro:
Adds an extra row to the end of resmon, which displays the total sum of all resources' tick time, memory usage, and streaming usage.
Tested, seems to works fine:
image


blattersturm:
hi; likely will be rejected as there's a bit different plan for cleaning this up (moving to dear imgui tables branch, adding relative times, adding warning elements, etc.)

SloPro:
Alright, no problem. I presume using the modified devtools-five.dll on a live server will not bode well with the anticheat, correct?

blattersturm:
> Alright, no problem. I presume using the modified devtools-five.dll on a live server will not bode well with the anticheat, correct?

if it runs it'll be fine probably but not very recommended

PR #557, opened 2020-10-22
Update locale-zh_Hans.json

1107128185:
Correct error


cerium-cfx:
Translations are done via https://translations.cfx.re

PR #559, opened 2020-10-23
add GET_AUDIOCONTEXT_FOR_SERVERID as per GET_AUDIOCONTEXT_FOR_CLIENT

isingrel:
add audio context native using server player id instead of client player id as argument

### DIFF
--- a/code/components/gta-net-five/src/MumbleVoice.cpp
+++ b/code/components/gta-net-five/src/MumbleVoice.cpp
@@ -615,7 +615,19 @@ static std::shared_ptr<lab::AudioContext> getAudioContext(int playerId)
 	std::string name = fmt::sprintf("[%d] %s",
 		FxNativeInvoke::Invoke<int>(getServerId, playerId),
 		FxNativeInvoke::Invoke<const char*>(getPlayerName, playerId));
+	return g_mumbleClient->GetAudioContext(name);
+}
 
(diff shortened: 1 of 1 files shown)


ntanis-dev:
Awesome! I was about to do this when I realized the despair us - infinity mode users are in.

mickeygeecom:
How is this native working - what does it, or can it do?

ntanis-dev:
> How is this native working - what does it, or can it do?

Its like `GET_AUDIOCONTEXT_FOR_CLIENT` but instead of `Player` it requires a server ID to be used for infinity (big mode) servers where players are split into "instances" and you cannot query a player outside of the client's player instance.

mickeygeecom:
> > How is this native working - what does it, or can it do?
> 
> Its like `GET_AUDIOCONTEXT_FOR_CLIENT` but instead of `Player` it requires a server ID to be used for infinity (big mode) servers where players are split into "instances" and you cannot query a player outside of the client's player instance.

I understand that, but what does the original  `GET_AUDIOCONTEXT_FOR_CLIENT` do? 🤔 Still lacking the purpose of use.

orhanasan:
It is particularly used for adding sound filters for a voice relay from a client indicated by that server ID to the client code is working. Other than that, you can use some "hidden" natives to create filters, just search for them.

PR #569, opened 2020-10-30
feat(statebags) - add JS definitions to v8index.d.ts

nimoalr:
Add definitions for https://github.com/citizenfx/fivem/pull/539 

Not sure if I should define GlobalState type to any or not, so did not assigned a type to it.
Feel free to comment and ask me for any changes.


blattersturm:
I somewhat doubt a proxy that takes string keys can be considered an any. Maybe an interface type with a `[key: string]: any` _inside of it_ is more suitable?

See a random [SO post](https://stackoverflow.com/a/54802162) I found.

lfshr:
@blattersturm this seems to have become stale. #638 implements the interfaces.

PR #577, opened 2020-11-12
make sv_endpoints convar persistent

tayandenga:
...right now each connection resets the value set in .cfg

I am running fxserver through GRE tunnel, using sv_endpoints to provide proper endpoint address. The value is being reset each getEndpoints request with current configuration.


tayandenga:
I am running fxserver through GRE tunnel, using sv_endpoints to provide proper endpoint address. The value is being reset each getEndpoints request with current configuration.

PR #579, opened 2020-11-14
tweak: increase MaxVehicleModelInfos

rubbertoe98:
This increases the number of vehicles that can be streamed. Tested locally multiple times with an increased number of vehicles that previously caused a crash on connection, no issues found.


blattersturm:
I suspect this is not needed anymore since the dynamic allocation fix, unless that got backed out.

rubbertoe98:
Unfortunately FiveM still crashes when going over the limit, have just tested with the dynamic allocation fix.

blattersturm:
Then that's likely an entirely different issue and you should provide a repro. This value is practically unused now.

PR #581, opened 2020-11-15
feat(onesync): add startProjectile event

PichotM:


blattersturm:
Is this valid across builds?

Disquse:
@blattersturm can confirm, there's no any difference comparing 1604 and 2189.

PR #583, opened 2020-11-21
feat(server): getEndpoints scripted handler

tayandenga:
Allow to dynamically response to getEndpoints method when sv_dynamicEndpoints ConVar is set to true - define final endpoints to client depending on user preferences


blattersturm:
I don't think it's a good idea to add more events that'll invoke the ScRT from an asynchronous HTTP handler. `playerConnecting` is already problematic in this regard (and is pending a rework), and I think it might be better to add some ability to target the player's future endpoints from that callback or its replacement, instead.

tayandenga:
Indeed, I've added a ConVar to not call this method for everyone, just those, who _might want_ use this handler. This was simpliest way to skip overhead.

Right now I am suffering 50Gbps DDoS attacks on daily - some of them are easily mitigated, some diverse my IP address. I would like to run on few ports and few IPs, but also not lead players to endpoints, that are right now down. So this is result of my best idea at this moment - generating the endpoints.
So I am open to any alternatives, also waiting for the playerConnecting rework :)

blattersturm:
Apparently dead PR. If this is still required by anyone we could make another pass at this and turn it into a `need-manual-work` PR.

PR #584, opened 2020-11-22
feat(server): drop getEndpoints method, let playerConnecting NETEV serve endpoints

tayandenga:
After reviewing what @blattersturm pointed in his reply to my previous pull request, more friendly solution today.

* Got no tools to check does it build, unfortunately.


blattersturm:
This change is not compatible with current backend services.

tayandenga:
> This change is not compatible with current backend services.

Damn, what I am missing?

dayidir:
fivem failed to fetch /info.json to obtain policy metadata , Although the Fivem server is open, we cannot login due to this error code.

blattersturm:
> fivem failed to fetch /info.json to obtain policy metadata , Although the Fivem server is open, we cannot login due to this error code.

Okay, but is that related to this pull request? What are you trying to accomplish here?

dayidir:
Since we do not know the cause of the problem, we are only trying to log into the server at the moment.

dayidir:
What is the main cause of this problem? I'm just curious about this. I could only find this page on Google search.

blattersturm:
Merge conflict and apparent incompatibility.

PR #585, opened 2020-11-23
gameconfig: Increased phInstGta

PERPGamer:
Raised phInstGta from 16384 to 32768 to allow more polygons to cut down on "texture lost" or "world lost" when using non-lore cars.

### DIFF
--- a/data/client/citizen/common/data/gameconfig.xml
+++ b/data/client/citizen/common/data/gameconfig.xml
@@ -382,7 +382,7 @@
 						</Item>
 						<Item>
 							<PoolName>phInstGta</PoolName>
-							<PoolSize value="16384"/>
+							<PoolSize value="32768"/>
 						</Item>
 						<Item>
 							<PoolName>PhysicsBounds</PoolName>


smallo92:
Yeah that's not going to magically fix the issue at hand, that's just sweeping it under the rug. Also this will just make it even worse for people with lower end PCs.

jaymo1011:
The solution, is to simply lower the size of textures being used. For example, having "4k" textures on a car is down right **wasteful** as unless you're rending at at that resolution (the vast *vast* majority of people are not) you're never going to notice the detail.

Lowering the detail of a texture also reduces the time for downloading the texture from the server as there is *less* data to download and then less data to stream into the game.

No matter what way you slice it, there isn't an need to raise this limit nor does doing so outweigh the negative effects that would come from raising it.

PERPGamer:
While I never said it was going to fix it or that 4k textured cars (or other large textured cars) are going to be fixed,  lowering your car textures should already be a owners/dev first thing they do. I do understand this. Just wanted to make this pull/get it added to help out, not as a be-all fix.

blattersturm:
phInstGta has nothing to do with "cars" or "texture loss" even.

PR #586, opened 2020-11-24
Help identify native execution in script host failure cause

R0JEA:
Add resource name to the error trace to help identify the cause of error.

### DIFF
--- a/code/components/citizen-scripting-lua/src/LuaScriptRuntime.cpp
+++ b/code/components/citizen-scripting-lua/src/LuaScriptRuntime.cpp
@@ -768,7 +768,7 @@ static int Lua_InvokeFunctionReference(lua_State* L)
 		char* error = "Unknown";
 		scriptHost->GetLastErrorText(&error);
 
-		lua_pushstring(L, va("Execution of native %016x in script host failed: %s", 0xe3551879, error));
+		lua_pushstring(L, va("Execution of native %016x in script (%s) host failed: %s", 0xe3551879, GetResourceName(), error));
 		lua_error(L);
 	}
 


blattersturm:
What?

The Lua error will _already_ be handled by something that contains the resource name.

blattersturm:
Also,

1. Lua_InvokeFunctionReference has _nothing_ to do with anything you'll normally execute.
2. Lua_InvokeFunctionReference doesn't even have GetResourceName in scope. This wouldn't even build.

R0JEA:
How would you then identify the script that causes this error if there is no futher information in trace?

blattersturm:
> How would you then identify the script that causes this error if there is no futher information in trace?

There always is. Literally.

R0JEA:
I have this happening at the moment and all console outputs is "InvokeNative: execution failed: Argument at index 1 was null.
SCRIPT ERROR: Execution of native 000000002f7a49e6 in script host failed: Argument at index 1 was null."

No line, no resource name. Neither above, nor below

blattersturm:
What server build are you using? Is this even on the server?

R0JEA:
3218 win build, of course it's on the server

blattersturm:
> 3218 win build

Odd. Any repro?

> of course it's on the server

Well, could be a client console print as well?

R0JEA:
No repro, unfortunately. Don't know the cause

> Well, could be a client console print as well?
Yes, but I think 000000002f7a49e6 stands for TriggerClientEvent which wouldn't be used on client?

blattersturm:
> No repro, unfortunately. Don't know the cause

Cool.

R0JEA:
Any ideas why that _could_ happen?

blattersturm:
> Any ideas why that _could_ happen?

yeah a few actually

R0JEA:
Good luck then. It really is painful going through all the files just to find the trigger with an invalid target

blattersturm:
yeah thankfully it's not me

Rubylium:
Little repo for an error like this. Use TriggerClientEvent with the target as nil like 

`TriggerClientEvent("test", nil)`

mickeygeecom:
> I have this happening at the moment and all console outputs is "InvokeNative: execution failed: Argument at index 1 was null.
> SCRIPT ERROR: Execution of native 000000002f7a49e6 in script host failed: Argument at index 1 was null."
> 
> No line, no resource name. Neither above, nor below

Sounds like your trying to trigger a server => client event with a source that's not online anymore. Maybe due to some async behavior, and simply not checking whether or not the source/player is (still) online.

R0JEA:
Or simply a target type, I know. But it's hard to trace where this is comming from.

PR #589, opened 2020-11-25
Add CORS Support to all Endpoints

codekrafter:
Add the "Access-Control-Allow-Origin" header to all endpoints, allowing them to be accessed by web clients.

Currently this is only present on the "/profileData.json" endpoint, when there is data. This means that if it returns an empty response (Line 418) it will trigger a CORS error on all web clients. It also triggers a CORS error on all other endpoints.


Explooosion-code:
Well... Not everyone may want to set it that way. I think a server.cfg flag would be a better solution than just hardcoding it to always  *

Spajker7:
You could setup a reverse proxy between your web clients and the server.

blattersturm:
Merge conflict as this code was refactored since. Also, indeed, since usually no direct TLS is used it may be more appropriate to use a reverse proxy.

PR #591, opened 2020-12-04
Update Debug.cs | Ability to enable/disable Debug

Kolopsel:
Enables developers to easily enable/disable implemented debug messages.

### DIFF
--- a/code/client/clrcore/Debug.cs
+++ b/code/client/clrcore/Debug.cs
@@ -11,8 +11,12 @@ namespace CitizenFX.Core
     public static class Debug
     {
         [SecuritySafeCritical]
+		
+		public static bool Enabled { get; set; }
+		
         public static void Write(string data)
         {
+			if(Enabled != true) { return; }
 			var channel = "script:" + ((InternalManager.GlobalManager?.ResourceName) ?? "mono");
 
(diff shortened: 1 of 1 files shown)


blattersturm:
Not sure about this.

What's the intended use case, and why expose it as API surface and not, say, something in a lower-level subsystem?

Kolopsel:
You do have a point, however I think it would be much more pleasent to have something like this already implented in the API. Of course you only need a couple of lines here and there to add something similiar to your script, but like I said before this just makes everything regarding debugging a little bit easier and is especially great for people who want to keep their script tidy.

blattersturm:
Huh?

Also, this wouldn't compile/run properly in the first place either, the formatting is all off, and again I'm not sure why this needs to be an API rather than a more global flag.

Kolopsel:
I do not know why the formatting is suddenly off. Last time I checked, it seemed perfectly fine. Anyway now that I think about it, it would probably make more sense to have something like this as a global flag but how would you go about doing that, if I may ask?

blattersturm:
Would be solved by the issues page/warning system once implemented.

manups4e:
for this i simply used the default DEBUG symbol present when you create a new project to print only when in debug compilation

PR #593, opened 2020-12-20
fix(onesync): add missing 2189 sync nodes bits

PichotM:


blattersturm:
uh, what?

how is this mergeable? you didn't add a new Is2189() if, you _unconditionally_ read the extra bit, wtf?

Korioz:
I confirm the `CPedGameStateDataNode` bit added to the node, just checked rn in ida

PichotM:
my bad I was confused why Is2189 was missing 🙃

PR #596, opened 2020-12-26
Optimizations to Citizen.Wait

niekschoemaker:
Don't use GetGameTimer if we know we only want to wait till the next frame. This saves a lot of native calls if there's a lot of threads.

Value of 5 is choosen as a reasonable value.


niekschoemaker:
This should work, another option could be to simply check if it is not 0.

Testing with this change it appears to save quite a lot of FPS (around 10+) on a server with a lot of resources which run threads that execute every tick.

PR #598, opened 2020-12-27
Fix(prebuild): Clear LUA_CPATH

niekschoemaker:
Using a custom LUA_CPATH which doesn't search the current directory would break the prebuild script.


niekschoemaker:
This was an issue which I was having for a long time which caused me to not be able to build the client and/or server at all.

Simply clearing LUA_PATH and LUA_CPATH avoids this issue entirely.

PR #599, opened 2020-12-27
add example that also fixes infinite recursion.

AvarianKnight:


blattersturm:
Should be taken in with the native review.

PR #600, opened 2020-12-28
fix: GET_GAME_TIMER

NeQYT:
At this moment, if we call a function without using a thread, wait function totally not work. After check, time must always be obtained, even if there is no thread.

### DIFF
--- a/data/shared/citizen/scripting/lua/scheduler.lua
+++ b/data/shared/citizen/scripting/lua/scheduler.lua
@@ -230,13 +230,14 @@ end
 SetTimeout = Citizen.SetTimeout
 
 Citizen.SetTickRoutine(function()
+	curTime = GetGameTimer()
+
 	if not hadThread then
 		return
 	end
 
 	-- flag to skip thread exec if we don't have any
 	local thisHadThread = false
-	curTime = GetGameTimer()
 
 	for coro, thread in pairs(newThreads) do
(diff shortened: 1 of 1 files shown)


blattersturm:
That doesn't make much sense. Wait only works within a thread context, do you have _any_ example of a script that behaves wrongly with the change?

NeQYT:
> That doesn't make much sense. Wait only works within a thread context, do you have _any_ example of a script that behaves wrongly with the change?

Problem occurs only, when script doesn't have any thread. Example:

```lua

-- Server:

RegisterCommand('test', function(source, args)
    TriggerClientEvent('test:test', source)
end)

-- Client:

RegisterNetEvent('test:test')
AddEventHandler('test:test', function()
    print('start')
    Citizen.Wait(10000) -- * Wait was skipped, 2 values were printed immediately. 
    print('end')
end)

```

NeQYT:
And example with a working wait:

```lua
-- Server:

RegisterCommand('test', function(source, args)
    TriggerClientEvent('test:test', source)
end)

-- Client:

    -- do anything
end)

RegisterNetEvent('test:test')
AddEventHandler('test:test', function()
    print('start')
    Citizen.Wait(10000) -- * Wait was skipped, 2 values were printed immediately. 
    print('end')
end)

```

PR #602, opened 2021-01-02
Removing possible XSS injection on names

MrSentex:
This is not FiveM fault and any resource who FiveM has uploaded is vulnerable to xss but community scripts like scoreboards or esx_spectate do not escape names correctly so attackers can inject everything they want to all clients on a server.

### DIFF
--- a/code/components/citizen-server-impl/src/InitConnectMethod.cpp
+++ b/code/components/citizen-server-impl/src/InitConnectMethod.cpp
@@ -444,6 +444,11 @@ static InitFunction initFunction([]()
 				name = name.substr(0, 200);
 			}
 
+			if (name.find("<") != std::string::npos || name.find(">") != std::string::npos) {
+				sendError("Your name contains illegal characters. Please remove '<' and/or '>' from your name and try again.\n\nWhy I can't have '<' or '>' on my name?\nSometimes names are not escaped correctly on community resources so servers can be affected by XSS attacks injecting javascript code on clients.");
+				return;
+			}
+			
 			TicketData ticketData;
 
 			if (!lanVar->GetValue())


blattersturm:
No. < and > are perfectly valid characters in names, and users should not be forced to have 'limited' names (or change their Steam/in-game name for stupid reasons).

MrSentex:
People are abusing changing their names to <scrtipt src="blabla"> and injecting porn images even child porn.
Thats is disgusting and not all servers are able to fix this issue so I thougth it was a good idea to introduce a block for that.

MrSentex:
well github block `<script src="blabla">` hehe

FlawwsX:
Couldn't you make your own script that rejects players with those symbols?

FlawwsX:
Couldn't you make your own script that rejects players with those symbols?

blattersturm:
> and not all servers are able to fix this issue

Then maybe those people should not run a server. All servers are 'able to fix this issue'. Nothing is preventing anyone from doing so.

technetium-cfx:
Resource authors should sanitize any form of input, including player names. It shouldn't matter what the source of the input is. Preventing certain characters from being used does not fix the underlying issue. What if the same resource author has a different input that's not a player name, but still vulnerable to any type of exploit? Long story short, report the issue to the resource author 🙂

MrSentex:
I already did but I can't reach every server. I think the same way @blattersturm but people just want to have fun, maybe if we set a variable who is by default true to block those chars?

technetium-cfx:
The issue remains that the resource author is responsible. They need to understand they are responsible for sanitizing input. If it's not player names, it'll be something else. 

Imagine it was some sort of chat resource, and people started embedding HTML in their messages. Would you ask FiveM to disallow any chat input with certain characters, or should the chat resource be fixed so characters are sanitized properly?

MrSentex:
I understand your position but even if authors update their resource many server will not update. So the problem is there a hole community is injecting shitty images on servers and asking for money. You could ban them  easily too.

blattersturm:
And would people update their server binaries, even, by that train of thought?

MrSentex:
Modify GetPlayerName too and autoescape the name

MrSentex:
This will help players and devs.

technetium-cfx:
You're missing the point. It will not help players or devs, it will simply ignore the underlying issue. The characters themselves are perfectly legal. The fact that resource authors don't do anything to sanitize input is the issue. As said, if it's not for player names, it'll be for something else. This needs to be addressed on that level, not give them more reason to ignore the real issue.

AvarianKnight:
> I understand your position but even if authors update their resource many server will not update. So the problem is there a hole community is injecting shitty images on servers and asking for money. You could ban them easily too.

The primary issue with this is why is it the Collectives problem when this is virtually the same thing as allow players direct access to the MySQL and people SQL Inject? They're using something that is known not to be escaped as an escaped variable. That's on them, they need to fix their code, forcing a 'mass fix' onto everyone won't fix the issues.

MrSentex:
But FiveM could help people too. You don't need to recompile or make any changes but sending a message with a snipped who block xss will help.

AvarianKnight:
> But FiveM could help people too. You don't need to recompile or make any changes but sending a message with a snipped who block xss will help.

You're missing the entire point, why would FiveM fix a problem that's not theirs to begin with.

MrSentex:
And you could ban players who are doing this too.

MrSentex:
I understand your point and is why I'm saying FiveM could help sending an alert to users talking about this problem and why is not FiveM fault. If players see that they will know FiveM is caring about them even when is not their fault.

`AddEventHandler("playerConnecting", function (name, kick, defer)
    defer.defer()

    Citizen.Wait(50)

	if (name:find("<") ~= nil or name:find(">") ~= nil) then
		defer.done("Your name is not allowed on this server. You have illegals caracters on it. Remove '<' and/or '>' an try again.\n")
		return
	end

    defer.done()
end)`

AvarianKnight:
> I understand your point and is why I'm saying FiveM could help sending an alert to users talking about this problem and why is not FiveM fault. If players see that they will know FiveM is caring about them even when is not their fault.
> 
> `AddEventHandler("playerConnecting", function (name, kick, defer)
> defer.defer()
> 
> ```
> Citizen.Wait(50)
> 
> if (name:find("<") ~= nil or name:find(">") ~= nil) then
> 	defer.done("Your name is not allowed on this server. You have illegals caracters on it. Remove '<' and/or '>' an try again.\n")
> 	return
> end
> 
> defer.done()
> ```
> 
> end)`

Cool, but everyone has previously said that the characters are perfectly legal, it's how the players are not sanitizing them that is the problem.

MrSentex:
I know they are legal. But you are giving a chance to server owner to patch the injection if they want to block the characters or if they are suffering the attack.

technetium-cfx:
We aren't giving anyone the chance. Resource authors do. Are you going to blame Google for making XSS attacks possible in Chrome too? No, websites have to make their site secure. I'm closing this.

d0p3t:
Blocking the characters isn't "patch the injection"

PR #603, opened 2021-01-06
Fix codegen for TS for natives that return values by pointers

DeForce:
This fixes native typing TS documentation:
Sample natives that has this issue:

https://runtime.fivem.net/doc/natives/?_0x2975C866E6713290
GetEntityPlayerIsFreeAimingAt
```
# C Declaration:
// GetEntityPlayerIsFreeAimingAt
BOOL GET_ENTITY_PLAYER_IS_FREE_AIMING_AT(Player player, Entity* entity);

# TS Declaration
/**
 * Returns TRUE if it found an entity in your crosshair within range of your weapon. Assigns the handle of the target to the *entity that you pass it.
 * Returns false if no entity found.
 */
declare function GetEntityPlayerIsFreeAimingAt(player: number, entity: number): number;
```

Other examples, just runtime links:
https://runtime.fivem.net/doc/natives/?_0x2E1202248937775C
https://runtime.fivem.net/doc/natives/?_0xDC16122C7A20C933

As you can see natives that has pointers in the declaration, are supposed to be return values, at least for natives that I saw.

The proposed fix fixes natives function/documentation generation so that it makes first pointers returnable.

Fixed Typing:
```
# TS Declaration
declare function GetEntityPlayerIsFreeAimingAt(player: number): [number, number];
```

This is only documentation generation fix, from what I saw - the function actually return array, so it's only documentation issue.

I am not sure about Lua/JS, if they have the same issue or not, but TS would be correctly typed after this PR. At least for the natives that I found have issues.

### DIFF
--- a/ext/natives/codegen_out_dts.lua
+++ b/ext/natives/codegen_out_dts.lua
@@ -157,11 +157,7 @@ local function printArgument(argument, native)
 
 	if argument.pointer then
 		if argument.type.nativeType == 'int' or argument.type.nativeType == 'float' then
-			if isSinglePointerNative(native) then
-				argType = 'number'
-			else
-				retType = 'number'
-			end
+			retType = 'number'
 		elseif argument.type.nativeType == 'vector3' then
 			retType = 'number[]'
 		else
--- a/ext/natives/codegen_out_dts.lua
+++ b/ext/natives/codegen_out_dts.lua
@@ -157,7 +157,12 @@ local function printArgument(argument, native)
 
 	if argument.pointer then
 		if argument.type.nativeType == 'int' or argument.type.nativeType == 'float' then
-			retType = 'number'
+			if isSinglePointerNative(native) then
+				argType = 'number'
+				retType = 'pointer'
+			else
+				retType = 'number'
+			end
 		elseif argument.type.nativeType == 'vector3' then
 			retType = 'number[]'
 		else
@@ -190,6 +195,8 @@ end
 local function formatDefinition(native)
 	local argsDefs = {}
 	local retTypes = {}
+	local markOptional = {}
+	local castUnknown = false -- We are not able to know whether a pointer is a return value or input, thus casting unknown
 
 	if native.returns then
 		table.insert(retTypes, printReturnType(native.returns))
@@ -201,6 +208,11 @@ local function formatDefinition(native)
 
 			if argType ~= nil then
 				table.insert(argsDefs, argumentName .. ': ' .. argType)
+				if retType == 'pointer' then markOptional[#argsDefs] = argumentName .. '?: ' .. argType end
+			end
+
+			if retType == 'pointer' then
+			    castUnknown = true
 			elseif retType ~= nil then
 				table.insert(retTypes, retType)
 			end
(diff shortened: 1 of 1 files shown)


blattersturm:
This breaks codegen for functions such as DeleteEntity:

```ts
declare function DeleteEntity(): number;
```

DeForce:
Okay, back to planning board then :)

blattersturm:
I suspect this might be hard to correctly implement due to there being no explicit data for `in`/`out`/`inout` arguments like such, which might be why the initial change was a tad weird.

DeForce:
I see two options that we can go for:
1. Make return to be unknown[]
```ts
declare function GetAmmoInClip(ped: number, weaponHash: string | number, ammo: number): unknown[];
```

That will make at least typescript not yell at you for wrong typing, because the typing is technically actually unknown.
Also could make the pointer a optional value:
```ts
declare function GetAmmoInClip(ped: number, weaponHash: string | number, ammo?: number): unknown[];
```

2.
Generate two or more functions for each pointer:
```ts
declare function GetAmmoInClip(ped: number, weaponHash: string | number): [number, number];
declare function GetAmmoInClip(ped: number, weaponHash: string | number, ammo:number): number;
```

That would solve the problem by making generation logic a little bit harder. But at least there would be correct typings, but one of the functions would be wrong, which will lead to ambiguity. At least it will have an option to return correctly with second one.

@blattersturm thoughs?

blattersturm:
At this time, optional values seem like the better option. It might be confusing for, say, DeleteEntity and break natives where a pointer _isn't_ the last argument, but until someone makes a pass over inout annotations that's probably the safest bet.

DeForce:
@blattersturm 
PR has been updated to what we have talked about above.
I have added a comment/warning, that is not really part of the doc-string, I hope that this is acceptable.
Actually, functions that have only one input parameter and doesn't have a set return value will be fine with what I implemented:
```ts
/**
 * Deletes the specified entity, then sets the handle pointed to by the pointer to NULL.
 * @param entity The entity to delete.
 */
declare function DeleteEntity(entity: number): void;


...
 */
declare function GetEntityPopulationType(entity: number): number;


// Return is unknown[] due to pointer value being input of the function
declare function GetAmmoInClip(ped: number, weaponHash: string | number, ammo?: number): unknown[];

/**
 * Returns TRUE if it found an entity in your crosshair within range of your weapon. Assigns the handle of the target to the *entity that you pass it.
 * Returns false if no entity found.
 */
// Return is unknown[] due to pointer value being input of the function
declare function GetEntityPlayerIsFreeAimingAt(player: number, entity?: number): unknown[];
```

gottfriedleibniz:
Another edge-case would be where a pointer isn't the only or last argument, e.g., `0xBBE7648349B49BE8`. Just ensure your changes don't clobber native(s) of that form.

DeForce:
> Another edge-case would be where a pointer isn't the only or last argument, e.g., `0xBBE7648349B49BE8`. Just ensure your changes don't clobber native(s) of that form.

0xBBE7648349B49BE8 is the same in my result and current.

PR #604, opened 2021-01-06
fix(native-decls): don't show return value twice

AvarianKnight:


blattersturm:
please, use proper category names (e.g. `fix(native-decls)` or so).

you can rebase -i and force-push to your branch usually

AvarianKnight:
Fixed

PR #609, opened 2021-01-09
Update SetMillisecondsPerGameMinute.md

JoshYouJust:


blattersturm:
I don't think the doc tooling will take in any content placed here.

ElPumpo:
You could also add that setting `NetworkOverrideClockTime` also resets this value to 2000

blattersturm:
Abandoned PR, will be taken in with the big native review.

PR #611, opened 2021-01-10
tool for anticheats

chrisskan:
Its a non local table called "acthreads" that stores all data that the "threads" table has. This table can be used from anticheats, if calculated properly


blattersturm:
what

PR #613, opened 2021-01-14
fix(clr): use GenAPI and GenFacades from .NET Arcade

tie:
This PR should fix the failing CI build by using GenAPI and GenFacades from [.NET Arcade](https://github.com/dotnet/arcade). It appears that old GenAPI and GenFacade packages were removed from [DotNet BuildTools at myget.org](https://dotnet.myget.org/gallery/dotnet-buildtools).

- GenAPI works with minor command line options syntax convention change.
- GenFacades requires a bit of XML programming… uh, build config? Anyway, I couldn’t figure out the right combination of options for shipped .targets so I just set up the task manually.

>…\GenAPI.targets(15,3): error MSB4019: The imported project "…\Microsoft.DotNet.BuildTools.GenAPI.targets" was not found. Confirm that the expression in the Import declaration "…\Microsoft.DotNet.BuildTools.GenAPI.targets" is correct, and that the file exists on disk. […\CitiMono.csproj]
>https://ci.appveyor.com/project/citizenfx/fivem/builds/37259981/job/sqs2a2n2t0ev8mb5


tie:
That’s weird, something isn’t working as expected on the CI machine.

tie:
OK, I can reproduce it locally.

tie:
Passing `-p:GenerateReferenceAssemblySource=true` seems to work.

tie:
Builds are passing, neat! Should be ready for review.

PR #614, opened 2021-01-15
fix: change GetMemoryUsage / Resource Monitor usage to KiB to avoid overflow

kn0pee:
This change was made to avoid the resource monitor displaying a negative value when a script has extreme memory usage. This was caused by the int64 overflowing.

It implements the following changes
- GetMemoryUsage function returns a value in kilobytes rather than bytes
- Resource monitor displays KiB as the lowest value, without decimal points
- Resource time warning logic changed to compensate for the new value

This was tested on a custom FX server with a C# and lua script, and it seemed to be working properly. If this should be implemented differently, or I have missed something please let me know. I looked through the repo to see if anything else relied on GetMemoryUsage, and personally couldn't find anything.

Note: As far as I know, JS scripting does not implement GetMemoryUsage, and therefore this was not changed.

### DIFF
--- a/code/components/citizen-devtools/src/ResourceTimeWarnings.cpp
+++ b/code/components/citizen-devtools/src/ResourceTimeWarnings.cpp
@@ -112,12 +112,12 @@ static ImVec4 GetColorForRange(float min, float max, float num)
 	}
 }
 
-static int64_t GetTotalBytes(const fwRefContainer<fx::Resource>& resource)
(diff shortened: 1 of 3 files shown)


kn0pee:
I thought about it some more, and there is no way I am hitting the max limit of int64.

Would be grateful for any advice on what could even cause that :O

blattersturm:
I'll merge this anyway since there doesn't seem to be any crazy obvious floating point math.

PR #615, opened 2021-01-15
feat(glue): forced streamer mode using `fivem://`

sharkykh:
Force streamer mode value per-session using:
`fivem://connect/127.0.0.1?streamerMode=true` (or `=1`)
Does **not** set the archived variable `ui_streamerMode`.

**Currently only supports FiveM 'cold start'.**
(calling fivem:// when FiveM is closed)

I am not sure how to correctly (and neatly) pass the `streamerMode` value [through the socket to the running FiveM instance](https://github.com/citizenfx/fivem/blob/4fb12d066857851f9619281e2eba7b0271fd0b38/code/components/glue/src/ConnectToNative.cpp#L1196).
Would appreciate guidance if anyone is up for it.

### DIFF
--- a/code/components/glue/src/ConnectToNative.cpp
+++ b/code/components/glue/src/ConnectToNative.cpp
@@ -1118,9 +1118,12 @@ void Component_RunPreInit()
 			nng_socket socket;
 			nng_dialer dialer;
 
+			auto j = nlohmann::json::object({ { "host", connectHost }, { "params", connectParams } });
+			std::string connectMsg = j.dump(-1, ' ', false, nlohmann::detail::error_handler_t::strict);
+
 			nng_push0_open(&socket);
 			nng_dial(socket, "ipc:///tmp/fivem_connect", &dialer, 0);
-			nng_send(socket, const_cast<char*>(connectHost.c_str()), connectHost.size(), 0);
+			nng_send(socket, const_cast<char*>(connectMsg.c_str()), connectMsg.size(), 0);
 
 			if (!hostData->gamePid)
 			{
@@ -1204,7 +1207,8 @@ static InitFunction connectInitFunction([]()
 			std::string connectMsg(buffer, buffer + bufLen);
 			nng_free(buffer, bufLen);
 
-			ConnectTo(connectMsg);
+			auto connectData = nlohmann::json::parse(connectMsg);
+			ConnectTo(connectData["host"], false, connectData["params"]);
 
 			SetForegroundWindow(FindWindow(L"grcWindow", nullptr));
 		}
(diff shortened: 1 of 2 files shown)


jaymo1011:
Is there an issue with the current implementation? Not that you explicitly said it but I gather that streamer mode does not persist/activate for the connect dialog after the first connection? If that is the case then you should fix the issue at its roots instead of implementing a weird hack in the fivem:// url handler. This only fixes it for the use case where someone uses the url handler to connect to servers rather than using the disconnect button and then picking another server from the list.

sharkykh:
That's on me, I guess I should have stated why I am proposing this addition.

It's not instead of having a persistent streamer mode setting,
but rather, an override that server owners can force upon their players, most notably using so-called launchers (I assume it's still frowned upon?)

I have also seen [this Discord message](https://discord.com/channels/192358910387159041/443079653624643595/797935607862132766) that pushed me to try and contribute something that IMO might make things more confusing for users (like the users who enabled streamer mode by mistake and didn't know what happened, as seen on the Discord server's support channels multiple times).
image

> The URL handler should not be used to mandate things like user preferences.

The reason for this feature aside, to my knowledge,
this is the only officially-sanctioned (?) way to start FiveM and connect to a server, so I felt like I had no other choice (again, the reason for the feature aside).

blattersturm:
This `connect-sm` thing seems extremely nasty and definitely not 'cleaner'.

Implementation-wise, I'd return to what you had before with query string parameters, and instead of adding a `streamerMode` argument propagated all over the place, pass the (parsed?) query string further along so cfx-ui and other bits can handle new arguments without also having to touch the native code.

sharkykh:
Thank you for reviewing!

> This `connect-sm` thing seems extremely nasty and definitely not 'cleaner'.
> 
> Implementation-wise, I'd return to what you had before with query string parameters, and instead of adding a `streamerMode` argument propagated all over the place, pass the (parsed?) query string further along so cfx-ui and other bits can handle new arguments without also having to touch the native code.

Did my best in that regard with the updated commit. 
I've also could not think of a better way to pass the parameters through the socket to an already-running client, other than using a delimiter.
It works, but I am unsure if that solution is clean enough.

> Also,
> 
> > most notably using so-called launchers
> 
> why, over, say, a website that won't require the user to be oddly trashy?

You're right. It does make much more sense.

blattersturm:
> I've also could not think of a better way to pass the parameters through the socket to an already-running client, other than using a delimiter.

Perhaps JSON encoding/parsing?

sharkykh:
> Perhaps JSON encoding/parsing?

Done.
For some reason using the "replace" error handling flag (when dumping the JSON) did not work, but "strict" works.
Even though it looks like the value is the same? I'm confused.

PR #616, opened 2021-01-18
fix: applies the same margin-right to 'fa' as 'fab'

lze3:
Noticed the lack of margin for the `fa-road` icon here, it uses `fa`, not `fab`, so that's why the `margin-right` doesn't get applied.

Before:

After:


blattersturm:
this pr is really fab

lze3:
i agree, very fab

Sent with [ProtonMail](https://protonmail.com/) Secure Email.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

PR #617, opened 2021-01-20
Small fixes in the settings menu

Jamelele:
- Fixes the 'show streaming progress' setting displaying as off when actually enabled
image

~~- Changed the way default values are set in `DisplaySetting` so that the low-performance mode correctly shows as 'off' by default.~~ _Just realised what I did is a complete non-solution so I've reverted it_
image


Jamelele:
Looks like although this was an improvement it didn't actually solve the problem I described with the checkbox, but #621 should.

PR #618, opened 2021-01-20
fix(scripting/lua): also set curTime in CreateThread

Frosty-Ice:


blattersturm:
why, exactly? this function doesn't really use curTime

Frosty-Ice:
I believe it’s what’s causing this bug here: https://forum.cfx.re/t/possible-bug-with-latest-server-versions-tested-with-3382-3394/2020351

Thread gets created, there’s no update to the time, it hits the Wait, yields the thread (0+2000), time still isn’t updated but is higher than 2000, so then it’s able to run right back into the routines without waiting the proper time.

blattersturm:
Event handlers use CreateThreadNow to execute internally, though.

Frosty-Ice:
Well, I have no clue then. It’s related to something with the CreateThread at the top with the Wait statement inside of it, because if I remove the Wait statement, it works, or if I remove the entire CreateThread, it works.

Frosty-Ice:
Maybe this check needs to be moved into the Wait function? I think that would fix it. I’ll have to try getting the build process to work again tonight to verify.

PR #626, opened 2021-01-26
fix(server/impl): Prevent m_cureNetId to become 0 on NetId rollover.

spoty123:


blattersturm:
This change doesn't look very atomic. Will likely implement differently.

PR #628, opened 2021-01-27
fix: use server url for getConfiguration

PichotM:
`getConfiguration` was the only client endpoint not using the server URL I don't know if it was intended 
This will make the `host` param in the request more reliable 👍


blattersturm:
No.

PR #629, opened 2021-01-29
Add new Ped/Vehicle Hashes

less-x:
and fixed some invalid ped hashes
Test:
```cs
API.RegisterCommand("check_pedhash", new Action<int, List<object>, string>((int source, List<object> args, string rawCommand) =>
{
    foreach (PedHash hash in (PedHash[])Enum.GetValues(typeof(PedHash)))
    {
        Model m = new Model(hash);
        if(!m.IsValid)
        {
            Debug.WriteLine($"{hash}/{m.Hash}/{(uint)m.Hash} is not valid");
        }
    }
}), false);
```


blattersturm:
One thing that's a risk here is that code like such will intermittently fail to work on 1604 if merged:

```csharp
	public static class VehicleList
	{
		static VehicleList()
		{
			foreach (var hash in Enum.GetValues(typeof(VehicleHash)).OfType<uint>().Select(a => unchecked(a)))
			{
				Hashes.Add(hash);
			}

			HashList = Hashes.ToList();
		}

        public static HashSet<uint> Hashes { get; } = new HashSet<uint>();

		public static List<uint> HashList { get; private set; }
	}
```

```csharp
				var thisHash = VehicleList.HashList[rng.Next(VehicleList.HashList.Count)];
```

I'm not sure how commonplace selection logic like this is, though.

blattersturm:
Only safe to apply when the big C# versioning rework is done. Also, outdated by now.

Thanks for contributing!

PR #631, opened 2021-01-31
feat(scripting/v8): add PerformHttpRequest

MRV6:


blattersturm:
I'm not sure if implementing the weird pre-ref-callback API here makes sufficient sense over, say, reworking the actual call (though from a compatibility perspective - maybe?). Also, the TS definition looks incomplete.

MRV6:
I don't think that the TS definition is incomplete, this seems fine. Can you check please ?

lze3:
> 
> 
> I don't think that the TS definition is incomplete, this seems fine. Can you check please ?

The function supports `(url, cb, method, data, headers, options)` as parameters, yet your TS definition is only expecting 2 at max.

You've also got some `let`s where they should be `const`s.

MRV6:
Fixed.

PR #633, opened 2021-01-31
Feat: Add remember channel convar

manzarek123:
When a user reconnects to a server they've already , by default the server will remember the last channel they were in and move them to it automatically. Toggling this setting to false will disable that feature.


blattersturm:
Didn't I tell you in the other PR you ravaged with this change that I won't accept this option?

PR #637, opened 2021-02-01
Fix: Don't try to connect if should not connect 

manzarek123:
when using MumbleSetServerAddress


manzarek123:
Or maybe this could be implmented differently by not calling Mumble_Disconnect(true); if not connected in the first place

PR #639, opened 2021-02-02
Fix: Actual fix for turkish translate

egerdnc:
Fixed wrong and weird words weblate added also added some of missing's translated by a Turkish people so it is much better now.


d0p3t:
You're supposed to propose fixes on weblate. https://translations.cfx.re 

Also, different line endings? Always see the diff before committing

PR #641, opened 2021-02-02
feat(onesync): add GET_PED_SOURCE_OF_DEATH/DAMAGE natives

less-x:
Even if I didn't notice anything wrong during my testing, note I'm not very comfortable with c++

### DIFF
--- a/code/components/citizen-server-impl/include/state/ServerGameState.h
+++ b/code/components/citizen-server-impl/include/state/ServerGameState.h
@@ -273,6 +273,7 @@ struct CPedHealthNodeData
 	int health;
 	int armour;
 	uint32_t causeOfDeath;
+	int lastDamageEntity;
 };
 
 struct CPedOrientationNodeData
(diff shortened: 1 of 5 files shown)


PichotM:
nice find
`GET_PED_LAST_DAMAGE_ENTITY` may be a better name?

Disquse:
`unk4` = Killed with headshot
`unk5` = Killed with Melee Damage

Not really related to this PR I guess, but might be useful for future :P

Also `GET_PED_SOURCE_OF_DEATH` probably can be implemented using `lastDamageEntity` I think (in case if after death this data node field isn't getting updated).

less-x:
Thanks for your feedbacks :)
It looks like `GET_PED_CAUSE_OF_DEATH` returns the last cause even when the ped is not dead and stop being updated when it is, so I think this can be named `GET_PED_SOURCE_OF_DEATH` as this acts exactly the same

_(Yes, I just discovered the Net object viewer ^^)_

PR #646, opened 2021-02-08
fix(gameconfig): increase AttachmentExtension pool

PichotM:
Many people are actually getting this crash.
Some people started to manually edit this field and are not experiencing any side-effects, I believe it could be fine to just increase it


Tinky124:
@PichotM Yea you right!

Tinky124 (approved):

PR #647, opened 2021-02-09
feat(onesync): add ptFxEvent + doc

Korioz:
tested on 1604 & 2189

### DIFF
--- a/code/components/citizen-server-impl/src/state/ServerGameState.cpp
+++ b/code/components/citizen-server-impl/src/state/ServerGameState.cpp
@@ -26,6 +26,7 @@
 #include <ServerEventComponent.h>
 
 #include <boost/range/adaptors.hpp>
+#include <boost/math/constants/constants.hpp>
 
 #include <OneSyncVars.h>
 #include <DebugAlias.h>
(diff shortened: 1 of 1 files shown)


code/components/citizen-server-impl/src/state/ServerGameState.cpp
```diff
@@ -4538,6 +4539,169 @@ struct CStartProjectileEvent
...
+
+		f111 = buffer.Read<uint8_t>(1);
+
+		if (f111)
+		{
+			f100 = buffer.Read<int>(8) / 255.0f;
+		}
+		else
+		{
+			f100 = -1;
```
Disquse, line 4654: `f100` is a float, if `f111` statement is false, `f100` is being set to ` 0xBF800000`, which is `-1.0f`.

code/components/citizen-server-impl/src/state/ServerGameState.cpp
```diff
@@ -4661,6 +4825,11 @@ static std::function<bool()> GetEventHandler(fx::ServerInstanceBase* instance, c
 	bool isReply = buffer.Read<uint8_t>(); // is reply
 	uint16_t eventType = buffer.Read<uint16_t>(); // event ID
 
+	if (Is2060() && eventType > 55) // patch for 1868+ game build as `NETWORK_AUDIO_BARK_EVENT` was added
```
Disquse, line 4828: I'm not sure if this is the best way to resolve events ordering issue, but you have to make that check to also work with 2189 game build. You can use `xbr::IsGameBuildOrGreater<2060>()`.

Disquse (approved):
Looks good to me!

PR #649, opened 2021-02-10
feat(onesync): add CTrainGameStateDataNode

DelphinusBeta:


code/components/citizen-server-impl/include/state/SyncTrees_Five.h
```diff
@@ -2062,7 +2062,58 @@ struct CPlaneGameStateDataNode { bool Parse(SyncParseState& state) { return true
...
+		float cruiseSpeed = state.buffer.ReadSignedFloat(8, 30.0f);
+
+		// 0 = Moving, 1 = Slowing down, 3 = Stopped, 5 = About to depart?
+		int status = state.buffer.Read<int>(3);
+
+		bool isStartCarriage = state.buffer.ReadBit();
+
+		bool isEndCarriage = state.buffer.ReadBit();
+
+		bool unk12 = state.buffer.ReadBit();
```
Disquse, line 2102: Might be "Is Mission Train"

code/components/citizen-server-impl/include/state/SyncTrees_Five.h
```diff
@@ -2062,7 +2062,58 @@ struct CPlaneGameStateDataNode { bool Parse(SyncParseState& state) { return true
...
+
+		bool direction = state.buffer.ReadBit();
+
+		bool unk14 = state.buffer.ReadBit();
```
Disquse, line 2106: Might be "Has Passenger Carriages"

code/components/citizen-server-impl/include/state/SyncTrees_Five.h
```diff
@@ -2062,7 +2062,58 @@ struct CPlaneGameStateDataNode { bool Parse(SyncParseState& state) { return true
...
+
+		bool isDerailed = state.buffer.ReadBit();
+
+		// Turns true when player enters metrotrain. Remains true after leaving.
+		bool unk16 = state.buffer.ReadBit();
```
Disquse, line 2111: Might be "Force Doors Open"

Disquse (approved):
Thanks for this, documented very good. If you want to improve this, here's code for this data node from x360 with debug strings: https://pastebin.com/QV75d0qC

Do note that params may vary and even my suggestions may be wrong, so you have to test them.

code/components/citizen-server-impl/include/state/SyncTrees_Five.h
DelphinusBeta: Seems to always be false, if randomly generated or the 24 variants of CreateMissionTrain

code/components/citizen-server-impl/include/state/SyncTrees_Five.h
DelphinusBeta: Seems to always be true, same test done as above

Disquse: Good to know, thanks for testing, as I said this might be wrong since it's not even different game builds, but different game builds for different platforms :P

PR #653, opened 2021-02-12
feat(extra-natives-rdr3): SET_MINIMAP_TYPE for both 1311.20 and 1355.

slashkeyvalue:
Not too comfortable with cpp, hopefully this is fine.


blattersturm:
Will merge this once `rdr3`-specific native decls will work, probably.

slashkeyvalue:
Sounds good

blattersturm:
Been a while and I think we have this support now, can you squash the PR into a single commit?

KadDarem:
Hi @slashkeyvalue 

Do you know if it's possible to add GET_MINIMAP_TYPE() native to get the current value of the minimap type ?

PR #654, opened 2021-02-12
Unlock update channel switch for RedM

Ktos93:
Nothing special but makes it very easy to change update channel


ext/cfx-ui/src/app/settings.service.ts
```diff
@@ -237,7 +237,7 @@ export class SettingsService {
 				colorizeValue: true,
 				category: '#SettingsCat_Account',
 			});
-
+		}
 			this.addSetting('updateChannel', {
```
Disquse, line 241: What about fixing tabulation?

PR #657, opened 2021-02-16
Patch decal def limits

ISFASI:
Increasing decal def (see decals.dat) limit to 512.
Tested on 1604, 2060 and 2189 using `replace_level_meta` for loading custom `decals.dat`


blattersturm:
data file when

ISFASI:
> data file when

small steps for big features

PR #658, opened 2021-02-17
feat(server/resources): Support logging of network events to console and file with convars

kn0pee:


blattersturm:
Unsure about the need for this one - I'll leave it around for now so I'll be able to take a better look somewhen.

kn0pee:
All good. The main reason I implemented this was to be able to easier diagnose net events causing issues originating from the server.

You mentioned in discord that you believed hitching of the network thread (in my case) was being caused by large or high amounts of net events being sent to the client. You can view this on each client, however this doesn't help diagnose crashes if the net events weren't going to everyone.

Having the ability to log to a file helps correlate net event usage to a crash. This enables server owners to fix net events which are occurring to often, or convert large ones to latent events. Otherwise it can be like looking for a needle in a haystack as to which events are causing issues, especially with high player counts.

mickeygeecom:
I'd second exactly what kn0pee adresses here, would indeed be helpful. In general is backlogging just a very neat tool for debugging and perf improvements. 🤗

blattersturm:
Likely not the best way to do this - closing this so it doesn't linger around.

PR #659, opened 2021-02-19
tweak(extra-natives-five): disable auto-repair on extra sync

DelphinusBeta:
Currently SET_VEHICLE_AUTO_REPAIR_DISABLED only disables auto-repair for the player that has network control of the vehicle. When the vehicle extras sync between players, auto-repair is called from a different location, which isn't disabled.


YBa2Cu3O7 (approved):
Lookin good.

PR #660, opened 2021-02-20
NUI_HAS_FOCUS native

Sadler2:
It allows to easily detect when some external resource already took NUI focus and not to start our own focus-aimed actions in that case.

### DIFF
--- a/code/components/nui-core/src/CefInput.cpp
+++ b/code/components/nui-core/src/CefInput.cpp
@@ -18,6 +18,8 @@
 #include <windowsx.h>
 #include <console/Console.VariableHelpers.h>
 
+using nui::HasFocus;
+
 extern nui::GameInterface* g_nuiGi;
 
 static bool g_hasFocus = false;
@@ -32,11 +34,6 @@ static ConVar<bool> uiLoadingCursor("ui_loadingCursor", ConVar_None, false);
 bool isKeyDown(WPARAM wparam)
 {
 	return (GetKeyState(wparam) & 0x8000) != 0;
-}
-
-bool HasFocus()
-{
-	return (g_hasFocus || g_hasOverriddenFocus);
 }
 
 #include <shared_mutex>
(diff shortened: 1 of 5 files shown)


ISFASI:
It may be useful to implement extra getter for ``SET_NUI_FOCUS_KEEP_INPUT`` too

Sadler2:
@ISFASI looks reasonable, I'll test it, and if it works ok, add `NUI_HAS_FOCUS_KEEP_INPUT` in addition to `NUI_HAS_FOCUS`.

blattersturm:
IS_NUI_FOCUSED
IS_NUI_INPUT_RETAINED

Might be better names?

blattersturm:
Also, descriptions in the native decls, maybe. :P

Sadler2:
@blattersturm Well, I have no special preferences in naming, only thought that the "keep_input" part should have definitely connect this getter with "SET_NUI_FOCUS_KEEP_INPUT" in programmers' mind. Ok, I'll change it and add descriptions.

blattersturm:
> only thought that the "keep_input" part should have definitely connect

Good point - IS_NUI_FOCUS_KEEPING_INPUT?

Sadler2:
Renamed and tested, it works ok and ready to merge.

R0JEA:
@blattersturm Why wasn't this merged yet??

visibait:
Merge this hahaha this is a great and useful native!

code/components/nui-core/src/CefInput.cpp
```diff
@@ -355,7 +360,7 @@ static HookFunction initFunction([] ()
 {
 	g_nuiGi->QueryMayLockCursor.Connect([](int& argPtr)
 	{
-		if (HasFocus() && !g_keepInput)
+		if (nui::HasFocus() && !g_keepInput)
```
blattersturm, line 363: Is this _really_ needed to be changed rather than e.g. `using nui::HasInput` somewhere out of the namespace decl?

blattersturm (changes requested):
One tiny thing, and ideally a squash-rebase.

code/components/nui-core/src/CefInput.cpp
Sadler2: I guess, it's not. I'll fix it, recheck this request on current codebase and send a commit.

Sadler2: Done. It's squash-rebased and tested on local server.

blattersturm (approved):
Seems fine. Queued for merge next push.

PR #661, opened 2021-02-20
feat(onesync): add SET_PLAYER_CULLING_RADIUS native

rubbertoe98:
This PR adds the ability to set a players culling radius on onesync infinity with the following native:

SET_PLAYER_CULLING_RADIUS(playerSrc,radius)

### DIFF
--- a/code/components/citizen-server-impl/include/state/ServerGameState.h
+++ b/code/components/citizen-server-impl/include/state/ServerGameState.h
@@ -466,10 +466,23 @@ struct SyncEntityState
 
 	virtual ~SyncEntityState();
 
-	inline float GetDistanceCullingRadius()
+
+	inline float GetDistanceCullingRadius(float playerCullingRadius)
 	{
-		// #TODO1S: figure out a good value for this
-		return overrideCullingRadius != 0.0f ? overrideCullingRadius : (424.0f * 424.0f);
+		//Use priority ordering
+		if (overrideCullingRadius != 0.0f) 
+		{
+			return overrideCullingRadius;
+		}
+		else if (playerCullingRadius != 0.0f) 
+		{
+			return playerCullingRadius;
+		}
+		else
+		{
+			// #TODO1S: figure out a good value for this
+			return (424.0f * 424.0f);
+		}
 	}
 
 	inline uint32_t GetScriptHash()
@@ -800,6 +813,13 @@ struct GameStateClientData : public sync::ClientSyncDataBase
 	std::shared_ptr<fx::StateBag> playerBag;
 
 	uint32_t routingBucket = 0;
+
+	float playerCullingRadius = 0.0f;
+	
+	inline float GetPlayerCullingRadius()
+	{
+		return playerCullingRadius;
+	}
 
 	GameStateClientData()
 		: syncing(false)
--- a/code/components/citizen-server-impl/include/state/ServerGameState.h
+++ b/code/components/citizen-server-impl/include/state/ServerGameState.h
@@ -258,6 +258,13 @@ struct CEntityOrientationNodeData
 	compressed_quaternion<11> quat;
 };
 
+struct CObjectOrientationNodeData
+{
+	bool highRes;
+	compressed_quaternion<11> quat;
+	float rotX, rotY, rotZ;
+};
+
 struct CPhysicalVelocityNodeData
 {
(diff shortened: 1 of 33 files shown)


code/components/citizen-server-impl/src/state/ServerGameState.cpp
```diff
@@ -840,7 +840,9 @@ void ServerGameState::Tick(fx::ServerInstanceBase* instance)
 					float distSquared = (diffX * diffX) + (diffY * diffY);
 					if (distSquared < entity->GetDistanceCullingRadius())
 					{
-						isRelevant = true;
+						if (playerEntity->playerCullingRadius == 0.0f || distSquared < playerEntity->playerCullingRadius) {
```
PichotM, line 843: It would be better to include this condition with the one above.
And read the value from entityClient/client data not the entity

code/components/citizen-server-impl/include/state/ServerGameState.h
```diff
@@ -801,6 +802,8 @@ struct GameStateClientData : public sync::ClientSyncDataBase
 
 	uint32_t routingBucket = 0;
 
+	float playerCullingRadius = 0.0f;
```
PichotM, line 805: This will not be used as you're only reading the value from the player entity, this would be a non-used field for all non-player entities.
In my opinion, it would be better to add this field in the client class or GameStateClientData as you did

code/components/citizen-server-impl/src/state/ServerGameState_Scripting.cpp
```diff
@@ -1333,4 +1333,41 @@ static InitFunction initFunction([]()
...
+				// get the owning server instance
+				auto instance = resourceManager->GetComponent<fx::ServerInstanceBaseRef>()->Get();
+
+				// get the server's game state
+				auto gameState = instance->GetComponent<fx::ServerGameState>();
+
+				auto [lock, clientData] = gameState->ExternalGetClientData(client);
+				clientData->playerCullingRadius = radius * radius;
+
+				fx::sync::SyncEntityPtr playerEntity;
```
PichotM, line 1357: All this part below is not needed if you stick to client data

PichotM (commented):
Also, I believe you may also need to include this custom culling radius in the migration distance check

rubbertoe98:
The above commit removes some unused code, uses the GameStateClientData to get the player culling radius instead getting it from the entity class as suggested. I've also added a function that decides the priority of which culling distance to use in the following order, thanks for @PichotM for the idea. example:

- entity radius if set by SET_ENTITY_DISTANCE_CULLING_RADIUS
- player radius if set by SET_PLAYER_CULLING_RADIUS
- default radius (424.0f)

code/components/citizen-server-impl/src/state/ServerGameState.cpp
```diff
@@ -838,16 +838,17 @@ void ServerGameState::Tick(fx::ServerInstanceBase* instance)
 					float diffY = entityPos.y - playerPos.y;
 
 					float distSquared = (diffX * diffX) + (diffY * diffY);
-					if (distSquared < entity->GetDistanceCullingRadius())
-					{
+
+					float cullingRadius = entity->GetDistanceCullingRadius(clientDataUnlocked->GetPlayerCullingRadius());
+					if (distSquared < cullingRadius) {
```
blattersturm, line 843: Why does this introduce a new variable and a formatting error?

code/components/citizen-server-impl/src/state/ServerGameState.cpp
```diff
@@ -838,16 +838,17 @@ void ServerGameState::Tick(fx::ServerInstanceBase* instance)
...
 						isRelevant = true;
 					}
 					else
 					{
 						// are we owning the world grid in which this entity exists?
 						int sectorX = std::max(entityPos.x + 8192.0f, 0.0f) / 150;
 						int sectorY = std::max(entityPos.y + 8192.0f, 0.0f) / 150;
-
+
```
blattersturm, line 851: Useful!

blattersturm (changes requested):
Please fix the formatting oddity and squash your commits. :)

rubbertoe98:
Fixed the formatting issues, but I think i've messed up the git squash, going to remake the PR

PR #662, opened 2021-02-21
fix(scripting): rework server-internal events 

jaymo1011:
~~playerJoining is emitted by the server and is considered a net event but wasn't added to the safe net events whitelist for Lua or JS.~~

Server only events which are internally considered net events (like playerConnecting, playerDropped and playerJoining) are now handled separately from proper network events and cannot be triggered by clients unless the event is explicitly registered as a network event.

This fixes issues with the playerJoining event and stops malicious clients from triggering playerConnecting or playerDropped themselves.

### DIFF
--- a/code/client/clrcore/EventHandlerDictionary.cs
+++ b/code/client/clrcore/EventHandlerDictionary.cs
@@ -12,6 +12,17 @@ namespace CitizenFX.Core
 {
     public class EventHandlerDictionary : Dictionary<string, EventHandlerEntry>
     {
+#if IS_FXSERVER
+        // We need this to ensure that server-internal events can't be triggered from the client as unlike 
+        // Lua and JS, C# has never had net event filtering so adding it now would break compatibility.
+        private static HashSet<string> m_serverInternalNetEvents = new HashSet<string>
+        {
+            "playerConnecting",
+            "playerDropped",
+            "playerJoining",
+        };
+#endif
+
         public new EventHandlerEntry this[string key]
         {
             get
@@ -48,6 +59,13 @@ namespace CitizenFX.Core
             
             if (TryGetValue(lookupKey, out entry))
             {
+#if IS_FXSERVER
+                if (m_serverInternalNetEvents.Contains(eventName) && !sourceString.StartsWith("internal-net"))
+                {
+                    return;
+                }
+#endif
+
                 await entry.Invoke(sourceString, arguments);
             }
         }
--- a/data/shared/citizen/scripting/lua/scheduler.lua
+++ b/data/shared/citizen/scripting/lua/scheduler.lua
@@ -265,11 +265,6 @@ end)
 
 ]]
 
-local alwaysSafeEvents = {
-	["playerDropped"] = true,
-	["playerConnecting"] = true
-}
-
 local eventHandlers = {}
 local deserializingNetEvent = false
 
@@ -280,21 +275,25 @@ Citizen.SetEventRoutine(function(eventName, eventPayload, eventSource)
 
 	-- try finding an event handler for the event
 	local eventHandlerEntry = eventHandlers[eventName]
-	
+
 	-- deserialize the event structure (so that we end up adding references to delete later on)
 	local data = msgpack.unpack(eventPayload)
 
 	if eventHandlerEntry and eventHandlerEntry.handlers then
 		-- if this is a net event and we don't allow this event to be triggered from the network, return
 		if eventSource:sub(1, 3) == 'net' then
-			if not eventHandlerEntry.safeForNet and not alwaysSafeEvents[eventName] then
+			if not eventHandlerEntry.safeForNet then
 				Citizen.Trace('event ' .. eventName .. " was not safe for net\n")
 
+				_G.source = lastSource
 				return
 			end
 
 			deserializingNetEvent = { source = eventSource }
 			_G.source = tonumber(eventSource:sub(5))
+		elseif IsDuplicityVersion() and eventSource:sub(1, 12) == 'internal-net' then
+			deserializingNetEvent = { source = eventSource:sub(10) }
+			_G.source = tonumber(eventSource:sub(14))
 		end
 
 		-- return an empty table if the data is nil
--- a/data/shared/citizen/scripting/lua/scheduler.lua
+++ b/data/shared/citizen/scripting/lua/scheduler.lua
@@ -5,6 +5,7 @@ local debug = debug
 local coroutine_close = coroutine.close or (function(c) end) -- 5.3 compatibility
 local hadThread = false
 local curTime = 0
+local isDuplicityVersion = IsDuplicityVersion()
 
 -- setup msgpack compat
 msgpack.set_string('string_compat')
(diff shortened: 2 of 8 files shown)


blattersturm:
Not sure if this is a good idea at all. In fact, not sure if it's a good idea to even register these as net events in the first place - this'd allow clients to incorrectly send such when these should be server-internal.

jaymo1011:
I assume you're talking about all of these always safe events? Whatever change is made needs to ensure compatibility so perhaps for these internal events with player sources, a different tag on the event source could be used.
`net:1` -> the event source is a client
`internal-net:1` -> the event source is the server but referencing a player (there's probably a better name for the tag)

If an event is triggered and the event source starts with `net:`, the event needs to be registered (as usual) but if the event source starts with `internal-net:`, the event does not need to be registered.

Would something like that work and be feasible to implement? I could just add that to this PR in that case.

blattersturm:
Could be a valid solution, yes - however keep in mind that _all_ ScRTs parsing `source` values need to be in on this.

jaymo1011:
temporarily closing while I work on this

data/shared/citizen/scripting/v8/main.js
```diff
@@ -396,10 +396,12 @@ const EXT_LOCALFUNCREF = 11;
...
 					return;
 				}
-
+        
 				global.source = parseInt(source.substr(4));
+			} else if (IsDuplicityVersion() && source.startsWith('internal-net')) {
+				global.source = parseInt(source.substr(13));
 			}
-
+
```
jaymo1011, line 404: No idea why this indenting stuff is still broken, I *just* resolved the merge conflict and now these random tabs and spaces just above have just appeared :(

data/shared/citizen/scripting/lua/scheduler.lua
```diff
@@ -280,23 +275,27 @@ Citizen.SetEventRoutine(function(eventName, eventPayload, eventSource)
...
 			end
 
 			deserializingNetEvent = { source = eventSource }
 			_G.source = tonumber(eventSource:sub(5))
+		elseif IsDuplicityVersion() and eventSource:sub(1, 12) == 'internal-net' then
+			deserializingNetEvent = { source = eventSource:sub(10) }
+			_G.source = tonumber(eventSource:sub(14))
 		end
 
+		-- deserialize the event structure (so that we end up adding references to delete later on)
```
blattersturm, line 296: Moving this seems undesirable with regards to the precarious case of broadcast funcrefs: every reference _submitted_ must be _deserialized_ in order to be GC'd at _some_ point. A more viable approach would be to annotate the ScRT event call with a set of funcref IDs to immediately discard upon ignoring an event, and _then_ one can move serialization to be only on-demand.

This was moved there in [b250df1cf1859e21d8a2078b23ade14e098a6ebd](https://github.com/citizenfx/fivem/commit/b250df1cf1859e21d8a2078b23ade14e098a6ebd#diff-27f3d1b4362eb7a4837bdb80d6a1a71c08e408e134ff43adae7116b4632c6466R146), and this comment was amended to indicate that it was up there for a reason.

Current broadcast funcrefs might be extremely harmful, for they will lead to extremely confusing GC scenarios if even _one_ ScRT hangs on to them for too long, but this change will lead to guaranteed memory leaks... I _think_. **Not sure in fact, would need explicit testing and hard to submarine in here.**

I do think the comment could be clearer in this regard!

data/shared/citizen/scripting/lua/scheduler.lua
```diff
@@ -280,23 +275,27 @@ Citizen.SetEventRoutine(function(eventName, eventPayload, eventSource)
...
+			if not eventHandlerEntry.safeForNet then
 				Citizen.Trace('event ' .. eventName .. " was not safe for net\n")
 
+				_G.source = lastSource
 				return
 			end
 
 			deserializingNetEvent = { source = eventSource }
 			_G.source = tonumber(eventSource:sub(5))
+		elseif IsDuplicityVersion() and eventSource:sub(1, 12) == 'internal-net' then
```
blattersturm, line 291: Do we not cache IsDuplicityVersion's result elsewhere already? Since this won't change at runtime, calling _this_ frequently is a bit redundant.

data/shared/citizen/scripting/v8/main.js
```diff
@@ -398,6 +398,8 @@ const EXT_LOCALFUNCREF = 11;
 				}
 
 				global.source = parseInt(source.substr(4));
+			} else if (IsDuplicityVersion() && source.startsWith('internal-net')) {
```
blattersturm, line 401: Same memoization request here as above for Lua.

code/client/clrcore/EventHandlerDictionary.cs
```diff
@@ -48,6 +59,14 @@ internal async Task Invoke(string eventName, string sourceString, object[] argum
             
             if (TryGetValue(lookupKey, out entry))
             {
+#if IS_FXSERVER
+
+                if (m_serverInternalNetEvents.Contains(eventName) && !sourceString.StartsWith("internal-net"))
+                {
+                    Debug.WriteLine($"Event {eventName} was not safe for net.");
```
blattersturm, line 66: Where's the rejection?

I consider it _somewhat_ unlikely that any script will be calling these and expecting them to work, and even a log line can be seen as a compatibility break.

blattersturm (changes requested):

code/client/clrcore/EventHandlerDictionary.cs
jaymo1011: Whoops, forgot to add the `return` statement 🤦‍♂️
Will include that with your other requested changes 😅

data/shared/citizen/scripting/lua/scheduler.lua
jaymo1011: No it is not cached anywhere however, as you said, it won't change at runtime so caching it will improve performance not just here but in other places like exports, RPC invocation and state bags. I'll make this into its own commit, separate from the other changes requested in this file but including the same requested changes to the JS ScRT, grouping the same type of changes together.

data/shared/citizen/scripting/v8/main.js
jaymo1011: Resolved above, will make the same change here of caching the result.

data/shared/citizen/scripting/lua/scheduler.lua
jaymo1011: I think the best thing for *this* PR is just to revert the change and create a seperate PR for this specifically. In hindsight, I was overstepping the scope of this PR with this change and should have just changed what needed to be changed to get the feature to work correctly.

Happy to move forward with that or should I include such changes here too?

blattersturm: Would indeed be suggested to put any such changes elsewhere, yes.

jaymo1011:
whoops, internet went out and my revert of the changes in the Lua ScRT didn't push

blattersturm (approved):
Looks fine. Hopefully there won't be some secretly-hidden compatibility concerns with the new prefix.

PR #663, opened 2021-02-22
feat(impl/endpoint): support IP authentication with CF

PichotM:
¯\_(ツ)_/¯
useful


blattersturm:
[shrug kaomoji]
exploitable

blattersturm:
honestly idk if this will check endpoints first but _really_ at code duplication even if so

maybe configurable?

PR #666, opened 2021-02-23
fix(clr): respect parameter arrays in event handlers and func refs

jaymo1011:


blattersturm:
Since we don't already allow `params object[] n` to be a receiver of an IList<object>, this implementation should be fine.

PR #667, opened 2021-02-23
feat(onesync): add SET_PLAYER_CULLING_RADIUS native

rubbertoe98:
This PR adds the ability to set a players culling radius on onesync infinity with the following native:

SET_PLAYER_CULLING_RADIUS(playerSrc,radius)

This also decides the priority of which culling distance to use in the following order, thanks to @PichotM for the idea. example:

- entity radius if set by SET_ENTITY_DISTANCE_CULLING_RADIUS
- player radius if set by SET_PLAYER_CULLING_RADIUS
- default radius (424.0f)

### DIFF
--- a/code/components/citizen-server-impl/include/state/ServerGameState.h
+++ b/code/components/citizen-server-impl/include/state/ServerGameState.h
@@ -478,10 +478,22 @@ struct SyncEntityState
 
 	virtual ~SyncEntityState();
 
-	inline float GetDistanceCullingRadius()
+	inline float GetDistanceCullingRadius(float playerCullingRadius)
 	{
-		// #TODO1S: figure out a good value for this
-		return overrideCullingRadius != 0.0f ? overrideCullingRadius : (424.0f * 424.0f);
+		//Use priority ordering
+		if (overrideCullingRadius != 0.0f) 
+		{
+			return overrideCullingRadius;
+		}
+		else if (playerCullingRadius != 0.0f) 
+		{
+			return playerCullingRadius;
+		}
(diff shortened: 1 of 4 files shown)


PichotM (approved):

mikeemoo:
@rubbertoe98 Do you have a simple example of this working? Maybe I'm misunderstanding how this should behave, but when I use SetPlayerCullingRadius I'm not detecting any kind of difference.

arthur-rl:
> @rubbertoe98 Do you have a simple example of this working? Maybe I'm misunderstanding how this should behave, but when I use SetPlayerCullingRadius I'm not detecting any kind of difference.

Can you send the code you are using to do this? Also you can check it out on the FiveM documentation https://docs.fivem.net/natives/?_0x8A2FBAD4

mikeemoo:
Sure - 

With a single player on the server,

```javascript
onNet("expand", () => {
  SetPlayerCullingRadius(source, 10000.00);
});

setInterval(() => {
  console.log(GetAllPeds().length);
}, 1e3);
```

I was hoping to see more entities stick around as I move around the map, however I see no difference no matter what I set this value to.

blattersturm:
This is to prevent the server from removing faraway remote entities - you might be thinking of a population range instead.

mikeemoo:
If my player moves away from the Peds it initially created, does that not make them faraway entities?

blattersturm:
You're forgetting the "remote" constraint there.

mikeemoo:
Gotcha. Thanks.

PR #672, opened 2021-03-02
feat(scripting/v8): global shared object, support for modern OOP-style frameworks

Sadler2:
This patch creates a `global.shared` object inside every v8 scripting context, that may be used to exchange any data and objects between different fivem resources by reference without any extra data serialization. We use this functionality to create our TypeScript OOP framework with straightforward easy to use API: [NovaEngine](https://gitlab.com/novalife/engine).

I have not so much experience with v8, any help integrating this feature into fivem would be appreciated.

### DIFF
--- a/code/components/citizen-scripting-v8/src/V8ScriptRuntime.cpp
+++ b/code/components/citizen-scripting-v8/src/V8ScriptRuntime.cpp
@@ -1924,6 +1924,9 @@ result_t V8ScriptRuntime::Create(IScriptHost* scriptHost)
 		context->Global()->Set(context, String::NewFromUtf8(GetV8Isolate(), "window", NewStringType::kNormal).ToLocalChecked(), context->Global());
 	}
 
+	static Global<Object> sharedObject = Global<Object>::Global(GetV8Isolate(), Object::New(GetV8Isolate()));
+	context->Global()->Set(context, String::NewFromUtf8(GetV8Isolate(), "shared", NewStringType::kNormal).ToLocalChecked(), sharedObject.Get(GetV8Isolate()));
+
 	std::string nativesBuild = "natives_universal.js";
 
 	{


blattersturm:
No.

While this might be viable at this point by means of implementation detail (shared isolate, non-shared context), this is merely an implementation detail, and one not mirrored in other ScRTs.

Exposing this implementation detail, outside of violating the 'all ScRTs are equal' policy the runtime holds, as well as the infinitesmal amount of scoping problems that can be caused here, also holds us to this shared-isolate construct _for all eternity_.

It's much more of a recommended solution, if your framework wants to use objects directly, to create your own equivalent of client_script and server_script for your framework, and handle resource start/stop events, loading scripts from this metadata (GET_RESOURCE_METADATA stuff) in your 'controlling' resource.

Sadler2:
I understand how to parse the fields like "client_script" with GET_RESOURCE_METADATA, but starting those scripts in the context of that resource looks harder. Any resmon profiling would stop working in the case of running it directly, the whole resource isolation should also be done manually.

blattersturm:
> but starting those scripts in the context of that resource looks harder

Don't. Just run in your own context.

> Any resmon profiling would stop working in the case of running it directly

Fixable by exposing APIs to allow resources to 'impersonate' others as a nested scope manually (i.e. manually setting a generic fx::PushEnvironment to be popped/reset later).

> the whole resource isolation should also be done manually.

Yes, but both Lua and JS have sufficient primitives to allow at least some of this.

Sadler2:
Thanks, I'll start digging in this direction :)

blattersturm:
If you need any help with this, don't hesitate to ask, of course. :)

Sadler2:
I can describe the current situation with the project, perhaps you would suggest me something.

Loading through GetResourceMetadata and LoadResourceFile went fine, test scripts run without errors. Although I have to also add my scripts to "files" metadata field to make LoadResourceFile work.

I tried to make an isolation layer through the 'vm' module of node.js, but the client side doesn't have this, so I switched to basic 'eval' to make it more universal between client and server, partly sacrificing the isolation. The next difficulty came from the necessity to dynamically stop scripts in accord with resource stopping. The only thing came to my mind was to wrap all the "setTimeout"/"clearTimeout", "setInterval"/"clearInterval", etc. functions, put them to the map and stop when resource stops. I still can't stop the main "thread", but it's not a big problem, it often dies very soon anyway.  It still doesn't seem reliable enough to put on my server, testing continues.

I also tried to make an impersonation through the pushEnvironment, but no luck for now, my level of understanding V8 code is far from ideal. [Here is the function I've tried](https://gist.github.com/Sadler2/45127dfcb19594bd2fd9f16777c6428f).

Next big problem would be NUI, because it uses the context of calling resource. I maybe can create a wrapper, that I would register proper callbacks, listen for the events and forward it to and from my shared resource. This solution has its drawbacks, I need to at least test the performance before I can use it inside loops and tick functions.

PR #676, opened 2021-03-05
perf(scheduler): thread queue/running thread stack

pogue-mahoney:
The `threads` `wakeTime` sorted queue and `newThreads` stack replace the coroutine-keyed threads table and `newThreads` meta table for the tick routine.  This allows not only faster index-based iteration through `threads` and `newThreads`, but also lets the `threads` loop short-circuit after encountering the first thread with a later wake time.  The `runningThreadStack` replaces the coroutine-keyed `threads` table for awaits.

### DIFF
--- a/data/shared/citizen/scripting/lua/scheduler.lua
+++ b/data/shared/citizen/scripting/lua/scheduler.lua
@@ -1,6 +1,6 @@
 local GetGameTimer = GetGameTimer
 local _sbs = Citizen.SubmitBoundaryStart
-local coresume, costatus = coroutine.resume, coroutine.status
+local coresume, costatus, tinsert, tremove = coroutine.resume, coroutine.status, table.insert, table.remove
 local debug = debug
 local coroutine_close = coroutine.close or (function(c) end) -- 5.3 compatibility
 local hadThread = false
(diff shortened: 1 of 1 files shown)


blattersturm:
Do you have any numbers on this? The sorted insertion mainly seems a little odd - though I guess it mainly optimizes for the case where most coroutines are left idle.

Also, compatibility concerns as usual, look at the history of the main execution loop for a few past gotchas that should be left compatible.

Got burned by attempts at optimizing here a few too many times. 😕

pogue-mahoney:
> Do you have any numbers on this? 
I do, though they're always going to be anecdotal.  If you keep the number if threads small in each execution environment, you won't see much.  It makes a big difference for resources with a lot of loops of varying waits, for example (6% to .6% on one case).

>though I guess it mainly optimizes for the case where most coroutines are left idle.
It mainly optimizes tick over insertion, yes.  That said you'll notice the queue is ordered so the highest index in the table is first out.  This optimizes large numbers of fast threads because the shifts will be smaller in those cases.
 
> Also, compatibility concerns as usual, look at the history of the main execution loop for a few past gotchas that should be left compatible.
> 
> Got burned by attempts at optimizing here a few too many times. 😕
I'll do that, and get back to you.

PR #679, opened 2021-03-09
fix(scripting/v8): exported arrays

cyrep:
[](https://discord.com/channels/192358910387159041/553225490186371082/818632616755462224)

### DIFF
--- a/data/shared/citizen/scripting/v8/main.js
+++ b/data/shared/citizen/scripting/v8/main.js
@@ -104,7 +104,7 @@ const EXT_LOCALFUNCREF = 11;
 
 	/**
 	 * Deletes ref function
-	 * 
+	 *
 	 * @param {int} ref
 	 */
 	Citizen.setDeleteRefFunction(function (ref) {
@@ -119,9 +119,9 @@ const EXT_LOCALFUNCREF = 11;
 
 	/**
 	 * Invokes ref function
-	 * 
-	 * @param {int} ref 
-	 * @param {UInt8Array} args 
+	 *
+	 * @param {int} ref
+	 * @param {UInt8Array} args
 	 */
 	Citizen.setCallRefFunction(function (ref, argsSerialized) {
 		if (!refFunctionsMap.has(ref)) {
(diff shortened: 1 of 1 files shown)


draobrehtom:
@cyrep described issue more detailed in discord channel before creating PR:

изображение

blattersturm:
Even if such an issue exists, that wouldn't be _fixed_ by this 'fix', that 'fix' would break JS calling into exports defined in Lua/C#, so isn't even a fix but much rather a break.

I can't copy the example snippets from this screenshot so I'm unable to test, no clue what 'Discord channel' you're talking about but there's no 'Discord channel' for development/bug reporting, so please provide said snippets as _not a screenshot_.

blattersturm:
> https://discord.com/channels/192358910387159041/553225490186371082/818632616755462224

This link leads to a login page, I don't use 'discord.com' so I wouldn't be able to read said message. Is there any more detail about the alleged issue on a public place like the forums or Matrix?

Anyway, thanks for the paste - I'll look into this and see why the line in the comment before isn't packing the array value correctly.

blattersturm:
OK, right, the hell, why are these being _double-unwrapped_?

https://github.com/citizenfx/fivem/blob/a1cd64b01533704dde6fe5b43d85321c3336fef0/data/shared/citizen/scripting/v8/main.js#L89-L96

This PR _could_ be merged, but only if it is stripped of its whitespace changes and then force-pushed, _and_ it might lead to compatibility issues if people have worked around this double-unwrapping manually.

Root cause analysis:

- The commit https://github.com/citizenfx/fivem/commit/8b4c32665f042f3494b844de326f6cb15ad08160 refactored unwrapping to be 'global' like in other runtimes, as CallRef was apparently not unwrapping.
- The commit https://github.com/citizenfx/fivem/commit/fd9419570130cd002c6250935c53251711b9175a#diff-e83d9af2b76da416d953a90bdea5c4a2d5b6e3ce4fa0d034cc4949c1030121d9R253 however had a special case for this for exports only, which then led to doubled unwrapping.

As above, this PR should act as a fix then, but I can't merge it in its current state as it has a crazy amount of unrelated changes.

cyrep:
sry fo the whitespace changes (my IDE automatically removes unnecessary whitespaces on save). should've double checked my changes, mb.
thx draobrehtom for making my case and thx blattersturm for looking into it.

PR #680, opened 2021-03-09
Make provide work with exports

Sampleeeee:
Providing for a resource will allow exports to work too.

Ex:
```lua
-- new_resource/fxmanifest
provide 'old_resource'

-- new_resource/shared.lua
exports('TestFunction', function()
    print('new test')
end)

-- some other random resource
exports['old_resource']:TestFunction()
```
Output:
> new test

### DIFF
--- a/data/shared/citizen/scripting/lua/scheduler.lua
+++ b/data/shared/citizen/scripting/lua/scheduler.lua
@@ -981,6 +981,13 @@ setmetatable(exports, {
 		AddEventHandler(getExportEventName(GetCurrentResourceName(), exportName), function(setCB)
 			setCB(func)
 		end)
+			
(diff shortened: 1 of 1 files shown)


blattersturm:
This has a few flaws (and I already had a local changeset doing this somewhere, just not placed in the main repo for some reason):

- Only for the Lua ScRT. Really?
- Only allows one provide statement as you're just not even checking for multiple such entries. 

Again, as with the other ScRT PR, these two reports were on my todo list for today and then I wake up with two half-assed PRs that both are incomplete and somewhat undesirable.

Sampleeeee:
Sorry, was just curious if I could figure out how to fix this. Also saw the issue where it only works if you're using the exports function instead of the exports in the resource manifest.

PR #685, opened 2021-03-13
fix(onesync-natives): compatibility with game seat indexes

PichotM:
For sync nodes, the driver seat index starts at 1, and so on
This fixes the compatibility with the game seat index that starts at -1 for the driver

### DIFF
--- a/code/components/citizen-server-impl/src/state/ServerGameState_Scripting.cpp
+++ b/code/components/citizen-server-impl/src/state/ServerGameState_Scripting.cpp
@@ -1001,7 +1001,7 @@ static InitFunction initFunction([]()
     {
         auto vn = entity->syncTree->GetVehicleGameState();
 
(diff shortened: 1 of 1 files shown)


spoty123:
> game seat index that starts at -1 for the driver
I don't really see why would this matter, it's noted in the native reference that it starts at index one?

Anyway, if those changes do get approved, it might also be a good idea to change the indexes that are noted inside of the native reference.

PichotM:
It is odd to use client side -1 for the driver and 1 server side

blattersturm:
> I don't really see why would this matter, it's noted in the native reference that it starts at index one?

So? If the native has the same name as a client-side native, it should behave the same way?

Statements are phrased like questions and end with a question mark?

spoty123:
> If the native has the same name as a client-side native, it should behave the same way?

Sure thing, it wasn't very clear to me until now.

PR #686, opened 2021-03-18
RedM Discord Presence

Ktos93:
Require to set RedM app id and app asset


Slypher (approved):

vinicius7199 (approved):

egerdnc:
Damn

sorambr (approved):

code/components/lovely-script/src/Lovely.cpp
```diff
@@ -1,177 +1,173 @@
-/*
```
blattersturm, line 1: Why does this remove and re-add the entire file?

blattersturm (changes requested):
Does whoever 'approves these changes' actually check the changeset at all?

code/components/lovely-script/src/Lovely.cpp
Ktos93: Looks like the PR was broken due to the time it was waiting for, I'll do it again and it should look good

PR #688, opened 2021-03-21
fix(dev-tools): increase used memory threshold

PichotM:


Ekinoxx0:
75M is kinda arbitrary 🤷‍♂️
But I suppose this is linked to gathering all resources of a server to a single one (aka not splitting them for modularity)

Are you caching this many values into your client-side script ?
No memory leak ?

PichotM:
that's not "gathering all resources of a server to a single one"
That's just my game mode that is a single resource and it works relatively fine.
I am pretty sure most servers are above the 50Mb limit in total so people with a single main resource are negatively impacted by this warning

Ekinoxx0:
I see 👍 
After some testing on random servers, can confirm most of them are running 30-60 resources with 1/2Mb so would lead to 60/120Mb in total

We could link the number of resources to the maximum limit per resource : 
 - so servers with bigger gamemode in one resource could benefit a bigger limit (like 75 like you suggest)
 - and servers with a lot of modularity (aka a lot of resources) would be more limited to 20/30Mb ?

So yeah, this limit value will benefit some tweaking.

PichotM:
please

Snowthisway:
I would like to see this change too, we have similiar issue.

manzarek123:
I would like this too :)

Tinky124:
I would like this too :)

manups4e:
I agree

PR #689, opened 2021-03-23
Update Lovely.cpp

awesomeness1jake:
should be a bit more performant, unsure if this would break when using onesync. I just don't see why there needs to be more than 256 native invocations every frame.


PichotM:
`NETWORK_GET_NUM_CONNECTED_PLAYERS` is not working with 1s

blattersturm:
meh pr

blattersturm:
- indentation change lol
- comment that should've been an out-of-band line comment
- wouldn't even work

reopen if need be

awesomeness1jake:
Okay thanks for the feedback :) I'll only reopen if I'm certain it works and wouldn't have negative affects for non 1s and 1s, as I should've done that beforehand.

PR #691, opened 2021-04-01
feat(extra-natives-five): SET_DEFAULT_*_MULTIPLIER natives

Walsheyy:
This pull request was initially created by DelphinusBeta before it was closed.
Just to reiterate, the natives allow you to go over a multiplier of 1.0 and do not need to be called on tick and has been tested on 1604 + 2189.
The PopulateNow native helps to show the increase in vehicles when using more than 1.0

### DIFF
--- a/code/components/extra-natives-five/src/DensityNatives.cpp
+++ b/code/components/extra-natives-five/src/DensityNatives.cpp
@@ -6,12 +6,17 @@
 #include <Resource.h>
 #include <fxScripting.h>
 
+#include <GameInit.h>
+
 static float* g_pedDensity;
 static float* g_pedScenarioDensity;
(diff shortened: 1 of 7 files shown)


Disquse:
What? What's the point of duplicating PR? You didn't even marked him as a commits author. 
Moreover, I'm still confused why @DelphinusBeta closed his PR, because it never was rejected, it only needed more verification.

Walsheyy:
The pull request that @DelphinusBeta created was closed so I would presume nothing further would happen with it. I will add him as a commits author now. I wouldn't want such a great feature go to waste hence why I created this PR. Hope this is okay?

Disquse:
Original author explained why he closed his PR. This can be closed as well.

PR #693, opened 2021-04-04
feat(scripting/scheduler): add ClearTimeout

AvarianKnight:
- takes the coroutine return value from SetTimeout
- allows you to clear a timeout early like in JS
- useful (obviously) if you need to cancel a timeout

### DIFF
--- a/data/shared/citizen/scripting/lua/scheduler.lua
+++ b/data/shared/citizen/scripting/lua/scheduler.lua
@@ -227,9 +227,22 @@ function Citizen.SetTimeout(msec, callback)
 	}
 
 	hadThread = true
+	return coro
+end
+
+function Citizen.ClearTimeout(coro)
+	local coroType = type(coro)
+	if coroType == 'thread' then
+		threads[coro] = nil
+		coroutine_close(coro)
+	else
+		-- don't want to silently error if they send a invalid type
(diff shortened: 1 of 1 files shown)


AvarianKnight:
Might be better to check the type of 'coro' to see if it's a thread as if they clear old coroutine we probably don't want to still error.

AvarianKnight:
Any update on this?

blattersturm:
i don't see any update on this, no, other than your comment

gottfriedleibniz:
Random thoughts: 

You should probably be more defensive about how `ClearTimeout` works. The threads table is indirectly being exposed, e.g., `ClearTimeout(coroutine.running())`. Also, what is the expected/desired behaviour if `ClearTimeout` is invoked after `coro` has already resumed its execution?

The current scheduler implementation makes the Lua coroutine API fairly transparent. I'm not entirely sure how I feel about exposing internally created coroutines (besides through `coroutine.running`). There should be no ambiguities in the scheduler, otherwise expect random framework XYZ to exploit it.

A few things I would consider first.

AvarianKnight:
This might be something I retry later as there were quite a few changes to the scheduler since the initial conception of this, plus the fact that this is on my master branch and I'm working on some other stuff.

Thanks for all of the feedback and it will be something I take into consideration on the next pull request.

PR #696, opened 2021-04-07
fix(voip/mumble): reactivate UDP mode when possible

GlenCFL:
This allows clients to re-enable UDP transfer mode following it being disabled due to a UDP timeout. The feature requires changes to the server, but simply wont be active when communicating with older server builds.

I tried to get this `// is this a Mumble ping?` section to trigger a breakpoint throughout testing, but never managed to. That's why it's been deleted.


blattersturm:
> I tried to get this // is this a Mumble ping? section to trigger a breakpoint throughout testing, but never managed to. That's why it's been deleted.

Uh, is that even a good reason? This is for a ping sent from the Mumble client itself.

GlenCFL:
Maybe not. The comment made me think that someone also questioned why it was even present.

GlenCFL:
I can amend the commit if necessary. I also just noticed that I accidentally removed a debug print.

nta:
Yeah, please do.

Generally, however, I'm curious as to what this protocol change tries to accomplish - is it to trigger an explicit crypt channel resync in case the automatic >X loss-triggered resync does not happen?

Would there not be another underlying bug, instead?

blattersturm:
Yeah, please do. 
 
Generally, however, I'm curious as to what this protocol change tries to accomplish - is it to trigger an explicit crypt channel resync in case the automatic >X loss-triggered resync does not happen? 
 
Would there not be another underlying bug, instead?

GlenCFL:
It's actually simpler. I was looking into another issue and thought that TCP was the cause, when it's actually related to #697. That's why I also made the changes in #694. I think both are overall good changes, but neither are must haves.

In that regard, you are right, but I didn't know that personally until I learned more.

PR #697, opened 2021-04-07
fix(voip/mumble): actively poll the jitter buffer in both modes

GlenCFL:
This is my quick fix for this issue:
https://forum.cfx.re/t/fivem-voice-chat-lags-behind/2579415/

I believe the nativeAudio end does actively poll, so that's what I've configured on the other end. This might be a candidate for utilizing multiple cores, but maybe that's not worth the effort.

### DIFF
--- a/code/components/voip-mumble/include/MumbleAudioOutput.h
+++ b/code/components/voip-mumble/include/MumbleAudioOutput.h
@@ -102,6 +102,11 @@ private:
 			return isTalking;
 		}
 
+		virtual bool ShouldManagePoll()
+		{
+			return false;
+		}
+
 		virtual bool ShouldPollAudio()
 		{
 			return false;
@@ -128,6 +133,7 @@ private:
 		uint32_t lastTime;
 
 		std::mutex jitterLock;
+		std::shared_mutex pollLock;
 
 		int iLastConsume = 0;
 		int iBufferFilled = 0;
@@ -183,6 +189,11 @@ private:
 		virtual bool Valid() override
 		{
 			return context && context->destination();
+		}
+
+		virtual bool ShouldManagePoll()
+		{
+			return true;
 		}
 
 		virtual bool ShouldPollAudio()
--- a/code/components/voip-mumble/include/MumbleAudioOutput.h
+++ b/code/components/voip-mumble/include/MumbleAudioOutput.h
@@ -102,7 +102,7 @@ private:
 			return isTalking;
 		}
 
-		virtual bool ShouldManagePoll()
+		virtual bool ShouldPollAudio()
 		{
 			return false;
 		}
@@ -185,9 +185,11 @@ private:
 			return context && context->destination();
 		}
 
-		virtual bool ShouldManagePoll()
+		virtual bool ShouldPollAudio()
 		{
-			return true;
+			XAUDIO2_VOICE_STATE vs;
+			voice->GetState(&vs);
+			return vs.BuffersQueued == 0;
 		}
 
 		virtual void PushSound(int16_t* voiceBuffer, int len) override;
--- a/code/components/voip-mumble/src/MumbleAudioOutput.cpp
+++ b/code/components/voip-mumble/src/MumbleAudioOutput.cpp
@@ -261,6 +261,11 @@ void MumbleAudioOutput::ClientAudioState::OnBufferEnd(void* cxt)
 	{
 		isTalking = false;
 	}
+
+	if (ShouldPollAudio())
+	{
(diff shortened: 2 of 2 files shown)


blattersturm:
I believe an earlier implementation already worked like this, but am not sure exactly what happened to that variant and why it went back to a push-based poll trigger.

Note that this should *not* be manually triggered in case of native audio (which is why ShouldManagePoll exists) as native audio will poll/push from the RAGE audio mixer thread, so should not be polled anywhere else for it will end up out of sync.

GlenCFL:
I've noticed that your transmission does start to go out of sync due to packet loss with this solution, but that's seemingly gone immediately after. This is still preferable to what otherwise happens on production right now.

code/components/voip-mumble/include/MumbleAudioOutput.h
```diff
@@ -185,11 +180,6 @@ class MumbleAudioOutput
 			return context && context->destination();
 		}
 
-		virtual bool ShouldManagePoll()
```
blattersturm, line 188: This should be kept because as said in the comment above, NuiAudioSink.cpp will handle polling for native audio so that it runs synchronously with the game mixer.

code/components/voip-mumble/src/MumbleAudioOutput.cpp
```diff
@@ -1331,6 +1326,24 @@ void MumbleAudioOutput::ThreadFunc()
 	CoInitialize(nullptr);
 
 	InitializeAudioDevice();
+
+	while (true) {
+		const int opusFrameTime = 40;
```
blattersturm, line 1331: I don't think polling more frequently in XA2 mode would harm too much, and it may lead to less dropouts in fact.

code/components/voip-mumble/src/MumbleAudioOutput.cpp
```diff
@@ -1331,6 +1326,24 @@ void MumbleAudioOutput::ThreadFunc()
...
+
+		{
+			std::shared_lock<std::shared_mutex> _(m_clientsMutex);
+
+			for (auto& client : m_clients)
+			{
+				if (client.second)
```
blattersturm, line 1338: Add ShouldManagePoll back here as well.

code/components/voip-mumble/src/MumbleAudioOutput.cpp
```diff
@@ -1331,6 +1326,24 @@ void MumbleAudioOutput::ThreadFunc()
...
+				{
+					client.second->PollAudio((48000 / 1000) * opusFrameTime);
+				}
+			}
+		}
+
+		std::this_thread::sleep_for(std::chrono::milliseconds(opusFrameTime));
```
blattersturm, line 1345: As to the above, note that when lowering sleep frequency below ~16ms care should be taken that Sleep() or wrappers for it will no longer suffice and OS scheduling APIs (waitable timers?) should be used instead.

blattersturm (changes requested):

blattersturm:
> This is still preferable to what otherwise happens on production right now.

Wasn't the reported issue about XA2, not about RAGE audio? RAGE audio has very peculiar buffer pulling to prevent any underruns:

https://github.com/citizenfx/fivem/blob/1d339e47183cf72178ffcd0de7c4c74d29dbc2e6/code/components/extra-natives-five/src/NuiAudioSink.cpp#L1033

GlenCFL:
I believe it was regarding XA2, yes. I did overlook nativeAudio in the quick removal of that conditional.

Would you just suggest halving the frame time here to 20 then? Still above the 16ms as you mentioned.

blattersturm:
20ms would hopefully work, yeah.

GlenCFL:
There's an audible cutout at some very short interval, which can be heard very clearly when whistling, but not particularly when talking.

blattersturm:
Right. A better implementation may directly use the XA2 API to trigger a poll on buffers being empty.

GlenCFL:
I think that kills this naïve attempt.

GlenCFL:
Reopening this, as the XA2 implementation seems simple enough to just tack on.

blattersturm (approved):
LGTM.

blattersturm:
Likely going to have to back this out as it leads to a flurry of user crashes that I can't reproduce as I don't have a server with "lots" of voice use nor did users provide any infi - did you even test this at all?

GlenCFL:
I'm in the same position as you then. I tested it locally pretty extensively, but not with many players.

Go ahead and back out of it for now.

PR #698, opened 2021-04-08
feat(scripting/clr): Expanded server-sided externals

cm8263:
So two things:

1. I don't really know a whole lot of C#, so if someone that actually does could look this over, I would appreciate that. I apologize in advance for any stupid mistakes; also I have no idea what the `SecuritySafeCritical` stuff is about, so you might want to scrutinize that area particularly heavily.

2. These are more or less a (checked over for stuff that is not compatible) copy of the "normal" Externals, mainly because you don't fix what ain't broke, and also because I figured it should match what people are used to using on the client.

I made this PR after trying out some of the new OneSync server stuff and realizing I didn't like all the API calls was making lol; hopefully, I've done this correctly. Cheers.


manups4e:
is this gonna be added anytime soon?

blattersturm:
Probably not. Too much code duplication from the 'client' externals, messy commit history, would've preferred this to be done via some #if stuff in the main External/ dir.

cm8263:
> I guess the original author is also long gone so this can be closed. If they're _not_ gone they can always update it so it can be reopened, or if anyone else wants to finish this they can open a new PR.

Still here 👋, however:

> would've preferred this to be done via some #if stuff in the main External/ dir.

This sounds a little out of my scope of knowledge, so I might have to leave it be; unless you're able to elaborate a bit more for me on what exactly needs to be done, so I can take another crack at it?

PR #699, opened 2021-04-09
fix(voip/mumble): add jitter buffer management to XA2 path

GlenCFL:
This is my second take on solving the problem within #697, [referencing this forum post](https://forum.cfx.re/t/fivem-voice-chat-lags-behind/2579415), whereby the jitter buffer hoards more and more packets over time in XA2 audio mode.

I've tested this locally, between my PC and my laptop.

### DIFF
--- a/code/components/voip-mumble/include/MumbleAudioOutput.h
+++ b/code/components/voip-mumble/include/MumbleAudioOutput.h
@@ -107,7 +107,7 @@ private:
 			return false;
 		}
 
-		virtual void PollAudio(int frameCount);
+		virtual void PollAudio(int frameCount, bool jitterLocked = false);
 
 		virtual void AfterConstruct()
 		{
--- a/code/components/voip-mumble/src/MumbleAudioOutput.cpp
+++ b/code/components/voip-mumble/src/MumbleAudioOutput.cpp
@@ -687,14 +687,28 @@ void MumbleAudioOutput::HandleClientVoiceData(const MumbleUser& user, uint64_t s
 	jbp.span = numSamples;
 	jbp.timestamp = (48000 / 100) * sequence;
 
+	std::unique_lock _(client->jitterLock);
+	jitter_buffer_put(client->jitter, &jbp);
+
+	if (!client->ShouldManagePoll())
 	{
-		std::unique_lock _(client->jitterLock);
-		jitter_buffer_put(client->jitter, &jbp);
+		return;
 	}
 
-	if (client->ShouldManagePoll())
-	{
-		client->PollAudio(numSamples);
+	int startingPackets;
+	jitter_buffer_ctl(client->jitter, JITTER_BUFFER_GET_AVAILABLE_COUNT, &startingPackets);
+
+	int availablePackets = startingPackets;
+	while (availablePackets != 0) {
+		client->PollAudio(numSamples, true);
+
+		// Give the jitter buffer up to four packets to work with prior to a flush.
+		if (startingPackets <= 4)
+		{
+			break;
+		}
+
+		jitter_buffer_ctl(client->jitter, JITTER_BUFFER_GET_AVAILABLE_COUNT, &availablePackets);
 	}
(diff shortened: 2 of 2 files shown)


blattersturm:
Sure, let's try putting this one in and seeing how stable it gets.

blattersturm:
And backed out yet again. People are reporting this looping on a single HandleClientVoiceData call for over 1 second until jitter buffers for said client empties out, hitching the main thread as well if it needs to enter a mutex.

GlenCFL:
I don't really know enough about this jitter buffer to know why that would be. It was introduced in https://github.com/citizenfx/fivem/commit/db7e4899a966c7d0fc4c7fe98adb8f50c0f4ca4b, so if all else fails you could just have the XA2 path simply decode the audio and push it out immediately once again. That's what it did prior to that commit:
https://github.com/citizenfx/fivem/blob/b4654e595a60577ac42ae6578dea63226f30b732/code/components/voip-mumble/src/MumbleAudioOutput.cpp#L624-L664

blattersturm:
Is there any easy way to replicate the original issue so I can try to make a pass at it myself?

GlenCFL:
The easiest way I found is to enable Mumble loopback on your server, which is preferably running on a different computer than the client you will test on. I just edited this line to always be true:
https://github.com/citizenfx/fivem/blob/904b6c1b93c89a35d44b250634d0d22686dd7b88/code/components/voip-server-mumble/src/client.cpp#L947

I then used [Clumsy](https://jagt.github.io/clumsy/index.html) (running alongside the FiveM client) to simulate varying levels of packet drop on UDP. The reason you kind of want two computers is because Clumsy doesn't work well on localhost UDP traffic. The filter I used was:

`udp and inbound and ip.DstAddr == 192.168.1.2 and ip.SrcAddr == 192.168.1.3`

The `ip.DstAddr` being local address of the computer you're testing the client on, with the `ip.SrcAddr` being the computer running the server.

---

The server should be using OneSync, with nativeAudio disabled. The issue should be very apparent if Clumsy is working correctly, as at 15% packet loss the desync (and lack of recovery) is almost immediate.

GlenCFL:
I've still been thinking over how this could ever take one second. I think it has to be the pointer timestamp (based on the sequence number) becoming outdated, so it has to loop a ton of times in order for `jitter_buffer_get()` to get into the span of a valid packet.

I just wonder when that could happen, but I assume it would be with proximity VOIP enabled. Someone talks, goes far away on the map, talks again there, returns, and talks nearby. If you only receive voice data for nearby clients and the client itself stays alive despite being across the map, then this might be the cause.

The solution would be something like:
```c++
uint32_t targetTimestamp = 48000 / 100 * (sequence - (PACKETS_AVAILABLE * 4)); // sequence increments by 4
int ts = jitter_buffer_get_pointer_timestamp(client->jitter);

while (ts < targetTimestamp)
{
    jitter_buffer_tick(client->jitter);
    jitter_buffer_get_pointer_timestamp(client->jitter);
}
```

GlenCFL:
There's actually no mystery to the 1+ second hitch. PollAudio just takes 125-225ms, with 70-120 of that being the Opus decode itself. Dump four Opus packets in a row, add in some other CPU congestion, and there you go.

blattersturm:
That's curious - what would make Opus decoding so slow?!

GlenCFL:
I'm looking into the manner I conducted the timings a bit more. It doesn't seem right given Opus frames are 40ms.

GlenCFL:
If you push a packet into the JitterBuffer at an early sequence and then another at a much higher sequence, `PollAudio` does actually get called a proportionate amount of times to the increase in the sequence number. For instance, I pushed one packet at sequence 5000, added 40000, and the print was:

`[     93156] [         FiveM]      UV loop: mumble/ Outputting packets took 10001 iterations and 4187ms.`

The throwaway code that does this: https://pastebin.com/k3LKKyrD

PR #700, opened 2021-04-10
changes in the explanation of the function

Hiroshi-Nagaza:
restricted was separated as an optional argument,
Added Handler Function Parameters.

PR #701, opened 2021-04-10
provide command handler arguments

AvarianKnight:
this takes from #700 and doesn't change the accepted naming convention of source.


blattersturm:
Both this and the other one would be a no-op, native-doc-tooling doesn't randomly insert top-level sections that aren't defined into output.

PR #703, opened 2021-04-12
tweak(conhost): revert console and developer tools censorship on prod

iQuadCore:
Canary update channel is inherently unstable a lot of the times / sometimes fully unusable, since it was made for testing changes into the FiveM mod itself.
That's not something we would like to use on daily basis for creating and testing our own stuff (by we I mean server developers, cars / clothes / maps mod creators - none of the people I've talked to were happy about this undesirable changes), but we are now forced to do so because tools we need were disabled.
It's also important to perform testing using the same client (update channel) players will later use, so this request reverts console and dev tools blocking on production client.

### DIFF
--- a/code/components/conhost-v2/src/ConsoleHostGui.cpp
+++ b/code/components/conhost-v2/src/ConsoleHostGui.cpp
@@ -754,6 +754,7 @@ static void EnsureConsoles()
 	}
 }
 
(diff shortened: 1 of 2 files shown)


blattersturm:
`+set moo 31337`?

PR #704, opened 2021-04-16
tweak(gamestate/server): lock grid mutex when it's really necessary

warxander:


blattersturm:
How often do you encounter entities out of this range?

warxander:
I don't know, but does it matter?
This tweak is logic-independent, I would say it's more about code style.

PR #710, opened 2021-05-01
increase vehicle limit

utkuali:
not sure if it works, would be great to be able to increase vehicle limit size to create real parking script


Technoblazed:
Spoiler alert, if it worked it would have already been done.

Disquse:
Why would you try to PR an untested changes? Do you want to break FiveM for everyone or what?

blattersturm:
> not sure if it works

We're not your personal testing crew.

PR #712, opened 2021-05-06
feat(vfs/posix): add make hard link control code support

darkshoz:


code/components/vfs-impl-server/src/PlatformDevice.Posix.cpp
```diff
@@ -219,6 +220,29 @@ bool LocalDevice::ExtensionCtl(int controlIdx, void* controlData, size_t control
 
 		return true;
 	}
+	else if (controlIdx == VFS_MAKE_HARDLINK && controlSize == sizeof(MakeHardLinkExtension))
+	{
+		auto data = reinterpret_cast<MakeHardLinkExtension*>(controlData);
+
+		auto filename = realpath(data->existingPath.c_str(), NULL);
```
blattersturm, line 227: >If resolved_path is specified as NULL, then realpath() uses
>malloc(3) to allocate a buffer of up to PATH_MAX bytes to hold
>the resolved pathname, and returns a pointer to this buffer.  **The**
>**caller should deallocate this buffer using free(3).**

This'd leak memory.

blattersturm (changes requested):
Looks sound, minus the memory leak.

darkshoz:
My bad, the memory leak should be resolved.

blattersturm (approved):

blattersturm:
Just found this doesn't even.. build. Missing the hard-link extension header.

Perhaps I should stop assuming people tested their PRs.

PR #713, opened 2021-05-07
feat(onesync): door and control data node research

Hendreik:


gottfriedleibniz:
Would you happen to have an Xbox dump of V? If so, you can ignore me. 

Otherwise, it contains debugging strings to most sync nodes (albeit not one-to-one with the current iteration of V). It will greatly reduce the amount of 'guessing' that needs to be done on field names and the naming convention of the node fields can follow those debug strings.

Hendreik:
Nope, although that sounds like it would be great to have :)

gottfriedleibniz:
I found this on the internet: https://pastebin.com/raw/iaehSAws. It is clearly not mine as I do not use Ghidra. It looks like most of the SyncNode code dumped. 

Also found on the same "internet" are the Xbox netevent definitions: https://pastebin.com/raw/svJ87khi. Clearly not mine, as, like above, I do not use Ghidra.

PR #719, opened 2021-05-21
Update GetNumPlayerIndices.md

KonstantinosKoufopoulos:


ext/native-decls/GetNumPlayerIndices.md
```diff
@@ -10,3 +10,4 @@ int GET_NUM_PLAYER_INDICES();
 
 
 ## Return value
+Returns the player count on the server.
```
jaymo1011, line 13: The grammar of this is incorrect and the explanation itself could be misleading. `GET_NUM_PLAYER_INDICES` returns the number of clients with assigned net ids. Clients who are still connecting to the server (including deferred connections) are **not** counted.

blattersturm:
Abandoned PR, will be taken in with the big native review.

PR #721, opened 2021-05-28
Fixed Lag Spikes on Loader Startup

ScopesDK:
Fixed Lag Spikes on Loader Startup
Also PrintFv Cointians much more PCU Usages then std cout


blattersturm:
This code isn't even valid nor does it make any sense. Maybe stick to UC?

PR #722, opened 2021-05-31
Removal of 'get ped stat index' function to return MP0 at all times

awesomeness1jake:


blattersturm:
Good job making a PR that'd break a lot of intended behavior? Now _nothing_ will return MP0 at _any_ time.

awesomeness1jake:
So its intended to breaks the stats for the sp characters? If you don't mind explaining how would nothing return MP0?

blattersturm:
> If you don't mind explaining

image

Since you do mind explaining things, I will opt to as well.

PR #723, opened 2021-05-31
feat(onesync): ped task tree node parsing and natives

Hendreik:
Add parsing for the CPedTaskTreeDataNode and adds three getter natives for tasks

### DIFF
--- a/ext/native-decls/GetPedScriptTaskCommand.md
+++ b/ext/native-decls/GetPedScriptTaskCommand.md
@@ -1,0 +1,17 @@
+---
+ns: CFX
+apiset: server
+---
+## GET_PED_SCRIPT_TASK_COMMAND
+
+```c
+Hash GET_PED_SCRIPT_TASK_COMMAND(Ped ped);
+```
+
+Gets the script task command currently assigned to the ped. Tasks: https://alloc8or.re/gta5/doc/enums/eScriptTaskHash.txt
+
+## Parameters
+* **ped**: The target ped.
+
+## Return value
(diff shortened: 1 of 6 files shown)


ext/native-decls/GetPedScriptTaskCommand.md
```diff
@@ -0,0 +1,17 @@
...
+ns: CFX
+apiset: server
+---
+## GET_PED_SCRIPT_TASK_COMMAND
+
+```c
+Hash GET_PED_SCRIPT_TASK_COMMAND(Ped ped);
+```
+
+Gets the script task command currently assigned to the ped. Tasks: https://alloc8or.re/gta5/doc/enums/eScriptTaskHash.txt
```
blattersturm, line 11: I'd prefer not externally sourcing such definitions. @gottfriedleibniz do you have any thoughts on this?

gottfriedleibniz: Agreed. The native repository has left in most links to alloc8or's `/doc/enum/` path. However, since the changes to doc-tooling for enums/annotations, the policy on external references should follow.

My only concern would citations and giving people credit for their work, e.g., while eScriptTaskHash is in game-code, those hashes still needed to be reversed. ... I very much like avoiding stepping on one's toes (:

Hendreik: Sure, I don't want to copy their research just in case of any issues with permission. If you want me to copy it internally, I can

PR #725, opened 2021-06-01
Change alpine integrated run.sh

routmoute:
For create a docker container from scratch with your alpine folder, this run.sh don't work anymore. This probably a default library path problem. Manually setted, it's ok.

### DIFF
--- a/code/tools/ci/run.sh
+++ b/code/tools/ci/run.sh
@@ -3,4 +3,6 @@
 cd /opt/cfx-server
 [ -d cache ] || mkdir cache
 
-exec /opt/cfx-server/FXServer $SERVER_ARGS $*
+exec /opt/cfx-server/ld-musl-x86_64.so.1 \
+    --library-path "/usr/lib/v8/:/lib/:/usr/lib/" -- \
+    /opt/cfx-server/FXServer +set citizen_dir /opt/cfx-server/citizen/ $SERVER_ARGS $*


blattersturm:
yeah the run.sh you're mentioning is literally the one here: https://github.com/citizenfx/fivem/blob/master/data/server_proot/run.sh

the one in code/tools/ci is used for entirely different purposes and should not be that one, it's meant for directly running on alpine

routmoute:
This run.sh is in fx.tar.xz/alpine/opt/cfx-server/run.sh

I don't want use distro on distro but I don't want modify files too and path in alpine external run.sh have not good path if not modified

routmoute:
the run.sh in tar archive force to run it at external folder to "alpine", if we modify this placement, the path is different, its content must be different.

blattersturm:
Yeah, 'if you modify this placement', why should _we_ as the project maintainers care about where _you_ 'modify this placement'?

If you're running in real Alpine, running `/opt/cfx-server/FXServer` by itself works fine.

routmoute:
It doesn't work: exit code 139 (segmentation fault)
I think it's problem with default library path..

blattersturm:
> I think it's problem with default library path..

You apparently suck at 'thinking'.

blattersturm:
Underlying issue is these two lines FWIW:

https://github.com/citizenfx/fivem/blob/master/code/server/launcher/src/ServerMiniDump.cpp#L487-L488

If running directly on Alpine in a non-debug build, this will have `tgtDyn` be NULL.

routmoute:
It work with this modified run.sh

blattersturm:
OK? That's still not a fix though, and that rather breaks other systems that use native Alpine.

routmoute:
Thank you very much for your aimability and very good help !

blattersturm:
Here's some working steps for making a container by the way:

```bash
mkdir -p /tmp/ducker
cd /tmp/ducker
curl -LO https://runtime.fivem.net/artifacts/fivem/build_proot_linux/master/4012-46d76f5d2b9b6752aebf365c334b46e7f0d9915f/fx.tar.xz
tar xvf fx.tar.xz
tar cv -C alpine/ . | docker import --change "CMD /opt/cfx-server/FXServer" - fxserver:4012
```

```
docker run -it --rm fxserver:4012
```

routmoute:
4012 works fine, thank you: https://github.com/routmoute/fxserver/blob/dev/Dockerfile

PR #726, opened 2021-06-01
perf(scheduler): performance improvements

pogue-mahoney:
reduced thread indexing using stack/closures

array iteration

thread object reuse/culling

lazy thread creation

### DIFF
--- a/data/shared/citizen/scripting/lua/scheduler.lua
+++ b/data/shared/citizen/scripting/lua/scheduler.lua
@@ -1,6 +1,6 @@
 local GetGameTimer = GetGameTimer
 local _sbs = Citizen.SubmitBoundaryStart
-local coresume, costatus = coroutine.resume, coroutine.status
+local coresume, costatus, tremove = coroutine.resume, coroutine.status, table.remove
 local debug = debug
 local coroutine_close = coroutine.close or (function(c) end) -- 5.3 compatibility
 local hadThread = false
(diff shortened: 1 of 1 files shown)


blattersturm:
Is there any explanation as to this? How did you verify the changes still work as expected with 100% compatibility? Any reasoning behind these changes?

This kind of change set is super risky to merge and therefore we would really prefer to have some more substantiation.

pogue-mahoney:
There were some issues with my previous strategy as you pointed out there might be.  This time I used a black box approach to thinking about the changes, and ensuring all entry points would experience no impact, as well as taking an extended period for evaluation.  

Using array iteration on tick was still a focus here, as was relying upon runningCoroutine and closures in Await to reduce the need to index threads at all there.

Thread reuse and culling, as well as lazy thread creation (in other words, waiting to create a thread object when we know we will need to resume it) both cut down own table creation and indexing and garbage collection.

blattersturm:
Ah, right, you had a PR before. Do you have any comparisons in regards to allocation count, time taken as well as GC delay on both 5.3 and 5.4?

Complexity increase might still be a bit odd for maintainability.

pogue-mahoney:
> Do you have any comparisons in regards to allocation count, time taken as well as GC pressure on both 5.3 and 5.4?
To be honest I'm not aware of that diagnostic capability.  My testing focused on reducing ticks in both greedy and passive components.
 
> Complexity increase might still be a bit odd for maintainability.
I would argue this represents reductions in complexity in terms of the need for newThreads and focuses the role of the threads table on tick iteration.  It's your code to maintain of course.

gottfriedleibniz:
Looking at this scheduler, the times are about the same as my MLQ implementation (a linked list). Similar idea to avoid the `lua_next` and function overhead of `pairs`.

...

Rephrasing the other problem: resources that create many shortly lived threads, e.g., esx/esx-reborn using `Citizen.SetTimeout`, are unnecessarily creating additional `objects` (wakeTime/boundary/name/etc) as those Lua tables can be recycled instead of being thrown to the garbage collector. This idea is also discussed/recommended in [Lua Performance Tips](https://www.lua.org/gems/sample.pdf).

Your approach: When a coroutine dies, the `object` is temporarily saved for some fixed time; intended to be recycled. After `time` has passed all dead `objects` in the threads array are removed (`table.remove`). My thoughts (bullet-point form):

* A (doubly) linked-list is more flexible and still offers O(1) access times. In addition, Lua 5.4 includes fast-pathing mechanisms for access tables with integers.

* Use a fixed-size pool: when a thread dies move the `object` to a fixed sized pool. If the pool is full throw it away (... to the garbage collector). This becomes trivial to implement with a linked-list.

* When it comes to scheduling (in general): amortizing costs across each 'tick' is superior than one tick having additional duties of unknown significance/overhead. You must really consider *worst* case scenario, the Lua ecosystem very much follows Murphy's law.

If you want to continue down this avenue, I can give you my old MLQ implementation. However, for this PR to be considered some data/justification/proof-of-worth will be required. For example, extract the scheduling code to your own Lua environment (emulate the few things required as needed, e.g., `GetGameTimer`), and then collect data that way.

PR #729, opened 2021-06-04
Updates for build instructions

Jamelele:


blattersturm:
Can you folks stop trying to touch this document with weird opinions?

Jamelele:
What is opinionated about these changes?

The first fix was to make the example consistent with the instructions above to use Boost 1.71.0 and obviously the symlink instructions are out of date as the 'cache' folder has since been renamed.

blattersturm:
I'm not even sure why the symlink instructions are there in the first place as they're unneeded and explicitly wrong. Most the document is wrong and outdated since the 'fxd' tool exists, in fact, and it's more effort to approve and merge a PR than would be worth it especially for 'tiny' and irrelevant changes.

Jamelele:
Never mind then, I wasn't aware this document was so outdated

PR #730, opened 2021-06-07
[webadmin]: Update bootsrap.min.css: little bit more space h1

Konijntje-x:
Under H1 (Console, FXServer, but unconfigered or other name etc.) there is a little bit more space now. 
It's more spacious.


blattersturm:
... uh?

PR #733, opened 2021-06-09
fix(dev/fxd): create missing dirs on download

QTBertchen:
On an fresh install the folder code/build/ doesn't exist and so curl needs to create the missing folders


QTBertchen:
When following the instructions from [fivem/building.md](https://github.com/citizenfx/fivem/blob/master/docs/building.md), they fail at this step:
```
:: downloads the right Chrome version for 64-bit projects
fxd get-chrome
```
With error:
grafik

As far as I know this is because the folder structure code/build/ doesn't exist for a newly cloned repository.
Possible fixes are either to pass curl the extra argument --create-dirs
Or something like this:
```
if (!(Test-Path "$SaveDir")) {
	mkdir "$SaveDir"
}
```

PR #736, opened 2021-06-11
feat(scheduler.lua): add `require` function to load scripts at runtime (Lazy Loading)   

0xWaleed:
This `require` function will load scripts at runtime, it works like lua's original `require` function that is being disable by CFX
for security reasons. this will use CFX natives to load scripts instead of adding them to `client_script` or `server_script` entries.
Also this will help to do lazy loading scripts on demand.

usage: will be the same as lua `require` function. we don't use `/` for directory separator not the .lua extension to make it same as lua on other platforms.
```lua
require('resource-name.dir.dir.script') 
```

As per the cfx documentation with **client** side, you need to add the file to `file` entry in fxmanifest in order to load it at runtime.

Thank you _blattersturm_ for your points for my previous PR, I changed what you mentioned in order to meet the requirements.
feel free to add changes that I need to work on.  

#735

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -194,5 +194,3 @@ privates_config.lua
 
 # python files
 *.pyc
-
-.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -194,3 +194,5 @@ privates_config.lua
 
 # python files
 *.pyc
+
+.idea/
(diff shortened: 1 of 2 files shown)


blattersturm:
_again?_

PR #737, opened 2021-06-12
feat(natives/five): network world state control for ropes

Hendreik:
Unlike other natives which implement world state, ADD_ROPE does not have a parameter to whether you want the rope networked. Instead it relies on a net game check whether to create an underlying CNetworkRopeWorldStateData. As network world state is currently not implemented, this leads to a memory leak and pool being filled upon multiple rope creations. This native aims to solve two problems:

- Prevent the issue aforementioned
- Maintain backwards compatibility for future, as if world state was to be implemented, all ropes (which people assumed to be non-networked) would suddenly be networked without this

### DIFF
--- a/code/components/extra-natives-five/src/NativeEdits.cpp
+++ b/code/components/extra-natives-five/src/NativeEdits.cpp
@@ -1,0 +1,27 @@
+#include <StdInc.h>
+#include <GameInit.h>
+#include <Hooking.h>
+#include <ScriptEngine.h>
+
+static bool* g_ropesCreateNetworkWorldState;
+
+static HookFunction hookFunction([]()
+{
+	g_ropesCreateNetworkWorldState = (bool*)hook::AllocateStubMemory(1);
+	*g_ropesCreateNetworkWorldState = false;
+
+	// replace ADD_ROPE native net game check whether to create CNetworkRopeWorldStateData
+	auto location = hook::get_pattern<uint32_t>("80 3D ? ? ? ? ? 74 71 E8 ? ? ? ? 48", 2);
+	*location = (intptr_t)g_ropesCreateNetworkWorldState - (intptr_t)location - 5;
+
+	fx::ScriptEngine::RegisterNativeHandler("SET_ROPES_CREATE_NETWORK_WORLD_STATE", [](fx::ScriptContext& context)
+	{
+		bool shouldCreate = context.GetArgument<bool>(0);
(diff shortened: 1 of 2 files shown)


blattersturm:
What, actually, is currently breaking network world state events? Every so often I believe someone hits one of these being problematic except I assumed the underlying issue for these was fixed quite some time ago. 

Do they not replay out of scope, or some similar obvious missed step? This is not too relevant in RDR, funnily, as a lot of formerly-world state/array handler entries are now minimal network objects. I'm hoping this won't bite back/be an issue in 2022 or whenever V's refactor hits PC.

Hendreik:
I believe it has some requirement for you to be a script host, though I have not entirely confirmed that. I had a go at trying to fix it but I didn't get too far, heres some of my research if it helps: https://gist.github.com/Hendreik/4c58274bcf39dd1d70de51e8b387a692

Hendreik:
Did some research, world state is reliant on there being a script host player. A few common functions show this:

Note: isEvent represents whether this came from an event, otherwise it came from the game.

`CNetworkWorldStateData::Change(bool isEvent) - 0x141003FDC`
Called when a state is to be changed (typically state creation)
If I am script host:
- isEvent: Change it locally. CNetworkWorldStateData::DoUpdate will send this to all players later
- !isEvent: Change it locally. CNetworkWorldStateData::DoUpdate will send this to all players later

If I am not script host:
- isEvent: Change it locally
- !isEvent: Change it locally then send an event to script host to change it. If no script host exists, add to pending array

`CNetworkWorldStateData::Revert(bool isEvent) - 0x141039958`
Called when a state is to be reverted (typically state deletion)
If I am script host:
- isEvent: Revert it locally then send event to all players to revert it
- !isEvent: Revert it locally then send event to all players to revert it. Also do some log

If I am not script host:
- isEvent: Revert it locally
- !isEvent: Send event to script host to revert it. If no script host exists, add to pending array

`static CNetworkWorldStateData::PlayerJoined(void* player) - 0x14102C2E0`
Called when a player joins - If I am script host then send event to joining player with my world state

`static CNetworkWorldstateData::DoUpdate() - 0x141045B50`
Called by rage::CNetworkObjMgr::DoUpdate(). Handles who has control, pending state and state expiry

blattersturm:
Ah, one of these derivative PlayerJoined listeners. That makes sense, perhaps some typical faking out can be done there.

PR #739, opened 2021-06-13
changed source type back to string

lurkmorr:
cool that source is number, but all the natives that require source, assume that it's string


blattersturm:
How does changing back a type definition change/fix this in any way? It still is a type mismatch.

blattersturm:
If you want this to be _correct_ and consistent, make the _native definitions_ correctly aware of the `number | string` property of player indices.

PR #740, opened 2021-06-15
Added: Dutch translations for the new friendly connection errors

DonTosti:
In this commit I've added Dutch translations for the new friendly connection errors.


blattersturm:
Translations go via the translation service which may be a bit behind. I've asked @technetium-cfx to look at this multiple times by now.

PR #742, opened 2021-06-18
Uptade native information

FreakOutModz:
Additional information about how to use the native and what does it return in value. Return a value between 0.0 and 10.0 have been tested on almost every GTA cars / planes / boats. Most common vehicle use by player will return 5.0 and some vehicles will have a 0.0 for some reason.

Information use:

- https://imgur.com/a/UTnBryk

Code use:

- https://pastebin.com/2cSy35zW

Car get from:

- https://www.gtabase.com/grand-theft-auto-v/vehicles/

### DIFF
--- a/ext/native-decls/GetVehicleOilLevel.md
+++ b/ext/native-decls/GetVehicleOilLevel.md
@@ -5,6 +5,7 @@ apiset: client
 ## GET_VEHICLE_OIL_LEVEL
 
 ```c
+// 0xFC7F8EF4
 float GET_VEHICLE_OIL_LEVEL(Vehicle vehicle);
 ```
 
@@ -13,3 +14,12 @@ float GET_VEHICLE_OIL_LEVEL(Vehicle vehicle);
 * **vehicle**: 
 
 ## Return value
+Return a value between 0.0 and 10.0 representing the current oil level.
+
+## Examples
+```lua
+-- A short example of how to print value of oil level in Lua
+local vehicle = GetVehiclePedIsIn(PlayerPedId(), false)
+local oilLvl = GetVehicleOilLevel(vehicle)
+print(oilLvl)
+```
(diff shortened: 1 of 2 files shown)


ext/native-decls/GetVehicleOilLevel.md
```diff
@@ -5,6 +5,7 @@ apiset: client
 ## GET_VEHICLE_OIL_LEVEL
 
 ```c
+// 0xFC7F8EF4
```
blattersturm, line 8: Why does this have to be there?

ext/native-decls/GetVehicleOilLevel.md
```diff
@@ -13,3 +14,12 @@ float GET_VEHICLE_OIL_LEVEL(Vehicle vehicle);
 * **vehicle**: 
 
 ## Return value
+Return a value between 0.0 and 10.0 representing the current oil level.
+
+## Examples
+```lua
+-- A short example of how to print value of oil level in Lua
```
blattersturm, line 21: 'A short example printing the oil level value in Lua', maybe?

blattersturm (changes requested):

ext/native-decls/GetVehicleOilLevel.md
FreakOutModz: If I take a look in https://github.com/citizenfx/natives/blob/master/.github/CONTRIBUTING.md I replace the information from:
```c
// 0x0123456789ABCDEF 0x1234ABCD
BOOL _EXAMPLE_NATIVE(int param);
```
so I replace by
```c
// 0xFC7F8EF4
float GET_VEHICLE_OIL_LEVEL(Vehicle vehicle);
```
If not necessary, I would make the change to remove it and take note of it in the future.

FreakOutModz:
First of all, thank you for the feedback. This is my first time improving a native. I just changed the file based on your comments made on the first proposal. #742

blattersturm:
Abandoned PR, same as the other 'uptade', may be used as reference material during the native decl review.

PR #743, opened 2021-06-18
Uptade native SetVehicleOilLevel

FreakOutModz:
Additional information about how to use the native and description. Set a value between 0.0 and 10.0 have been tested on almost every GTA cars / planes / boats and it's the most common way used by default in game. Even with 0.0, vehicle can still work perfectly. 

Code use:
https://pastebin.com/3mHiYmTn
Car tested from:
https://www.gtabase.com/grand-theft-auto-v/vehicles/
Tested server:
build 2189 | Canary More
FiveM last client Uptade from 20 h 58 / 2021-17-06 (UTC−4)


blattersturm:
Why is this a separate PR from your other 'uptade'?

PR #744, opened 2021-06-18
WebAssembly scripting runtime

ZOTTCE:
# Summary

The goal of this project is make WebAssembly runtime for FiveM. With this runtime we add ability for end-users to make fast scripts with languages supporting compiling in WASM (like Rust, Go, C, AssemblyScript). It is an opportunity to bring in new developers and programming paradigms.

# Explanation

Current functionality:

- Invoke native functions
- Subscribe to events
- Call ref functions
- Communicate with another runtimes / scripts.

There is already the v8 engine that is capable of using WebAssembly modules. When we develop a script we mostly use native functions, so, if we choose to write bindings for JS it makes no sense, we will spend all the time in JS runtime and convert objects from JS to our RT and back. It is costly and ugly method. Also most of the mentioned languages is strongly typed and this requires to make some hack for functions like `console.log` (just look at [`web-sys`](https://rustwasm.github.io/wasm-bindgen/api/web_sys/console/index.html) Rust crate), `emit`, callbacks in `on/onNet`. If I got it right currently FiveM contains old version of Node.js (~12) and v8 itself so it is better to have one shared runtime that can be updated by change one line Cargo.toml.

With this implementation of WASM scripting runtime we remove the JS (v8) layer and deal directly with FiveM. We don't forced to make some wrappers to work with JS functions, no variadic functions, no additional cost. Just raw values and message pack encoder / decoder. It is fast as well optimized v8 (in many years) and outperforms with event handling.
https://imgur.com/gbq0th5

The runtime exposes small set of API to work with FiveM functions and expect some exports from scripts.

## API

More details can be found in [the implementation](https://github.com/ZOTTCE/fivem-wasm/blob/master/runtime/src/lib.rs) of runtime and [bindings for Rust](https://github.com/ZOTTCE/fivem-wasm/tree/master/bindings/core).

### Host

Module name: `host`

- `log` - logs a message in the console
- `invoke` - invokes native function
- `canonicalize_ref` - performs name transformation
- `invoke_ref_func` - calls ref functions

### Script

- `_start` - entry point (default WASM name)
- `__cfx_on_event` - callback for incoming events
- `__cfx_on_tick` - runtime calls scripts on each tick
- `__cfx_call_ref` - another script wants to call execute our function
- `__cfx_duplicate_ref`
- `__cfx_remove_ref`
- `__cfx_alloc` - runtime allocates some memory in script memory
- `__cfx_dealloc` - deallocate that has been allocated my us

## About the runtime implementation

So, why Rust if there is already C bindings for `wasmtime`?

- I have experience with it. For me it's best to write in Rust than make a billion mistakes in C/C++
- We can choose any WASM intepreter that is available for Rust (`wasmtime`, `wasmer`, `wasmi`)
- FiveM has a Rust crate and builds it
- Cargo makes working with dependencies easier

In the end we can simply ship prebuilt static libraries to remove Rust compiler as it would be if use `wasmtime` bindings with its libraries.

# Future work
- [ ] Select crates that should be in the main repo
    - [x] ~~Runtime~~
    - [x] ~~Glue~~
    - [ ] Bindings?
- [x] Split the runtime crate from bindings
- [ ] Decide to allow or disallow access to filesystem from scripts (we can select directories)
- [x] ~~Include a crate in build process (I am not familiar with it)~~
- [ ] Make some changes in component code to make sure it is idiomatic
- [ ] Write more bindings (or even framework-like library) in Rust (I want to support it). I haven't done much because don't want to spend time on this project if it will not be merged


blattersturm:
Hi,

While the effort is appreciated, it's a shame to see you've not done much with the comments when you first mentioned in chat you'd try to work on this: I don't think it's a good idea to add a dependency to a random submodule written in a random language with custom code as someone will have to maintain it, and we can't rely on you or anyone else being available longer-term if any issues occur with this.

Since the `wasmtime` library you used does have C ABI exports, it's a shame to still see you _not using them_.

Similarly, even if we'd add binding generators for Rust to the main project, we'd like to see everything project-specific be in-tree and in a scope _we_ can control, not some submodule owned by someone external which may vanish or otherwise change at any time, nor writing things in a way inconsistent with the rest of the project (non-opaque library import in an unrelated language not used by the project at this time, API binding generation via some other means than `ext/natives/` codegen and therefore inconsistent with any future system changes).

blattersturm:
As an addendum to the last comment: I see we still have the `jexl-eval` thing in-tree, this was done mainly as the `jexl-rs` library was the only library of its class and it didn't already expose C bindings. It's currently still an optional dependency and the build process will fail gracefully if this component didn't build, which might be more difficult to ensure for a component implementing an ScRT.

I'm not sure how clean any C++ ABI support for Rust is at this time either as I've not used it much at all (other than for writing the `jexl-eval` binding), but at least importing _actual_ code rather than relying on a submodule would be a core priority for this to be workable. :)

ZOTTCE:
Thanks for the response @blattersturm

As I said I haven't used C / C++ because I have no experience with it and it will be a bad decision to write this component in C++ because of this (I have no clue how to manipulate memory safe and idiomatic).

Yes, I got your point about maintaining, but what can we do about it?
My code can be a part of your codebase. Here is options:
* It can be in this repo like `jexl-eval`
* I can transfer ownership to @citizenfx organization and it will remain like a git submodule

About bindings and native generator I am not sure if they should be here. It is an option for people choosing write scripts in Rust.
It could be nice to have a primary language with first-class support for the WebAssembly runtime, but if you have no Rust developers I don't think we can consider current bindings as a part of the runtime. The generator itself can be rewritten in lua like the rest, but it don't solve that I wrote above.

I don't know do you have like experimental features feature in FiveM, I think the runtime could live there for some time.

blattersturm:
> As I said I haven't used C / C++ because I have no experience with it and it will be a bad decision to write this component in C++ because of this (I have no clue how to manipulate memory safe and idiomatic).

Understandable in a way, I see you've tried to make the C ABI part as minimal as would be workable.

> * It can be in this repo like `jexl-eval`

This would be generally preferred, yes.

> About bindings and native generator I am not sure if they should be here. It is an option for people choosing write scripts in Rust.
> It could be nice to have a primary language with first-class support for the WebAssembly runtime, but if you have no Rust developers I don't think we can consider current bindings as a part of the runtime. The generator itself can be rewritten in lua like the rest, but it don't solve that I wrote above.

I still feel that generators could be best off in-tree as well, so no matter what you end up doing in the future, if you give up or move to other projects, it'll still be able to be maintained by others or us.

I'm not inherently opposed to using Rust, but it's more of a chicken-and-egg problem: if I don't have any use for Rust, I won't have any place to use it or do anything with it. If something comes up with an in-tree bit of code, I at least have some reason to look into Rust code and get more habituated with the language and its tooling.

> I don't know do you have like experimental features feature in FiveM, I think the runtime could live there for some time.

Generally the `ext/` subdirectory is for things like this - there's no more advanced feature flag system or build-time toggling/incubator as there hasn't been much need for such yet, but anything that doesn't fit well in the normal code/ tree usually goes in ext/.

Since the whole project is generally considered experimental and subject to change, I'd feel any code would be fine living within the main repo, perhaps with a disclaimer of 'this will change in the future, it's not settled yet' on top. At least ScRTs would only harm the build process and servers which use them, as instantiation is done only after a HandlesFile call succeeds.

ZOTTCE:
So, should I consider your messages like you're ok with this idea and I can continue working on this?

I may propose the next plan:
* Move `runtime` crate to `ext/wasm-runtime`
* Add steps to build the crate in build tools (but, I guess, I will need some help)
* Rewrite generator using lua and put it along with others

Still ... In my opinion, binding developers should be responsible for generating natives for their languages (it is not a part of the runtime, it's just wrappers around `host::invoke` runtime function). When I wrote this crate I thought it would be cool if FiveM distributes JSON file with list of natives and their arguments. I've seen `codegen_out_json.lua` but it loses some data if I remember it right.

ZOTTCE:
I've made some changes. 

- Add check list to resolve a few questions
- Now source code is in the main repo
- Remove bindings for now

I think `bindings` crate should have dedicated repo and live there. Because as I said it is not a part of the runtime. The advantage I see to include it in the main repo is to have fresh bindings to natives. It can be resolved by uploading *.lua definitions on your servers with every release.

I need your opinion about points in Future work.

@blattersturm

ZOTTCE:
I guess, I've done my work here. `ext/wasm-runtime` now it looks like it has one responsible and can be used from another language. That's the reason why I haven't included it in this repo.

The last part is to add the Rust compiler in build tools.

btw I still waiting for your response

blattersturm:
I'll see about a response soon, this is a bit non-trivial to review so it may take some time.

kngrektor:
IMO the ABI needs to have a rigorous specification. I believe that it currently breaks Rust assumptions (since it reuses event names). Also employing tricks like having strings both be null terminated and sized would be nice.

ZOTTCE:
> IMO the ABI needs to have a rigorous specification. 

Most runtimes and compilers don't implement Reference Types WASM proposal yet. We can use only primitives.

> I believe that it currently breaks Rust assumptions (since it reuses event names). 

Which assumptions and what part of the runtime breaks it? Don't get about reusing event names.

> Also employing tricks like having strings both be null terminated and sized would be nice.

Why should we have both? C strings are fine.

kngrektor:
> Most runtimes and compilers don't implement Reference Types WASM proposal yet. We can use only primitives.
You've already implicitly defined an ABI in your implementation, might as well write a proper-er spec for it
> Which assumptions and what part of the runtime breaks it? Don't get about reusing event names.
Let's say you're in the event handler for an event with a longer name (e.g. `longer`) and trigger the an event with a shorter name (e.g. `short`). If the resource that triggered the event you'll have entered it again. After the shorter event has been fully triggered you'll return to the original event handler, except now it'll see the shorter event name (e.g. `short` instead of `longer`). irc this is / can lead to undefined behavior in Rust.
> Why should we have both? C strings are fine.
As then you could use https://doc.rust-lang.org/std/string/struct.String.html#method.from_raw_parts directly which is faster. Both length and null terminated is the best of both worlds. Length could even be useful in c world (strlen is slow).

ZOTTCE:
> You've already implicitly defined an ABI in your implementation, might as well write a proper-er spec for it

Yeah, my bad. It is in `readme.md` now.

> Let's say you're in the event handler for an event with a longer name (e.g. longer) and trigger the an event with a shorter name (e.g. short). If the resource that triggered the event you'll have entered it again. After the shorter event has been fully triggered you'll return to the original event handler, except now it'll see the shorter event name (e.g. short instead of longer). irc this is / can lead to undefined behavior in Rust.

Maybe you're right. Will look at it. There are two options: allocate memory for each event (like I did the first time). Or leave it and make a note about it. It is poor design to call your own events in scripts (isn't it?), but yes there can be situations like this (trigger an external event that calls your event in a response). Anyway, that is a good catch, thank you.

> As then you could use https://doc.rust-lang.org/std/string/struct.String.html#method.from_raw_parts directly which is faster. Both length and null terminated is the best of both worlds. Length could even be useful in c world (strlen is slow).

The runtime itself doesn't use passed strings. It just gets a pointer in the host memory space and gives it to the CitizenFX (log, invoke).

ZOTTCE:
Any updates?

Yum1x:
Waiting for this merge 😍

ZOTTCE:
ok gg

blattersturm:
_You're_ the one who didn't reply to feedback. What are you expecting?

blattersturm:
Also please unpublish/otherwise disown your hijacked un-namespaced 'crate' names (`cfx-*`) on this 'crates.io' system since you're violating the ToS, term 24:

> The Users shall not represent themselves, any entity they represent, or a product they provide as being endorsed by or affiliated with CitizenFX (or related brands such as FiveM), either in person, or on a uniquely-namespaced service, including but not limited to domain names (“fivem.tld”, “citizenfx.tld”, “fivem-plugins.tld”), social media (“socialspace.com/users/citizenfx”, “social.net/fivem”), or product branding (“FiveM Admin Tool”, “FXInstaller”, “CitizenFX”). Correct alternatives would be similar to “scripts-for-fivem.tld”, “socialspace.com/users/cool-rp- fivem”, “Admin Tool for FiveM” or “Installer for FXServer”.

ZOTTCE:
> You're the one who didn't reply to feedback. What are you expecting?
what feedback?

Roms1383:
damn that's sad that it turns out this way guys, this PR is a super initiative IMHO

blattersturm:
> damn that's sad that it turns out this way guys, this PR is a super initiative IMHO

Hm? This PR doesn't allow you to do anything more than you already can do.

Roms1383:
Well, I'm gonna be very careful with what I answer here because despite being a career developer IRL :
- I don't come from a low-level background (and yes, mingling around with pointers is unknown land to me)
- I'm not a game developer (and yes I have very low knowledge about this, my field being about software development)

Also, I by no means want people to think that I disregard their favorite language / tool, or the tremendous effort that FiveM community has already put in making it the awesome tool that it has become.

So first I would reply that : because everybody love to develop with his/her favorite tool. That's probably why there's support for C# and JS in the first place, right ? Otherwise it could as well have kept being a C++ only ecosystem.

IMHO Rust could bring a lot of benefit to the FiveM ecosystem. Its design, rich types system and features can allow to :
- experiment very low-level and unsafe things in a **safe way** while doing (almost) no concession regarding speed and performances
- iterate faster on building more and more features : because Rust is like this, it can be hard to roll out the foundations but the more libs and traits you have the easier you can reuse and build on top of these (especially talking about Rust `traits` here)
- interop with other languages, at least more easily (allowing to take advantage of existing codebase in different languages while using Rust)

On the other hand, WASM is made to be interoperable with JS right from the start, so onboarding it would allow for existing JS developers to take advantages of libs built in WASM, while still enjoying coding in their favorite language.

I know my answer will probably not convince everybody, but I personally have the feeling that these are definitely a nice addition to the FiveM ecosystem, without even speaking about the effort that ZOTTCE put in doing this PR that deserves some consideration.

blattersturm:
Instead of writing giant walls of text to convince us to 'review this PR', again, keep in mind that **it does not add any new capabilities**. You can already use WASM as much as you desire _via the V8 runtime_ without too many obvious drawbacks.

The entire premise of this PR (adding a WASI-only runtime) is only to satisfy some random technological purity in a somewhat unneeded way, and especially not justifying adding actual code in a weird novel language nothing else of such complexity in the codebase seems to use.

blattersturm:
Saying it a third time before someone goes and harasses elsewhere: **this PR does not add any new capabilities for users to write code, since WASM is already supported via the V8 runtime**.

blattersturm:
Clarifying since people asked anyway: we're welcome to a PR like this, just _not_ one that implements all logic in a language that we still don't have the capacity to maintain longer-term.

I tried to be lenient here and was hoping on a team member to have Rust skills, but that didn't really pan out as expected.

PR #746, opened 2021-06-21
Allow events to be logged on console

Jonirulah:
Hello there, first pull ever

I've been watching scheduler.lua for a while and I didn't see any kind of event logging in the server console, I made one on my own, but it would be great for script developers to identify any undesired behaviour in the scripting also to identify performance issues even though we have profilers, this is not a priority.

Have a good night everyone.

### DIFF
--- a/data/shared/citizen/scripting/lua/scheduler.lua
+++ b/data/shared/citizen/scripting/lua/scheduler.lua
@@ -458,6 +458,11 @@ end
 
 function TriggerEvent(eventName, ...)
 	local payload = msgpack.pack({...})
+	Citizen.CreateThread(function()
+		local EventData = ({"Server Event Registered: " .. eventName})
+		local payload_a = msgpack.pack(EventData)
+		return TriggerEventInternal("trigger_log", payload_a, payload_a:len())
+	end)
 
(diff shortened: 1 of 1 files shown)


wowjeeez:
at least add some toggle option that is OFF by default, because im sure that not every developer wants to see their console littered up with random event logs, also the text is wrong too, because u put ur logger into the `TriggerEvent` function.

Cephy314:
Developers should be adding their own logging where needed. Completely unnecessary to spam every single event into the log.

jaymo1011:
While this is certainly well intentioned, it's completely unnecessary as developers should be adding their own logs for events if required in the debugging process and also, the [profiler](https://docs.fivem.net/docs/scripting-manual/debugging/using-profiler) exists!

blattersturm:
Duplicate of #658 and also written in a very much _WTF_ way.

blattersturm:
In fact - what does this accomplish that you can't accomplish within your resource by simply doing this:

```lua
local oldTriggerEvent = TriggerEvent

TriggerEvent = function(name, ...)
    -- log login
    return oldTriggerEvent(name, ...)
end
```

PR #747, opened 2021-06-23
Replace GET_PLAYER_PED(-1) with PLAYER_PED_ID

awesomeness1jake:
PLAYER_PED_ID is shown to be significantly faster compared to GET_PLAYER_PED(-1)


awesomeness1jake:
I had an old branch on this

PR #748, opened 2021-06-23
Replace GET_PLAYER_PED(-1) with PLAYER_PED_ID

awesomeness1jake:
GET_PLAYER_PED(-1) was an old way of getting the player ped


mickeygeecom:
You do realize that GetPlayerPed() takes player "source" as an argument, for other players - and not only "yourself"?

This would be a breaking change!

TomGrobbe:
> This would be a breaking change!

Is it though? Because in the changes only -1 was being used, and that was a hardcoded value.

PR #749, opened 2021-06-24
tweak(ext/cfx-ui): active route style for nav buttons

itschip:
I've done some changes for the nav buttons. 

- Added active route styles.
- Removed focus from the nav buttons

Here's how the active style looks like:
image

Reason for removing the focus was because if you did not clickaway, the selected nav button would overlap the other ones you tried to select, hiding some of the text.

**Before:**
image

**Now:**
image

### DIFF
--- a/ext/cfx-ui/src/app/nav/app-nav.component.scss
+++ b/ext/cfx-ui/src/app/nav/app-nav.component.scss
@@ -139,6 +139,13 @@
 			font-size: var(--fs-2);
 		}
 
+        .active {
+            @include theme() using ($theme) {
+                color: gtv($theme, accentContrastColor);
+                background-color: gtv($theme, accentColor1);
+            };
+        }
+
 		&>* {
 			position: relative;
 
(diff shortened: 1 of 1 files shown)


Cephy314:
That’s pretty cool.  Though perhaps the inactive nav buttons shouldn’t highlight or they should maybe be pushed out by the text.

itschip:
@Cephy314 What I'd really like is to just use a tooltip, as it would be most clean look. Not sure if people would like it though. Regarding 'maybe be pushed out by the text', could to explain this a bit further? Not sure if fully understand 😄

blattersturm:
We have already had this design discussion when this was first implemented. A tooltip is not readable enough, accordion-style moving-out would be extremely confusing with input.

itschip:
@blattersturm Thanks for the clarification. Any other thoughts around the changes? I've seen the same behavior on the 'servers' page, but I thought it'd be better to fix that in another PR 😄

PR #750, opened 2021-06-24
feat(component/citizen-server-gui): Added convar to allow custom window title

1A3Dev:
Adds a new `disable_wintitle` convar that can be set to `1` to disable FXServer interfering with the console window title (to allow people to use batch files to set it).

### DIFF
--- a/code/components/citizen-server-gui/src/ServerGui.cpp
+++ b/code/components/citizen-server-gui/src/ServerGui.cpp
@@ -181,9 +181,9 @@ private:
 				auto hostNameVar = conCtx->GetVariableManager()->FindEntryRaw("sv_hostname");
 				auto iconVar = conCtx->GetVariableManager()->FindEntryRaw("sv_icon");
 				auto gameNameVar = conCtx->GetVariableManager()->FindEntryRaw("gamename");
-				auto changeConsoleTitle = conCtx->GetVariableManager()->FindEntryRaw("consoletitle");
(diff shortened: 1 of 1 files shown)


Cephy314:
Have you actually tested this yet?  I can’t imagine anyone who would accept a pr for untested code.

1A3Dev:
> Have you actually tested this yet? I can’t imagine anyone who would accept a pr for untested code.

No since I couldn't get my VS to work however I don't see why it wouldn't work

blattersturm:
>to allow people to use batch files to set it

but why

blattersturm:
also I won't merge any patch series that's just 5 different 'Update ServerGui.cpp' commits.

1A3Dev:
> > to allow people to use batch files to set it
> 
> but why

Because I use the title to detect which server is which and the new title makes it more awkward

blattersturm:
... but it still contains the hostname? Also, 'detect'?

blattersturm:
Also, one-liner to put in a JS resource to set whatever window title you want:

```js
process.stdout.write(String.fromCharCode(27) + "]0;" + 'fuck you' + String.fromCharCode(7));
```

1A3Dev:
> Also, one-liner to put in a JS resource to set whatever window title you want:
> 
> ```js
> process.stdout.write(String.fromCharCode(27) + "]0;" + 'fuck you' + String.fromCharCode(7));
> ```
> 
> image

Ok, thanks for the help.

PR #751, opened 2021-06-25
Fix overlapping style for filters in servers page

itschip:
Another PR similar to [#749 ](https://github.com/citizenfx/fivem/pull/749). This fixes the current style where selected filters overlaps  other filters when hovering on them. All it does is remove the text when you select one.

Accidentally forgot to remove the 'z-index' when I was testing something, so it became two commits. Might just want to squash it. My bad 😄 

**When selected**
image

**Before (when hovering)**
image

**Now**
image

### DIFF
--- a/ext/cfx-ui/src/app/servers/components/filters/server-filter.component.scss
+++ b/ext/cfx-ui/src/app/servers/components/filters/server-filter.component.scss
@@ -283,6 +283,12 @@
 				}
 			}
 
+			&:focus-within {
+				.label {
+					display: none;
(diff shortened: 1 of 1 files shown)


blattersturm:
> Might just want to squash it.

why don't you?

blattersturm:
Requesting explicit design review for this one however.

blattersturm:
Seems merge conflict (already fixed?) and will be obsoleted by cfx-ui rewrite that's currently being worked on.

PR #756, opened 2021-07-02
fix mastodon.social server overload

mkljczk:


mkljczk:
you can set up your own instance https://docs.joinmastodon.org/user/run-your-own/
or host pleroma, which has significantly lower requirements https://docs.pleroma.social/backend/installation/debian_based_en

blattersturm:
This doesn't seem to fix much, except for breaking use cases where users who have set up 'mastodon.social' accounts instead of running their own ActivityPub instances will be having their feed be broken, again.

mkljczk:
@blattersturm they blocked the referrer for a reason.

blattersturm:
Are you representing 'them' here? If there has been a reason, then why haven't they nor you engaged in constructive conversation in this regard?

mkljczk:
no, I'm not. have you at least tried to contact them?

blattersturm:
If they wish to block us from their open service, common courtesy says they should contact *us* first.

Also, I couldn't find any contact details at all. 

Since you seem to be doing this for.. some.. reason unknown, I'm closing this PR.

mkljczk:
@blattersturm but how could they find you if all they knew about you is that somebody started ddosing mastodon.social?

blattersturm:
If you can even call legitimate use of a service 'DDoSing', a quick Google search for "nui-game-internal" goes a long way.

BlueRaccoonTech:
Y'know, you could just... send the operators of mastodon.social a message and try to come to some agreement, instead of just crossing your arms and being pissy about the fact a webadmin somewhere doesn't want to facilitate an application sending massive amounts of requests to their server without prior warning.

If you hosted a popular, public webserver somewhere, and I made some web browser that pinged your server every minute and it became popular, wouldn't you be upset over the massive amounts of overhead my application would cause to your server? Do you feel like it'd be fair that you'd have to research what my application is and contact me in the hopes that I'm cooperative, just so that your server could stop being spammed with requests, while the users of your service experience slowdowns and a bad experience? Or would you just simply block the tens of thousands of requests you're getting per minute from this mystery application?

mkljczk:
Technically, as it uses the ActivityPub S2S routes for this, itʼs just an incompliant, incomplete implementation of AP and thatʼs a good enough reason for blocking this mod

technetium-cfx:
Well this turned into a unproductive discussion in record time. We do in fact believe that reaching out is fair, yes. We've done so ourselves too, multiple times, and that has worked for us. I also believe you are misinterpreting what's happening here; there are no "tens of thousands of requests", in fact, it's all moderate. The use-case here is entirely legitimate, so I don't see why it warrants this discussion. We're open to solutions, we're open to discussions, but at least let it be productive.

BlueRaccoonTech:
@technetium-cfx  To clarify, I came here from a thread discussing what was going on from the server side, it's where I got the tens of thousands number. https://mastodon.social/@Gargron/106512564764833865

I was unaware that attempts were made to reach out. If you all did attempt to reach out about this issue and received no response, then I can understand why a workaround was put in place. I just personally find frustrations with the concept of changing useragents, hence why I left my two cents.

PR #757, opened 2021-07-02
i made the switch to self-hosted easier for you

mkljczk:
i didn't write a single line of c++ since a few years so hope that's working.
now all you need to do is set up pleroma and create all the accounts there.

you're welcome

### DIFF
--- a/code/components/nui-core/src/NUIClient.cpp
+++ b/code/components/nui-core/src/NUIClient.cpp
@@ -307,15 +307,9 @@ auto NUIClient::OnBeforeResourceLoad(CefRefPtr<CefBrowser> browser, CefRefPtr<Ce
 		}
 	}
 
-	// to whoever may read this in charge of blocking `https://nui-game-internal*` as referrer on the `mastodon.social` instance:
-	// -> instead of outright blocking, it would've been possible to contact us as well (hydrogen@fivem.net et al.) before resorting to
-	//    outright oddity affecting actual users. even if server load is the issue.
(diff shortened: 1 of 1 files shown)


blattersturm:
These accounts are not created or provisioned by us. They are independently created by server owners on our platform. 

Are you even aware of the concept of dedicated servers in gaming? I'm unsure of what you are looking to accomplish by these patch series.

FireTrump:
[Looks like you'll have to work on another workaround soon](https://mastodon.social/@Gargron/106512681367353166)

technetium-cfx:
Welp, wasn't aware of that discussion. I'll shoot them a message.

Gargron:
Look, I'm not actually against game mods integrating with Mastodon to show some social content, but the way this is implemented was actively impacting mastodon.social's performance. I haven't read your code so I don't know if you attempt some caching but how it looks like in my logs, you have the same clients hitting the same endpoints over and over again multiple times per second, non-stop. At its peak we were looking at 240,000 nui-game-internal requests per minute. The accounts you're checking so frequently, I looked at them, and they haven't posted anything in months!

Also, we would have gotten in touch with you if you had issues enabled on your repo. We did not find a way to contact you until you put the snarky comment in your code, and even then we only found it when somebody else linked it to us.

technetium-cfx:
Hey, I actually _just_ sent you a message on Mastodon! Let's have a chat, I'm sure we can resolve this issue. We never meant any harm.

PR #761, opened 2021-07-08
Some useful utilities that CFX-based developers wants to use / see.

jvsc3:
These are the codes that I need a lot when I write code. I think it can meet the needs of other developers too. I'll upload the utility list as much as I can. There is a lot of utility thing on my computer that I never finished.

### DIFF
--- a/data/shared/citizen/scripting/lua/utilities.lua
+++ b/data/shared/citizen/scripting/lua/utilities.lua
@@ -1,0 +1,16 @@
+-- Beginning of the Switch and case statements
+
+_G.switch = function(pParameter, pCases)
+  if (not pParameter) then return false, "pParameter: Cannot be 'null'" end
+
+  local case = pCases[pParameter]
+  if case then return case() end
+    
+  local define = pCases['default']
+  return define and define() or nil
+end
+
+-- Ending of the Switch and case statements
+
+
+  
--- a/data/shared/citizen/scripting/lua/utilities.lua
+++ b/data/shared/citizen/scripting/lua/utilities.lua
@@ -1,3 +1,8 @@
+local fmath = math
+local math = {}
+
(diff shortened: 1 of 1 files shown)


data/shared/citizen/scripting/lua/utilities.lua
```diff
@@ -13,4 +18,98 @@ end
...
+			else
+				print(tabsCache .. k .. " = "..tostring(v));
+			end
+		end
+	else
+		print(tabsCache .. tostring(pTable))
+	end
+end
+
+table.has = function(pTable, pValue)
```
arthur-rl, line 49: 🤔

data/shared/citizen/scripting/lua/utilities.lua
```diff
@@ -0,0 +1,115 @@
...
+  if (pTable == nil) then return end
+  if (pValue == nil) then return pTable[1] end
+
+  for key, value in ipairs(pTable) do
+    if (value ~= pValue) then
```
Korioz, line 54: This condition shouldn't exists, return true if `value == pValue` and at the end of the function return false.

data/shared/citizen/scripting/lua/utilities.lua
```diff
@@ -0,0 +1,115 @@
...
+	else
+		print(tabsCache .. tostring(pTable))
+	end
+end
+
+table.has = function(pTable, pValue)
+  if (pTable == nil) then return end
+  if (pValue == nil) then return pTable[1] end
+
+  for key, value in ipairs(pTable) do
```
Korioz, line 53: Doing the `for i=1, #pTable` iterating method is faster in this case.

Korioz (changes requested):

PR #765, opened 2021-07-14
Update GetVehicleNumberPlateText.md

cyberpunkleigh:
Update information regarding 0x20 space character appended

### DIFF
--- a/ext/native-decls/GetVehicleNumberPlateText.md
+++ b/ext/native-decls/GetVehicleNumberPlateText.md
@@ -8,8 +8,10 @@ apiset: server
 char* GET_VEHICLE_NUMBER_PLATE_TEXT(Vehicle vehicle);
 ```
 
-
 ## Parameters
 * **vehicle**: 
+
+## Note
+If the number plate is less than 8 character return value will have a space at the end to fill the remaining characters
 
 ## Return value


ext/native-decls/GetVehicleNumberPlateText.md
```diff
@@ -8,8 +8,10 @@ apiset: server
 char* GET_VEHICLE_NUMBER_PLATE_TEXT(Vehicle vehicle);
 ```
 
-
 ## Parameters
 * **vehicle**: 
 
+## Note
+If the number plate is less than 8 character return value will have a space at the end to fill the remaining characters
```
Korioz, line 15: Probably better 🤔 :
`If the number plate text is less than 8 characters the result will be appended with spaces until there is 8 characters in it.`

Korioz (commented):
Hi 👋,

It seems you're *note* have a typo.

ext/native-decls/GetVehicleNumberPlateText.md
AvarianKnight: I also think that the notes are supposed to come before the ##Paramaters, and wont take a `##` markup

AvarianKnight (changes requested):

blattersturm:
Invalid syntax, user ignored changes requested.

PR #767, opened 2021-07-17
fix(native-decls/sdk): fix paramater mismatch

AvarianKnight:
This causes `fxd natives` to fail.


blattersturm:
... huh? When was this even supposedly added?

AvarianKnight:
For some weird reason, it only causes building the C# files to fail to build, the rest will still somehow manage to work.

PR #769, opened 2021-07-19
fix(scripting/v8): Default omitted timeout to 0

cactysman:
... otherwise the [check](https://github.com/citizenfx/fivem/blob/6de33031bddd96fd0cec08a4b0bf34e22836d303/data/shared/citizen/scripting/v8/timer.js#L120) fails because nothing is greater than `undefined`.

This breaks interop with incorrect usage (as found in some npm modules) of `setTimeout` and `setInterval` where omitted timeouts are meant to default to 0 ([spec § 8.6.10](https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html#timers)), but instead just don't run the function at all due to the missing default/min value of 0.

I also renamed the property for convenience (`timeout` instead of `interval`).

### DIFF
--- a/data/shared/citizen/scripting/v8/timer.js
+++ b/data/shared/citizen/scripting/v8/timer.js
@@ -14,7 +14,7 @@
     function setTimer(timer, callback, interval) {
         timers[timer.id] = {
             callback,
-            interval,
+            interval: Math.max(0, interval) || 0,
             lastRun: gameTime
         };
     }


data/shared/citizen/scripting/v8/timer.js
```diff
@@ -11,10 +11,10 @@
 
     let animationFrames = [];
 
-    function setTimer(timer, callback, interval) {
+    function setTimer(timer, callback, timeout) {
```
nihonium-cfx, line 14: Given how `setTimeout` is in fact just `setInterval` but will die after first run this should be kept as `interval`.

nihonium-cfx (changes requested):

blattersturm:
Can you actually squash the commits in your branch?

cactysman:
> Can you actually squash the commits in your branch?

Done

blattersturm:
btw do you even test the code you pr? nice syntax error 🙃

cactysman:
> btw do you even test the code you pr? nice syntax error 🙃

Of course I did and honestly I don't see what's supposed to be wrong here.

blattersturm:
You're missing `interval:`.

cactysman:
God damn it. That's what you get from rebasing on your phone.
It's fixed now.

PR #770, opened 2021-07-26
feat(natives/five): gamer tags visibility and behavior natives

Hendreik:
Adds:
SET_MP_GAMER_TAGS_VISIBLE_DISTANCE to be able to set the maximum distance at which gamer tags are displayed.
SET_MP_GAMER_TAGS_USE_VEHICLE_BEHAVIOR to toggle whether gamer tags should act differently when in a vehicle


blattersturm:
I wonder what happened to the CI build to be randomly.. canceled?

PR #775, opened 2021-08-02
Correct return type for getGameVersion export in ViSH

Disquse:
Should resolve startup crashes introduced in #773, caused by incorrect return type.


zirdic:
ok

E66666666:
:cry: sorry

Disquse:
> 
> 
> 😢 sorry

np!

PR #777, opened 2021-08-03
fix(native-decls): fix ePopulationType not being in the description

AvarianKnight:


AvarianKnight:
Gotta re open this with fixed wording

PR #778, opened 2021-08-03
fix(native-decls): fix ePopulationType not being in the description

AvarianKnight:
This fixes native doc showing this weirdly.


blattersturm:
I feel this rather should be fixed in native-doc-tooling/native-reference-ng.

blattersturm:
for reference, https://docs.fivem.net/natives/?_0xFC30DDFF has weird formatting:

image

PR #779, opened 2021-08-03
feat(resources/core): statebag trigger event on changes

Sadler2:
Triggers __cfx_internal:stateBagSetKey event on every StateBag update (local or replicated). I put it inside cfx_internal, because in my opinion, the final api written on top of this event, should be bound to StateBag objects in user scripts. 

**Use case**: imagine we have a nui status bar with hunger, thirst, and a whole bunch of buffs/conditions player may have. It would be much easier to update it from state on change, rather than creating a loop that would compare current values to previous ones  every second, detect changes and send values to nui.

Tested it up locally for an hour, hope I didn't mess up with pointers and concurrency.

### DIFF
--- a/code/components/citizen-resources-core/include/StateBagComponent.h
+++ b/code/components/citizen-resources-core/include/StateBagComponent.h
@@ -1,6 +1,8 @@
 #pragma once
 
 #include <ComponentHolder.h>
+#include <ResourceManager.h>
+#include <ResourceEventComponent.h>
 
 #include <optional>
 #include <string_view>
@@ -13,7 +15,6 @@
 
 namespace fx
 {
-class ResourceManager;
 
 class StateBagGameInterface
 {
@@ -114,6 +115,7 @@ public:
 
 public:
 	static fwRefContainer<StateBagComponent> Create(StateBagRole role);
+	fwRefContainer<fx::ResourceEventManagerComponent> eventManager;
 };
 }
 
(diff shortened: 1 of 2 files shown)


blattersturm:
Not a good idea to implement in this way, see discussion regarding event elision in a PR before.

Sadler2:
Could you please explain? The PR before is about logging events, I don't get how it is connected with this PR, and what's the better way to implement it you're talking about.

blattersturm:
Not that PR. Having the same event name for a hot path event is a guaranteed way to get performance issues in HLL code. There's a reason that this feature - especially with it appearing to be this trivial - isn't implemented yet.

blattersturm:
Also you seem to be insinuating this event is for use in HLL code but yet aren't proposing any way that HLL code would look at all. 😕

Sadler2:
Ok, what if we include StateBag id inside event name? For example __cfx_internal:stateBagSetKey:player:1 ? The only high-level code I might imagine is a subscription model `mystate.onchange((key, val) => { console.log(``${key}: ${val}``) })`.

blattersturm:
Potentially so, but there's a few other things that I'd like to see included here (ability to cancel a state change on server including rollback on the originating client, wildcard filtering for prefixes/suffixes - which would need some more advanced logic in event registration, and potentially bag scoping as well as key scoping), and exposing internal APIs meant for ScRT HLL code to user code is also wrong so that'd need some sort of prevention thought up.

blattersturm:
Implemented as 'state bag change handlers' by now.

PR #780, opened 2021-08-04
tweak(game/five): increase CWeaponInfoBlob limit

TheVortexX:


TheVortexX:
Value increase is to help fix weapon crashes when going over this limit

cmg-divined:
Would love to see this happen unless there's a reason not to. Yes.

PR #784, opened 2021-08-08
feat(v8): add PerformHttpRequest for js

AvarianKnight:
This allows for people to perform HTTP requests without needing an external library.

### DIFF
--- a/data/shared/citizen/scripting/v8/index.d.ts
+++ b/data/shared/citizen/scripting/v8/index.d.ts
@@ -91,6 +91,7 @@ declare function TriggerLatentServerEvent(eventName: string, bps: number, ...arg
 declare function getPlayerIdentifiers(player: number|string): string[]
 declare function getPlayerTokens(player: number|string): string[]
 declare function getPlayers(): string[]
+declare function PerformHttpRequest(url: string, cb: Function, method?: string, data?: string, headers?: { [key: string]: string }, options?: object): void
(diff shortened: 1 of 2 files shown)


blattersturm:
I do not think this is an appropriate solution. Rather, there should be a new ref-callback-based version of the function which should be directly exposed as 'PerformHttpRequest' (as a native), keeping the old method for compatibility with user code that inadvertently relied on it.

blattersturm:
Secondly... require('http')? Not very friendly but it _does_ exist.

AvarianKnight:
> I do not think this is an appropriate solution. Rather, there should be a new ref-callback-based version of the function which should be directly exposed as 'PerformHttpRequest' (as a native), keeping the old method for compatibility with user code that inadvertently relied on it.

Are you talking something along the lines of REGISTER_RAW_NUI_CALLBACK?

blattersturm:
> Are you thinking about something along the lines of REGISTER_RAW_NUI_CALLBACK?

If that uses ref callbacks, yes.

l1ru:
you can use node-fetch for this

AvarianKnight:
> you can use node-fetch for this

> This allows for people to perform HTTP requests **without needing an external library**.

blattersturm:
Abandoned PR, was apparently undesirable as opposed to unifying the native's implementation.

PR #785, opened 2021-08-08
feat(server/state): update fireEvent variables

AvarianKnight:
Unsure of what v1 is, it gets set to 3 when its a ped is on fire, and 5 when a vehicle is (though it doesn't seem to be very consistent)


Disquse (approved):
Tested on 1604 and 2545, works as expected.

Issue #787, opened 2021-08-13
Win32k user crit slowness detection


blattersturm:
A player reported frequent hitches while playing the game. ETW tracing showed the main thread blocking on the render thread, with one frame spending 300ms+ on the message loop in there:

![image](https://user-images.githubusercontent.com/24576130/129400566-d03494db-e445-4b74-9012-40f357b1745a.png)

Deeper nesting checks showed a lot of nested calls into uxtheme, dinput8 or win32k directly, mostly landing on win32kbase.sys!EnterCrit:

![image](https://user-images.githubusercontent.com/24576130/129400668-cd570ea1-0bec-43e6-b933-e6cf7c34e13e.png)

There should be a warning when this occurs in a sufficiently egregious fashion.

PR #788, opened 2021-08-13
feat(game/five): win32k usercrit slowness warnings

blattersturm:
An attempted implementation for #787 in the existing framework used for time warnings.

image

### DIFF
--- a/code/components/citizen-devtools/src/ResourceTimeWarnings.cpp
+++ b/code/components/citizen-devtools/src/ResourceTimeWarnings.cpp
@@ -12,6 +12,8 @@
 
 #include <chrono>
 
+#include <CfxLocale.h>
+
 #include "ResourceMonitor.h"
 
 DLL_IMPORT ImFont* GetConsoleFontTiny();
@@ -68,6 +70,72 @@ constexpr auto tuple_slice(Cont&& t)
 	std::make_index_sequence<I2 - I1>{});
 }
 
+static std::chrono::microseconds lastHitch;
+
+#ifdef GTA_FIVE
+#include <Hooking.h>
+#include <InputHook.h>
+
+static decltype(&SetThreadExecutionState) origSetThreadExecutionState;
+static decltype(&PeekMessageW) origPeekMessageW;
+
+static std::chrono::microseconds currentTotal;
+static bool currentWasFocusEvent = false;
+
+static EXECUTION_STATE WINAPI SetThreadExecutionState_Track(EXECUTION_STATE esFlags)
+{
+	if (esFlags == 3)
+	{
+		currentTotal = std::chrono::microseconds{ 0 };
+		currentWasFocusEvent = false;
+	}
+
+	return origSetThreadExecutionState(esFlags);
+}
+
+static BOOL WINAPI PeekMessageW_Track(LPMSG lpMsg, HWND hWnd, UINT wMsgFilterMin, UINT wMsgFilterMax, UINT wRemoveMsg)
+{
+	// a focus event may take long, so we ignore it
+	if (currentWasFocusEvent)
+	{
+		return origPeekMessageW(lpMsg, hWnd, wMsgFilterMin, wMsgFilterMax, wRemoveMsg);
+	}
+
+	// track just the PeekMessage call
+	// this will include internal wndproc invocations as well once other events run out
+	auto thisStart = usec();
+	auto rv = origPeekMessageW(lpMsg, hWnd, wMsgFilterMin, wMsgFilterMax, wRemoveMsg);
+	auto thisEnd = usec();
+
+	currentTotal += (thisEnd - thisStart);
+
+	// if we're out of events, and didn't get a focus event generated anyway
+	if (!rv && !currentWasFocusEvent)
+	{
+		if (currentTotal > 30ms)
+		{
+			lastHitch = thisEnd;
+		}
+	}
+
+	return rv;
+}
+
+static HookFunction hookFunctionGameTime([]()
+{
+	InputHook::DeprecatedOnWndProc.Connect([](HWND, UINT uMsg, WPARAM, LPARAM, bool&, LRESULT&)
+	{
+		// we want to ignore both focus-in and focus-out events
+		if (uMsg == WM_ACTIVATEAPP)
+		{
+			currentWasFocusEvent = true;
+		}
+	}, INT32_MIN);
+
+	origSetThreadExecutionState = hook::iat("kernel32.dll", SetThreadExecutionState_Track, "SetThreadExecutionState");
+	origPeekMessageW = hook::iat("user32.dll", PeekMessageW_Track, "PeekMessageW");
+});
+#endif
 
 static InitFunction initFunction([]()
 {
@@ -119,7 +187,7 @@ static InitFunction initFunction([]()
 	ConHost::OnDrawGui.Connect([]()
 	{
 #ifndef IS_FXSERVER
-		if ((usec() - warningLastShown) < 20s)
+		auto displayWarningDialog = [](auto cb)
 		{
 			ImGui::PushFont(GetConsoleFontTiny());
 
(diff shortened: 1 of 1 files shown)


code/components/citizen-devtools/src/ResourceTimeWarnings.cpp
```diff
@@ -12,6 +12,8 @@
 
 #include <chrono>
 
+#include <CfxLocale.h>
```
blattersturm, line 15: When we revive localization, we may want to make this localizable. @technetium-cfx are you still able to look into configuring Weblate or finding an alternative solution with similar Git support?

code/components/citizen-devtools/src/ResourceTimeWarnings.cpp
```diff
@@ -67,7 +69,67 @@ constexpr auto tuple_slice(Cont&& t)
 	return detail::slice_impl<I1>(std::forward<Cont>(t),
 	std::make_index_sequence<I2 - I1>{});
 }
+
+static std::chrono::microseconds lastHitch;
+
+#ifdef GTA_FIVE
```
blattersturm, line 75: This does not affect RDR3 as it does off-tread windowing, and NY may use a different loop on render anyway.

code/components/citizen-devtools/src/ResourceTimeWarnings.cpp
```diff
@@ -67,7 +69,67 @@ constexpr auto tuple_slice(Cont&& t)
...
+	}
+
+	auto trackStart = usec();
+	auto rv = origPeekMessageW(lpMsg, hWnd, wMsgFilterMin, wMsgFilterMax, wRemoveMsg);
+	auto trackEnd = usec();
+	trackTotal += (trackEnd - trackStart);
+
+	if (!rv && !trackIsInvalid)
+	{
+		if (trackTotal > 30ms)
```
blattersturm, line 110: This threshold is high enough to only affect me spamming Win+E to open Explorer windows, but not high enough to trip accidentally during normal usage like alt-tabbing.

code/components/citizen-devtools/src/ResourceTimeWarnings.cpp
```diff
@@ -67,7 +69,67 @@ constexpr auto tuple_slice(Cont&& t)
...
+
+static BOOL WINAPI PeekMessageW_Track(LPMSG lpMsg, HWND hWnd, UINT wMsgFilterMin, UINT wMsgFilterMax, UINT wRemoveMsg)
+{
+	if (trackIsInvalid)
+	{
+		return origPeekMessageW(lpMsg, hWnd, wMsgFilterMin, wMsgFilterMax, wRemoveMsg);
+	}
+
+	auto trackStart = usec();
+	auto rv = origPeekMessageW(lpMsg, hWnd, wMsgFilterMin, wMsgFilterMax, wRemoveMsg);
```
blattersturm, line 104: To prevent trapping on intentional DispatchMessage, we only account for PeekMessage time. Querying perf counter may however be slow, but granularity is of the essence here.

code/components/citizen-devtools/src/ResourceTimeWarnings.cpp
```diff
@@ -67,7 +69,67 @@ constexpr auto tuple_slice(Cont&& t)
...
+	return rv;
+}
+
+static HookFunction hookFunctionGameTime([]()
+{
+	InputHook::DeprecatedOnWndProc.Connect([](HWND, UINT uMsg, WPARAM, LPARAM, bool&, LRESULT&)
+	{
+		if (uMsg == WM_ACTIVATEAPP)
+		{
+			trackIsInvalid = true;
```
blattersturm, line 125: An attempt to reduce any time taken when alt-tabbing in. Maybe dead code, should be checked.

code/components/citizen-devtools/src/ResourceTimeWarnings.cpp
```diff
@@ -67,7 +69,67 @@ constexpr auto tuple_slice(Cont&& t)
...
+		}
+	}, INT32_MIN);
 
+	origSetThreadExecutionState = hook::iat("kernel32.dll", SetThreadExecutionState_Track, "SetThreadExecutionState");
```
blattersturm, line 129: The message loop on _render_ looks as follows:

```c
SetThreadExecutionState(3);

while (PeekMessage(...)) {
  // dispatch
}
```

code/components/citizen-devtools/src/ResourceTimeWarnings.cpp
```diff
@@ -139,6 +201,27 @@ static InitFunction initFunction([]()
 			}
 			ImGui::End();
 			ImGui::PopFont();
+		}
+		else if ((usec() - lastHitch) < 5s)
+		{
+			ImGui::PushFont(GetConsoleFontTiny());
+
+			const float DISTANCE = 10.0f;
```
blattersturm, line 209: A bit of code dup - maybe would be worthwhile to make a helper function for this kind of warning popup?

I'd have reused the resource warning string stuff but timing as well as phrasing seem ingrained there.

code/components/citizen-devtools/src/ResourceTimeWarnings.cpp
```diff
@@ -139,6 +201,27 @@ static InitFunction initFunction([]()
...
+			ImVec2 window_pos = ImVec2(ImGui::GetMainViewport()->Pos.x + ImGui::GetIO().DisplaySize.x - DISTANCE, ImGui::GetMainViewport()->Pos.y + ImGui::GetIO().DisplaySize.y - DISTANCE);
+			ImVec2 window_pos_pivot = ImVec2(1.0f, 1.0f);
+			ImGui::SetNextWindowPos(window_pos, ImGuiCond_Always, window_pos_pivot);
+
+			ImGui::SetNextWindowBgAlpha(0.3f); // Transparent background
+			if (ImGui::Begin("Time Warning", NULL, ImGuiWindowFlags_NoMove | ImGuiWindowFlags_NoTitleBar | ImGuiWindowFlags_NoResize | ImGuiWindowFlags_AlwaysAutoResize | ImGuiWindowFlags_NoSavedSettings | ImGuiWindowFlags_NoFocusOnAppearing | ImGuiWindowFlags_NoNav))
+			{
+				ImGui::Text(va("/!\\ %s", gettext("Slow system performance detected")));
+				ImGui::Separator();
+				ImGui::Text("%s", gettext("A call into the Windows API took too long recently and led to a game stutter.").c_str());
```
blattersturm, line 219: Maybe too technical?

blattersturm (commented):
Explaining my choices made a little bit.

blattersturm:
... oh, we don't have a reviewers group yet.

code/components/citizen-devtools/src/ResourceTimeWarnings.cpp
aiden-z-source: Possibly, but also straight to the point and makes it pretty clear it isn't an issue with FiveM or the server, so I think it should be fine.

blattersturm: Yeah - that's the point I was trying to make. 'Game stutter' hopefully is interpretable enough, people make up weird names for this.

code/components/citizen-devtools/src/ResourceTimeWarnings.cpp
```diff
@@ -67,7 +69,67 @@ constexpr auto tuple_slice(Cont&& t)
...
+
+#ifdef GTA_FIVE
+#include <Hooking.h>
+#include <InputHook.h>
+
+static decltype(&SetThreadExecutionState) origSetThreadExecutionState;
+static decltype(&PeekMessageW) origPeekMessageW;
+
+static std::chrono::microseconds trackTotal;
+static bool trackIsInvalid = false;
```
LWSS, line 83: maybe name `trackIsValid` instead

code/components/citizen-devtools/src/ResourceTimeWarnings.cpp
```diff
@@ -67,7 +69,67 @@ constexpr auto tuple_slice(Cont&& t)
...
+	}
+
+	return rv;
+}
+
+static HookFunction hookFunctionGameTime([]()
+{
+	InputHook::DeprecatedOnWndProc.Connect([](HWND, UINT uMsg, WPARAM, LPARAM, bool&, LRESULT&)
+	{
+		if (uMsg == WM_ACTIVATEAPP)
```
LWSS, line 123: isn't this uMsg sent for both tabbing in/out?

blattersturm: Correct, however that should only happen for a single chain as it's reset next message loop invocation.

code/components/citizen-devtools/src/ResourceTimeWarnings.cpp
blattersturm: Good point, same as `trackTotal` being a tad vague (currentTotal?)

blattersturm:
Another attempt at the changeset. Notable changes:
- more comments
- variable names renamed as requested
- got rid of the code duplication

PR #790, opened 2021-08-14
Add tooltips explaining premium levels in cfx-ui

blattersturm:
## Motivation

Users can't be expected to understand the periodic table or the fact server subscriptions are named after elements.

See the following (real) support questions:

> what do all the little tag abbreviation means in the server list? like Pt

> In the server browser some servers have "AG" or a "PT" icon in front of it... does PT means that they have the patreon sub?

> another thing, why some servers aren't portuguese but they show as PT list

## Solution

This might be helpable with minimal impact by adding a tooltip:

image

This changeset adds such a tooltip to the server list as well as the detail page.

### DIFF
--- a/ext/cfx-ui/src/app/servers/components/detail/servers-detail.component.html
+++ b/ext/cfx-ui/src/app/servers/components/detail/servers-detail.component.html
@@ -44,6 +44,8 @@
 								[class.ag]="server.premium === 'ag'"
 								[class.au]="server.premium === 'au'"
 								[class.pt]="server.premium === 'pt'"
+								data-balloon-pos="right"
+								[attr.aria-label]="premiumName"
 							>
 								<span>{{ server.premium }}</span>
 							</div>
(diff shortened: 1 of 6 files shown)


ext/cfx-ui/src/app/servers/components/detail/servers-detail.component.html
```diff
@@ -44,6 +44,8 @@
 								[class.ag]="server.premium === 'ag'"
 								[class.au]="server.premium === 'au'"
 								[class.pt]="server.premium === 'pt'"
+								data-balloon-pos="right"
```
blattersturm, line 47: This tooltip will only show if there's a premium value already, which at this point will only be ag/au/pt.

ext/cfx-ui/src/app/servers/components/list/servers-list-item.component.scss
```diff
@@ -92,6 +92,7 @@
 		bottom: 2px;
 
 		width: var(--q4);
+		pointer-events: none;
```
blattersturm, line 95: The 'premium' gradient was positioned over the badge for some reason - by dropping pointer events for it, the tooltip at least does work.

Issue #794, opened 2021-08-14
Jump list icons no longer work


blattersturm:
![image](https://user-images.githubusercontent.com/24576130/129438877-7a4fa555-b323-4ccd-a0d2-a6a9d11ad3c7.png)

Probably related to the big cache renamening _or_ connection flow changes, needs to be checked.

blattersturm:
https://github.com/citizenfx/fivem/blob/f16054580843fb8c277db55b410035c70d8028ce/code/components/glue/src/ConnectToNative.cpp#L365

Seemingly these try to store in `data/cache/browser/`, which is a directory we no longer ensure exists (as it was replaced by `data/nui-storage/` variants).

Should be moved to `data/cache/servers/` perhaps which is a directory we do ensure in `nui:gsclient`.

blattersturm:
Fixing this gave _some_ servers their icons back, but connecting from console to `connect https://fivem.net/` didn't add the correct icon anyway, but rather an icon-less entry, this means the connection flow needs checking as well.

Also, the 16x16 resizing seems to be _extremely ugly_ as if there's no interpolation being done, this is a separate issue however.

blattersturm:
![image](https://user-images.githubusercontent.com/24576130/129441672-de7886ce-ddd5-4458-a018-510607df45a0.png)

We seem to not be requesting or saving icons here given this example `lastServers` layout - if we fetch `/info.json` anyway later on in native code however we might be able to tell UI to backfill an icon (and token)?

Would need a way to ensure correct correlation however, maybe `connectTo` should get a nonce for that.

PR #803, opened 2021-08-14
Jump list icon fixes

blattersturm:
This is a series of two changes for #794, as well as potentially fixing some other UI history-related issues by means of coincidence in the second commit.


code/components/glue/src/ConnectToNative.cpp
```diff
@@ -882,6 +917,15 @@ static InitFunction initFunction([] ()
 		{
 			UpdatePendingAuthPayload();
 		}
+		else if (!_wcsicmp(type, L"backfillDone"))
```
blattersturm, line 920: Slightly risky to rely on this event to be present and handled by UI, but hopefully this shouldn't lead to any user issues.

PR #804, opened 2021-08-14
feat(glue): add build number to jump list shortcuts

### LINKED ISSUE #796: Jump list entries should save a build argument
Since we're starting the game anyway from these entries, we should save the last build this server was seen with and launch with the `-bX` argument directly so users don't have to switch from the last used build.

blattersturm:
Closes #796.

### DIFF
--- a/code/components/glue/src/ConnectToNative.cpp
+++ b/code/components/glue/src/ConnectToNative.cpp
@@ -322,6 +322,7 @@ struct ServerLink
 	std::string rawIcon;
 	std::string hostname;
 	std::string url;
+	nlohmann::json vars;
 };
 
 #include <wrl.h>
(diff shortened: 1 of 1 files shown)


code/components/glue/src/ConnectToNative.cpp
```diff
@@ -340,6 +341,21 @@ static WRL::ComPtr<IShellLink> MakeShellLink(const ServerLink& link)
...
 	if (SUCCEEDED(hr))
 	{
+		std::wstring buildArgument;
+		try
+		{
+			if (link.vars.is_object())
+			{
+				if (auto it = link.vars.find("sv_enforceGameBuild"); it != link.vars.end())
+				{
+					buildArgument = fmt::sprintf(L"-b%s ", ToWide(it.value().get<std::string>()));
```
blattersturm, line 351: Potential for argument injection, will fix to parse as int and then add.

PR #819, opened 2021-08-14
fix(gamestate/client): client RPC SEH wrapper

### LINKED ISSUE #798: ClientRPC native invocation is lacking SEH
https://forum.cfx.re/t/setvehicledoorslocked-rpc-causing-client-crashes-full-dump-included/4477043

https://github.com/citizenfx/fivem/blob/cbe56f78f86bebb68d7960a38c3cdc31c7d76790/code/components/gta-net-five/src/ClientRPC.cpp#L519-L527

If we directly invoke the handler from ScriptEngine, we don't have a SEH wrapper for worst-case failures in game code.

blattersturm:
Fixes #798, hopefully.

Untested, should do this before merging.

### DIFF
--- a/code/components/gta-net-five/src/ClientRPC.cpp
+++ b/code/components/gta-net-five/src/ClientRPC.cpp
@@ -28,6 +28,33 @@ extern NetLibrary* g_netLibrary;
 
 #include <concurrentqueue.h>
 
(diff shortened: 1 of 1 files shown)


blattersturm:
Tested SET_ENTITY_COORDS and it didn't fail on a _valid_ call, so should be fine.

PR #839, opened 2021-08-16
Update MumbleIsConnected.md

draobrehtom:
When player is connected to Mumble the native `MumbleIsConnected` returns `1` (number).  When player is not connected the native returns `false` (boolean). I am not sure if this is an intended behaviour, therefor I propose to update documentation first.


AvarianKnight:
In C++ 1 and 0 are intepreted as `true` and `false`

Issue #845, opened 2021-08-17
Clean up server names/descriptions even more


blattersturm:
The screenshot in #844 showed a few more edge cases that should be cleaned/handled:

![image](https://user-images.githubusercontent.com/24576130/129750008-d8fc5147-1b6b-4b8a-b3e9-28e4753cc8c0.png)

1. Server names with 'tags' in their name.
    See the '[LIVE] Next Stage RP [DK] [16+] [https:` listing.
2. Suffixed language codes, with spaces. Should be handled by 1.
3. Language codes/prefixes in the description.
    See the `[DK] Danish World` description.
4. Servers using `+` as separator in the description. This should be handled the same way as `,` since it may be present in a normal sentence.
5. `[PlumeESXLegacy]` odd tag.
6. `{DK}`: `{` and probably `(` should be treated as `[` is now.
7. [bullet point] [2021] listing. What?
8. It seems emoji with a space on only _one_ end aren't treated as a separator.
    ![image](https://user-images.githubusercontent.com/24576130/129750837-bb7afaf1-c1c8-488d-bf6e-d0b6fdacdb0c.png)

A few other points, too:
1. ~~Some of the 'tags' people use in listings/descriptions _could_ be added to the tag array. We should be careful with this e.g. not add these to 'raw' term search as people often use this to show up in 'competitor' searches.~~ Maybe at a later time.
2. The premium badges should be moved to the right of the country flags for alignment reasons.

blattersturm:
No putting the badge before. It's ugly.

![image](https://user-images.githubusercontent.com/24576130/129761529-fb7b985d-95c7-4c68-aa89-349393c7c59d.png)

hoyame:
I like ham, do you?

PR #847, opened 2021-08-17
Server list cleanup stuff

### LINKED ISSUE #844: Server list search should search locales if it looks like a language/country code
From Discord:

image

The user in this case was concerned 'not all servers showed up', as they were searching for servers where 'DK' showed up in the filtered name.

blattersturm:
Closes #844 and #845.


hoyame (approved):
good progress, keep it up, you are on the right track

Issue #824, opened 2021-08-15
Server history may lose backfilled server with console connect


blattersturm:
When doing a manual connection to a server, but the connection gets rejected (for example for having to change a game build), the server history entry gets replaced with a server entry _without_ the backfill info.

Instead of just replacing the existing server, history-add logic should take backfill and icon info from the existing entry if the new entry has none.

Issue #795, opened 2021-08-14
Jump list icons duplicate when reconnected to (e.g. build switching)


blattersturm:
![image](https://user-images.githubusercontent.com/24576130/129438890-80bdb1c1-d6bb-4985-82db-b61fd670345d.png)

After the game restarted I had two TT history entries. Probably needs to be canonicalized.

PR #853, opened 2021-08-18
tweak(ext/cfx-ui): Don't show more than 1 record for server history.

LWSS:
https://github.com/citizenfx/fivem/issues/795
https://github.com/citizenfx/fivem/issues/824

fixes some quirks with using `connect` and the server history

### DIFF
--- a/ext/cfx-ui/src/app/game.service.ts
+++ b/ext/cfx-ui/src/app/game.service.ts
@@ -1,4 +1,4 @@
-import { Injectable, EventEmitter, NgZone, Inject } from '@angular/core';
+import { Injectable, EventEmitter, NgZone, Inject, PLATFORM_ID } from '@angular/core';
 import { DomSanitizer } from '@angular/platform-browser';
 
 import { Server, ServerHistoryEntry } from './servers/server';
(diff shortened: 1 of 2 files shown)


ext/cfx-ui/src/app/game.service.ts
```diff
@@ -358,8 +360,12 @@ export class CfxGameService extends GameService {
 	private profileList: Record<string, string> = {};
 	card: boolean;
 
-	constructor(private sanitizer: DomSanitizer, private zone: NgZone, private translation: L10nTranslationService) {
+	private serversService: ServersService;
+
+	constructor(private sanitizer: DomSanitizer, private zone: NgZone, private translation: L10nTranslationService,
+				private httpClient: HttpClient, @Inject(PLATFORM_ID) private platformId: any) {
 		super();
+		this.serversService = new ServersService(httpClient, sanitizer, zone, this, platformId);
```
blattersturm, line 368: This seems fishy. Even if ServersService depends on GameService, might it make more sense instead to factor out the dependencies needed by both into a separate service (e.g. ServersCacheService), potentially forwarding calls from ServersService into this new service to be compatible with downstream consumers, instead of being weirdly recursive or doing this _completely weird hack_?

ext/cfx-ui/src/app/game.service.ts
```diff
@@ -730,6 +736,78 @@ export class CfxGameService extends GameService {
...
+								historyEntry.hostname = entry.hostname;
+								historyEntry.icon = entry.icon;
+							}
+						}
+					}
+				}
+			}
+		} else {
+			// The new address is a direct IP
+			if (entry.token && entry.token !== '' ) {
```
blattersturm, line 775: superfluous space

blattersturm (changes requested):
Hacky DI hack is hacky.

blattersturm:
Didn't reply to review comments, and cfx-ui is pending a rewrite.

PR #881, opened 2021-08-20
feat(server/impl): add GetVehicleWithType

AvarianKnight:
- useful if you only want to get specific types of vehicles.


blattersturm:
No, this is not something that I want to see implemented this way - intent is to expose more generic filtering functionality than having 'with_x' natives all over the place.

PR #883, opened 2021-08-21
fix(extra-natives/five): render cutout object outlines correctly

### LINKED ISSUE #874: Outline shader doesn't work for some trees or stt tubes
The stunt tube and the trees in the place I placed them in #872 have no outlines when selected - it could be these shaders don't have a basic drawunlit.

blattersturm:
Fixes #874 for now, minus the tree portion: this is a later concern.


blattersturm:
Clean diff: https://github.com/citizenfx/fivem/pull/883/commits/c11a39a3097a4ee0e6ec681bb96872b710a76e75?w=1

Issue #888, opened 2021-08-22
'No object by ID' prints and similar need script attribution


blattersturm:
https://github.com/citizenfx/fivem/blob/e60f75f741b5c22bac10f05627cc004e47aec301/code/components/gta-streaming-five/src/EntityDeletionHacks.cpp#L113

Ideally we'd have some sort of 'script warning' function for these which would get the current script invocation source, use it as a channel, and get the current call site too so we can show a line of code or function name that's responsible for the warning.

These functions can't really be made into errors as that'd break compatibility, sadly.

blattersturm:
There seems to be a script warning system now.

PR #892, opened 2021-08-23
feat(resources): dependency constraints

blattersturm:
Meta-dependencies starting with `/` to check for environment constraints before starting a resource.

Example:

```lua
dependencies {
    '/server:4000',
    '/server:999999',
    '/server:696969',
    '/policy:subdir_file_mapping',
    '/policy:meow',
    '/onesync',
    '/gameBuild:h4',
    '/native:0xE27C97A0',
    '/native:0xDEADBEEF',
}
```

Output example:
```text
[    c-resources-core] Resource 'constraint' can't run:
[    c-resources-core]   1. server needs to be 999999 or higher
[    c-resources-core]   2. server needs to be 696969 or higher
[    c-resources-core]   3. policy(meow): you do not have access to the 'meow' policy
[    c-resources-core]   4. gameBuild(h4): sv_enforceGameBuild needs to be at least 2189 (current is 1604)
[    c-resources-core]   5. native(0xDEADBEEF): native 0xDEADBEEF isn't supported
[ citizen-server-impl] Couldn't start resource constraint.
```

Docs are pending as per citizenfx/fivem-docs#269.

### DIFF
--- a/code/components/citizen-resources-core/include/ResourceManagerConstraintsComponent.h
+++ b/code/components/citizen-resources-core/include/ResourceManagerConstraintsComponent.h
@@ -1,0 +1,46 @@
+#pragma once
+
+#ifdef COMPILING_CITIZEN_RESOURCES_CORE
+#define RESOURCES_CORE_EXPORT DLL_EXPORT
+#else
+#define RESOURCES_CORE_EXPORT DLL_IMPORT
+#endif
+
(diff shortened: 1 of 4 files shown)


blattersturm:
Note that this logic should _also_ feed into whatever warning/issue feed proposed in #888.

PR #894, opened 2021-08-23
Add camera shake override setting

catmino:
Allows users to turn camera shake off, both while driving a vehicle and running (third person only).
Mainly a QOL change for people suffering from motion sickness due to camera shake.
Currently requires a game restart to take effect if changed manually in F8 console (eg. seta "cam_disableCameraShake" "true") as forcing value changes on address while game is running caused crashes (unsure why, works fine in base game), however, changing it in FiveM launcher settings works properly, so I went with that solution and hook on first load. Haven't had any issues doing it that way.

### DIFF
--- a/code/components/gta-core-five/src/DisableCameraShake.cpp
+++ b/code/components/gta-core-five/src/DisableCameraShake.cpp
@@ -1,0 +1,45 @@
+/*
+ * This file is part of the CitizenFX project - http://citizen.re/
+ *
+ * See LICENSE and MENTIONS in the root of the source tree for information
+ * regarding licensing.
+ */
+
+#include <StdInc.h>
+#include <GameInit.h>
+#include <Hooking.h>
+#include <CoreConsole.h>
+#include <nutsnbolts.h>
+
+static std::shared_ptr<ConVar<bool>> g_cameraShakeConvar;
+
+static void CameraShakeOverride()
+{
+	if (g_cameraShakeConvar->GetValue())
+	{
+		//Vehicle high speed camera shake
+		//48 8B D9                  mov     rbx, rcx
+		//48 81 C7 88 04 00 00      add     rdi, 488h     <--------------
+		//8B 6F 08                  mov     ebp, [rdi+8]
+
+		hook::nop(hook::get_pattern("48 8B D9 48 81 C7 ? ? ? ? 8B ? ? 85 ?", 3), 7);
+
+		//Ped running camera shake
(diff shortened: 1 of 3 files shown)


blattersturm:
image

uhh, that doesn't really look like the file went in the right place

blattersturm:
Crashing is probably as you're nop'ing code that's currently being executed, perhaps defer it to a point where you can be sure the game isn't running that code or don't change code at runtime.

catmino:
> 
> 
> Crashing is probably as you're nop'ing code that's currently being executed, perhaps defer it to a point where you can be sure the game isn't running that code or don't change code at runtime (e.g. use a jitasm stub as seen in a few other places).

Yeah, crashes were only an issue when trying to make changes while game is running, both while trying to nop it or just change the values, couldn't figure out the reason whatsoever - I'm probably far too new to hooking etc. to figure it out easily. Had zero crashes with nop'ing it before the game loads though (which is how it currently works). I'll keep working on it and see if I can figure a more proper way to do it though.

Also, no idea what the heck happened with file ending up there in the end, had some issues with git.

blattersturm:
> Also, no idea what the heck happened with file ending up there in the end, had some issues with git.

I suspect it's VS' 'add file' dialog, it has a tendency to fill the wrong path - move it to the right component and rerun `fxd gen`/premake.

code/components/gta-core-five/src/DisableCameraShake.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+		hook::nop(hook::get_pattern("48 8B D9 48 81 C7 ? ? ? ? 8B ? ? 85 ?", 3), 7);
+
+		//Ped running camera shake
+		//0F 29 70 E8				movaps xmmword ptr[rax - 18h], xmm6
+		//0F 29 78 D8				movaps  xmmword ptr [rax-28h], xmm7
+		//48 81 C7 08 08 00 00		add rdi, 808h		 <--------------
+		//48 8B F1					mov rsi, rcx
+
+		hook::nop(hook::get_pattern("57 48 81 EC ? ? ? ? 48 8B B9 ? ? ? ? 0F 29 70 E8 0F 29 78 D8 48 81 C7 ? ? ? ?", 23), 7);
```
blattersturm, line 33: This pattern exists _twice_ in 2372, but you want the first one anyway. In addition, this seems like a big hack, you're relying on the value w/o offset add being `0` so it'll fail the condition below.

blattersturm:
I'll tentatively add this to the merge queue for someone to fix at a later time.

PR #933, opened 2021-08-28
tweak(vendor): update a number of dependencies

blattersturm:


blattersturm:
Unsure why this was left as a draft - I'll apply this somewhen in a normal commit later.

PR #934, opened 2021-08-29
tweak(shared): modernize fwEvent

blattersturm:
1. Removes a bunch of old-C++ dependencies.
2. Adds operator bool, const operator() and Disconnect methods.

The second part is required for a future changeset adding state bag events.


blattersturm:
Marking as draft as some internal dependencies need updating too.

PR #964, opened 2021-09-06
fix(scheduler): don't error on rejected promise

NotSomething0:


blattersturm:
This seems like an incompatible behavior change, and all you've got to say is...

> No description provided.

What does this even 'fix'? In JS, rejecting a promise leads to `await` throwing as well, same for a Task in C#.

NotSomething0:
I completely forgot about Lua error handling using pcall/xpcall a simple Google search would have solved all my problems 🤦

PR #687, opened 2021-03-21
WIP Async JS Refs

DiscworldZA:
This adds the beginning of async/await functionality to exports.

Creating async functions as exports and objects with async functions should now be possible.

PR #975, opened 2021-09-08
feat(scripting/v8): Add support for __cfx_async_retval

dunak-debug:
Author @bladecoding and @DiscworldZA (Inactive PR: https://github.com/citizenfx/fivem/pull/687)

**This PR should implement __cfx_async_retval into V8 runtime.**

Testing:
- [x] Javascript -> Javascript - bit more testing
- [x] Javascript -> Lua - bit more testing
- [x] Javascript -> C#

https://github.com/citizenfx/fivem/blob/83cb017b2023d333654ea28996b92cc8baa3f32c/data/shared/citizen/scripting/lua/scheduler.lua#L1039
Lua tends to unpack awaited result so therefore I put result as array `cb([v], null)` inside V8.

Would be glad for any help with this.

### DIFF
--- a/data/shared/citizen/scripting/v8/main.js
+++ b/data/shared/citizen/scripting/v8/main.js
@@ -90,7 +90,26 @@ const EXT_LOCALFUNCREF = 11;
 					case 0:
 						return undefined;
 					case 1:
-						return retvals[0];
+						const rv = retvals[0];
+						if (rv && rv['__cfx_async_retval']) {
+							return new Promise((res, rej) => {
+								rv['__cfx_async_retval']((v, e) => {
+									if (e != null)
+										rej(e);
+									else {
+										switch(v.length) {
+											case 0:
+												res(undefined);
+											case 1:
+												res(v[0]);
+											default:
+												res(v);
+										}
+									}
+								});
+							});
+						}
+						return rv;
 					default:
 						return retvals;
 				}
@@ -98,6 +117,8 @@ const EXT_LOCALFUNCREF = 11;
 		};
 	}
 
+	const AsyncFunction = (async () => {}).constructor;
+	codec.addExtPacker(EXT_FUNCREF, AsyncFunction, refFunctionPacker);
 	codec.addExtPacker(EXT_FUNCREF, Function, refFunctionPacker);
 	codec.addExtUnpacker(EXT_FUNCREF, refFunctionUnpacker);
 	codec.addExtUnpacker(EXT_LOCALFUNCREF, refFunctionUnpacker);
@@ -132,7 +153,19 @@ const EXT_LOCALFUNCREF = 11;
 
 		try {
 			return runWithBoundaryStart(() => {
-				return pack([refFunctionsMap.get(ref).callback(...unpack(argsSerialized))]);
+				const rv = refFunctionsMap.get(ref).callback(...unpack(argsSerialized));
+				if (rv instanceof Promise) {
+					return pack([{'__cfx_async_retval': (cb) => {
+						rv.then(v => {
+							if (cb != null)
+								cb([v], null);
+						}).catch(err => {
+							if (cb != null)
+								cb(null, err);
+						});
+					}}]);
+				}
+				return pack([rv]);
 			});
 		} catch (e) {
 			global.printError('call ref', e);


data/shared/citizen/scripting/v8/main.js
```diff
@@ -90,14 +90,27 @@ const EXT_LOCALFUNCREF = 11;
...
 					case 1:
-						return retvals[0];
+						const rv = retvals[0];
+						if (rv && rv['__cfx_async_retval']) {
+							return new Promise((res, rej) => {
+								rv['__cfx_async_retval']((v, e) => {
+									if (e != null)
+										rej(e);
+									else
+										res(v);
```
blattersturm, line 100: This doesn't seem to transparently unwrap array values like above (1 = return directly, 0 = undefined, 2 = array).

blattersturm (changes requested):

data/shared/citizen/scripting/v8/main.js
dunak-debug: Should be resolved in next commit. Also I´ve encountered issue while catching errors from __cfx_async_retval. After serializing the error object is sent as:
image

dunak-debug: Actually I could get error message by returning `err.toString()` or `err.message` inside `setCallRefFunction` but unhandled promise rejection won´t throw error this way.

blattersturm:
Now, what if there's _no_ value?

dunak-debug:
> Now, what if there's _no_ value?

serialize `undefined`, unpack as `null` 

Am I missing something important?

blattersturm:
Right, I forgot this is always packed in an array.

dunak-debug:
Done.

Also do you have any idea about mentioned promise error handling?

blattersturm (approved):

DiscworldZA:
This is what me and @bladecoding did back in the day. Glad to see this works!

Nexxed:
FYI @dunak-debug `Lua -> JS` doesn't seem to work, as an exported and async JS function will always return an empty table:
```lua
local res = exports.resource:hello(1)
print(json.encode(res)) -- returns "[]", no matter what value `hello` resolves from a promise, and with no metatable either
```
I'm unsure if this is something that isn't supported or just a bug, given that it says in the PR that `JavaScript-> RT` is working, not `RT -> JavaScript` (if that's meant to mean anything specific, anyway)

dunak-debug:
> FYI @dunak-debug `Lua -> JS` doesn't seem to work, as an exported and async JS function will always return an empty table:
> 
> ```lua
> local res = exports.resource:hello(1)
> print(json.encode(res)) -- returns "[]", no matter what value `hello` resolves from a promise, and with no metatable either
> ```
> 
> I'm unsure if this is something that isn't supported or just a bug, given that it says in the PR that `JavaScript-> RT` is working, not `RT -> JavaScript` (if that's meant to mean anything specific, anyway)

I am sure that I have tested exports. What the export looks like?

Nexxed:
> > FYI @dunak-debug `Lua -> JS` doesn't seem to work, as an exported and async JS function will always return an empty table:
> > ```lua
> > local res = exports.resource:hello(1)
> > print(json.encode(res)) -- returns "[]", no matter what value `hello` resolves from a promise, and with no metatable either
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I'm unsure if this is something that isn't supported or just a bug, given that it says in the PR that `JavaScript-> RT` is working, not `RT -> JavaScript` (if that's meant to mean anything specific, anyway)
> 
> I am sure that I have tested exports. What the export looks like?

We tested it with a regular async function (not explicitly returning a promise) and also one that returns a promise. Async functions either returned an error (with a 'see above for details' type of message, yet no info was 'above') or returned an empty array:
```js
function sleep(ms) {
  return new Promise(resolve => setTimeout(resolve, ms));
}

exports("helloAsync", async () => {
  await sleep(1000)
  return "hi"
})

exports("helloPromise", () => {
  return new Promise((res, rej) => {
    sleep(1000).then(res)
  })
})```

Do you have any written examples from when you were testing by any chance?
The above exports are not 1:1 to what we were using, but both use the same returns that we've tried that were unsuccessful.

Nexxed:
> > FYI @dunak-debug `Lua -> JS` doesn't seem to work, as an exported and async JS function will always return an empty table:
> > ```lua
> > local res = exports.resource:hello(1)
> > print(json.encode(res)) -- returns "[]", no matter what value `hello` resolves from a promise, and with no metatable either
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I'm unsure if this is something that isn't supported or just a bug, given that it says in the PR that `JavaScript-> RT` is working, not `RT -> JavaScript` (if that's meant to mean anything specific, anyway)
> 
> I am sure that I have tested exports. What the export looks like?

Ok so, I tested a little more and it turns out I was a build or two behind the commit that this was merged 🙃 
It's working fine now! 😅

PR #985, opened 2021-09-11
feat(scripting/lua): implement Citizen.SetInterval

thelindat:
People often use recursive SetTimeout functions and as I understand it isn't great for performance. Otherwise the proper way is with CreateThread and a while loop - which this essentially does ~~while returning the thread~~ while returning the interval id so you can easily close it from elsewhere in your resource.

---
- ~~To do~~
   - ~~I'd like to be able to easily adjust the wait on an active interval.~~
   - ~~A third optional argument for a function to call on close.~~

Confirmation if this is something needed or wanted would be nice. If there's anything else I need to consider then I appreciate the feedback.

---
Edit:
I've updated the function to keep things a bit cleaner and easier to utilise thanks to the metatable. SetInterval is, itself, a table tracking each interval's timer and the next id to grant when an interval is created (currentId).

Utilising the __call metamethod we can of course call the table, as before. I've formatted it to be more-or-less like the JS function `Citizen.SetInterval(callback, timer)`; leaving the timer undefined will just default to 0 so we can perform the callback on each tick.

The return value is the current wait time, allowing for different actions to be performed depending on the current time; clearing the interval just sets the ms to `-1`, which we can utilise for an "onclose" callback.

`local interval = self[id]` should prevent any nil index errors when clearing the interval.

```lua
local id = Citizen.SetInterval(function(ms)
    if ms == 1000 then
        print('hello world')
    elseif ms == -1 then
        print('goodbye cruel world')
    end
end)

RegisterCommand('test', function()
    Citizen.SetInterval[id] = 1000
    Citizen.Wait(0)
    Citizen.ClearInterval(id)
end)
```

### DIFF
--- a/data/shared/citizen/scripting/lua/scheduler.lua
+++ b/data/shared/citizen/scripting/lua/scheduler.lua
@@ -435,6 +435,35 @@ Citizen.SetTickRoutine(function(tickTime, profilerEnabled)
 	end
 end)
 
+Citizen.SetInterval = setmetatable({currentId = 0}, {
+	__call = function(self, callback, timer)
+		local di = debug_getinfo(callback, 'S')
+		local id = self.currentId + 1
+		self.currentId = id
+		self[id] = timer or 0
+		SLQ_AppendTail(thread_queue[1 --[[SLQ_T_NEW]]], thread_createRecord(function()
+			repeat
+				local interval = self[id]
+				coroutine_yield(curTime + interval)
+				callback(interval)
+			until interval == -1
+			self[id] = nil
+		end, ('interval [%s[%s]]'):format(di.short_src, id)))
+		hadThread = true
+		return id
+	end
+})
+
+SetInterval = Citizen.SetInterval
+
+function Citizen.ClearInterval(id)
+	if Citizen.SetInterval[id] then
+		Citizen.SetInterval[id] = -1
+	end
+end
+
+ClearInterval = Citizen.ClearInterval
+
 --[[
 
 	Event handling


blattersturm:
This syntax is still _fairly_ verbose, but I'm not sure if there's anything that'd be more friendly without leading to issues with Lua IDE/editor support needing to be adapted.

Exposing raw coroutines to close seems a bit silly as well, however: how does current SetTimeout handle this?

thelindat:
> Exposing raw coroutines to close seems a bit silly as well, however: how does current SetTimeout handle this?  

It seemed like the easiest way to close the coroutine if desired - the current method I use stores an argument in a table with its value set to the wait time. Not sure what you mean with SetTimeout; it just sets a trigger time then dies off.

I've cleaned up and rewritten the function that I currently use, but I don't know if it's necessarily better?
Perhaps a little more than necessary for implementation and just sticking to Set/Clear Interval.
```lua
local Intervals = {}
function Citizen.SetInterval(id, msec, callback, onclear)
	if Intervals[id] and msec then
		Intervals[id] = msec
	else
		local di = debug_getinfo(callback, 'S').short_src
		local record = thread_createRecord(function()
			Intervals[id] = msec
			repeat
				coroutine_yield(curTime + Intervals[id])
				callback(Intervals[id])
			until Intervals[id] == -1 and (onclear and onclear() or true)
			Intervals[id] = nil
		end)
		SLQ_AppendTail(thread_queue[1 --[[SLQ_T_NEW]]], record, Intervals[id])
		hadThread = true
	end
end

SetInterval = Citizen.SetInterval

function Citizen.ClearInterval(id)
	if Intervals[id] then Intervals[id] = -1 end
end

ClearInterval = Citizen.ClearInterval
```

```lua
SetInterval(1, 1000, function(ms)
	print(ms)
end, function()
	print('broke interval')
end)

RegisterCommand('adjust', function(_, args)
	SetInterval(1, args[1])
end)

RegisterCommand('break', function()
	ClearInterval(1)
end)
```

thelindat:
I've been looking through the scheduler a fair bit over the last few days and trying (struggling) to figure out how the whole thing works. I don't want to mess around with it too much in order to not break functionality, but maybe initialising the interval function as part of thread_createRecord (less function definitions) and identifying it as an interval in the records would be better than an additional table.

Is defining the interval id as an argument vs returning its thread reference from the scheduler preferable - or perhaps sending its bid and handling the break elsewhere?

PR #988, opened 2021-09-12
RedM sampler state fix

blattersturm:
Fixes this little issue reported to occur on D3D12:

GIF


MoosheTV:
I built and tested `blattersturm:fix/rdr3-sampler-state` and can confirm it fixes the issue.

PR #993, opened 2021-09-12
Update scheduler.lua

niekschoemaker:
I have no idea if this would be anything interesting.

Could be usefull to save a native call to GetGameTimer if you want to get tick time.

So this would allow you to do

```lua
local time = Citizen.Wait(0)
```

Instead of
```lua
Citizen.Wait(0)
local time = GetGameTimer()
```

### DIFF
--- a/data/shared/citizen/scripting/lua/scheduler.lua
+++ b/data/shared/citizen/scripting/lua/scheduler.lua
@@ -278,7 +278,7 @@ local function resumeThread(thread, coro) -- Internal utility
 		Citizen_SubmitBoundaryStart(thread[4 --[[SLQ_BOUNDARY]]], coro)
 	end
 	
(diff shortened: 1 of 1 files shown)


blattersturm:
Nice attempt to be clever, but by then one could also just implement a `curTime` getter in the ScRT itself, optimize GET_GAME_TIMER OAL-style, or _anything_ else almost.

niekschoemaker:
> 
> 
> Nice attempt to be clever, but by then one could also just implement a `curTime` getter in the ScRT HLL code itself, optimize GET_GAME_TIMER OAL-style, or _anything_ else almost.

Yeah as I said, just a small idea I had, seems like a logical return value for Citizen.Wait, but if not just feel free to close this.

blattersturm:
PR doesn't apply anymore after tickless.

PR #997, opened 2021-09-13
Server game state improvements

blattersturm:
A little memory/allocator performance saving.

Needs testing on a real server.


blattersturm:
Sizes are from 2372 and may not be valid elsewhere.

Log replacement pattern:

`\[     .*?\] \[         FiveM\]             MainThrd/ `

Issue #950, opened 2021-09-05
imgui/cursor/tooltips in cfx-ui look small at 4k-scaled resolution


blattersturm:
This probably should be scaled assuming a baseline, maybe can get rid of the different font sizes for imgui too (and the 'nothing fits on 800x600; thing).

PR #999, opened 2021-09-14
tweak(conhost): Scale imgui fonts/styles and citizen_cursor

LWSS:
https://github.com/citizenfx/fivem/issues/950

Unfortunately imgui scaling is pretty rigged, so instead of removing the different fonts, I added more :P 

I did notice that the menu cursor(citizen_cursor) was very small when super scaling to 4k, so I changed it to follow the monitor's scale % setting. This includes doubling the .png itself because stretching it leads to jaggies on the right.


blattersturm:
While this implementation made sense, I believe it's more rational to just take a more naive approach of calculating the size relative to a reference height (e.g. 1080) and setting the ImGui::IO global scaling option.

PR #1000, opened 2021-09-14
Try to parse valid tickets anyway when sv_lan is set to 1

ArchieeLL:
This is a simple change to allow all identites being exposed of clients who provive a valid ticket.

### DIFF
--- a/code/components/citizen-server-impl/src/InitConnectMethod.cpp
+++ b/code/components/citizen-server-impl/src/InitConnectMethod.cpp
@@ -505,40 +505,40 @@ static InitFunction initFunction([]()
 
 			TicketData ticketData;
 
-			if (!lanVar->GetValue())
-			{
-				auto ticketIt = postMap.find("cfxTicket");
+			auto ticketIt = postMap.find("cfxTicket");
 
-				if (ticketIt == postMap.end())
+			if (ticketIt == postMap.end() and !lanVar->GetValue())
+			{
+				sendError("No authentication ticket was specified.");
+				return;
+			}
+
+			try
+			{
+				std::string ticketError;
+
+				if (!VerifyTicket(guid, ticketIt->second, &ticketError) and !lanVar->GetValue())
 				{
-					sendError("No authentication ticket was specified.");
+					sendError(fmt::sprintf("Ticket authorization failed. %s", ticketError));
 					return;
 				}
 
-				try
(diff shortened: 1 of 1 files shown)


blattersturm:
What is the use case of this? sv_lan 1 is not meant to be used and will rather be removed at some point in the future due to common misconfiguration.

ArchieeLL:
Use case:
I have a server set up at work within the local network. Some PCs don't have internet access (only LAN), some do, thus the server needs to be sv_lan. This patch allows those who do have internet access to have all their identifiers available.

blattersturm:
Again, without internet access you already can't refresh the Rockstar ticket, so that already makes no actual sense.

ArchieeLL:
Yes, we do have to connect with internet like every 2-3 days, but that's not really that much of a problem, we plug mobile routers for a minute to achieve this.

blattersturm:
Anyway, it would indeed make more sense to just allow this pass only for a _final_ (as in, after proxy address checks) IP within local unroutable ranges.

ArchieeLL:
Well, I understand that, but as long as you provide the possibility to easily host/join an sv_lan server (it's as easy as deleting adhesive from both, some time ago it caused crashes etc. but now it doesn't), I think you should allow a little improvement. I think if some server owner wanted to allow cheaters to join, it wouldn't be much of a problem for them to make a smart proxy to disable sv_lan once a user with valid ticket is joining and enable it when a 'cheater' joins.

blattersturm:
The concern here is people setting sv_lan to 'circumvent downtime' without being aware of the consequences. If we remove one more visible consequence, it'll be somewhat problematic if people unintentionally enable this.

A local range check would only be for the check/accept identifiers stuff as proposed in this PR, and similarly I think it might make sense to auto-enable a sv_lan if no svadhesive is present.

ArchieeLL:
I agree, but I think someone else would have to code it, because all my skills let me do was finding the code line which is responsible for not giving the full identifiers when sv_lan is on. Also, another use case is 'circumventing downtime' as you said, because when FiveM services are down, both server owners and players sometimes get really pissed, so that's another reason why sv_lan is sometimes useless not for actually playing on LAN.

blattersturm:
Abandoned PR. Each of the use cases this solves has a few better ways of handling these.

Issue #956, opened 2021-09-06
TempBan Native Not Working


EayZed:
TempBanPlayer() or Citizen.InvokeNative(0x1E35DBBA) (https://docs.fivem.net/natives/?_0x1E35DBBA) Native Not Exists On FxServer

blattersturm:
Yeah, it doesn't, and this is intended. Can't be removed from native decls for compatibility.

PR #1001, opened 2021-09-15
Update TempBanPlayer.md

AvarianKnight:
This lets the user(s) know that this native is no longer available

Related to https://github.com/citizenfx/fivem/issues/956


blattersturm:
This should rather have a way to 'hide' natives from the doc listing, perhaps in the pass of built-in native decl reviews.

Issue #960, opened 2021-09-06
Server should support directly setting data on nodes


blattersturm:
Instead of all entity-setter natives being RPC to the owning client, some obvious calls (such as SET_ENTITY_COORDS) should directly be able to be set on the server, writing into the data node.

This'd require a bit of work, however:

1. Owning clients to accept clone-sync packets from the server for whatever entity this command is executed on. These should probably be scoped to only the nodes that have been set by the server.
2. The server to reject these nodes from any sequence before a clone-sync with a state update has reasonably been accepted (e.g. in NAK mode, the RTT/variance, unless a NAK got sent, though a simpler solution would be to send the last read sequence with any clone sync packet).
3. Post-processing functions (like reading of cached positions) to be handled correctly.
4. More nodes to be using the generic Serialize function rather than specific Parse/Unparse.

Spajker7:
This would be awesome. A couple of questions tho.

> 2. The server to reject these nodes from any sequence before a clone-sync with a state update has reasonably been accepted (e.g. in NAK mode, the RTT/variance, unless a NAK got sent, though a simpler solution would be to send the last read sequence with any clone sync packet).
If this gets implemented, would we be able to get some sort of event where we can reject a node update and revert the previous value to the owner client? Use cases would be for example static objects which we don't want "hacked" clients moving around, only accepting player health update if the new health is lower then old, etc.

Also, would it make more sense to have some more "direct" access to set/get data in data nodes instead of trying to emulate client setters which aren't always as "direct"?

AvarianKnight:
Is there anything stopping most data nodes from using Serialize over the Parse/Unparse, if nothing then would it be good to update them all in one PR or put them in different PRs (1 pr per node)

blattersturm:
Incomplete data nodes shouldn't be moved to Serialize at all as that'd be a little misleading.

PR #1005, opened 2021-09-18
feat(server/impl): update data node structs for RDR3

Ktos93:
Idk if it's the right way to do this. I was used fivem SyncTrees as example.I did this becouse i wanted to use entityCreating event to block spawning some vehicles but GET_ENTITY_MODEL and GET_ENTITY_POPULATION_TYPE wasnt here

```
AddEventHandler('entityCreating', function(entity)
    print(entity)
    print(GetEntityTypeName(GetEntityType(entity)))
    print(GetPopulationTypeName(GetEntityPopulationType(entity)))
    print(GetEntityModel(entity))  
end)

function GetPopulationTypeName(type)
    for k, v in pairs(entityPopulationTypes) do
        if v == type then
            return k
        end
    end
end

function GetEntityTypeName(type)
    for k, v in pairs(entityTypes) do
        if v == type then
            return k
        end
    end
end

```

image

### DIFF
--- a/code/components/citizen-server-impl/include/state/SyncTrees_RDR3.h
+++ b/code/components/citizen-server-impl/include/state/SyncTrees_RDR3.h
@@ -1200,7 +1200,7 @@ struct SyncTree : public SyncTreeBase
 			*modelHash = objectCreationNode->m_model;
 			return true;
 		}
-#if 0
+
 		auto[hasPan, playerAppearanceNode] = GetData<CPlayerAppearanceDataNode>();
 
 		if (hasPan)
@@ -1209,7 +1209,6 @@ struct SyncTree : public SyncTreeBase
 			return true;
 		}
 
-#endif
 		return false;
 	}
 
--- a/code/components/citizen-server-impl/include/state/SyncTrees_RDR3.h
+++ b/code/components/citizen-server-impl/include/state/SyncTrees_RDR3.h
@@ -14,7 +14,22 @@
 
 namespace fx::sync
 {
-struct CVehicleCreationDataNode { bool Parse(SyncParseState& state) { return true; } };
+struct CVehicleCreationDataNode
+{ 
+	uint32_t m_model;
+	ePopType m_popType;
+
+	bool Parse(SyncParseState& state)
+	{
+		uint32_t model = state.buffer.Read<uint32_t>(32);
+		m_model = model;
+
+		uint8_t popType = state.buffer.Read<uint8_t>(4);
+		m_popType = (ePopType)popType;
+		
+		return true; 
+	} 
+};
 
 struct CAutomobileCreationDataNode { bool Parse(SyncParseState& state) { return true; } };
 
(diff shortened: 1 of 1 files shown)


code/components/citizen-server-impl/include/state/SyncTrees_RDR3.h
```diff
@@ -241,7 +256,25 @@ struct CSectorPositionDataNode
...
-struct CPedCreationDataNode { bool Parse(SyncParseState& state) { return true; } };
+struct CPedCreationDataNode : GenericSerializeDataNode<CPedCreationDataNode>
+{ 
+	uint32_t m_model;
+	ePopType m_popType;
+
+	template<typename TSerializer>
+	bool Serialize(TSerializer& s)
+	{ 
+		// 7(?)
```
Disquse, line 267: That's 4, not 7. All good.

code/components/citizen-server-impl/include/state/SyncTrees_RDR3.h
```diff
@@ -367,7 +400,22 @@ struct CDoorScriptGameStateDataNode { bool Parse(SyncParseState& state) { return
...
 struct CHeliControlDataNode { bool Parse(SyncParseState& state) { return true; } };
 
-struct CObjectCreationDataNode { bool Parse(SyncParseState& state) { return true; } };
+struct CObjectCreationDataNode
+{
+	uint32_t m_model;
+
+	bool Unparse(SyncUnparseState& state)
+	{
+		state.buffer.Write<uint32_t>(32, m_model);
```
Disquse, line 409: Why would you implement unfinished `Unparse`? And why not implement `Parse` then?
Also, are you sure that this is 32 bits of model? That's not so obvious according to data node's code. It's most likely a single bit, if this bit is true, it read something with a 32 bits size.

code/components/citizen-server-impl/include/state/SyncTrees_RDR3.h
```diff
@@ -14,7 +14,22 @@
 
 namespace fx::sync
 {
-struct CVehicleCreationDataNode { bool Parse(SyncParseState& state) { return true; } };
+struct CVehicleCreationDataNode
+{ 
+	uint32_t m_model;
+	ePopType m_popType;
+
+	bool Parse(SyncParseState& state)
```
Disquse, line 22: According to #960, we should stick to generic serialize instead of Parse/Unparse. Same for other data nodes.

Ktos93:
Changed to serialize but idk how to make it work for object node

blattersturm:
f to review

code/components/citizen-server-impl/include/state/SyncTrees_RDR3.h
```diff
@@ -367,7 +403,20 @@ struct CDoorScriptGameStateDataNode { bool Parse(SyncParseState& state) { return
...
-struct CObjectCreationDataNode { bool Parse(SyncParseState& state) { return true; } };
+struct CObjectCreationDataNode  : GenericSerializeDataNode<CObjectCreationDataNode>
+{
+	uint32_t m_model;
+	bool m_hasInitPhysics;
+
+	template<typename Serializer>
+	bool Serialize(Serializer& s)
+	{
+		s.Serialize(32, m_model);
```
Disquse, line 414: This part is invalid. First field is a bool. It seems model hash is serialized somewhere later in the data node.

Disquse (approved):
Since `CObjectCreationDataNode` is not actually used in this PR, we can merge it. Tested a bit on local server.

PR #1006, opened 2021-09-18
feat(discord): rdr3 support for identity

Ktos93:
I have the impression that I have missed something important. I created a PR so that someone who knows what exactly needs, to guide me. Doesn't require any action on your part to add it?


WooAf:
can we add that please

blattersturm:
Seems to have a merge conflict and assumes we have OAuth handling for the 'RedM' app ID (we don't) and doesn't add any way for this to be signaled.

Will be done manually sometime as using the 'FiveM' app ID should be fine, i.e. only removing the `#ifdef` is sufficient.

PR #1017, opened 2021-09-19
feat(mumble): Add MUMBLE_REMOVE_VOICE_TARGET_CHANNEL

AvarianKnight:
This should be considered a draft, the primary reason I'm making this without adding the same natives for _PLAYER and _PLAYER_BY_SERVER_ID is to receive feedback and see if this is something I should pursue further.

The primary use case for this is to get rid of the grid system in pma-voice (see: https://github.com/AvarianKnight/pma-voice/pull/148) and just add players to each others voice target without having to constantly call `MUMBLE_CLEAR_VOICE_TARGET_CHANNELS` and re-add every user back to the players voice target.


code/components/gta-net-five/src/MumbleVoice.cpp
```diff
@@ -705,6 +705,29 @@ static HookFunction hookFunction([]()
...
 
+        fx::ScriptEngine::RegisterNativeHandler("MUMBLE_REMOVE_VOICE_TARGET_CHANNEL", [](fx::ScriptContext& context)
+        {
+            auto id = context.GetArgument<int>(0);
+            auto channel = context.GetArgument<int>(1);
+
+            if (id >= 0 && id < 31)
+            {
+                auto targetChannel = fmt::sprintf("Game Channel %d", channel);
+                std::vector<VoiceTargetConfig::Target>& targets = vtConfigs[id].targets;
```
blattersturm, line 716: Does vtConfigs get accessed by any other thread here i.e. potentially need locking?

Also, long types sound like they'd be better off with `auto&`.

code/components/gta-net-five/src/MumbleVoice.cpp
```diff
@@ -705,6 +705,29 @@ static HookFunction hookFunction([]()
...
+                for (size_t i = targets.size(); i--;)
+                {
+                    VoiceTargetConfig::Target& target = targets[i];
+                    if (!target.channel.empty() && target.channel.compare(targetChannel) == 0)
+                    {
+                        targets.erase(targets.begin() + i);
```
blattersturm, line 722: std::vector removal is nasty (rather use std::list as a replacement, maybe? especially if these can get big) and this is not a very idiomatic way to do removal (erase should remove another iterator, worst-case)

code/components/gta-net-five/src/MumbleVoice.cpp
```diff
@@ -705,6 +705,29 @@ static HookFunction hookFunction([]()
...
+            auto channel = context.GetArgument<int>(1);
+
+            if (id >= 0 && id < 31)
+            {
+                auto targetChannel = fmt::sprintf("Game Channel %d", channel);
+                std::vector<VoiceTargetConfig::Target>& targets = vtConfigs[id].targets;
+                for (size_t i = targets.size(); i--;)
+                {
+                    VoiceTargetConfig::Target& target = targets[i];
+                    if (!target.channel.empty() && target.channel.compare(targetChannel) == 0)
```
blattersturm, line 720: Why `.compare()`?

ext/native-decls/MumbleRemoveVoiceTargetChannel.md
```diff
@@ -0,0 +1,17 @@
...
+ns: CFX
+apiset: client
+---
+## MUMBLE_REMOVE_VOICE_TARGET_CHANNEL
+
+```c
+void MUMBLE_REMOVE_VOICE_TARGET_CHANNEL(int targetId, int channel);
+```
+
+Removes the voice channel from the users voice targets.
```
blattersturm, line 11: `the voice channel` -> `a` or `the specified` voice channel
`users` -> `user's`

ext/native-decls/MumbleRemoveVoiceTargetChannel.md
```diff
@@ -0,0 +1,17 @@
...
+
+Remover for [MUMBLE_ADD_VOICE_TARGET_CHANNEL](#_0x4D386C9E)
```
blattersturm, line 13: 'Remover' sounds a bit weird, 'Performs the opposite operation of ..`?

ext/native-decls/MumbleRemoveVoiceTargetChannel.md
```diff
@@ -0,0 +1,17 @@
...
+
+## Parameters
+* **targetId**: A Mumble voice target ID, ranging from 1..30 (inclusive).
+* **channel**: The game voice channel ID to remove.
```
blattersturm, line 17: `to remove from the target`?

code/components/gta-net-five/src/MumbleVoice.cpp
AvarianKnight: Doesn't look like it, vtConfigs is only ever modified by natives, which is sent to `MumbleClient::UpdateVoiceTarget` locks the client mutex and updates `m_pendingVoiceTargetUpdates`

code/components/gta-net-five/src/MumbleVoice.cpp
AvarianKnight: I'll switch this to std::list as this has the possibility to add/remove players constantly, and depending on other peoples use cases it could be quite a bit larger than just 128 players

code/components/gta-net-five/src/MumbleVoice.cpp
AvarianKnight: There wasn't really a reason, though now that I look back it looks bad syntactically

code/components/gta-net-five/src/MumbleVoice.cpp
```diff
@@ -700,30 +700,119 @@ static HookFunction hookFunction([]()
...
+				if (g_mumble.connected)
+				{
+					vtConfigs[id] = {};
+					g_mumbleClient->UpdateVoiceTarget(id, vtConfigs[id]);
+				}
 			}
 		});
 
-		fx::ScriptEngine::RegisterNativeHandler("MUMBLE_CLEAR_VOICE_TARGET_CHANNELS", [](fx::ScriptContext& context)
+        fx::ScriptEngine::RegisterNativeHandler("MUMBLE_REMOVE_VOICE_TARGET_CHANNEL", [](fx::ScriptContext& context)
```
blattersturm, line 711: How did this end up indented with spaces, ignoring the .editorconfig?

code/components/gta-net-five/src/MumbleVoice.cpp
```diff
@@ -700,30 +700,119 @@ static HookFunction hookFunction([]()
...
+
+            if (id >= 0 && id < 31)
+            {
+				if (g_mumble.connected)
+				{
+					auto targetChannel = fmt::sprintf("Game Channel %d", channel);
+					auto& targets = vtConfigs[id].targets;
+					targets.remove_if([targetChannel](auto& target)
+					{
+						if (target.channel == targetChannel)
```
blattersturm, line 724: Can be simplified to just a `return`.

code/components/gta-net-five/src/MumbleVoice.cpp
```diff
@@ -700,30 +700,119 @@ static HookFunction hookFunction([]()
...
+				{
+					VoiceTargetConfig::Target ch;
+					std::wstring targetName = g_mumbleClient->GetPlayerNameFromServerId(serverId);
+
+					if (!targetName.empty())
+					{
+						auto& targets = vtConfigs[id].targets;
+						targets.remove_if([targetName](auto& target)
+						{
+							if (target.users.size() > 0 && std::find(target.users.begin(), target.users.end(), targetName) != target.users.end())
```
blattersturm, line 781: size() > 0 check is a little unneeded, std::find handles begin == end fine

also, is each user a separate users target?

code/components/gta-net-five/src/MumbleVoice.cpp
```diff
@@ -700,30 +700,119 @@ static HookFunction hookFunction([]()
...
+			auto id = context.GetArgument<int>(0);
+
+			if (id >= 0 && id < 31)
+			{
+				if (g_mumble.connected)
+				{
+					auto& targets = vtConfigs[id].targets;
+					targets.remove_if([](auto& target)
+					{
+						if (target.channel.empty())
```
blattersturm, line 805: Why does this only remove all non-channels?

code/components/gta-net-five/src/MumbleVoice.cpp
```diff
@@ -700,30 +700,119 @@ static HookFunction hookFunction([]()
...
+							return false;
+						});
 					}
 				}
 
 				g_mumbleClient->UpdateVoiceTarget(id, vtConfigs[id]);
 			}
 		});
+
+		fx::ScriptEngine::RegisterNativeHandler("MUMBLE_CLEAR_VOICE_TARGET_CHANNELS", [](fx::ScriptContext& context)
```
blattersturm, line 794: Missing in codegen?

blattersturm (changes requested):

code/components/gta-net-five/src/MumbleVoice.cpp
AvarianKnight: I actually have no idea why, I'll try and fix it when I force push

code/components/gta-net-five/src/MumbleVoice.cpp
AvarianKnight: No, it is one of the original mumble natives, not sure why it's showing it as a new entry

code/components/gta-net-five/src/MumbleVoice.cpp
AvarianKnight: This was left in because `MUMBLE_ADD_VOICE_TARGET_CHANNEL` also pushes onto vtConfigs[id].targets, so this was left as a fast denial

> also, is each user a separate users target?

I'm not entirely sure what you're asking

code/components/gta-net-five/src/MumbleVoice.cpp
AvarianKnight: Because I forgot a exclamation point, its supposed to only clear _if_ its a voice channel

code/components/gta-net-five/src/MumbleVoice.cpp
blattersturm: If it is, I'd guess it's duplicate?

AvarianKnight: It's not, `MUMBLE_REMOVE_VOICE_TARGET_PLAYER_BY_SERVER_ID` keeps some of the code so I guess it detects it as a rewrite?

PR #1039, opened 2021-09-26
add more onesync server sided natives

JohnathanUK:
First time committing something here so, apologies if I do not understand!


blattersturm:
That's not how cached getters work at all. These have to be defined with actual get routines as well, the caching in RPC stuff is only for when they have been used to set.

JohnathanUK:
Oh, thanks!

PR #1061, opened 2021-10-05
Update GetInvokingResource.md

Greek-Gamer:
Added more info about the native!

### DIFF
--- a/ext/native-decls/GetInvokingResource.md
+++ b/ext/native-decls/GetInvokingResource.md
@@ -9,4 +9,20 @@ char* GET_INVOKING_RESOURCE();
 ```
 
 
-## Return value
+## Example
+
+```lua
+AddEventHandler("event", function(cb)
+    cb(GetInvokingResource())
+end)
+
+--Script1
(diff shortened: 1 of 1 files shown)


jaymo1011:
Really isn't that informative nor intuitive to a first time reader. This lacks a description and your examples are not well commented, including a weird "cb" pattern.

jaymo1011:
Of course you can try again but you're going to want to understand the *implementation* of the native before you write documentation about it. Read the source for what you're documenting! (If you can't understand the source code however, *you* are probably not the one who should be documenting it.)

Greek-Gamer:
> Of course you can try again but you're going to want to understand the _implementation_ of the native before you write documentation about it. Since this is a CFX native, a native implemented by FiveM itself rather than the game, read the source code for what you're documenting! (If you can't understand the source code however, _you_ are probably not the one who should be documenting it.)

Understood !

blattersturm:
This addition does not add enough value at this time, so we'll be closing this idle PR. See Jaymo's reply above.

PR #1088, opened 2021-10-20
fix(server/state): fixes playerLeftScope not emiting on player dropped

### LINKED ISSUE #1046: playerLeftScope does not emit when player leaves the server
Description: `playerLeftScope` does not emit when a player leaves the server
Expected behavior: I guess `playerLeftScope` should emit as well as `playerEnteredScope` does emit for each player who is on that scope when a player joins the server

AvarianKnight:
This closes #1046

### DIFF
--- a/code/components/citizen-server-impl/src/state/ServerGameState.cpp
+++ b/code/components/citizen-server-impl/src/state/ServerGameState.cpp
@@ -2636,7 +2636,10 @@ void ServerGameState::HandleClientDrop(const fx::ClientSharedPtr& client, uint16
 
 	if (fx::IsBigMode())
 	{
-		clientRegistry->ForAllClients([this, &client, netId](const fx::ClientSharedPtr& tgtClient)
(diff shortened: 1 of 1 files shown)


dev-callum:
Ive looked over it and it looks good i reckon this should be merged

blattersturm:
Are you sure this change won't lead to any compatibility concerns with scripts that rely on a workaround for this?

AvarianKnight:
> Are you sure this change won't lead to any compatibility concerns with scripts that rely on a workaround for this?

I'm pretty sure it most definitely would, as many people use playerDropped for a workaround.

I can't think of a solution for this either. I'm also not entirely sure of the use cases for this event either, as it's mostly used for "scoped events" iirc, which you can just use state bags + state bag listeners for too.

AvarianKnight:
Going to close this PR as this can't be done with maintaining backwards compatibility, plus I'm pretty sure using these events is already discouraged due to performance reasons.

PR #1109, opened 2021-11-01
feat(nui): allow using texture cursor instead of host cursor

Snowthisway:
Offers the ability to use the texture cursor instead of the host cursor with SET_NUI_CURSOR_TYPE;

I think the texture cursor looks alot better than the default OS cursor most people will have.


Snowthisway:
Having to add css to every single resource for existing projects to consistently change the cursor is very tedious.
I will be looking forward to your bitmap-to-HCURSOR solution.

PR #1124, opened 2021-11-07
feat(conhost): load extra glyph ranges to allow localized console output

Sadler2:
Console font has much more glyphs than FiveM loads. It's enough to enable them to allow fully localized console output for many European languages:
fivemlocale


LegendProjects:
Love you!

PR #1127, opened 2021-11-09
fix(sdk): show selected updateChannel when changing

itschip:
Fixed an issue when you selected a updateChannel nothing would happen in the ui. I believe this was a due to that project.manifest only loads when the sdk is initalized.

If you want this to be updated in a client storage, please let me know :D

Other: Improves ui feedback. You would have to reload sdk to see the change previously


nihonium-cfx:
Thank you for spotting this bug! While this approach may look valid it is not due the fact it is supposed to propagate manifest changes back to the shell rather than having diverged state.

PR #1128, opened 2021-11-10
feat(scripting/lua): added netEvent

Nova-px:
Triggers the event "netEvent" if any net event is coming in from client or server.


blattersturm:
... why? That's just weird.

This implementation is also all wrong, are you sure you're OK?

PR #1137, opened 2021-11-15
tweak(mumble-voip): New noise suppression

Vinipux322:
[Showcase](https://www.youtube.com/watch?v=VT0-JdRbgHg) 
🐌


.gitmodules
```diff
@@ -277,3 +277,6 @@
 [submodule "ext/sdk/resources/sdk-root/fxcode"]
 	path = ext/sdk/resources/sdk-root/fxcode
 	url = https://github.com/citizenfx/fxcode.git
+[submodule "vendor/rnnoise"]
+	path = vendor/rnnoise
+	url = https://github.com/almoghamdani/rnnoise-cmake
```
blattersturm, line 282: why weird fork

Vinipux322: it's just rnnoise for cmake, because by default it's supports only make(

Vinipux322: so, i can fork it, if it changes something

Vinipux322: i change it for my fork from original repository, now it's should be fine :)

blattersturm: >it's just rnnoise for cmake, because by default it's supports only make(

the original repo has the same layout, doesn't it? you're not even using either build system here. why?

Vinipux322: so, now i deleted useless cmake, but by default rnnoise can't compiled on windows, my fork have some changes to fix it

blattersturm:
Lots of commits. :/

Vinipux322:
> Lots of commits. :/
> 
> Also still uses own fork instead of xiph.org, why?

Hi, sorry of a lot of commits( by default xiph.org rnnoise can't compiled on windows, my fork have some changes to fix it

blattersturm:
Merge conflict as well by now, this is in my queue to manually fix at some point (as the author seems unable or unwilling to fix some requirements, and even if the fork is 'needed' this should be put in the `citizenfx/` org as with any non-upstream forks).

PR #1139, opened 2021-11-17
Assorted cursor fixes

blattersturm:
This PR includes two changes:

- b015dcc5433560287261deaec27cd49645aa44b0 implements something similar to #1109, replaces the default arrow cursor in NUI contexts with the citizen_cursor, though keeping the hardware cursor rendering for it.
- 3f8a1d129426b551cc9c9c8a7b0d5e11b9182dee *may* help with https://forum.cfx.re/t/cursor-invisible-after-opening-f8-console/4768705/27, I'm unable to actually repro the issue there so it might not fix it, however this flag being added doesn't break any existing behavior.

## Rationale
Bug fix and user request.

## Potential impact
No breaking issues are expected. This does not change any public API either.


thers (approved):
Looks like it should not affect FxDK, looks good to me then!

PichotM:
Fixed for me :)

PR #1144, opened 2021-11-22
Better looking adaptive cards (CSS)

MrSentex:
Removing the gray background makes using cards more viable for custom defer messages.
And removing the padding makes a possible background image more integrated.

Before:
before

After:
after

### DIFF
--- a/ext/cfx-ui/src/app/servers/connecting-popup.component.scss
+++ b/ext/cfx-ui/src/app/servers/connecting-popup.component.scss
@@ -116,11 +116,8 @@
 
 				width: calc(100% + var(--q2));
 
-				padding: var(--q1);
-			}
-
-			@include theme() using ($theme) {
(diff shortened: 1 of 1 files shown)


nihonium-cfx:
This was done on purpose, to emphasize UGC nature of this element, not to blend it with main UI. Apparently it's not doing it's job perfectly, heh. That said, if you have a better solution - would be great to see it!

MrSentex:
It may suffice to simply make the background transparent or not set it at all. Alternatively, with my solution and a link to https://adaptivecards.io/ in muted text on the bottom right to refer to the technology used could be great too.

LxegidijusxL29x:
tu ji palik o kita turi istrynti ir kitius i kelti yra negalima

blattersturm:
Idle PR, closing. `cfx-ui` is in the process of being rewritten and this feedback will hopefully be taken along.

PR #1146, opened 2021-11-22
Fix IconColor being set to Color attribute of Checkpoint

DatGuy1:
Probably originally due to conflicting definitions, IconColor was set to the checkpoint's Color regardless of what value was passed to the setter

### DIFF
--- a/code/client/clrcore/External/Checkpoint.cs
+++ b/code/client/clrcore/External/Checkpoint.cs
@@ -359,7 +359,7 @@ namespace CitizenFX.Core
 				{
 					return;
 				}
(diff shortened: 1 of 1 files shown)


blattersturm:
I think this already does nothing, however, since MemoryAccess/address isn't implemented here.

blattersturm:
https://github.com/citizenfx/fivem/blob/28053240879254d6ad62406d63c5861be6785d93/code/client/clrcore/External/Checkpoint.cs#L214

Yeah.

blattersturm:
This PR seems unneeded as such. :/

PR #1150, opened 2021-11-23
feat(scripting/lua): register msgpack table extension

thelindat:
Allow packing of tables without losing metatables (within the Lua ScRT). Maybe there's a reason not to do this 🤷

```lua
players[source] = setmetatable(self, {
	__index = mt,
})
	
exports('getPlayer', function(source)
	return players[source]
end)


local player = exports.resource:getPlayer(1)
player:set('name', 'not an rp name'
local name = player:get('name') -- you get the idea
```


blattersturm:
Doing this would break **all** Lua tables (aka arrays/maps) sent to any other ScRT and over the network.

Please, don't even think of this. There's no 'simple' solution to doing what you're trying to do.

PR #1152, opened 2021-11-23
Add mumble channel natives for rdr3 compatibility

Ekinoxx0:
After some testing with the recent Disquse PR, it seems there are no equivalent for some natives present in gta5 :
 - NetworkClearVoiceChannel (0xE036A705F989E049)
 - NetworkSetVoiceChannel (0xEF6212C2EFEF1A23)
 
These natives simply don't exist in RDR3 so there is no way to move the mumble client to any other channel (from my understanding).

These natives can be added back simply with this PR.
If Disquse has other plans for these natives, or if you think this is not the right way to implement this, feel free to close this PR as I was just trying to help with my minimal knowledge :)


Disquse:
I simply forgot about those natives 😅

PR #1154, opened 2021-11-25
feat(extra-natives/five): add portal entities natives

manzarek123:


blattersturm:
What's an example use case for these?

manzarek123:
I am using them for native audio occlusion dev tool.
They can also be useful for MLO developers I guess

PR #1161, opened 2021-11-28
'Tickless' resources

### LINKED ISSUE #1153: Unify tick scheduling to be done outside of HLLs
On a server such as [Transport Tycoon](https://servers.fivem.net/servers/detail/2epova), ETW traces show we can potentially save 30%+ in _total_ tick time if not using Lua for scheduling logic anymore.

I'm currently investigating a system of 'bookmarks' that a ScRT can schedule in the script host, which should refer to e.g. a Lua 'thread' state to be resumed.

Unrelatedly (for a later work item), it seems 70-90% of ref call time is _literal_ nonsense overhead in e.g. making temporary coroutines and inefficient calls for (de)serialization/allocation/etc.

blattersturm:
Removes `OnTick` for resources itself, instead handling this in manager scope. Also, implements native 'bookmarks' for Lua scheduling instead of doing this in the HLL.

Closes #1153 once merged - perhaps-WIP currently.

# Rationale

Servers often have hundreds of (mostly idle) Lua resources. Profiling showed around 30-40% of script frame time spent on servers like Tycoon (see #1153) was overhead.

# Validation

## Performance
* Earlier versions of this patch set went down from 1.5 msec of script time to <0.2 msec of script tick time in a contrived setup (200 resources each having a single Wait(5000) no-op loop + chat/spawnmanager/mapmanager as used in basic-gamemode)
* Tycoon went down from 5.8 msec of script time to 4.0 msec of script time.
* The queue bits aren't micro-optimized, needs profiling later on.

## Correctness
* Can join and play on Tycoon and a default server without issues. vMenu (a C# resource) also still executes and is tracked.
* Stack trace stitching was tested using `stest` series of resources.
* Profiler was tested to still record general resource data.
* Resource monitor shows data, uncertain if it is entirely correct.
* Citizen.Await still works.

# Risks

* Execution order doesn't exactly match old logic, sometimes resources may run interleaved.
* ~~Profiler has only had minimal testing,~~ some other diag tools (resmon?) may break too in rare scenarios (resource start/stop?).
* ~~`StartResource` on the server from a bookmark may be unsafe, this hasn't been tested.~~ Works fine.
* ~~RPC hasn't been tested, yet.~~ Also works, and correctly attributes to script.

### DIFF
--- a/code/client/shared/EventCore.h
+++ b/code/client/shared/EventCore.h
@@ -311,6 +311,11 @@ public:
 private:
 	size_t ConnectInternal(TFunc func, int order)
 	{
+		if (!func)
+		{
+			return -1;
+		}
+
 		auto cookie = m_connectCookie++;
 		auto cb = std::unique_ptr<callback>(new callback(func));
 		cb->order = order;
@@ -341,6 +346,11 @@ private:
 public:
 	void Disconnect(size_t cookie)
 	{
+		if (cookie == -1)
+		{
+			return;
+		}
+
 		callback* prev = nullptr;
 
 		for (auto cb = m_callbacks.get(); cb; cb = cb->next.get())
@@ -383,7 +393,7 @@ public:
 
 		for (auto cb = m_callbacks.get(); cb; cb = cb->next.get())
 		{
-			if (cb->function && !std::invoke(cb->function, args...))
+			if (!std::invoke(cb->function, args...))
 			{
 				return false;
 			}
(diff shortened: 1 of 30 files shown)


blattersturm:
Detailed profiler still works:

image

Agnew0:
A friend and I noticed this morning that our resmons were being a bit finnicky, have just come to the realization that it shows resource CPU Msec's incorrectly until you turn resmon off and then back on again.

Both of the following images are taken after joining my dev server.

### Resmon displaying incorrectly after loading in, with resmon already enabled:
img
### Resmon displaying correctly after disabling and reenabling:
 img

PR #1179, opened 2021-12-06
Social Club instructional button

freedy69:
Added a SCBUTTON_OFF context for the IB_SOCIALCLUB_UI button, making it possible for scripts to remove it with PauseMenuActivateContext. Currently, you'd have to disable x64 context, which works, but that method also removes other menu tabs.

Example: 
```
    -- before: https://i.imgur.com/IMJMnnb.png
    PauseMenuActivateContext(GetHashKey("SCBUTTON_OFF"))
    N_0x4895bdea16e7c080() -- refresh buttons
    -- after: https://i.imgur.com/oVH75yl.png```


blattersturm:
The _what_ button? Where? What's the use case here? Why would 'scripts' need to remove a button from somewhere?

freedy69:
> 

the social club button in the main menu, it doesn't really have any use in fivem

blattersturm:
Then... why would you need 'scripts' to be able to 'remove it'? Why not just remove it outright?

freedy69:
ahh you're right actually my bad, just thought there could be a scripted use for it

PR #1182, opened 2021-12-09
feat(scripting/lua): implement SetInterval

thelindat:
Pulled the last request down due to implementation of tickless resources. Still just a convenient wrapper as it was, and maybe the type checking is excessive. Should be 1:1 with the v8 setInterval function, though we can send the id back as an argument to alter the yield time.

I used `CreateThreadNow` so we could immediately return the runtime reference id instead of counting up in Lua.

```lua
local t = Citizen.SetInterval(function(...)
	GetEntityCoords(...)
end, 0, PlayerPedId())

Citizen.SetInterval(t, 500)

Citizen.ClearInterval(t)
```

### DIFF
--- a/code/components/citizen-scripting-lua/src/LuaScriptRuntime.cpp
+++ b/code/components/citizen-scripting-lua/src/LuaScriptRuntime.cpp
@@ -1081,8 +1081,18 @@ static int Lua_CreateThreadNow(lua_State* L)
 
 static int Lua_SetTimeout(lua_State* L)
 {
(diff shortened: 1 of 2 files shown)


blattersturm:
Should be safe to merge, at least as long as we don't have any new tick API.

To match `SetTimeout` in Lua, however, the timeout should be moved to the first argument.

thelindat:
> both should dynamically take an argument in _either_ order.

I would probably prefer that approach; I'll take a look at SetTimeout and make sure I don't break the game.

PR #762, opened 2021-07-11
Updated to correct indexes for the seat indexes..

Cmoen11:

PR #1190, opened 2021-12-14
Update GetPedInVehicleSeat.md

matthias-codes:
Just wanting to update this Native Docs as the old vehicle seat index numbers are wrong. Could be confusing for newbies or just people who don't know about those numbers.


SQKo:
Dupes #762

PR #1194, opened 2021-12-19
fix(docs): update vs2019 to vs2022

AvarianKnight:
This comes from a conversation with someone in another discord having issues trying to build the files because they were using vs2019 instead of vs2022

### DIFF
--- a/docs/building.md
+++ b/docs/building.md
@@ -4,7 +4,7 @@
 
 To build FiveM, RedM or FXServer on Windows you need the following dependencies:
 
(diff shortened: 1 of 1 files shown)


PichotM:
Btw you can also add that you cannot build FiveM using python 3.10 (only 3.8/3.9)

blattersturm:
> Btw you can also add that you cannot build FiveM using python 3.10 (only 3.8/3.9)

that's not even intended so why would he 'add' that?

blattersturm:
> having issues trying to build the files because they were using vs2019 instead of vs2022

that's also not intended, shit should build fine on 2019 still. 'issues'?

PR #1196, opened 2021-12-22
Indicate invalid resource categories for start/stop/ensure

### LINKED ISSUE #1159: `ensure [category]` doesn't explicitly indicate when there's no resource in this category
When trying to run `ensure [aaaaaaaaaaaa]`, nothing happens. This may be confusing.

ekralc:
start/stop/ensure accepts non-existent resource category names without a response.

To fix this, the category is now retrieved using `find`. If it doesn't exist, a message is returned.

Closes #1159

### DIFF
--- a/code/components/citizen-server-impl/src/ServerResources.cpp
+++ b/code/components/citizen-server-impl/src/ServerResources.cpp
@@ -814,7 +814,15 @@ static InitFunction initFunction([]()
 
 			if (resourceName[0] == '[' && resourceName[resourceName.size() - 1] == ']')
 			{
-				for (const auto& resource : g_resourcesByComponent[resourceName])
+				auto category = g_resourcesByComponent.find(resourceName);
+
+				if (category == g_resourcesByComponent.end())
+				{
+					trace("^3Couldn't find resource category %s.^7\n", resourceName);
+					return;
+				}
+
+				for (const auto& resource : category->second)
 				{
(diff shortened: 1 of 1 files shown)


Fortahr:
Looks good. Also, did you forget a return after line 863?

ekralc:
Yes you're right, I'll fix that right away

code/components/citizen-server-impl/src/ServerResources.cpp
```diff
@@ -814,7 +814,14 @@ static InitFunction initFunction([]()
 
 			if (resourceName[0] == '[' && resourceName[resourceName.size() - 1] == ']')
 			{
-				for (const auto& resource : g_resourcesByComponent[resourceName])
+				auto category = g_resourcesByComponent.find(resourceName);
+
+				if (category == g_resourcesByComponent.end()) {
```
blattersturm, line 819: Please, keep the indentation settings (that _should_ be defined in the .clang-format or .editorconfig - why is your editor ignoring these?) in mind. `{` goes on a new line.

blattersturm (changes requested):

code/components/citizen-server-impl/src/ServerResources.cpp
ekralc: Apologies, this should be fixed now.

ekralc:
I force pushed again because the commit message was wrong, it was from another commit that I've since squashed.

PR #1198, opened 2021-12-23
tweak(ext/natives): link seatIndex to eSeatPosition

borianm:
The previous description in the seat indexes part was incorrect.


blattersturm:
... _another one of these?_

This one might finally _not_ be weird however, unlike the last three PRs.

borianm:
I just stumbled on this native, followed the documentation and didn't get the expected results (-1 vs 1 for drivers seat).
If you think it's not useful, I'll close the PR.

PR #1202, opened 2021-12-26
tweak(scripting/lua): Citizen.Await promise handling

thelindat:
I was having a look at Citizen.Await / promises and noticed some weird hack, despite promises exposing state and value (maybe based on some old code?). I didn't experience any unexpected behaviour with this change to the implementation.

Change to main.js returns the actual error message, rather than an empty object in the Lua runtime.
image
image

### DIFF
--- a/data/shared/citizen/scripting/lua/scheduler.lua
+++ b/data/shared/citizen/scripting/lua/scheduler.lua
@@ -115,33 +115,19 @@ Citizen.AwaitSentinel = nil
 
 function Citizen.Await(promise)
 	local coro = coroutine_running()
-	if not coro then
-		error("Current execution context is not in the scheduler, you should use CreateThread / SetTimeout or Event system (AddEventHandler) to be able to Await")
-	end
+	assert(coro, "Current execution context is not in the scheduler, you should use CreateThread / SetTimeout or Event system (AddEventHandler) to be able to Await")
 
-	-- Indicates if the promise has already been resolved or rejected
-	-- This is a hack since the API does not expose its state
-	local isDone = false
-	local result, err
-	promise = promise:next(function(...)
-		isDone = true
-		result = {...}
-	end,function(error)
-		isDone = true
-		err = error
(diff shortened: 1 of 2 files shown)


data/shared/citizen/scripting/lua/scheduler.lua
```diff
@@ -849,10 +835,10 @@ local function lazyEventHandler() -- lazy initializer so we don't add an event w
 end
 
 -- Handle an export with multiple return values.
-local function exportProcessResult(resource, k, status, ...)
+local function exportProcessResult(resource, exportName, status, ...)
 	if not status then
 		local result = tostring(select(1, ...))
-		error('An error occurred while calling export ' .. k .. ' of resource ' .. resource .. ' (' .. result .. '), see above for details')
+		error(('An error occured while calling export %s of resource %s\n%s'):format(exportName, resource, result))
```
blattersturm, line 841: How'd you end up introducing a typo?

blattersturm (changes requested):
not gonna merge a typo :D

Odd how the person who PR'd this long ago (#75) did this when apparently the library they imported _did_ make state accessible.

data/shared/citizen/scripting/lua/scheduler.lua
thelindat: 😊 general incompetence

blattersturm:
i don't think our commit categorization likes `refactor` much. we need a `internal` category or so someday but for now `tweak`

thelindat:
> i don't think our commit categorization likes `refactor` much. we need a `internal` category or so someday but for now `tweak`

Amended to tweak. Is there somewhere I can view the acceptable categories?

blattersturm:
> Is there somewhere I can view the acceptable categories?

No, I don't think so. This may be in the changelog code but I'm not sure if this is on the client or server side.

blattersturm (approved):

PR #1206, opened 2021-12-27
Add missing strings in the Polish language

Robi321:
As title, i added missing string in Polish translation.


blattersturm:
We don't handle translation updates via individual GitHub PRs, sorry.

aiden-z-source:
Adding to the above, I believe https://translations.cfx.re/projects/citizenfx/cfx-ui/ is the correct place, however, correct me if I am wrong.

PR #1207, opened 2021-12-30
feat: add trigger event native for multiple targets

PichotM:
The goal of this native is to be able to send an event to a specific list of players ensuring that only the relevant players will get it, without having to serialize the payload at every call or sending it to all players.

Example (lua):
```lua
local targets = { "1", "3" }
TriggerClientsEvent("cancelTrade", targets, { example = 0 })
```
If you have a better name to suggest for this native share it :)

### DIFF
--- a/code/components/citizen-server-impl/include/ServerEventComponent.h
+++ b/code/components/citizen-server-impl/include/ServerEventComponent.h
@@ -16,6 +16,9 @@ namespace fx
 	public:
 		virtual void TriggerClientEvent(const std::string_view& eventName, const void* data, size_t dataLen, const std::optional<std::string_view>& targetSrc = std::optional<std::string_view>());
 
+		virtual void TriggerClientsEvent(const std::string_view& eventName, const void* data, size_t dataLen, const std::vector<std::string>& targets = std::vector<std::string>());
+
+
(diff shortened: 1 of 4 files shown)


GameMango36:
A better name could be something like `TriggerMultiClientEvent`? or maybe just accepting tables as the source for TriggerClientEvent would be a cleaner implementation.

freedy69:
Great idea, but as GameMango36 said TriggerMultiClientEvent would be a better name, hope this gets merged

blattersturm:
Not sure why this has to be a dedicated 'weird' native with another weird implementation.

Similarly, why is this needed _at all?_ Have you actually found this to be a performance bottleneck? And, if so, why aren't you sharing that data, since I guess that substantiates this specific implementation and not e.g. just single serialization to the same native multiple times..?

The lack of cross-ScRT support is also worrying. I won't accept this without any more information as to why and how you ended up with this specific implementation.

blattersturm:
Another implementation idea: use the string argument to the existing TRIGGER_CLIENT_EVENT_INTERNAL to encode multiple client IDs - it should be a string already so e.g. a separator could be a perfectly valid option here.

Parsing the string should be done in the native itself, and then the internal call and overloads could use a target list.

PR #1208, opened 2021-12-30
Update locale-fi.json

oSeneX:
Fixed and added some translations.


titanium-cfx:
Please use https://translations.cfx.re :)

PR #1210, opened 2022-01-01
feat(server): support target list for trigger client event

PichotM:
Old PR: https://github.com/citizenfx/fivem/pull/1207


marciogalvao89:
awesome! need this! 👍 👍 👍 👍 👍

RockySouthpaw:
> awesome! need this! 👍 👍 👍 👍 👍

What even is the need for this?

Fortahr:
> What even is the need for this?

Looks like this makes `TriggerClientEvent` also allow an array of player IDs to be passed and send the event to all those players, instead of just one.

blattersturm:
> > What even is the need for this?
> 
> Looks like this makes `TriggerClientEvent` also allow an array of player IDs to be passed and send the event to all those players, instead of just one.

wow, it does????? i need this right now 👍👍 👍 👍 👍 

---

Not merged yet by the way as there's a few small changes I want to do to it myself for, uh, sanity reasons. They were in my comment to the last one too.

manups4e:
what if i set -1 to one of the array positions?

Xinerki:
```lua
for i,target in pairs(targets) do
	TriggerClientEvent("cancelTrade", target, { example = 0 })
end
```
khaby

blattersturm:
This PR is in my manual edit queue, it needs to have some changes made to prevent API misuse.

PichotM:
Any update on this PR?
Also, can you tell me what you wanted to do so I might help if you're busy

Korioz:
Up,

Sending a client event to a huge list of players (400+) is not viable without the implementation of this PR, the consecutive native calls leads to bad performance (like packing the same data over and over again), for example this would be useful to send to all players in a bucket, which can have a lot of players inside it.

manups4e:
Can confirm even with FxEvents and it's implementation would benefit from this

thelindat:
I'm not sure what changes were needed to prevent misuse, but this, or something similar, is badly needed.

Calling `TriggerClientEvent` just once instead of for a dozen or so targets provides a non-negligible performance boost - not just because of the reduced native calls but thanks to all the arguments only being msgpacked once.

Some specific use-cases:
- sending events to all players in a "group"
- sending events to all players in a bucket (#1656)
- sending events to all players in range of some position

PichotM:
The new commit includes:
- A new function `TriggerMulticastClientEvent` in lua/js/c# instead of relying on `TriggerClientEvent` parameters.
We still use the same logic for the target list as the initial PR.

outsider31000:
agreed, should only accept arrays, other wise might as well implement in the TriggerClientEvent

code/client/clrcore/BaseScript.cs
IllInuz, line 1: Shouldn't this contain nullchecks, and validate the eventName to prevent empty strings?

Player[] to IEnumerable<Player> for more flexbility etc?

I have made some changes, feel free to use/not use.


```
public static void TriggerMulticastClientEvent(IEnumerable<Player> players, string eventName, params object[] args)
{
    if (players == null)
        throw new ArgumentNullException(nameof(players));
    
    if (string.IsNullOrEmpty(eventName))
        throw new ArgumentNullException(nameof(eventName));

    try
    {
        var argsSerialized = MsgPackSerializer.Serialize(args);
        var targetNetIds = string.Join(" ", players.Select(p => p?.Handle ?? throw new ArgumentException("Player cannot be null", nameof(players))));

        unsafe
        {
            fixed (byte* serialized = argsSerialized.Length > 0 ? &argsSerialized[0] : null)
            {
                Function.Call(Hash.TRIGGER_CLIENT_EVENT_INTERNAL, 
                    eventName,
                    targetNetIds,
                    serialized,
                    argsSerialized.Length);
            }
        }
    }
    catch (Exception ex)
    {
        throw new InvalidOperationException($"Failed to trigger multicast event '{eventName}'", ex);
    }
}```

n3xuuu:
any news on this?

PichotM:
> any news on this?

The PR is not finished

n3xuuu:
> The PR is not finished

And when do you expect this pull request to be finished? It's been a while since it was made, and there's been 0 visible progress.

Whisperisgod:
Will there be any progress on this matter?

Z3rio:
I would love to take a shot at helping out with this based on the feedback, if thats fine.
Or do you think you're going to continue working on this?

PR #1195, opened 2021-12-20
feat(streaming-five): Add support for other Game Events (Emitting and Reacting)

Fortahr:
This update will re-enable the hooking for the missing game events and retrieves all entities associated with it.

I've created separate scripting events that are being called, this, to allow fitting arguments to be passed to scripts and not to break existing listeners. They are called "gameEventEmit" and "gameEventReact" respectively as there are 3 game event versions; 1 that's already handled with "gameEventTriggered", 1 that sends an event to nearby entities (like when an event just occured), and the last 1 by who entities react to previous events (e.g.: CEventShockingDeadBody).

Events like CEventShocking*, CEventPotentialWalkIntoVehicle, and CEventPotentialGetRunOver already retrieve the world position.
To support more events' data retrieval we would need to analyze the event data and setup proper conversions.

Adding include/dependency : "vendor:msgpack-c" to 2 projects: gta-streaming-five and handling-loader-five. To allow this new event parser to send more types to scripts (e.g.: uint32 and float);

PR #1214, opened 2022-01-06
feat(streaming-five): Add support for other Game Events (fixed)

Fortahr:
This update will re-enable the hooking for the missing game events and retrieves all entities associated with it.

- Better design that follows the game's event system, which fixes the crashing of previous version https://github.com/citizenfx/fivem/pull/1195.
- All events will now get the associated entity, even those missing in previous version.
- Previous Emit/React versions merged, event names are now the same as the game event name itself, e.g.: "CEventShockingSeenNiceCar".
- Right now only CEventShocking* reads the position of the event and puts it in the args parameter.

Example usage (Lua):
```lua
AddEventHandler("CEventShockingGunshotFired", function(entities, eventEntity, args)
	print("CEventShockingGunshotFired", json.encode(entities), eventEntity, json.encode(args));
end);
```

### DIFF
--- a/code/components/citizen-resources-gta/src/GameEventFunctions.cpp
+++ b/code/components/citizen-resources-gta/src/GameEventFunctions.cpp
@@ -9,6 +9,27 @@
 
 static InitFunction initFunction([]
 {
+	OnTriggerGameEventExt.Connect([](const GameEventData& data)
+	{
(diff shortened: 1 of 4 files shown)


Fortahr:
Also: tested on 1604, 2060, 2189, 2372, and 2545.

PR #1225, opened 2022-01-13
feat(scripting/lua): Add Citizen.RegisterErrorHandler

Jarrrk:
Allows a resource to register a callback to receive stack traces, useful for sentry implementations

```lua
Citizen.RegisterErrorHandler(function(data)
	print('error', data.err)

	print('formatted stack trace', data.fst)
end)
```

### DIFF
--- a/data/shared/citizen/scripting/lua/scheduler.lua
+++ b/data/shared/citizen/scripting/lua/scheduler.lua
@@ -64,6 +64,8 @@ json.setoption('with_hole', true)
 -- temp
 local _in = Citizen.InvokeNative
 
+local errorHandlers = {}
+
 local function FormatStackTrace()
 	return _in(`FORMAT_STACK_TRACE` & 0xFFFFFFFF, nil, 0, Citizen.ResultAsString())
 end
@@ -128,6 +130,18 @@ function Citizen.Await(promise)
 	end
 
 	return promise.value
+end
+
+function Citizen.RegisterErrorHandler(callback)
+	local invokingResource = GetCurrentResourceName()
+
+	-- check if a table exists for the calling resource to insert a callback
+	if errorHandlers[invokingResource] == nil then
+		errorHandlers[invokingResource] = {}
+	end
+
+	-- insert the callback
+	table.insert(errorHandlers[invokingResource], callback)
 end
 
 Citizen.SetBoundaryRoutine(function(f)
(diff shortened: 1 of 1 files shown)


blattersturm:
nah

blattersturm:
This has been discussed quite a few times and each time the answer was relatively close to 'not going to happen', forgot why but could search it up if need be.

blattersturm:
Found at least one past discussion: https://matrix.to/#/!rUiwxlJvpdAOiTowFn%3Acfx.re/%24LzUswGkPdmhlkzClwgVTWB2bjjVyqqf55AmbPbvBxIg

blattersturm:
That discussion also surfaces a few concerns:

1. cross ScRT support. this PR doesn't handle any of that. 
2. this PR just passes a string including for errors caught later, it doesn't help at all with raw stack data or splitting error and text, requiring the consuming caller to handle parsing themselves and requiring us to keep the text format stable -> no go
3. no consideration for e.g. 'cheat detection' scripts abusing this functionality against the intended use, as it's a very lazy uninformed implementation

Jarrrk:
Seems oddly illogical to not allow developers to track errors they otherwise will (for the most part) never have access to at client level, there were 'workarounds' of instancing `Citizen.Trace` and what not to replicate this pr but since have been squashed, for good reason.

> This implementation also has some silly decisions

As per the linked discussion and your first comment, it's relatively irrelevant no? This behavior is frowned upon given it's intended link to Sentry, what's the alternative? I know it's a niche developer profile in the cfx.re community that a server developer cares if their players get errors/issues, but some of us do and want to create a more fluid player experience, we can't do that if we're missing core information to build upon this.

> this PR clearly doesn't meet the requirements for this kind of feature.

Given the amount of dormant pr's, I didn't expect a response in a few months let alone 27 seconds, so, mind my 5 minute pr to guage interest.

> Even discounting the need for or desirability of such a feature

Sure, because only a handful of people would want to use it, care for it or even know why they'd want it.

blattersturm:
> Seems oddly illogical to not allow developers to track errors

Developers are 'allowed to track errors' fine. This is an odd extrapolation from a specific PR with a specific implementation being rejected.

Similarly, I will not reply to the jumping-to-conclusions in your second paragraph, since this was _also_ not implied anywhere.

> Given the amount of dormant pr's, I didn't expect a response in a few months let alone 27 seconds

'Dormant PRs' are usually backlogged for a variety of reasons, nor is anyone complaining about any 'expected response'. Another weird extrapolation.

That, and GitHub doesn't have a way to mass-close PRs with a message, and last time we did so _without_ a message it led to numerous angry replies just like yours.

Skipping the last paragraph of jumping to conclusions as well, one final **important** part I want to add:

> so, mind my 5 minute pr to guage interest.

A PR Ipull request, i.e. a _request_ to _pull_ your branch into `master`), especially if not marked as draft, is not a means to 'gauge interest'. A draft PR would be, but that's rather a 'review my code to see if it makes sense', but yours also wasn't, nor did you mention it was an unfinished 'gauging interest' patch in the description, so pardon us for not reading your mind.

Acceptable means of gauging interest, for example, would be, in order of preference:

1. A chat discussion in any of the usual places (Discord, 'releng' Discord, or the Matrix chat).
2. A forum topic, e.g. in 'Platform Suggestions'.
3. An issue, but ideally only once there's an actionable work item, as the issues here are only meant for anything that is actionable.
4. A draft PR, with a bit more descriptive text than you have provided here.

This PR, for all intents and purposes, looked like a code dump of some unfinished code, and these _never_ have winners, as evidenced by the fact that we are right here, discussing this, with a lot of blame-placing and odd conclusions implied in response.

PR #1231, opened 2022-01-14
feat(onesync): more ped game state research

MRV6:
Adds 3 natives : 
- `IS_FLASH_LIGHT_ON`
- `IS_PED_USING_ACTION_MODE`
- `IS_PED_HANDCUFFED`


gottfriedleibniz:
Unsure how much you care about correctness, but `bool bool7 = state.buffer.ReadBit();` is actually an object ID. It just happens to correspond to the mount netobj, so it will never be parsed.

PR #1234, opened 2022-01-16
fix(rage-device/five): Error out on cloud files

4mmonium:
**The issue:**
The application seems to get stuck in a loop when cloud files are
accessed, this is caused by users installing their GTA on the cloud,
causing the game to neither load, nor 'gracefully' exit.

**The solution:**
To correct this, we simply get the file attributes and check if
the following flag is set: `FILE_ATTRIBUTE_RECALL_ON_DATA_ACCESS`. If
it's set, we return an invalid handle, so it's passed down the stack.
Once this errors, `FatalError` under `OpenArchiveWrap` will be
triggered. I have modified the message so any affected clients are
notified in the event this occurs.


jakub-cfx (changes requested):
Left a small suggestion + a question about skipping hook usage, looking good otherwise 👍

jakub-cfx (approved):
Looks reasonable to me, PR description matches the changes perfectly.

PR #1235, opened 2022-01-16
feat(onesync): add requestControlEvent + doc

Korioz:
Hi 👋,

Tested on 2545.

Use case: cancel the event to prevent abusers from manually taking control of an entity.


blattersturm:
nah

blattersturm:
Also, _please_, stop the pull requests without asking.

PR #1236, opened 2022-01-19
feat(fxdk): Add Typescript resource template

CodeWithBryan:
Typescript template that uses ESBuild and clean builds using `.fxdkignore`

Updated the README.md to current build steps

Removed unnecessary imports from existing templates

### DIFF
--- a/ext/sdk/resources/sdk-root/shell/src/resource-templates/lua/descriptor.tsx
+++ b/ext/sdk/resources/sdk-root/shell/src/resource-templates/lua/descriptor.tsx
@@ -1,4 +1,3 @@
-import * as React from 'react';
 import { SiLua } from "react-icons/si";
 import { ResourceTemplateDescriptor } from '../types';
 
--- a/ext/sdk/resources/sdk-root/shell/src/resource-templates/ts/scaffolder.ts
+++ b/ext/sdk/resources/sdk-root/shell/src/resource-templates/ts/scaffolder.ts
@@ -1,0 +1,183 @@
+import { inject, injectable } from "inversify";
+import { FsService } from "backend/fs/fs-service";
+import { concurrently } from "utils/concurrently";
+import * as cp from "child_process";
+import { ResourceTemplateScaffolder, ResourceTemplateScaffolderArgs } from "../types";
+
+@injectable()
+export default class TsScaffolder implements ResourceTemplateScaffolder {
+  @inject(FsService)
+  protected readonly fsService: FsService;
+
+  async scaffold({ manifest, resourceName, resourcePath }: ResourceTemplateScaffolderArgs) {
+    manifest.clientScripts.push('dist/client.ts');
+    manifest.serverScripts.push('dist/server.ts');
+
+    manifest.fxdkWatchCommands.push([
+      "yarn",
+      [ "watch" ]
+    ]);
+    manifest.fxdkBuildCommands.push([
+      "yarn",
+      [ "build" ]
+    ]);
+
+    const promises: Promise<any>[] = [];
+    const relativePath = (to: string) => this.fsService.joinPath(resourcePath, to);
+
+    // Client Script
+    promises.push(this.fsService.mkdirp(relativePath('client'))
+      .then(() => concurrently(
+          this.fsService.writeFile(relativePath('client/client.ts'), `console.log("[${resourceName}] Client Resource Started");`),
+          this.fsService.writeFile(relativePath('client/tsconfig.json'), getClientTSConfig()),
+        )));
+
+    // Server Script
+    promises.push(this.fsService.mkdirp(relativePath('server'))
+      .then(() => concurrently(
+          this.fsService.writeFile(relativePath('server/server.ts'), `console.log("[${resourceName}] Server Resource Started");`),
+          this.fsService.writeFile(relativePath('server/tsconfig.json'), getServerTSConfig()),
+        )));
+
+    // Build Tools
+    promises.push(this.fsService.writeFile(relativePath('package.json'), getPackageJson(resourceName)));
+    promises.push(this.fsService.writeFile(relativePath('build.js'), getBuildJS()));
+    promises.push(this.fsService.writeFile(relativePath('.fxdkignore'), getIgnore()));
+
+    await Promise.all(promises);
+
+    await this.installModules(resourcePath);
+  }
+
+  private async installModules(path: string) {
+    return new Promise<void>((resolve, reject) => {
+      cp.exec("yarn", { cwd: path, windowsHide: true }, (err) => {
+        if (err) {
+          return reject(err);
+        }
+
+        return resolve();
+      });
+    });
+  }
+}
+
+function getIgnore(): string {
+  return `client/
+server/
+node_modules/
+package.json
+build.js
+`;
+}
+
+function getClientTSConfig(): string {
+  return `{
+  "compilerOptions": {
+    "baseUrl": ".",
+    "noImplicitAny": true,
+    "module": "commonjs",
+    "target": "es2021",
+    "lib": [
+      "es2021"
+    ],
+    "types": [
+      "@citizenfx/client"
+    ],
+    "moduleResolution": "node",
+    "resolveJsonModule": true,
+    "esModuleInterop": true,
+    "noEmit": true,
+  },
+  "include": [
+    "./",
+    "../common"
+  ]
+}`;
+}
+
+function getServerTSConfig(): string {
+  return `{
+  "compilerOptions": {
+    "baseUrl": ".",
+    "noImplicitAny": true,
+    "module": "commonjs",
+    "target": "es2021",
+    "lib": [
+      "es2021"
+    ],
+    "types": [
+      "@citizenfx/server",
+      "@types/node"
+    ],
+    "moduleResolution": "node",
+    "resolveJsonModule": true,
+    "esModuleInterop": true,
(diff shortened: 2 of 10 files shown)


TheCoolestPaul:
The more support for TypeScript I see the happier I get. I appreciate the work @CodeWithBryan 👍I hope this gets merged right quick.

ext/sdk/resources/sdk-root/shell/src/resource-templates/ts/scaffolder.ts
```diff
@@ -0,0 +1,171 @@
...
+  "name": "${resourceName}",
+  "version": "1.0.0",
+  "main": "index.js",
+  "author": "you",
+  "license": "MIT",
+  "scripts": {
+    "build": "cross-env MODE=production node build.js",
+    "watch": "cross-env MODE=development node build.js"
+  },
+  "dependencies": {},
```
itschip, line 134: To be removed?

ext/sdk/resources/sdk-root/shell/src/resource-templates/ts/scaffolder.ts
```diff
@@ -0,0 +1,171 @@
...
+    "@citizenfx/client": "latest",
+    "@citizenfx/server": "latest",
+    "@types/node": "^16.3.1",
+    "cross-env": "7.0.3",
+    "esbuild": "^0.14.10"
+  }
+}`;
+}
+
+function getBuildJS(): string {
```
itschip, line 145: Are you sure this will work properly for both client/server? I would've assumed we needed to define a platform for the server-side scripts? And can be done only with the CLI perhaps? 

Like `esbuild serverr/server/js --bundle --platform=node --target=someTarget --outfile=dist/server.js`

We also might not want to minify server-side, or what do you think?

CodeWithBryan: Realistically it would be up to the end user if they wanna minify their server, I could see the justification for not minifying it by default.

As far as defining the platform as node, I've never had issues using this setup. Are there specific use cases that you think won't be covered?

ext/sdk/resources/sdk-root/shell/src/resource-templates/ts/scaffolder.ts
```diff
@@ -0,0 +1,181 @@
...
+  "license": "MIT",
+  "scripts": {
+    "build": "cross-env MODE=production node build.js",
+    "watch": "cross-env MODE=development node build.js"
+  },
+  "devDependencies": {
+    "@citizenfx/client": "latest",
+    "@citizenfx/server": "latest",
+    "@types/node": "^16.3.1",
+    "cross-env": "7.0.3",
```
nihonium-cfx, line 138: Please use CLI option passing instead of adding this dep. For example `node build.js --mode=production`.

ext/sdk/resources/sdk-root/shell/src/resource-templates/ts/scaffolder.ts
```diff
@@ -0,0 +1,181 @@
...
+
+const server = {
+  platform: 'node',
+  target: ['node16'],
+  format: 'cjs',
+};
+
+const client = {
+  platform: 'browser',
+  target: ['es2018'],
```
nihonium-cfx, line 165: Use `chrome93` instead, as it matches v8 version being used, `9.3`.

ext/sdk/resources/sdk-root/shell/src/resource-templates/ts/scaffolder.ts
```diff
@@ -0,0 +1,181 @@
...
+}`;
+}
+
+function getBuildJS(): string {
+  return `const esbuild = require('esbuild');
+
+const onRebuild = (context) => {
+  return async (err, res) => {
+    if (err) {
+      return console.log(\`[\${context}]: Rebuild failed\`, err);
```
nihonium-cfx, line 150: If this is error, why not `console.error`?

ext/sdk/resources/sdk-root/shell/src/resource-templates/ts/scaffolder.ts
```diff
@@ -0,0 +1,181 @@
...
+  format: 'cjs',
+};
+
+const client = {
+  platform: 'browser',
+  target: ['es2018'],
+  format: 'iife',
+};
+
+for (let context of [ 'client', 'server' ]) {
```
nihonium-cfx, line 169: Nitpick: why `let` and not `const`? 🙃

ext/sdk/resources/sdk-root/shell/src/resource-templates/ts/scaffolder.ts
```diff
@@ -0,0 +1,181 @@
...
+  esbuild.build({
+    bundle: true,
+    entryPoints: [\`\${context}/\${context}.ts\`],
+    outfile: \`dist/\${context}.js\`,
+    minify: process.env.MODE === 'production' ? true : false,
+    watch: process.env.MODE === 'production' ? false : {
+      onRebuild: onRebuild(context),
+    },
+    ...(context === 'client' ? client : server),
+  }).then(() => console.log(\`[\${context}]: Built successfully!\`)).catch(() => process.exit(1))
```
nihonium-cfx, line 179: Nitpick: where's semicolon in the end? Looks weird 🙃

nihonium-cfx (changes requested):
Both `tsconfig.json` files are poorly written, w/o checking for JS features supported by scripting runtimes on client and server and unnecessary options.

Proper `tsconfig.json` files:

**Server**
```json
{
    "compilerOptions": {
        "baseUrl": ".",
        "noImplicitAny": true,
        "module": "commonjs",
        "target": "es2021",
        "lib": [
            "es2021"
        ],
        "types": [
            "@citizenfx/server",
            "@types/node"
        ],
        "moduleResolution": "node",
        "resolveJsonModule": true,
        "esModuleInterop": true,
        "noEmit": true
    },
    "include": [
        "./",
        "../common"
    ]
}
```

**Client**
```json
{
    "compilerOptions": {
        "baseUrl": ".",
        "noImplicitAny": true,
        "module": "commonjs",
        "target": "es2021",
        "lib": [
            "es2021"
        ],
        "types": [
            "@citizenfx/client"
        ],
        "moduleResolution": "node",
        "resolveJsonModule": true,
        "esModuleInterop": true,
        "noEmit": true,
    },
    "include": [
        "./",
        "../common"
    ]
}
```

Also note that `DOM` lib in client `tsconfig.json` was inevitable albeit being wrong due because of `console` interface missing in `@citizenfx/*` types. This has already been addressed and fix is pending push to the latest branch.

**Minification is also a no go for now** as no native source map support, so it will be hard to map errors to source.

ext/sdk/resources/sdk-root/shell/src/resource-templates/ts/scaffolder.ts
nihonium-cfx: While it is much more appreciated to not include "junk" build files in resource itself, there's only two options to achieve parallel build of both server and client parts: extra dep like `concurrently` or programmatically with such file, I prefer the latter.

ext/sdk/resources/sdk-root/shell/src/resource-templates/lua/descriptor.tsx
```diff
@@ -1,4 +1,3 @@
-import * as React from 'react';
```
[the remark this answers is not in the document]
CodeWithBryan: Gotcha, Yeah I was going to do a separate PR, but it seemed almost too small of a change. In the future I'll be sure to keep them seperate

ext/sdk/resources/sdk-root/shell/src/resource-templates/ts/scaffolder.ts
CodeWithBryan: For the reason of not wanting to pull in a process argument library to parse args, I did the simplest implementation of this suggestion via

`node build.js production` and `node build.js development`

and then just checked
`const production = process.argv[2] === 'production' ? true : false;`

CodeWithBryan: I'm still not totally happy with this solution though, and am open to suggestions

nihonium-cfx: Can just reuse this code: https://github.com/citizenfx/fivem/blob/master/ext/sdk-build/utils.js#L54-L63

Usage be like: https://github.com/citizenfx/fivem/blob/master/ext/sdk-build/build.js#L4-L5

nihonium-cfx:
Please don't forget to undo changes to `yarn.lock`!

ext/sdk/resources/sdk-root/shell/src/resource-templates/ts/scaffolder.ts
```diff
@@ -0,0 +1,183 @@
...
+  "author": "you",
+  "license": "MIT",
+  "scripts": {
+    "build": "node build.js production",
+    "watch": "node build.js development"
+  },
+  "devDependencies": {
+    "@citizenfx/client": "latest",
+    "@citizenfx/server": "latest",
+    "@types/node": "^16.3.1",
```
nihonium-cfx, line 139: This is no longer needed.

nihonium-cfx (approved):

PR #1237, opened 2022-01-20
feat(mumble): resolve push-to-talk not working inside menu/map

Ktos93:
usage IS_DISABLED_CONTROL_PRESSED fix problem when we cant talk with open menu/map but still we can disable this control by DISABLE_ALL_CONTROL_ACTIONS etc.

### DIFF
--- a/code/components/gta-net-five/src/MumbleVoice.cpp
+++ b/code/components/gta-net-five/src/MumbleVoice.cpp
@@ -476,9 +476,14 @@ static void Mumble_RunFrame()
 #endif
 
 		constexpr const uint32_t PLAYER_CONTROL = 0;
+#ifdef GTA_FIVE
 		static auto isControlPressed = fx::ScriptEngine::GetNativeHandler(0xF3A21BCD95725A4A);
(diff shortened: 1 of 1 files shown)


ghostrecom:
We need this change to be 100% functional

dkden7e:
Yes! this is really important!

Disquse:
Sadly this solution is very rough and have very annoying side effects. For example, if you press `N` (default push-to-talk control) and keep talking, while trying to open in-game chat (or do anything that would block the controls in general, open map, alt-tab and etc), the state of pressed `N` would be frozen until you press it again. So player would keep talking even when he doesn't want to. That's why FiveM have a bit more complex realization of such a feature. I would prefer to avoid such annoying "side effect", for me the benefits of this PR aren't comparable to the problems that this solution makes.

AvarianKnight:
Would it not make sense to just use `IsDisabledControlPressed` for GTA and RDR?

This would fix an issue for pma-voice where you can't use radio (or simulate PTT) while in the map because every control action gets disabled

outsider31000:
its been 3 years since this been here @Ktos93 ,  just bumping it, assuming is still needed?

PR #1250, opened 2022-01-27
feat(extra-natives/five): add IS_GAME_WINDOW_FOCUSED

skyrossm:
Add a simple function for checking game window focus.

### DIFF
--- a/code/components/extra-natives-five/src/GameWindowNatives.cpp
+++ b/code/components/extra-natives-five/src/GameWindowNatives.cpp
@@ -1,0 +1,20 @@
+#include <StdInc.h>
+#include <windowsx.h>
+#include <ScriptEngine.h>
+#include <CrossBuildRuntime.h>
+
+static InitFunction initFunction([]()
+{
+	fx::ScriptEngine::RegisterNativeHandler("IS_GAME_WINDOW_FOCUSED", [](fx::ScriptContext& context)
+	{
+		bool result = false;
+
+		auto gameWindow = CoreGetGameWindow();
+		if (gameWindow)
+		{
+			result = gameWindow == GetForegroundWindow();
+		}
+
+		context.SetResult<bool>(result);
+	});
+});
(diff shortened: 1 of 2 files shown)


blattersturm:
... use cases? Impact assessment? Are we just supposed to take this and deal with it?

skyrossm:
Personal use case is going to be checking for focus to pause taskbars, while looking for a native similar some people had uses for "Away" statuses, and holding off on cursor manipulation if the game was not focused.
I figured a native like this would be better than trying to hackily work around with cursor position, last input, etc.

blattersturm:
Flexible - sure. However, for example game pad input works fine if the game isn't focused.

MatinDark:
About abuse, I have no idea but it would be handy especially for RP servers we can let other players know others are tab out or not, so they don't expect a response from them

jaymo1011:
> About abuse, I have no idea but it would be handy especially for RP servers we can let players know, others are tab out or not, so they don't expect a response from them

Potential abuse in this very example is RP servers **banning/kicking users for being "tabbed out" for too long** (via staff instruction or automatically) as some sort of AFK prevention or a naive "anti cheat" measure that gets bypassed by said cheats offering to patch this native to always return true. Seems silly? Remember that widely used "protections" are to ban/kick users for pressing the delete key or console key.

If ruining the user experience wasn't eventual outcome of most newly implemented *neat* features, the worry for abuse wouldn't be so significant.

Sampleeeee:
In my opinion, worrying about abuse is silly. Why give up a feature that has multiple use cases just because some servers will kick / ban their players. A player can always find another server to play on.

callum1337:
Bump, Can this get merged soon, this would also help servers see if players are

callum1337:
any update

pachxl:
hello can you add this now

FabianTerhorst (approved):
Tested and works well. Thanks for the contribution.

PR #1254, opened 2022-01-29
Support CCarHandlingData fields for GetVehicleHandlingFloat/SetVehicleHandlingFloat

Frosty-Ice:
I wasn't sure how to deal with cars that didn't have the CCarHandlingData entry, so I just don't allow them to be get/set.


blattersturm:
How about all the _other_ sub handling types? This should be able to be implemented more generically.

PR #1255, opened 2022-01-30
feat(convars): add GET_CONVAR_FLOAT

AvarianKnight:
Currently, there's no way to get a convar as a float

This leads to some really hacking solutions (like [here](https://github.com/AvarianKnight/pma-voice/blob/f908bcc628d0cebc9668f933a200808cd2921e2c/client/main.lua#L6))

### DIFF
--- a/code/components/citizen-resources-client/src/ConsoleScriptFunctions.cpp
+++ b/code/components/citizen-resources-client/src/ConsoleScriptFunctions.cpp
@@ -60,4 +60,25 @@ static InitFunction initFunction([]()
 			context.SetResult(atoi(var->GetValue().c_str()));
 		}
 	});
+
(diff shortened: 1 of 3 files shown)


blattersturm:
I'm slightly confused as to why we apparently have two sets of convar getter implementations that only differ in _slight_ ways.

AvarianKnight:
Doesn't look like client convars was originally a thing (https://github.com/citizenfx/fivem/commit/af718ea4f67f9d7853ca8920a94d5bd2d259619e#diff-41ce82e7751a7ca63df92aaa62f2a4bd92f9b513ef487fb3b36837f6f187974b and https://forum.cfx.re/t/get-convar-and-get-convar-int-client-side/164859)

PR #1280, opened 2022-02-04
tweak(core/rdr3): PoolManagement - add the vtable list extracted form ps4 and all the proper pools found in gameconfig.xml

0x-FADED:
form my tests now the game is able to find more than 500+ pools


itsumiYasuka:
@blattersturm I think this should be proper PR now...... XD

PR #1292, opened 2022-02-10
feat(extra-natives/five): DEFINE_DRY_VOLUME

HeadshotsOps:
This adds a new native that allows specifying a volume where water collision effects do not apply. This functionality is based on R*' way of disabling water collision under the casino.

This native allows volumes to be defined once and they will persist trough the whole game lifetime.

One question would be if it would be better to move the casino dry volume to this new collection of volumes too?

### DIFF
--- a/code/components/extra-natives-five/src/WaterExtraNatives.cpp
+++ b/code/components/extra-natives-five/src/WaterExtraNatives.cpp
@@ -1,0 +1,128 @@
+#include "StdInc.h"
+#include <Hooking.h>
+#include <ScriptEngine.h>
+#include <ICoreGameInit.h>
+
+struct sDepthBound
+{
+	__m128 min, max;
+	bool isRemoved;
+
+	sDepthBound(std::array<float, 4> min, std::array<float, 4> max)
+		: min(_mm_load_ps(min.data())), max(_mm_load_ps(max.data())), isRemoved(false)
+	{
+	}
+
+	bool operator==(const sDepthBound& bound) const
+	{
+		bool isEqualMin = ((_mm_movemask_ps(_mm_cmpeq_ps(this->min, bound.min))) & 7) == 7;
+		bool isEqualMax = ((_mm_movemask_ps(_mm_cmpeq_ps(this->max, bound.max))) & 7) == 7;
+		return isEqualMax && isEqualMin;
+	}
+
+	void MarkAsRemoved()
+	{
+		isRemoved = true;
+		max = _mm_setzero_ps();
+		min = _mm_setzero_ps();
+	}
+};
+
+std::vector<sDepthBound> g_dryAreas = {}; // maybe the r* casino volume should be moved to this vector?
+
+inline bool IsInBounds(__m128* entityCoords, sDepthBound bounds)
+{
+	if (bounds.isRemoved)
+	{
+		return false;
+	}
+	bool isInsideMin = (_mm_movemask_ps(_mm_cmple_ps(*entityCoords, bounds.max)) & 7) == 7; // Check every coordinate component to be less than the max and create a mask per component -1 = true, 0 = false
+	bool isInsideMax = (_mm_movemask_ps(_mm_cmple_ps(bounds.min, *entityCoords)) & 7) == 7; // Convert the result to an integer mask, all 3 bits need to be set aka 7.
+
+	return isInsideMin && isInsideMax;
+}
+
+static hook::cdecl_stub<void*(__m128*, bool, uint64_t)> _TestForUnderWaterVisuals([]()
+{
+	return hook::get_pattern("48 8B C4 48 89 58 08 48 89 68 18 48 89 70 20 57 48 81 EC ? ? ? ? 83 48 10 FF");
+});
+
+static hook::cdecl_stub<void*(void*, __m128*, uint64_t)> _TestForWaterPhysics([]()
+{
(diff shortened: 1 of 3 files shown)


Disquse:
Hey, after looking a bit at your code it seems there's no way to remove those volumes? Maybe you can return array index as a handle so people could remove those volumes in scripts.

It seems there's no resetting code anywhere? So if player leave a server that used this native, he would also have those volumes on any other server that he join later, until he restart his game. You can use `OnShutdownSession` for example to clear volumes.

Also, `DEFINE_` sounds weird to be honest, there's no any single R* native with this prefix, would make more sense to use `CREATE_` or `ADD_` imo.

HeadshotsOps:
Hi, thanks for your review and useful feedback. 
I did indeed not think about resetting between sessions and I have also added a native to remove volumes now.
Also the array index is now returned when creating volumes.
The naming was changed to CREATE_ , it was indeed sounding a bit unusual.
I have implemented removals by marking the volumes as removed and setting their bounds to zero, this to keep previously defined volume handles valid instead of their index being shifted with a full removal from the vector. I'm not sure if this would be the most clean way.

ext/native-decls/RemoveDryVolume.md
```diff
@@ -0,0 +1,15 @@
...
+apiset: client
+---
+## REMOVE_DRY_VOLUME
+
+```c
+void REMOVE_DRY_VOLUME(int handle);
+```
+
+Removes a dry volume from the game session.
+See CREATE_DRY_VOLUME for more info
```
austinsasko, line 12: Should link back to the md file
```suggestion
See [CREATE_DRY_VOLUME.md](CREATE_DRY_VOLUME) for more info
```

blattersturm: no, it shouldn't?!

blattersturm:
@Disquse @HeadshotsOps was there anything else blocking this?

code/components/extra-natives-five/src/WaterExtraNatives.cpp
```diff
@@ -0,0 +1,148 @@
...
+			g_dryAreas.push_back(*newVolume);		
+		}
+	});
+
+	fx::ScriptEngine::RegisterNativeHandler("REMOVE_DRY_VOLUME", [=](fx::ScriptContext& context)
+	{
+		int index = context.GetArgument<int>(0);
+		if (index > g_dryAreas.size() || index < 0)
+		{
+			trace("Attempting to delete an invalid dry volume handle (%d)", index);
```
Disquse, line 137: Please add `\n`.

code/components/extra-natives-five/src/WaterExtraNatives.cpp
```diff
@@ -0,0 +1,148 @@
...
+		{
+			context.SetResult<int>(g_dryAreas.size());
+			g_dryAreas.push_back(*newVolume);		
+		}
+	});
+
+	fx::ScriptEngine::RegisterNativeHandler("REMOVE_DRY_VOLUME", [=](fx::ScriptContext& context)
+	{
+		int index = context.GetArgument<int>(0);
+		if (index > g_dryAreas.size() || index < 0)
```
Disquse, line 135: I believe there should be `index >= g_dryAreas.size()`.
You can remove dry volume indexed as `0` when you actually have no dry volumes created, this will cause client crash.
When you have at least 1 dry volume, if you call `REMOVE_DRY_VOLUME` for next index (i.e. for next volume, when it wasn't created yet) and then call `CREATE_DRY_VOLUME`, you will get crash.

code/components/extra-natives-five/src/WaterExtraNatives.cpp
```diff
@@ -0,0 +1,148 @@
...
+		context.GetArgument<float>(1),
+		context.GetArgument<float>(2),
+		},
+		new float[]{
+		context.GetArgument<float>(3),
+		context.GetArgument<float>(4),
+		context.GetArgument<float>(5),
+		});
+
+		std::vector<sDepthBound>::iterator v = std::find(g_dryAreas.begin(), g_dryAreas.end(), newVolume);
```
Disquse, line 118: It seems AABB values needs to be limited, there's some weird side effects. If you're creating a dry volume with too high values (in my case I was testing -9999 to 9999), and your local ped is swimming, after creation this region some big part of the map would work like there's some water region (only physically). Wonder why "ped is swimming" part is very important for repro'ing this bug.
Here's repro video: https://gfycat.com/PlaintiveFaithfulAffenpinscher

Disquse (changes requested):
Found some issues, please take a look.

HeadshotsOps:
Thanks for the review, I will make the changes asap and look into that weird behavior

Disquse:
Also, please take care of possible memory leaks `new float[3]` needs to be destructed after using. Same as `newVolume` data, since you're storing a copy and not a reference in `g_dryAreas`.

code/components/extra-natives-five/src/WaterExtraNatives.cpp
HeadshotsOps: It seems to be the case that creating a dry volume in an area where your ped is swimming, causes this. I cannot seem to figure out how to force the game to somehow reset the swimming state for now. However, dry volumes are likely created when the client joins since its a pretty static map feature. I don't see any usecase for removing water effects if a player is already in that pool anyways. Unless you think this is problematic enough to stop going into production?

HeadshotsOps:
Added @nikez ' edits #1612  and updated native decl. Anything else that needs to be changed?

PR #1296, opened 2022-02-14
Update SetVehicleAutoRepairDisabled.md

TomStilson:
Added explanation on the effect of this native.

### DIFF
--- a/ext/native-decls/SetVehicleAutoRepairDisabled.md
+++ b/ext/native-decls/SetVehicleAutoRepairDisabled.md
@@ -8,7 +8,7 @@ apiset: client
 void SET_VEHICLE_AUTO_REPAIR_DISABLED(Vehicle vehicle, BOOL value);
 ```
 
-Disables the vehicle from being repaired when a vehicle extra is enabled.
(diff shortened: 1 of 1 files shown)


TheKronoss:
This is not how the native works. When set to true it won't repair the vehicle when you enable extra's. It won't stop you from enabling extra's on damaged vehicles.

4mmonium:
All this native does is prevent vehicle repairs from happening when extras are enabled on a vehicle. There's a collection that takes vehicle entities to skip `g_skipRepairVehicles` and once the repair method comes along it's checked whether a vehicle should be repaired or not [by checking if said pointer is in the collection](https://github.com/citizenfx/fivem/blob/b5c87bf731c721400b336637b0a886f5be726657/code/components/extra-natives-five/src/VehicleExtraNatives.cpp#L1271).

Then it will just jump the repair method if set to [`true`](https://github.com/citizenfx/fivem/blob/b5c87bf731c721400b336637b0a886f5be726657/code/components/extra-natives-five/src/VehicleExtraNatives.cpp#L1295), meaning said vehicle ended up in the `g_skipRepairVehicles` collection.

It doesn't really mean you cannot enable vehicle extras on damaged vehicles per se.

PR #1299, opened 2022-02-17
Update SetMpGamerTagsUseVehicleBehavior.md

TomStilson:
More precise function name.

### DIFF
--- a/ext/native-decls/SetMpGamerTagsUseVehicleBehavior.md
+++ b/ext/native-decls/SetMpGamerTagsUseVehicleBehavior.md
@@ -1,11 +1,12 @@
 ---
 ns: CFX
 apiset: client
+aliases: ["0x6DD05E9D83EFA4C9","_SET_MP_GAMER_TAGS_USE_VEHICLE_BEHAVIOR"]
 ---
-## SET_MP_GAMER_TAGS_USE_VEHICLE_BEHAVIOR
+## SET_MP_GAMER_TAGS_DISABLE_VEHICLE_BEHAVIOR
 
 ```c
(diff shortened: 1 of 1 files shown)


ext/native-decls/SetMpGamerTagsUseVehicleBehavior.md
```diff
@@ -1,14 +1,15 @@
 ---
 ns: CFX
 apiset: client
+aliases: ["0x6DD05E9D83EFA4C9","_SET_MP_GAMER_TAGS_USE_VEHICLE_BEHAVIOR"]
```
4mmonium, line 4: This second alias you added `_SET_MP_GAMER_TAGS_USE_VEHICLE_BEHAVIOR` will never work since it won't match the one registered by the native handler. 

Also not quite sure where you got that first alias from `0x6DD05E9D83EFA4C9`, looks like you got that from `CREATE_MP_GAMER_TAG_WITH_CREW_COLOR`.

4mmonium (changes requested):
Hi, there's a chance this PR may be closed, but I left you some feedback anyway so you know how contributions should be done when creating a Native PR:

- Remove `0x6DD05E9D83EFA4C9` from the alias list.
- Change `_SET_MP_GAMER_TAGS_USE_VEHICLE_BEHAVIOR` to `SET_MP_GAMER_TAGS_USE_VEHICLE_BEHAVIOR` in the alias list.

Ultimately, I'm not really sure why this is really needed since the documentation for this native seems fine as is.

Thank you for your contribution! 😊

PR #1307, opened 2022-02-23
Create codeql-analysis.yml

Curiosamentes:


technetium-cfx:
What purpose has this?

blattersturm:
I'm suspecting GH has some button that drops in this template and makes a PR.

PR #1313, opened 2022-03-02
Fix null reference exception in model.cs

Keyinator:
A bug was causing null reference exception if null was == (or !=) compared to another model.
The issue has been fixed by firstly checking if the left side is null, and if so performing a null check on the right side.

This is a pr related to:
#1312

### DIFF
--- a/code/client/clrcore/External/Model.cs
+++ b/code/client/clrcore/External/Model.cs
@@ -404,11 +404,14 @@ namespace CitizenFX.Core
 
 		public static bool operator ==(Model left, Model right)
 		{
-			return left.Equals(right);
+  			if (left == null)
+    			return right == null;
+
+  			return left.Equals(right);
 		}
 		public static bool operator !=(Model left, Model right)
 		{
-			return !left.Equals(right);
+			return !(left==right);
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


Keyinator:
This has a 1+ year anniversary now xD
(Just stumbled upon this when looking at issues)

Keyinator:
> Assuming that's what you were asking about, this likely wasn't merged as the commit messages were somewhat unsuitable (and it was somehow _two_ commits).
> 
> If not that, I wouldn't know what the comment is meant to imply either, it doesn't really seem actionable.

Sorry for the late reply. This was not meant as a snarky comment but rather a fun-fact with the slight idea of "hey, maybe can take a look at this again".

As for the multiple commits I don't remember either what happened there.
Looking at the current main-branches code the issue should still occur when null as a "Model" is being compared to another value since in both lines 432 and 436 of https://github.com/citizenfx/fivem/blob/master/code/client/clrcore/External/Model.cs#432 left will equal null which does not have the function "Equals".

I don't know how best to handle this issue. I could open a more descriptive pull-request if you'd like me to but feel free to just implement the changes if that works better.

PS: Looking at the code using ``==`` instead of ``ReferenceEquals`` should work and be more readable should you chose the later option of implementing yourself.

blattersturm:
A rebase + force push to the same PR branch should be sufficient to correct the commit messages.

PR #1315, opened 2022-03-07
gameconfig: increase FragmentStore

exodusprime-rp:
This increases FragmentStore from 48000 to 96000 so more vehicle modkits can be streamed.


d22tny:
This should be closed as it has already been raised to 59000 and also added into the increase_pool_size command.

radium-cfx:
Closing as invalid due to the above comment.

PR #1316, opened 2022-03-09
feat(server): parse NETWORK_PLAY_SOUND_EVENT

wdesgardin:
This adds parsing for NETWORK_PLAY_SOUND_EVENT.

Since a few days some cheaters are spamming networked sounds using PlaySoundFromCoord and/or PlaySoundFromEntity natives on my servers.
This would provide the ability to filter wich sounds are allowed base on the sound name hash etc..

Tested on 2372 and 2545

### DIFF
--- a/code/components/citizen-server-impl/src/state/ServerGameState.cpp
+++ b/code/components/citizen-server-impl/src/state/ServerGameState.cpp
@@ -5086,6 +5086,70 @@ struct CNetworkPtFXEvent
 
 	MSGPACK_DEFINE_MAP(effectHash, assetHash, posX, posY, posZ, offsetX, offsetY, offsetZ, rotX, rotY, rotZ, scale, axisBitset, isOnEntity, entityNetId, f109, f92, f110, f105, f106, f107, f111, f100);
 };
+
+/*NETEV playSoundEvent SERVER
+/#*
+ * Triggered when a player starts playing a sound
+ *
+ * @param sender - The ID of the player that triggered the event.
+ * @param data - The event data.
+ #/
+declare function playSoundEvent(sender: number, data: {
+	isOnEntity: boolean, 
+	audioHash: number, 
+	audioRefHash: number,
+	soundId: number, 
+	unk1: boolean, 
+	entityNetId: number, 
+	posX: number, 
+	posY: number, 
+	posZ: number,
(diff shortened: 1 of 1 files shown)


LWSS:
👍  seems to work on 1604 too

```lua
RegisterCommand('sound', function()
    local soundId = GetSoundId()
    local localped = GetPlayerPed(-1)
    local coords = GetEntityCoords(localped, true)
    PlaySoundFromCoord(soundId, "DRYER", coords.x, coords.y, coords.z, "CARWASH_SOUNDS", true, 200, true)
    Wait(5000)
    StopSound(soundId)
    ReleaseSoundId(soundId)
end)
```

donkras:
yeah, can confirm that "sound spamming" is being abused by cheaters

blattersturm:
I don't think we should add more 'script-based' event handlers for stuff where the _only_ valid use case would be to not route these events (unless explicitly enabled).

wdesgardin:
I didn't notive any crashes, but some off this sounds are just really anoying. 

I'll try to submit another PR where theses events are not routed unless they are explicitly enabled with a convar

marviniwurscht:
Heyo why has this been closed? @wdesgardin @blattersturm 
I can confirm that this is needed, since a lot of modders are spamming sounds and doing stuff like ear rap1ng serverwide.
I dont know if there even is another way to detect something like this, apart from adding an event for this...
What happend to the idea "disabled by default, enable it with a convar"?
Or how about adding a convar to deactivate networked sounds in general?

marviniwurscht:
> I don't think we should add more 'script-based' event handlers for stuff where the _only_ valid use case would be to not route these events (unless explicitly enabled).
> 
> Similarly, if this _crashes clients_, this isn't something that'd be fixed by a server event handler people have to add to scripts.

I think you might have misunderstood the issue we are having. 
There are no crashes caused by spamming this Native. The only issue we have is that Players (especially Streamers) get annoyed when a random modder hops on the server and plays an insanely loud "BEEP" sound on every player online.
We would like to be able to "recieve" this event to identify the hackers / sound abusers and then ban them.
So the whole purpose of this event would be to identify cheaters.

blattersturm:
jesus christ the BUMP

PR #1318, opened 2022-03-12
tweak(gameconfig): up TxdStore limit to 105500

AvarianKnight:
This fixes a crash for NVE users who most likely installed every mod provided, alongside having a server with a lot of streamed content

This was tested on 1604 and 2545 and didn't cause any crashes while fixing the Pool crash

### DIFF
--- a/data/client/citizen/common/data/gameconfig.xml
+++ b/data/client/citizen/common/data/gameconfig.xml
@@ -458,7 +458,7 @@
 						</Item>
 						<Item>
 							<PoolName>TxdStore</PoolName>
-							<PoolSize value="85500"/>
+							<PoolSize value="105500"/>
 						</Item>
 						<Item>
 							<PoolName>CNetObjVehicle</PoolName>


blattersturm:
... that's a big bump. @duk-37 did your magical shadow pool thing ever work out for asset stores?

AvarianKnight:
The bump _likely_ doesn't need to be this big but seeing as it's getting hit I thought it was a good idea to add some leeway.

If you want me to reduce this just let me know and I'll lower it to 95500.

donkras:
"NVE users who most likely installed every mod provided" 10/10 couldn't be more realistic

duk-37:
> ... that's a big bump. @duk-37 did your magical shadow pool thing ever work out for asset stores?

Response is a bit late since I've been busy with stuff and missed this mention.
Last attempt was super generic, but fine tuning per-pool may lead to better results. Will give it a shot.

duk-37:
Pool resize hacks are tracked [here](https://github.com/citizenfx/fivem/pull/1419) now.

PR #1320, opened 2022-03-13
Fix NUI resources loading on top of the loading screen

### LINKED ISSUE #940: NUI resources loading get stacked on top of the loading screen
If using `loadscreen_manual_shutdown`, the loading screen will lose keyboard/mouse focus after a NUI resource has loaded and been added to the frame stack.

VPagani:
Fixes #940 

**Test resources:** [resources.zip](https://github.com/citizenfx/fivem/files/8238793/resources.zip)

### DIFF
--- a/ext/ui-build/data/root.html
+++ b/ext/ui-build/data/root.html
@@ -23,6 +23,11 @@ iframe
 	width: 100%;
 	height: 100%;
 }
+	
+iframe[name="loadingScreen"]
+{
+	z-index: 1;
+}
 </style>
 <script type="text/javascript">
 window.citFrames = {};


AvarianKnight:
I don't think this fixed the specified issue, only fixes Nui loading over the loading screen.

VPagani:
Have you tested it, @AvarianKnight ? It worked here for me. I've uploaded test resources with a button interaction on the loadscreen and a dummy resource with a nui that fills the screen visually, so it tests exactly what you're talking about.

VPagani:
> when another resource with Ui loads

In addition to this, I've discovered this bug only happens when the other resource with a NUI has also a client_script, if it doesn't have a client_script it creates the NUI frame before the loadingScreen frame is created which keeps the loadingScreen frame above everything as it should be. Maybe this info helps fixing other part of the code, but I think the change I made is more straightfoward, since it doesn't depend on the frame creation order.

Slypher (commented):
i tested it and it worked for me 👌

blattersturm:
This doesn't take into account cases where someone may be using manual shutdown, and _expecting_ to be able to draw something interactive on top of the loading screen.

VPagani:
@blattersturm Yes, it does. I've tested it here just to make sure. The `focusFrame` function sets the `z-index` of the frame to 99999 thus the focused NUI will be above anything including the loadingScreen frame, so if someone wants to draw interaction above the loadscreen they just need to call `SetNuiFocus(true, true)` which is already the expected way to start an interaction with a NUI.

https://github.com/citizenfx/fivem/blob/599fe9142abb90b2c967fb3a0525f23c3b0013a1/ext/ui-build/data/root.html#L172

blattersturm:
Right, makes sense.

However, I'm still worried that somewhere, some server may be incorrectly relying on the loading screen inherently setting UI focus, but a new page drawing on top -> getting focus implicitly. I'm also not sure how one would even recognize that programmatically, but this seems like a great way to get some angry-user-support-hell to occur at some point if some popular resource relies on that.

VPagani:
Yeah, I understand the pain of dealing with people that rely on bugs and workarounds instead of reporting or trying to help fixing. My intention here was to make the use case of "activate the manual shutdown of the loadscreen to add interaction to it and have an early-game UI that can smoothly transition from the loadscreen" work as expected, because currently you can't do that as the example "Test resources" i've uploaded shows.

Also I've seen some servers that, while loading, renders the hud or character selection NUI above the loadscreen some time before shutting it down, which seems kinda buggy and it is not exactly their fault because by default the loadscreen is expected to be above everything else.

VPagani:
I think there will be no way of fixing this bug if we want to prevent breaking possible servers relying on it. Maybe the only way of knowing this would be sending this change to Canary release and watching for any bug report related to stucked loadscreens on it.

blattersturm:
> because currently you can't do that

You _can_ do such fine, you just have to make the entire page `display: none`, `visibility: hidden` or `opacity: 0` until some user interaction is supposed to occur.

blattersturm:
Sorry for the wall of text above, again. Same for the dismissive reply assuming it 'doesn't take [case] into account'. The original PR description was missing any statement of that and usually we find more PRs that _don't_ do what they're supposed to vs. those that _do_, and there's no way to tell just looking at code if people didn't do a risk analysis/use case statement in their PR description.

VPagani:
> You _can_ do such fine, you just have to make the entire page `display: none`, `visibility: hidden` or `opacity: 0` until some user interaction is supposed to occur (and you'd still run into behavior as described in #940)

Well, the behavior described in #940 is exactly the reason why you cannot do that fine, because if the interaction is interrupted by any nui frame created, you get stuck in the loadscreen not able to continue the interaction flow.

But your reasoning makes sense, I've seen this bad reaction against you when you're just trying to improve the platform before. So, unfortunately, it's better to not mess with this.

> There's probably ways to more explicitly make intended/desired behavior clearer

Do you think that maybe adding support for something like a `loadscreen_keep_focus "yes"` fxmanifest metadata is a good "solution" to #940 or would it make things worse/confusing? This metadata could invoke the `focusFrame` for the `loadingScreen` frame which would set it's `z-index` to 99999 and keep it in focus even after other nui frames are created, kinda similar to what I intended to do with this PR.

> Sorry for the wall of text above
 
No need to be sorry, I like to undestand your POV and experience of managing this huge project as it teaches me a lot.

> The original PR description was missing any statement

Yeah, sorry for that. I'm still learning how to better contribute to other projects.

This PR can be closed then as it will most likely become a compatibility issue for people using workarounds.

blattersturm:
Yes, a metadata field to assign (and keep?) focus, while hacky, may be a viable solution there.

VPagani:
Ok, I'll see if I can do that.

PR #1326, opened 2022-03-16
tweak(data/rdr3): Increase PortalInst pool size

Identity-labs:
Hi,

With modding we are able to go out of actual size.
Tested with few peoples no side effect with this new value.

:)

### DIFF
--- a/data/client_rdr/citizen/common/data/gameconfig.xml
+++ b/data/client_rdr/citizen/common/data/gameconfig.xml
@@ -1485,7 +1485,7 @@
 				</Item>
                 <Item>
 					<PoolName>PortalInst</PoolName>
-					<PoolSize value="300" />
+					<PoolSize value="600" />
 				</Item>
 				<Item>
                   <PoolName>PoseMatcherStore</PoolName>
--- a/data/client_rdr/citizen/common/data/gameconfig.xml
+++ b/data/client_rdr/citizen/common/data/gameconfig.xml
@@ -1051,7 +1051,7 @@
                 </Item>
                 <Item>
                   <PoolName>Dummy Object</PoolName>
-                  <PoolSize value="13500"/>
+                  <PoolSize value="30000"/>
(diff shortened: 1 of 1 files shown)


blattersturm:
Was `PortalInst` this pool of scene graph nodes that is related to some other static sizes, too, or?

Identity-labs:
We have done many tests, we have custom 37 ytyp, 15 custom MLO, with only 30 ytyp active we don't crash, we can disable an MLO in other side of the map to solve this error, ne need to have all portals in same area to crash.
All values are tested with different people.
I don't know what other information is needed ? crash dump ?

Identity-labs:
Previous crash dump need to be analyzed, it's an unknown pool with size limit of 400, not present in gameconfig.xml
But with all of this pool increased this pool crash is not happening.

PortalInst pool crash happen near last ressource sorted by name (start order ?) if we disable a MLO at 5000 meters we don't crash anymore, so i think this pool is global pool, and not rendered portals in scene
Dummy Object is not a crash but console being flooded by max pool size reach and random vanilla MLO miss lot of entities
CLightEntity pool crash happen in st denis by adding light entities in MLO and light entities in area

Those change have been applied on high number of players of our communities without any side effect.

blattersturm:
> it's an unknown pool with size limit of 400, not present in gameconfig.xml

Right, _that's_ the other portal limit that's not likely to be changeable easily.

Identity-labs:
Ok so enable/disable all MLOs in saint denis dynamically with distance of 60 meter solve portal pool issue, so it's a workaround for now.
But we still have Dummy Object pool issue.
I make another PR only with Dummy Object and MapTypesStore ?

Disquse (approved):

Disquse:
I believe `PortalInst` pool is adjustable without tons of hacks. At least the pool allocation code looks ok after quick analysis.

PR #1346, opened 2022-03-25
feat(asi/five): add sh mode

donkras:
This will allow anticheat team to do some "stricter" checks regarding
LoadLibrary based ( and similar ) injectors, cause when sh mode is
disabled, we won't load any "unwanted" files that might interact with
fivem stuff - a while ago duk said this would be good to add, so i did ;p

reopened it to cleanup commit history and other stuff


donkras:
image

donkras:
resolved by https://github.com/citizenfx/fivem/commit/61424642f7fabc7952d589de0de86f5356cac289

PR #1352, opened 2022-03-28
feat(onesync): Added entityOwnerChanged event

Nova-px:
Added entityOwnerChanged event if an entity get a new network owner


blattersturm:
- Weird commit message (the PR title is not the place to put a commit message).
- Not remotely enough info - no use case mention, no anything.
- Weird spelling/grammar in docs.

Yeah, nope.

PR #1354, opened 2022-03-28
tweak(converter/rdr3): Add more bounds type to rdr3->five converter

Identity-labs:
Allow us to edit and edit bounds, but five->rdr3 make game crash, somehow triangles are corrupted when converted back in rsc8


Ktos93:
This works? I tested this and i dont see any benefits

PR #1370, opened 2022-04-04
Block entity control requests

### LINKED ISSUE #963: Disable entity control requests in entity lockdown mode
requestControlEvent is currently not being parsed at all, but this also should _probably_ get restricted in entity lockdown mode by default.

Will need to check if the game uses this anywhere by itself, however.

blattersturm:
This (relatively untested) changeset will block REQUEST_CONTROL_EVENT routing based on a configurable policy.

Supported modes for the `sv_filterRequestControl` variable are as follows:
* **-1**: Default, equivalent to `2` at this time, but will also warn in console so users can fix stuff/be told to disable it.
* **0**: Off. Also disables the routing bucket/entity lockdown-based policy.
* **1**: Blocks control requests to entities controlled by players (currently, occupied vehicles only) that have existed for more than `sv_filterRequestControlSettleTimer` milliseconds (default 30000) - hereafter referred to as 'settled'.
* **2**: Blocks control requests to _all_ entities controlled by players.
* **3**: Blocks control requests to all entities controlled by players, and any 'settled' non-player entities.
* **4**: Does not route REQUEST_CONTROL_EVENT whatsoever.

In addition, any mode but 'off' will have some additional checks as well:
- Control request events can't be routed across routing buckets. This likely wasn't a thing anyway, but just a double-check.
- Control request events will _always_ be blocked if the sender is in 'strict' entity lockdown mode, either by the global mode setting, or their routing bucket being set to such.

Closes #963.

### DIFF
--- a/code/components/citizen-server-impl/include/state/ServerGameState.h
+++ b/code/components/citizen-server-impl/include/state/ServerGameState.h
@@ -653,6 +653,7 @@ struct SyncEntityState
 	// #IFARQ this means lastFramesSent
 	std::array<uint64_t, MAX_CLIENTS> lastFramesPreSent;
 
+	std::chrono::milliseconds createdAt{ 0 };
 	std::chrono::milliseconds lastReceivedAt;
 	std::chrono::milliseconds lastMigratedAt;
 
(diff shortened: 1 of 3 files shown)


blattersturm:
This requires a docs update as well as someone to write a change notification (like the [last one](https://forum.cfx.re/t/change-notification-embedded-mumble-server-maximum-slot-count-configuration/4796074)) that'll be linked from the `rcmitigation` alias.

blattersturm:
Going to merge this with an addition of the default _not being changed yet_. Someone should still write the post, though. :(

PR #1374, opened 2022-04-05
fix(scripting): no-op REGISTER_SCRIPT_WITH_AUDIO

donkras:
According to fivem native docs - "This native does absolutely nothing,
just a nullsub" so we'll just disable it, as it's proven to be useful
for shady "anticheats"


donkras:
nvm im soo dumb - dont do coding stuff when ur tired LMAO

PR #1379, opened 2022-04-06
Fix conceptual error

XenoS-ITA:
There is a conceptual error in the scheduler, Lua doesnt support the `not table[key]` for arrays


blattersturm:
Affirmative. This seems to not have affected anything, but it's indeed silly.

PR #1382, opened 2022-04-07
tweak(server): state bags can now be created by clients and replicated

### LINKED ISSUE #1304: AddStateBagChangeHandler, entity does not exists when event is triggered
Hello,

When using the following state bag handler on, a vehicle

```
AddStateBagChangeHandler('bubble', nil, function(bagName, key, value, reserved, replicated)
    local netId = tonumber(string.match(bagName, 'entity:(%d+)'))
    local vehicle = NetToVeh(netId)
    print(vehicle)
end)
```

I get the following error:
GetNetworkObject: no object by ID 27
0

But the following works fine:
```
AddStateBagChangeHandler('bubble', nil, function(bagName, key, value, reserved, replicated)
    Citizen.Wait(0)
    local netId = tonumber(string.match(bagName, 'entity:(%d+)'))
    local vehicle = NetToVeh(netId)
    print(vehicle)
end)
```

I would expect the entity to exist when the handler is triggered.

Repro:

server:
```
for _, veh in pairs(GetAllVehicles()) do
    Entity(veh).state.bubble = 4
end
```

client:
```
AddStateBagChangeHandler('bubble', nil, function(bagName, key, value, reserved, replicated)
    local netId = tonumber(string.match(bagName, 'entity:(%d+)'))
    local vehicle = NetToVeh(netId)
    print(vehicle)
end)
```

Then leave the onesync zone of the vehicle and enter it again

### LINKED ISSUE #1335: Entity state bag doesn't get replicated when set by client
When I try to set a state bag key on an entity (vehicle) and set `replicated` to `true`, the key won't get replicated to the server or other clients.
I've created a [sample resource](https://github.com/BlackFlash5/StateBagRepro) to replicate this issue:
- get in any vehicle (as driver)
- run command `bagclient` / `bagserver` to verify the key `TestBag` is not set
- run `setbagclient test test` to save a dictionary with kvp `{ test, test }` to replicated key `TestBag`
- key `TestBag` is now set on the client, but not on the server or other clients (verify with `bagclient` / `bagserver`)

```
> bagclient
Bag is empty!
> setbagclient test test
OnSetBagClient | Args: test, test
- StateBagChangeHandler -
BagName: entity:124 | Key: TestBag | Replicated: True
Key: test | Value: test
> bagclient
Key: test | Value: test
> bagserver
OnBagServer | Player: Black Flash
Bag is empty!
```

Server: 5412
Client update branch: Latest (Unstable)

thorium-cfx:
* Any client that owns an entity can now create its replicating state bags on the server.
* Clients will queue their state bag changes and send them after getting an ACK from the server (meaning the server now knows of the entity).
* State bag updates to clients are now being sent after the clone creation packets.
* State bag updates are not being sent to clients that just got a call to remove the associated entities.

resolves #1335 
resolves #1304

### DIFF
--- a/code/components/citizen-resources-core/include/StateBagComponent.h
+++ b/code/components/citizen-resources-core/include/StateBagComponent.h
@@ -78,6 +78,19 @@ public:
 	// Sets the owning peer ID.
 	//
 	virtual void SetOwningPeer(std::optional<int> peer) = 0;
+	
+	//
+	// Enable or disable immediate replication of this state bag's changes.
+	// If enabled it'll send the state bag update immediatly (old behavior).
+	// If disabled then new replicating state bag changes are queued, use SendQueuedUpdates() to send them.
+	// #TODO: potentially remove this once the throttling system is in place
+	//
+	virtual void EnableImmediateReplication(bool enabled) = 0;
+
+	//
+	// Will send all queued replicating state bag updates to all targets, see EnableImmediateReplication(bool).
+	//
+	virtual void SendQueuedUpdates() = 0;
 
 	//
 	// Gets data for a key.
@@ -97,8 +110,9 @@ public:
 
 	//
 	// Should be called when receiving a state bag control packet.
+	// arg: outBagNameName; if given (!= nullptr) and if the state bag wasn't found then this string will contain the bag name, otherwise outBagNameName is unchanged.
 	//
-	virtual void HandlePacket(int source, std::string_view data) = 0;
+	virtual void HandlePacket(int source, std::string_view data, std::string* outBagNameName = nullptr) = 0;
 
 	//
 	// Gets a state bag by an identifier. Returns an empty shared_ptr if not found.
--- a/code/components/citizen-resources-core/src/StateBagComponent.cpp
+++ b/code/components/citizen-resources-core/src/StateBagComponent.cpp
@@ -6,7 +6,7 @@
 #include <shared_mutex>
 #include <unordered_set>
 
-#include <EASTL/fixed_set.h>
+//#include <EASTL/fixed_set.h>
 #include <EASTL/fixed_vector.h>
 
 #include <SharedFunction.h>
@@ -27,7 +27,7 @@ public:
 
 	virtual void Reset() override;
 
-	virtual void HandlePacket(int source, std::string_view data) override;
+	virtual void HandlePacket(int source, std::string_view data, std::string* outBagNameName = nullptr) override;
 
 	virtual std::shared_ptr<StateBag> GetStateBag(std::string_view id) override;
 
@@ -105,9 +105,13 @@ public:
 	std::optional<int> GetOwningPeer();
 	virtual void SetOwningPeer(std::optional<int> peer) override;
 
-	void SendKey(int target, std::string_view key);
+	// #TODO: potentially remove this once the throttling system is in place
+	virtual void EnableImmediateReplication(bool enabled) override;
+	virtual void SendQueuedUpdates() override;
 
-	void SendKeyAll(std::string_view key);
+	void SendKeyValue(int target, std::string_view key, std::string_view value);
+
+	void SendKeyValueToAllTargets(std::string_view key, std::string_view value);
 
 	void SendAll(int target);
 
@@ -128,10 +132,14 @@ private:
 
 	std::shared_mutex m_dataMutex;
 	std::map<std::string, std::string, std::less<>> m_data;
+
+	std::atomic<bool> m_replicationEnabled; //< #TODO: potentially remove this once the throttling system is in place
+	std::mutex m_replicateDataMutex;
+	std::unordered_map<std::string, std::string> m_replicateData;
 };
 
 StateBagImpl::StateBagImpl(StateBagComponentImpl* parent, std::string_view id)
-	: m_parent(parent), m_id(id)
+	: m_parent(parent), m_id(id), m_replicationEnabled(true)
 {
 	
 }
@@ -189,8 +197,7 @@ void StateBagImpl::SetKey(int source, std::string_view key, std::string_view dat
 			std::string keyStr{ key };
 			std::string dataStr{ data };
 
-			gameInterface->QueueTask(make_shared_function([
-				parent,
+			gameInterface->QueueTask(make_shared_function([parent,
 				source,
 				replicated,
 				id,
@@ -220,27 +227,37 @@ void StateBagImpl::SetKey(int source, std::string_view key, std::string_view dat
 
 void StateBagImpl::SetKeyInternal(int source, std::string_view key, std::string_view data, bool replicated)
 {
-	std::string lastValue;
-
 	{
 		std::unique_lock _(m_dataMutex);
 
-		if (replicated)
+		if (auto it = m_data.find(key); it != m_data.end())
 		{
-			if (auto it = m_data.find(key); it != m_data.end())
+			if (data != it->second)
 			{
-				lastValue = std::move(it->second);
+				it->second = data;
+			}
+			else
+			{
+				replicated = false; // nothing changed
 			}
 		}
-
-		m_data[std::string{ key }] = std::string{ data };
+		else
+		{
+			m_data.emplace(key, data);
+		}
 	}
 
 	if (replicated)
 	{
-		if (data != lastValue)
+		if (m_replicationEnabled)
 		{
-			SendKeyAll(key);
+			SendKeyValueToAllTargets(key, data);
+		}
+		else
+		{
+			// store it for now
+			std::unique_lock lk(m_replicateDataMutex);
+			m_replicateData.emplace(key, data);
 		}
 	}
 }
@@ -255,22 +272,38 @@ void StateBagImpl::SetOwningPeer(std::optional<int> peer)
 	m_owningPeer = peer;
 }
 
+void StateBagImpl::EnableImmediateReplication(bool enabled)
+{
+	m_replicationEnabled = enabled;
+}
+
+void StateBagImpl::SendQueuedUpdates()
+{
+	decltype(m_replicateData) replicationData;
+
+	{
+		std::unique_lock lk(m_replicateDataMutex);
+		m_replicateData.swap(replicationData);
+	}
+
+	for (auto& [key, value] : replicationData)
+	{
+		SendKeyValueToAllTargets(key, value);
+	}
+}
+
 void StateBagImpl::SendAll(int target)
 {
-	eastl::fixed_set<std::string_view, 50> keys;
+	// string_view's target memory can get removed in the mean time, so let's keep locking it instead
+	//eastl::fixed_set<std::string_view, 50> keys;
 
 	{
 		std::shared_lock _(m_dataMutex);
 
 		for (auto& [ key, value ] : m_data)
 		{
-			keys.insert(key);
+			SendKeyValue(target, key, value);
 		}
-	}
-
-	for (auto key : keys)
-	{
-		SendKey(target, key);
 	}
 }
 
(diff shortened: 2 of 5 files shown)


code/components/citizen-resources-core/include/StateBagComponent.h
```diff
@@ -97,8 +106,9 @@ class CRC_EXPORT StateBagComponent : public fwRefCountable, public IAttached<Res
 
 	//
 	// Should be called when receiving a state bag control packet.
+	// return an optional int of the entity id, it's only given if the packet was not processed and can be retried if required
 	//
-	virtual void HandlePacket(int source, std::string_view data) = 0;
+	virtual std::optional<uint32_t> HandlePacket(int source, std::string_view data) = 0;
```
blattersturm, line 111: A bit ugly. I'd rather see a success/fail return value + an out pointer argument that would get (a move-copy of?) the bag name put there, if rejected for such reason.

code/components/citizen-resources-core/src/StateBagComponent.cpp
```diff
@@ -503,7 +503,7 @@ bool StateBagComponentImpl::IsSafePreCreateName(std::string_view id)
 
 	for (const auto& prefix : m_preCreatePrefixes)
 	{
-		if (id.find(prefix) == 0)
+		if (id.rfind(prefix, 0) == 0) // equivalant of a starts with
```
blattersturm, line 506: `equivalent`, with an `e`.

I think C++20 has an explicit starts-with but we're still not in that compiler mode.

code/components/citizen-resources-core/src/StateBagComponent.cpp
```diff
@@ -619,10 +620,36 @@ void StateBagComponentImpl::HandlePacket(int source, std::string_view dataRaw)
 
 	if (!bag)
 	{
+#ifdef IS_FXSERVER
+		if (strncmp(bagName.data(), "entity:", 7) == 0)
```
blattersturm, line 624: This file _should_ be generic: no specific server logic, and ideally no specific parser for `entity:` either. The game interface calls should reflect this.

code/components/citizen-resources-core/src/StateBagComponent.cpp
```diff
@@ -619,10 +620,36 @@ void StateBagComponentImpl::HandlePacket(int source, std::string_view dataRaw)
...
+		{
+			try
+			{
+				int entityId = std::stoi(bagName.data() + 7);
+
+				uint32_t owner = m_gameInterface->GetEntityOwner(entityId);
+				if (owner == uint32_t(-1))
+				{
+					return entityId;
```
blattersturm, line 633: If this case represents 'deferred creation', it might make more sense instead of a return value to have another (optional) callout in the game interface to 'store for later', or similarly, store it _oneself_ in case such a bag gets created.

code/components/citizen-server-impl/src/state/ServerGameState.cpp
```diff
@@ -4041,20 +4114,92 @@ auto ServerGameState::GetEntityLockdownMode(const fx::ClientSharedPtr& client) -
...
 {
 	m_instance = instance;
 
 	m_lockdownModeVar = instance->AddVariable<fx::EntityLockdownMode>("sv_entityLockdown", ConVar_None, m_entityLockdownMode, &m_entityLockdownMode);
 
 	auto sbac = fx::StateBagComponent::Create(fx::StateBagRole::Server);
+	sbac->AddSafePreCreatePrefix("entity:");
 	sbac->SetGameInterface(this);
 
+	uv_timer_init(Instance<net::UvLoopManager>::Get()->GetOrCreate("svMain")->GetLoop(), &g_unhandledStateBagTimer);
```
blattersturm, line 4176: Generally, using `uvw::` so we can correctly delete. In this case it has 'global' lifetime, but eh.

code/components/citizen-server-impl/src/state/ServerGameState.cpp
```diff
@@ -4041,20 +4114,92 @@ auto ServerGameState::GetEntityLockdownMode(const fx::ClientSharedPtr& client) -
...
+			{
+				{
+					std::unique_lock lk(g_unhandledStateBagMutex);
+					auto& updates = g_unhandledStateBagUpdates[unhandled.value()];
+					updates.emplace_back(std::chrono::system_clock::now() + g_unhandledStateBagExpiration, slotId, packetData);
+				}
+
+				if (uv_timer_get_due_in(&g_unhandledStateBagTimer) == 0)
+				{
+					constexpr uint64_t delay = g_unhandledStateBagExpiration.count() + 16; // + 16 as there was an occurance where it triggered early (7ms left)
```
blattersturm, line 4199: [weird magic number]

blattersturm (changes requested):
Quick cursory review on style.

code/components/citizen-server-impl/src/state/ServerGameState.cpp
thorium-cfx: I had that one set up, then I noticed that `uvw::TimerHandle` took in the data member for itself. And the custom data storage they offered was of type `std::shared_ptr<T>` so I couldn't store `fx::StateBagComponent*` without reinterpret_cast, which in turn would've broken the `std::shared_ptr` logic, sigh...

code/components/citizen-resources-core/src/StateBagComponent.cpp
thorium-cfx: Agreed, also on the above comments about this approach.

On the latter: storing it *oneself* would also mean that systems using this system *must* call its clean-up function to get rid of outdated updates, I normally don't do this by design/principle (make it hard to use it incorrectly). Unless, of course, we turn the State Bag System agnostic of all the other systems or if we can tell the State Bag system it can store packets for later. Your thoughts?

code/components/citizen-resources-core/src/StateBagComponent.cpp
thorium-cfx: 👍

fkk-cz:
> Any client that owns an entity can now create its replicating state bags on the server.

Couldn't this cause issues where clients would be able to arbitrarily create/set state bag values that are unneeded or unwanted by the server? Or would the server still have to initialize the state bag on the entity first (current behavior) before it can be modified by the owning client?

thorium-cfx:
> Couldn't this cause issues where clients would be able to arbitrarily create/set state bag values that are unneeded or unwanted by the server? Or would the server still have to initialize the state bag on the entity first (current behavior) before it can be modified by the owning client?

Current conditions on creation of state bags are unchanged, i.e.: the entity must exist, and the requester must be the owner of said entity. But yes, this would allow clients to create state bags on the server without the server initializing these state bags first, still above conditions apply.

code/components/citizen-server-impl/src/state/ServerGameState.cpp
blattersturm: A holding wrapper could work.

code/components/citizen-resources-core/src/StateBagComponent.cpp
blattersturm: The queue being _internal_ seems like a valid idea, too, so any consumers wouldn't have to deal with added complexity other than being able to approve a pending change for later.

blattersturm:
> Couldn't this cause issues where clients would be able to arbitrarily create/set state bag values that are unneeded or unwanted by the server?

As above, not in any different way than is the case now. Filtering is still a later concern, including I guess some way to preemptively reject any client-side updates.

Also, I guess this feature should check for 'entity lockdown' being enabled as well, since in those cases the specific race this part tries to solve wouldn't ever apply.

thorium-cfx:
Whole different setup, no more timers, and a potential base to include throttling later

blattersturm:
'd be nice if someone'd test these with 'real' use cases before I'd merge. :)

thorium-cfx:
That would be great.

This is what I kept on testing while developing this:
1. A `StateBagChangeHandler` to see changes the moment they are applied, directly getting the entityId by `NetToVeh(netId)` or `NetworkGetEntityFromNetworkId(netId)`, where `netId = tonumber(string.match(bagName, "entity:(%d+)"))`, on both client and server.
2. Spawned vehicle on client, then straight after the create call: `Entity(entityId).state:set("apple", "banana", true)` and other fruit pairs, fruit is important.
3. Spawned vehicle on server, applied state bag changes, then on client; move far away, then come back, seeing state bag changes being applied (doing this a few times for good measure).
4. Retrieve state bag values on both server and client at different moments.

BuddiesTV:
> That would be great.
> 
> This is what I kept on testing while developing this:
> 
> 1. A `StateBagChangeHandler` to see changes the moment they are applied, directly getting the entityId by `NetToVeh(netId)` or `NetworkGetEntityFromNetworkId(netId)`, where `netId = tonumber(string.match(bagName, "entity:(%d+)"))`, on both client and server.
> 2. Spawned vehicle on client, then straight after the create call: `Entity(entityId).state:set("apple", "banana", true)` and other fruit pairs, fruit is important.
> 3. Spawned vehicle on server, applied state bag changes, then on client; move far away, then come back, seeing state bag changes being applied (doing this a few times for good measure).
> 4. Retrieve state bag values on both server and client at different moments.

So, i've tested these things with scripts from a production server, everything worked fine. Also i tested it with a script that was just made to test these things. I'll link a video that everything is working as it should.

[Video](https://streamable.com/w6srou)

Ekinoxx0:
I also tested these changes on a dev server and it fixes many issues we encountered when using state bags.

Before :
 - A lot of issues with applying vehicle properties that are stored in state bags (Entity doesn't exist when receiving the state bag, need to wait multiple frames before it sync correctly, not ideal)
 - Server cant verify model of entities that it doesn't know of when receiving the state bag
 
 Using this fix :
  - Vehicle properties are applied as soon as the state bag is received (Entity exist correctly on creation)
 
 Thanks for the good work thorium 👍

mythicalbro:
Did anyone test it with more than one client? I understand that the client that sets the state needs to be the owner of the entity, but the state change should be applied, received for all other clients no matter if they own that entity or not, right?
I tested it and currently the state changes are not being applied, received for client if other client owns it. As soon as this client becomes the owner of that entity the latest state applies. Also, it seems like the state changes are being applied if client is not owner but has had ownership of that entity.
One more thing I noticed when spawning vehicle and setting its state right after spawning it on client, is that this state change is received by other clients even if it doesn't exist for them/is outside their culling radius.

thorium-cfx:
> ... but the state change should be applied, received for all other clients no matter if they own that entity or not, right?
If they are in the vicinity then yes.

> I've tested it and currently the state changes are not being applied, received for client if other client owns it.
Could you give us a step-by-step repro?

> ... is that this state change is received by other clients even if it doesn't exist for them/is outside their culling radius.
Could you elaborate on what culling you are referring to? Need to make sure we talk about the same thing.

I wonder if this issue wasn't there before, as these changes shouldn't change who is getting the packets and who isn't, it only enables clients to create the initial state bag and moves the sending of state bag updates packets until after the entity's creation.

Also when you got a repro can you create a separate issue on this?

AvarianKnight:
This seems very easy to reproduce:

1. Spawn a vehicle with two clients in the vicinity
2. Try to set the state bag on the owning client
3. Only the owning client gets the replication back

If the other client gets ownership of the vehicle it starts to get the sync requests.

It mainly looks like you don't get sync requests until you've owned the entity, after you've owned the entity you will get the sync requests (its no longer limited to just the owning player)

mythicalbro:
> Could you give us a step-by-step repro?
@AvarianKnight provided repro is the way I was able to reproduce it myself.

> Could you elaborate on what culling you are referring to? Need to make sure we talk about the same thing.
Onesync infinity player, player vehicle culling I suppose. It does not happen when the vehicle is spawned from server using CreateVehicleServerSetter native.

> Also when you got a repro can you create a separate issue on this?
@AvarianKnight did - #1691

PR #1387, opened 2022-04-09
tweak(native-decl): set the proper return type

AvarianKnight:
After the introduction of https://github.com/citizenfx/fivem/commit/ce964dd76086bcfb2d4e8e6f7f36433c8d35b199 SET_STATE_BAG_VALUE will always return 0 instead of its proper return type.

### DIFF
--- a/ext/native-decls/SetStateBagValue.md
+++ b/ext/native-decls/SetStateBagValue.md
@@ -5,7 +5,7 @@ apiset: shared
 ## SET_STATE_BAG_VALUE
 
 ```c
-void SET_STATE_BAG_VALUE(char* bagName, char* keyName, char* valueData, int valueLength, BOOL replicated);
+BOOL SET_STATE_BAG_VALUE(char* bagName, char* keyName, char* valueData, int valueLength, BOOL replicated);
 ```
 
 Internal function for setting a state bag value.


blattersturm:
NAK: ABI compatibility concerns, and this native shouldn't have ever returned any value in the first place unless weirdly directly invoked.

Closing for now as not actionable.

mythicalbro:
Looking at its native handler it does return boolean. False if state bag does not exist and true if it exists.
image

blattersturm:
Yeah, but the declaration can't be changed as C# compatibility and such, + anything relying on it would already be wrong (and I'm surprised that the JS native wrapper even returned this value! as a `void` native should always return nothing even from the wrapper).

What I'm rather surprised by is that neither of you provided the context for this change either here or in an issue report/forum post/**anything** other than a Discord message in a _private_ chat that I didn't read until now and that I'm usually not even in:

image

I wasn't even saying it didn't return any value, I'm explicitly mentioning _that it shouldn't to script due to the declaration_.

You'd have gotten a better reply if instead of trying to 'prove me wrong!!' and deleting the branch on my reject-close (as this PR's solution would've been incorrect) you'd have actually provided context.

PR #1393, opened 2022-04-11
Add devtools-rdr3 component and net event log for RedM

Disquse:


blattersturm:
Couldn't this be moved to citizen:devtools (w/ a non-server check) given that it's likely not game-dependent?

Disquse:
Reworked, now net event log code is shared between games.

blattersturm (approved):

PR #1401, opened 2022-04-13
fix(statebags): Fix metatables on __unpack

niekschoemaker:
The playerTM and entityTM tables weren't available in the scope, since they were used in the definition of the variable itself.

Defining it as local one line before using it fixes this.


blattersturm:
`TM` -> `MT` while you're at it, perhaps.

niekschoemaker:
> `TM` -> `MT` while you're at it, perhaps. Also, `scripting/lua` as component name in commit message maybe.

I amended the changes, changed name to MT and also changed the component name in the commit.

PR #1403, opened 2022-04-14
tweak(bootstrap): show firewall issue and name (win8+)

### LINKED ISSUE #792: 'Bad access' errors from Winsock when connecting need more friendly messages
https://forum.cfx.re/t/redm-wont-even-download/4426000

image

The user claimed this was fixed when they changed their firewall configuration, which indeed matches prior mentions of this error.

Code could detect this specific error condition and notify the user to disable their firewall, potentially using the WSC API to tell them the name of the firewall software they have installed so they can check these applications.

thorium-cfx:
* Bootstrap will now show if the issue is related to a firewall block + name of firewall.
* Function to get the firewall's name on Windows 8 and higher, can be used on server and client.

closes #792


code/client/launcher/Bootstrap.cpp
```diff
@@ -73,6 +74,19 @@ bool VerifyViability();
 
 extern void ResetUpdateChannel();
 
+void MessageBox_OnFailedRequestURL(int result)
+{
+	switch(result)
+	{
+	case 7: // CURL ERROR 7 Failed to connect, request is blocked by a firewall or similar.
+		MessageBox(NULL, va(L"An error (%i, %s) occurred while checking the bootstrapper version. " CONTENT_URL_WIDE L" is being blocked by your firewall (%s), to fix this you need to allow FiveM through your firewall or disable it.", result, ToWide(DL_RequestURLError()), os::GetActiveFirewallName()), L"O\x448\x438\x431\x43A\x430", MB_OK | MB_ICONSTOP);
```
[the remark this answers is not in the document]
thorium-cfx: Ah good catch!

code/client/launcher/Bootstrap.cpp
```diff
@@ -73,6 +74,19 @@ bool VerifyViability();
 
 extern void ResetUpdateChannel();
 
+void MessageBox_OnFailedRequestURL(int result)
+{
+	switch(result)
+	{
+	case 7: // CURL ERROR 7 Failed to connect, request is blocked by a firewall or similar.
```
blattersturm, line 81: Can we perhaps amend this to only show the 'custom' message for known-firewall-y _detailed_ error strings? I've mainly seen the 'Bad access' variant show this early, but there might have been others (e.g. TLS stuff).

code/shared/os/Firewall.h
```diff
@@ -0,0 +1,63 @@
...
+std::wstring GetActiveFirewallName(std::wstring defaultName = L"Unknown")
+{
+#ifdef _WIN32_WINNT >= 0x0602 // Need at least windows 8 to compile this
+
+	// IWSCProductList and IWscProduct are Windows 8+ Desktop
+	if (IsWindows8OrGreater())
+	{
+		CoInitializeEx(0, COINIT_APARTMENTTHREADED);
+
+		IWSCProductList* productList = nullptr;
```
blattersturm, line 23: `Microsoft::WRL::ComPtr<T>` from `wrl.h` as in other places might be useful instead.

code/shared/os/Firewall.h
```diff
@@ -0,0 +1,63 @@
...
+		if (SUCCEEDED(CoCreateInstance(__uuidof(WSCProductList), NULL, CLSCTX_INPROC_SERVER, __uuidof(IWSCProductList), reinterpret_cast<LPVOID*>(&productList))))
```
blattersturm, line 24: _uuidof + reinterpret_cast -> IID_PPV_ARGS macro

code/shared/os/Firewall.h
```diff
@@ -0,0 +1,63 @@
...
+		{
+			LONG productCount = 0;
+			if (SUCCEEDED(productList->Initialize(WSC_SECURITY_PROVIDER_FIREWALL)) && SUCCEEDED(productList->get_Count(&productCount)))
+			{
+				IWscProduct* product;
```
blattersturm, line 29: Another WRL-ifyable place.

code/shared/os/Firewall.h
```diff
@@ -0,0 +1,63 @@
...
+				WSC_SECURITY_PRODUCT_STATE state;
+				BSTR bstrName = nullptr;
+
+				for (ULONG i = 0; i < productCount; ++i)
+				{
+					if (SUCCEEDED(productList->get_Item(i, &product)))
```
blattersturm, line 35: If WRL-ified placing it here might be better, especially if `&` doesn't ReleaseAndGetAddressOf.

code/client/launcher/Bootstrap.cpp
```diff
@@ -73,6 +74,19 @@ bool VerifyViability();
 
 extern void ResetUpdateChannel();
 
+void MessageBox_OnFailedRequestURL(int result)
+{
+	switch(result)
+	{
+	case 7: // CURL ERROR 7 Failed to connect, request is blocked by a firewall or similar.
+		MessageBox(NULL, va(L"An error (%i, %s) occurred while checking the bootstrapper version. " CONTENT_URL_WIDE L" is being blocked by your firewall (%s), to fix this you need to allow " PRODUCT_NAME L" through your firewall or disable it.", result, ToWide(DL_RequestURLError()), os::GetActiveFirewallName()), L"O\x448\x438\x431\x43A\x430", MB_OK | MB_ICONSTOP);
```
blattersturm, line 82: If firewall is 'Unknown' (or if Windows [Defender] Firewall shows up as a result here) maybe not showing such in the message could be better - 'your firewall' seems better than 'your firewall (Unknown)'

blattersturm (changes requested):

code/client/launcher/Bootstrap.cpp
thorium-cfx: Do you know of a list of these error strings and do we also need to take in account localization with these?

We could also choose to add such instructions (check their firewall or network) as an optional but potential fix. I looked around for error code 7 and it's often associated with a wrong IP, port, or mostly a firewall (or equivalent) on PC/network that's blocking it.

thorium-cfx:
Message will now explain that error code 7 is an issue due to a blocked request by either their firewall or somewhere else on their network and shows 2 possible solutions. Firewall name will now only be shown when it has been found + other requested changes are applied.

Example
>An error (7, "Description") occurred while checking the bootstrapper version. A request to \<address\> is being blocked by your firewall or somewhere else on your network.
>
>Possible solutions:
>  1. Allow [FiveM, RedM, etc.] through your firewall [(Firewall Name)] or disable it.
>  2. Contact your network administrator.

thorium-cfx:
Note @thorium-cfx (me) resolve conflict

code/shared/os/Firewall.cpp
```diff
@@ -0,0 +1,50 @@
...
+namespace fx::os
+{
+std::wstring Firewall::GetActiveName(std::wstring_view defaultName)
+{
+#ifdef _WIN32_WINNT >= 0x0602 // Need at least windows 8 to compile this
+
+	using namespace Microsoft;
+
+	// IWSCProductList and IWscProduct are Windows 8+ Desktop
+	if (IsWindows8OrGreater())
```
FabianTerhorst, line 12: We only support windows 10+ afaik.

FabianTerhorst (commented):
Looks much better then the current one line error message.

code/shared/os/Firewall.cpp
thorium-cfx: Correct, I left it in as it was from the old code

thorium-cfx:
1 small instructional mistake when I was reordering some parts

PR #1408, opened 2022-04-17
Add administrative rights info

0xLuca:
To create symlinks on windows, the cmd.exe needs admin rights.

### DIFF
--- a/docs/building.md
+++ b/docs/building.md
@@ -17,7 +17,7 @@ To build FiveM, RedM or FXServer on Windows you need the following dependencies:
 * [MSYS2](https://www.msys2.org/) at `C:\msys64\` which is where the installer places it.
 * [Node.js](https://nodejs.org/en/download/) and [Yarn](https://classic.yarnpkg.com/en/docs/install/) in your PATH as `node` and `yarn`.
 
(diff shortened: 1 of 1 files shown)


blattersturm:
Incorrect. It doesn't 'need administrative rights', and it shouldn't get such either.

Do you have Windows configured wrongly so it doesn't allow symlinks for normal users, or are you using an ancient version?

0xLuca:
> Incorrect. It doesn't 'need administrative rights', and it shouldn't get such either.
> 
> Do you have Windows configured wrongly so it doesn't allow symlinks for normal users, or are you using an ancient version?

I am using 10 Pro 21H2, no changes to configuration for creating symlinks.
Cloning does not work and end when trying to create the symlinks with the message "permission denied".

blattersturm:
Do you have 'developer mode' disabled or such? Might make more sense to mention that if it's not enabled by installing VS than asking people to run random code as admin.

0xLuca:
> Do you have 'developer mode' disabled or such? Might make more sense to mention that if it's not enabled by installing VS than asking people to run random code as admin.

That was indeed the problem, it was not being turned on automatically by installing vs.

PR #1416, opened 2022-04-21
Expanded C# API + Blip client/server checks (pending review).

manups4e:
- Added a new Input class to handle game input in a more advanced mode.
- The camera fix is mainly because actually the GameplayCamera MemoryAddres is forced to IntPtr.Zero that leads to errors for many class members.
- Added some Math methods for Vectors and position in circle/sphere
- Added Colors consts in Debug class to help using colors in console.
- Added ScreenToWorld to point a Raycast into a given direction from a screen coord (using cursor position)
- Changed Blip class based on the comments [Here ](https://github.com/citizenfx/fivem/pull/698).

### DIFF
--- a/code/client/clrcore/Debug.cs
+++ b/code/client/clrcore/Debug.cs
@@ -10,7 +10,17 @@ namespace CitizenFX.Core
 {
     public static class Debug
     {
-        [SecuritySafeCritical]
+		public const string LIGHT_RED = "^1";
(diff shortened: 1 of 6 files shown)


manups4e:
Of course these changes can be kept as reference for future changes in the API :D

manups4e:
Closed for a more cleaner future PR

PR #1418, opened 2022-04-22
Blip.cs set client / server checks for review

manups4e:
Based on this quote:
`Probably not. Too much code duplication from the 'client' externals, messy commit history, would've preferred this to be done via some #if stuff in the main External/ dir.`
i tried to change Blip.cs to match it.


donkras:
you should fix two things tbh:
- commits name are not alike others ( others as in those made by bubble or other cfx members ) 
- would be the best if you squashed those two commits into one

manups4e:
This is a draft made to see if this is what d-bub wanted.. it's not meant to be merged. if I get his blessing I'd close this PR and work on all the classes in a new PR.
I'll keep your suggestions in mind for the next one 😊 thanks!

code/client/clrcore/External/Blip.cs
```diff
@@ -155,11 +155,11 @@ public enum BlipSprite
 		ArmWrestling,
 		AmmuNationShootingRange = 313,
 		RaceAir,
-		RaceCar,
+		RaceLand,
```
blattersturm, line 158: This'd break source compatibility with old scripts. Adding a duplicate entry would work fine, instead (i.e. `RaceLand = 315, RaceCar = 315,`).

code/client/clrcore/External/Blip.cs
```diff
@@ -227,7 +227,136 @@ public enum BlipSprite
 		Stopwatch = 430,
 		DollarSignCircled,
 		Crosshair2,
-		DollarSignSquared = 434
+		DollarSignSquared = 434,
+		RaceStunt,
+		HotProperty,
+		UrbanWarfare,
+		Crown = 439,
```
blattersturm, line 234: For new entries, instead of/in addition to 'made up names' as in SHVDN, I'd prefer the original names as seen in `hud.gfx` be used, these are also on https://docs.fivem.net/docs/game-references/blips/

code/client/clrcore/External/Blip.cs
```diff
@@ -498,7 +639,20 @@ private void _Delete()
 				Handle = handle;
 			}
 		}
+#elif IS_FXSERVER
+		[SecuritySafeCritical]
+		private void _Delete()
+		{
+			int handle = Handle;
+			unsafe
```
blattersturm, line 647: Unrelatedly, I'm not sure if this needs an `unsafe` block as such.

Also, server blip natives do need some extra 'love' but I guess this is a simple file to check if the `#if` stuff made sense.

code/client/clrcore/External/Blip.cs
```diff
@@ -498,7 +639,20 @@ private void _Delete()
 				Handle = handle;
 			}
 		}
+#elif IS_FXSERVER
```
blattersturm, line 642: `#else` would do fine

blattersturm (changes requested):
Affirmative, see comments however.

code/client/clrcore/External/Blip.cs
manups4e: i see 🤔

code/client/clrcore/External/Blip.cs
manups4e: all right i'll take a look at it!

code/client/clrcore/External/Blip.cs
manups4e: yeah it was a draft to see if the #if stuff would make sense in this context

PR #1419, opened 2022-04-23
feat(gta/core): dynamic pool resizing

duk-37:
A very preliminary attempt at abusing virtual memory to allow for dynamically-"resizable" pools. At the moment eligible pools need to be manually specified with a hash and size/growth policies.

Can be enabled by passing `-poolhacks` as a command line option.

Note that there's still a fixed upper limit, but it can be raised much higher because the actual memory is only committed on an as-needed basis.

Ideally we get this working with the `TxdStore` pool but that needs me to either:
1. Hardcode its location somehow
2. Figure out how to exclude pre-allocated pools from pool hackery and have reasonable defaults

### DIFF
--- a/code/components/gta-core-five/src/PoolManagement.cpp
+++ b/code/components/gta-core-five/src/PoolManagement.cpp
@@ -7,6 +7,9 @@
 #include <MinHook.h>
 
 #include <Error.h>
+#include <sysAllocator.h>
+
+#include <unordered_set>
 
 class RageHashList
 {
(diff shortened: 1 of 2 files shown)


donkras:
duk doing stuff again!!!! no way

duk-37:
🙂 
image

duk-37:
note: above is not fully working yet, streaming stuff also needs hashmap resize stuff

antigonos7:
> note: above is not fully working yet, streaming stuff also needs hashmap resize stuff

2802 has little space for extra txd
Is it possible to have 105500 txd poolsize?

duk-37:
sincere congratulations to everyone on the fivem team :)

code/components/gta-core-five/src/PoolManagement.cpp
```diff
@@ -237,15 +401,131 @@ static void* PoolAllocateWrap(atPoolBase* pool)
...
+
+					// populate new free entries in hashmap
+					if (nextFree == -1)
+					{
+						assetMap->nextFreeEntry = currentSize;
+						nextFree = currentSize;
+					}
+					else
+					{
+						assetMap->data[nextFree].link = currentSize;
```
duk-37, line 508: todo: pretty sure this is wrong,`nextFree = currentSize;` needs to be on both branches

Legacy-TacticalGamingInteractive:
:(

prikolium-cfx:
Thanks, duk. We appreciate your contribution and passion, but we decided to implement it in another way.

PR #1420, opened 2022-04-24
Update GetVehicleAlarmTimeLeft.md

HamidRez4:


ext/native-decls/GetVehicleAlarmTimeLeft.md
```diff
@@ -7,9 +7,11 @@ apiset: client
 ```c
 int GET_VEHICLE_ALARM_TIME_LEFT(Vehicle vehicle);
 ```
-
+Returns vehicle alarm time left(Millisecond).<br />
```
blattersturm, line 10: Why the `<br />`? Similarly, weird phrasing - is English not your native language?

blattersturm (changes requested):
Worrying PR - providing a description as to _why_ would've helped as so far it only adds more worry than anyone doing this manually...

HamidRez4:
> Worrying PR - providing a description as to _why_ would've helped as so far it only adds more worry than anyone doing this manually...

I will write a description in the future

ext/native-decls/GetVehicleAlarmTimeLeft.md
HamidRez4: I prefer to use this to break lines but if it is unusual for you I will use double space, no is not my native language.

Issue #1421, opened 2022-04-24
Restarting current resource within itself results in server crash


mythicalbro:
```js
const currentResourceName = GetCurrentResourceName();

RegisterCommand(`restart:${currentResourceName}`, () => {
  // Works
  //   ExecuteCommand(`restart ${currentResourceName}`);

  // Crashes
  setImmediate(() => ExecuteCommand(`restart ${currentResourceName}`));
});

// Crashes
// setTimeout(() => ExecuteCommand(`restart ${currentResourceName}`), 1e3);
```

blattersturm:
This isn't fixable for obvious reasons and there's _already_ logic preventing this for the _intended_ natives such as STOP_RESOURCE.

The best we can do is just prevent a stop/start of a currently-executing resource, but you haven't provided much.. details.

mythicalbro:
Alright, so I have to add the command to separate resource to fix that.
The reason why I need separate command for restarting specific resource is because only players with specific ace permission should be able to restart this specific resource. I don't want to give players permission for restart command as they could restart any resource which should not be restarted.

PR #1441, opened 2022-04-29
Resolve crash reported in #1421

donkras:
Couldn't reproduce the crash on client side, that's why i made it only exist in FXServer, tested it in every way i could think of and it didn't cause any issues


thorium-cfx:
Potential fix to #1421

PR #1442, opened 2022-04-29
feat(httphandler): add serverstarted to dynamic.json

pachxl:


blattersturm:
> _No description provided._

_No closing comment provided._

pachxl:
true

aiden-z-source:
You have added what this does now to your comment (which most, especially bub could have understood). But what is the use case for this?

pachxl:
> You have added what this does now to your comment (which most, especially bub could have understood). But what is the use case for this?

My personal use case is for a website status for any server and also for a discord bot to show the servers status and also alert users when a server is starting up.

mickeygeecom:
If you can't fetch data from the already existing .json's, the server is assumlingly not started and running. This seems unnecessary.

pachxl:
> If you can't fetch data from the already existing .json's, the server is assumlingly not started and running. This seems unnecessary.

I thought that as well but this is not the case

blattersturm:
... then try finding out what goes wrong with whatever you're doing instead of adding a weird 'status' indicator.

PR #1450, opened 2022-05-09
Add getter for SET_ENTITY_ONLY_DAMAGED_BY_RELATIONSHIP_GROUP

bukan-siapa-siapa:
While looking through some cheat menus so I can detect some features, I discovered they all mostly use SetEntityOnlyDamagedByRelationshipGroup to give themselves Invincibility, and apparently they are claiming it as 'stealth' invincible. There is no native yet for Server developers to check for this flag, so I decided to make it and ask for a pull request.

### DIFF
--- a/code/components/extra-natives-five/src/EntityExtraNatives.cpp
+++ b/code/components/extra-natives-five/src/EntityExtraNatives.cpp
@@ -36,4 +36,23 @@ static InitFunction initFunction([]()
 
 		context.SetResult<bool>(result);
 	});
+
+	/* original setter is SET_ENTITY_ONLY_DAMAGED_BY_RELATIONSHIP_GROUP, perhaps this is for entities too? */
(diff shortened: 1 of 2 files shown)


blattersturm:
No 'getters' that only make sense for 'anti-cheat' use cases will be added until the certification/'limited-access feature' system for such is implemented.

Keyinator:
> No 'getters' that only make sense for 'anti-cheat' use cases will be added until the certification/'limited-access feature' system for such is implemented.

I totally get that for "extended"(/intrusive) privileges.
But knowing variables like "critical damage" are basic functionalities they should not require certification imo.
Especially for server-owners who develop the anticheat themselves and will be unlikely to get verification.

While cheats themselves can be detected by fivem, such values (like crit-damage and rel-group-damage) will never be detectable directly as they depend on the scripts the server is using. So requiring certification (and restricting access) would be really bad for a lot of people (not only "big" anticheats").

PR #1453, opened 2022-05-12
Client Interface keys added and translated (Turkish)

itsBLOK:


technetium-cfx:
Please use https://translations.cfx.re/ for client interface translations.

PR #1462, opened 2022-05-27
feat(client/clrcore): Updated C# EventHandlerAttribute support

ktmitton:
* Updated how the return type for the event is determined so it isn't hardcoded to void, but uses the actual return type of the function. This allows event handlers that perform async actions.
* Added the ability to apply multiple EventHandlerAttribute tags to the same function for instances where you want the same action to be applied on multiple events.

Here's an example of how this looks in usage:
```csharp
[EventHandler("mittons:setspawnwantedlevel")]
[EventHandler("mittons:resetspawnwantedlevel")]
public void SetSpawnWantedLevel(int? level = default)
{
    _spawnWantedLevel = level;
}
```

### DIFF
--- a/code/components/citizen-server-impl/include/state/ServerGameState.h
+++ b/code/components/citizen-server-impl/include/state/ServerGameState.h
@@ -1215,6 +1215,10 @@ public:
 
 	void ReassignEntity(uint32_t entityHandle, const fx::ClientSharedPtr& targetClient);
 
+private:
+	void ReassignEntityInner(uint32_t entityHandle, const fx::ClientSharedPtr& targetClient);
+
+public:
 	void DeleteEntity(const fx::sync::SyncEntityPtr& entity);
 
 	void ClearClientFromWorldGrid(const fx::ClientSharedPtr& targetClient);
--- a/code/components/citizen-server-impl/src/state/ServerGameState.cpp
+++ b/code/components/citizen-server-impl/src/state/ServerGameState.cpp
@@ -2515,7 +2515,7 @@ void ServerGameState::SetPopulationDisabled(int bucket, bool disabled)
 }
 
 // make sure you have a lock to the client mutex before calling this function!
-void ServerGameState::ReassignEntity(uint32_t entityHandle, const fx::ClientSharedPtr& targetClient)
+void ServerGameState::ReassignEntityInner(uint32_t entityHandle, const fx::ClientSharedPtr& targetClient)
 {
 	auto entity = GetEntity(0, entityHandle);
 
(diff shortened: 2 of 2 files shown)


ktmitton:
This is essentially code that I've run in my Client class that extends BaseScript, but I haven't found the right documentation on how to build and test the change in the actual sdk.

Can somebody point me at the right docs and/or provide tips on how to test a change like this?

blattersturm:
I think @thorium-cfx was working on a rework of the .NET runtime, so any changes to the existing one may complicate matters a bit.

ktmitton:
That's convenient, because one of the things I was considering was figuring out how to update to .net 6 :D

@thorium-cfx any way I can help contribute to the .net runtime revamp?

SQKo:
>     * Added the ability to apply multiple EventHandlerAttribute tags to the same function for instances where you want the same action to be applied on multiple events.

This is nice, can the same thing be done with Command attribute? e.g. to create multiple command names/shortcuts on same action

ktmitton:
> > ```
> > * Added the ability to apply multiple EventHandlerAttribute tags to the same function for instances where you want the same action to be applied on multiple events.
> > ```
> 
> This is nice, can the same thing be done with Command attribute? e.g. to create multiple command names/shortcuts on same action

I don't see why not, I'll take a look in a few minutes

thorium-cfx:
Interesting, I can think of reasons for multiple EventHandler attributes on 1 function. Code looks fine but haven't tested it.
To compile you would need to compile the client itself https://github.com/citizenfx/fivem/blob/master/docs/building.md

> I think @thorium-cfx was working on a rework of the .NET runtime, so any changes to the existing one may complicate matters a bit.
This is true and no this won't complicate anything, the reworked version doesn't use any core files of this version, so feel free.

> That's convenient, because one of the things I was considering was figuring out how to update to .net 6 :D
> 
> @thorium-cfx any way I can help contribute to the .net runtime revamp?
Yeah that would be nice. Though not any time soon, we do want to look into CoreCLR. Especially sandboxing and performance are big concerns to us.

The rework however was/is focused on increasing performance overall, plus some extra features. In short (unless you hate performance) you won't be disappointed ;) As to contribute; all the hard work is already done. Though functionality/features like these can help.

ktmitton:
@SQKo I want to write some tests to verify functionality and performance, but here's a preview of the changes for your request:
https://github.com/citizenfx/fivem/compare/master...ktmitton:addMultipleCommandAttributeSupport

Feel free to take the conversation over to my github or dm's, but I'm gonna stop cluttering up this repo with comments on this until I have a PR, seems the polite thing for me to do 😄. Just wanted to keep you in the loop!

ktmitton:
@thorium-cfx partly because I'm still having trouble getting a build to run, and partly because there are different compiler directives to account for, so manual testing won't be fun, I created this PR:
https://github.com/citizenfx/fivem/pull/1470

Can you take a look at that and let me know if you think I _should_ commit those tests, or should I just keep them in my local branches?

PR #1463, opened 2022-05-27
Added missing SetVehicleCurrentGear native declaration

BerkieBb:
I found this native in the code but not on the native documentation, so this pull request adds it.


BerkieBb:
I'm going to make a new PR, my local history is messed up on this branch so I don't want to add anything random to this

BerkieBb:
I'm not going to make a new one because I understand why this was never added, the automatic transmission of gta overwrites this setter, better yet, it just ignores it completely.

PR #1465, opened 2022-05-29
tweak(mumble/server): increase some limits

Korioz:
This is a fix mainly addressing this issue: https://forum.cfx.re/t/native-audio-manzareks-investigations/3776013/14

It has been tested on a mumble server with 1100+ players, the test proved this PR fixes :
- mumble-server rejecting authentication of some people.
- loss of voice with 32+ players in the same area.


blattersturm:
Are you sure _all_ of these changes are helpful? Some of them would indicate issues elsewhere... or some optimization needed somewhere (for example, to not transfer members of channels the user isn't currently in).

They make sense, still, but still indicate another issue to fix somewhere.

Korioz:
I'm sure that increasing clients udp queue limit from 5 to 25 fix the issue this PR is about (link above).

Increasing `DEFAULT_MAX_BANDWIDTH` from 76000 to 144000 and clients tcp queue limit from 30 to 150 reduced the number of reports of clients having trouble to connect to mumble-server.

I'm not sure increasing `listeningUsers` is helpful.

I cannot guarantee *all of these changes are helpful* and since most of these issues appears only under very specific conditions it only make the task of trying another changeset more difficult.

manzarek123:
> This is a fix mainly addressing this issue: https://forum.cfx.re/t/native-audio-manzareks-investigations/3776013/14

This is not related at all by the way

PR #1467, opened 2022-05-29
feat(extra-natives/five): GetPedDecorations

nikez:
Intended usage is to retrieve the currently applied decorations and count if needed.

Tested patterns on 160x, 2372, 2628.
Tested with custom decorations (tattoos)

### DIFF
--- a/code/components/extra-natives-five/src/EntityExtraNatives.cpp
+++ b/code/components/extra-natives-five/src/EntityExtraNatives.cpp
@@ -4,6 +4,53 @@
 #include <scrBind.h>
 
 #include <Streaming.h>
+#include <Hooking.h>
+
+static int PedIdOffset;
+static uintptr_t CReplayPedExtensionPtr;
+
+class OverlayEntry
+{
+public:
+	uint32_t collection_hash; // 0x0000
+	uint32_t overlay_hash; // 0x0004
+	char pad_0008[8]; // 0x0008
+	float unk_float; // 0x0010
+}; // Size: 0x0014
+
+class PedEntry
+{
+public:
+	char pad_0000[184]; // 0x0000
+	OverlayEntry entries[87]; // 0x00B8
+	uint32_t overlay_count; // 0x0784
+	char pad_0788[80]; // 0x0788
+}; // Size: 0x07D8
+
+class CReplayPedExtension
+{
+public:
+	PedEntry entries[1]; // Indexed by a value from CPed-class (0x308)
+};
+
+class ptrCReplayPedExtension
+{
+public:
+	CReplayPedExtension* ptr; // 0x0000
+}; // Size: 0x0008
+
+struct DecorationResult
+{
+	DecorationResult(uint32_t m_collection_hash, uint32_t m_overlay_hash)
+		: collection_hash(m_collection_hash), overlay_hash(m_overlay_hash)
+	{
+	}
+
+	uint32_t collection_hash;
+	uint32_t overlay_hash;
+
+	MSGPACK_DEFINE_ARRAY(collection_hash, overlay_hash)
+};
 
 static const char* GetEntityArchetypeName(int entityHandle)
 {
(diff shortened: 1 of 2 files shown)


nikez:
Don't know if wanted, but here's a test script to verify that it works.
Ensure your ped is a freemode character and start script. If script output is `All test completed successfully.` it completed successfully.

Flow is:
Remove all existing ped decorations
Loop through a list consisting of two different overlays and tattoos
On each loop, check if the count (`GET_PED_DECORATIONS_COUNT`) of applied decorations is equal to the loop count and that the hashes reported from (`GET_PED_DECORATIONS`) are equal to the hashes of the decorations we've applied.

code/components/extra-natives-five/src/EntityExtraNatives.cpp
```diff
@@ -18,6 +62,72 @@ static const char* GetEntityArchetypeName(int entityHandle)
...
 }
 
+static HookFunction hookFunction([]()
+{
+	PedIdOffset = *hook::get_pattern<int>("00 00 66 85 C0 78 ? 0F BF C8 0F B7 05 ? ? ? ? 3B C8 7D ? 44 8B C1", -0x2);
+	CReplayPedExtensionPtr = hook::get_address<uintptr_t>(hook::get_pattern<uintptr_t>("4C 03 05 ? ? ? ? EB 03 4D 8B C3"), 3, 7);
+
+	fx::ScriptEngine::RegisterNativeHandler("GET_PED_DECORATIONS", [](fx::ScriptContext& context)
+	{
+		if (PedIdOffset == 0x0 || CReplayPedExtensionPtr == 0x0)
```
blattersturm, line 72: Would these ever be zero?

code/components/extra-natives-five/src/EntityExtraNatives.cpp
```diff
@@ -4,6 +4,50 @@
...
+};
+
+class ptrCReplayPedExtension
+{
+public:
+	class CReplayPedExtension* ptr; // 0x0000
+}; // Size: 0x0008
+#pragma pack(pop)
+
+static struct DecorationResult
```
blattersturm, line 44: `static` is a bit weird for something that's not declaring anything.

code/components/extra-natives-five/src/EntityExtraNatives.cpp
```diff
@@ -4,6 +4,50 @@
...
+class CReplayPedExtension
+{
+public:
+	PedEntry entries[1]; // Indexed by a value from CPed-class (0x308)
+};
+
+class ptrCReplayPedExtension
+{
+public:
+	class CReplayPedExtension* ptr; // 0x0000
```
blattersturm, line 40: `class` here is redundant.

code/components/extra-natives-five/src/EntityExtraNatives.cpp
```diff
@@ -4,6 +4,50 @@
 #include <scrBind.h>
 
 #include <Streaming.h>
+#include <Hooking.h>
+
+static int PedIdOffset;
+static uintptr_t CReplayPedExtensionPtr;
+
+#pragma pack(push, 1)
```
blattersturm, line 12: `#pragma pack` here seems unneeded as there's no 'broken' alignment.

code/components/extra-natives-five/src/EntityExtraNatives.cpp
```diff
@@ -18,6 +62,72 @@ static const char* GetEntityArchetypeName(int entityHandle)
...
+		if (entity && entity->IsOfType<CPed>())
+		{
+			uint16_t bufferIndex = *(uint16_t*)((uintptr_t)entity + PedIdOffset);
+			ptrCReplayPedExtension* decorationSystem = (ptrCReplayPedExtension*)CReplayPedExtensionPtr;
+			if (decorationSystem && decorationSystem->ptr)
+			{
+				PedEntry* decorationPedEntry = &decorationSystem->ptr->entries[bufferIndex];
+				if (decorationPedEntry)
+				{
+					decorationList.resize(decorationPedEntry->overlay_count);
```
blattersturm, line 90: Perhaps prefer `.reserve()` and `.push_back` in the loop (or even `.emplace_back` without the `{}` initializers).

code/components/extra-natives-five/src/EntityExtraNatives.cpp
```diff
@@ -18,6 +62,72 @@ static const char* GetEntityArchetypeName(int entityHandle)
...
+					{
+						decorationList[i] = { decorationPedEntry->entries[i].collection_hash, decorationPedEntry->entries[i].overlay_hash };
+					}
+				}
+			}
+		}
+		context.SetResult(fx::SerializeObject(decorationList));
+	});
+
+	fx::ScriptEngine::RegisterNativeHandler("GET_PED_DECORATIONS_COUNT", [](fx::ScriptContext& context)
```
blattersturm, line 101: What's the use for a native to get the count without getting the data itself? Lot of code duplication as such.

ext/native-decls/GetPedDecorations.md
```diff
@@ -0,0 +1,24 @@
...
+
+```c
+object GET_PED_DECORATIONS(Ped ped);
+```
+
+Returns a list of decorations applied to a ped.
+
+The data returned adheres to the following layout:
+```
+[{collectionHash1, overlayHash1}, ..., {collectionHashN, overlayHashN}]
```
blattersturm, line 15: `{}` would refer to an object, which this isn't.

ext/native-decls/GetPedDecorations.md
```diff
@@ -0,0 +1,24 @@
...
+```
+
+Only works on the invoking client. Calling it on another client results in undefined data.
```
blattersturm, line 18: What would 'invoking client' mean? It's a client-side native, the client invoking the native would always be the invoking client.

I'm guessing _remote players_ won't have accurate data, instead?

For example:

> This command will return undefined data if invoked on a remote player ped.

ext/native-decls/GetPedDecorations.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Parameters
+* **ped**: The ped you want to check
```
blattersturm, line 21: `you want to check` -> `to retrieve data for.`?

code/components/extra-natives-five/src/EntityExtraNatives.cpp
nikez: Monkey see, monkey do. I used existing FiveM code as a reference. But can change up.

https://github.com/citizenfx/fivem/commit/bc2f29bbc57da75a423c77fb760380f230b07b24

ext/native-decls/GetPedDecorations.md
nikez: Monkey see, monkey do. I used existing FiveM documentation as a reference. But can change up.

https://github.com/citizenfx/fivem/commit/bc2f29bbc57da75a423c77fb760380f230b07b24

code/components/extra-natives-five/src/EntityExtraNatives.cpp
nikez: Honestly; this was more-so a nice-to-have. I don't see a huge use case for it, so could honestly refactor and remove it. Will do.

nikez:
🧋There.

PR #1470, opened 2022-05-31
Add clr tests

ktmitton:
This is my first attempt to write some tests before making some more complex changes. I stubbed out CitizenFX.Core.Native.API with a spy I could use to monitor the side-effects.

I also stubbed out some stuff in CitizenFX.Core, the bare minimum to get BaseTests.InitializeOnAdd() to run, but long-term I can see those stubs being replaced with their real implementations. That would involve mocking out more of API, though, which is why I avoided it for this first pass.


ktmitton:
@blattersturm I realize I may have tagged the wrong person before, but I'm not sure what the process is here. What's the best way to go about getting this PR reviewed/approved? I'm holding off on making additional changes until I know this is a good path to take, avoid putting in a bunch of time for something that may not be accepted.

blattersturm:
I'm not sure myself _either_ - perhaps best to discuss changes you want to actually make *beforehand* as otherwise it ends up with stuff like this that otherwise looks very out of place and doesn't help anyone.

thorium-cfx:
Uh finger slipped :/ phone...

FabianTerhorst:
Not sure if adding tests for mono rt1 still make sense now.

PR #1471, opened 2022-05-31
[Voip Mumble] Add stereo - pan / noise and compressor node for audio context filters

joelwurtz:
Hey,

This pull request intend to add more audio node / filters to audio context natives : 

  *  StereoPannerNode : Allow to do some balance between left / right spekers with a "pan" parameter
   * NoiseNode : Allow to add white / pink / brown noise to the audio 
   * PeakCompNode : Add a compression node, which should be really useful to smooth volume between low and high mic and a big curve with the waveshaper node
   
There is also a new native "AUDIOCONTEXT_GET_NODE_PARAM" which should be useful to get any parameter from any audio node instead of writing every binding for each parameter

### DIFF
--- a/code/components/voip-mumble/src/MumbleAudioOutput.cpp
+++ b/code/components/voip-mumble/src/MumbleAudioOutput.cpp
@@ -1867,6 +1867,30 @@ struct scrBindArgument<lab::FilterType>
 	}
 };
 
+template<>
+struct scrBindArgument<lab::NoiseNode::NoiseType>
+{
+	static lab::NoiseNode::NoiseType Get(fx::ScriptContext& cxt, int i)
+	{
+		std::string_view a = cxt.CheckArgument<const char*>(i);
+
+		if (a == "pink")
+		{
+			return lab::NoiseNode::NoiseType::PINK;
+		}
+		else if (a == "brown")
+		{
+			return lab::NoiseNode::NoiseType::BROWN;
+		}
+		else if (a == "white")
+		{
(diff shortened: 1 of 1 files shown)


blattersturm:
Unrelatedly to the actual changes:

I'm not sure if the LabSound/XA2 code path even was reliable enough to add any new functionality to - does anyone actually use this code path on their server still? 😮 I thought people moved on to the game's audio subsystem handler which I again _think_ was more reliable nowadays?

joelwurtz:
It works, but yeah it's not so reliable : 

 * You have to wait between node connection (we wait for 200ms) otherwise effect / node is not applied (or it's random, i believe it depends on the client load, it it's fast enough ^^)
 * You have to be sure to destroy those node once there are not used, otherwise there are stille processed (even if not used) and the cpu consommation can boom
 
We try them, as they allow to do one filter per listening person in the mumble (so we can have a different effect per person, like crushing the audio for a radio if distance is too long, removing frequence add some effect if someone is in the water, etc ....)

But finally we reswitch to submix, (we have both implementation for submix or audiocontext) as like you said there are not reliable enough, and use a pool system for submix (i don't think there will be more than 40 effects in parallels so it's good enough)

This PR can be closed then.

From what i understand submix effect directly comes from rage engine ? does the 40 max effect is an internal limitation ?

blattersturm:
With submix stuff I believe you only need a different index if there's a different filter setup, not for different sources.

PR #1476, opened 2022-06-04
feat(extra-natives/five): Add GET_ALL_VEHICLE_MODELS native

Lucas7yoshi:
This native returns all vehicle models that are registered, it also will be unregister any vehicles in resources that get stopped.

This is my first PR relating to C++. I can't promise its perfect.

Test command:

```lua
RegisterCommand("spawnrandomcar", function()
	local vehicles = GetAllVehicleModels()
	local veh = vehicles[math.random(1, #vehicles)]
	RequestModel(veh)
	repeat Wait(0) until HasModelLoaded(veh)
	local veh = CreateVehicle(veh, GetEntityCoords(PlayerPedId()), GetEntityHeading(PlayerPedId(), true, false)
	SetPedIntoVehicle(PlayerPedId(), veh, -1)
end)
```

### DIFF
--- a/code/components/extra-natives-five/src/ModelListNatives.cpp
+++ b/code/components/extra-natives-five/src/ModelListNatives.cpp
@@ -1,0 +1,51 @@
+#include <StdInc.h>
+#include <Hooking.h>
+
+#include <ScriptEngine.h>
+#include <ScriptSerialization.h>
+
+static uint32_t(*initVehicleArchetype)(const char* name, bool a2, unsigned int a3);
+
+static void(*g_origArchetypeDtor)(fwArchetype* at);
+
+// Use a map of hashes and names for optimization purposes
+// We just hash it on registration instead of needing to rehash it when checking for it every deregistration
+static std::unordered_map<uint32_t, std::string> g_vehicles;
+
+static uint32_t initVehicleArchetype_stub(const char* name, bool a2, unsigned int a3)
+{
+	g_vehicles.insert({ HashString(name), name });
+	return initVehicleArchetype(name, a2, a3);
+}
+
+static void ArchetypeDtorHook(fwArchetype* at)
+{
+	g_vehicles.erase(at->hash);
+	g_origArchetypeDtor(at);
+}
+
+static HookFunction hookFunction([]()
+{
+	//hook for vehicle registering
+	{
+		auto location = hook::get_pattern<char>("E8 ? ? ? ? 48 8B 4D E0 48 8B 11");
+		hook::set_call(&initVehicleArchetype, location);
+		hook::call(location, initVehicleArchetype_stub);
+	}
+	// hook to catch deregistration of vehicles
+	{
+		auto location = hook::get_pattern("E8 ? ? ? ? 80 7B 60 01 74 39"); // Taken from StreamingFreeTests.cpp
+		hook::set_call(&g_origArchetypeDtor, location);
+		hook::call(location, ArchetypeDtorHook);
+	}
+	fx::ScriptEngine::RegisterNativeHandler("GET_ALL_VEHICLE_MODELS", [](fx::ScriptContext& context)
+	{
+		// its redundant and possibly confusing to give both hashes and names, so just give the names.
+		std::vector<std::string_view> models;
+		for (const auto& pair : g_vehicles)
+		{
+			models.push_back(pair.second);
+		}
+		context.SetResult(fx::SerializeObject(models));
+	});
+});
--- a/ext/native-decls/GetAllVehicleModels.md
+++ b/ext/native-decls/GetAllVehicleModels.md
@@ -1,0 +1,32 @@
+---
+ns: CFX
+apiset: client
+---
+## GET_ALL_VEHICLE_MODELS
+
+```c
+object GET_ALL_VEHICLE_MODELS();
+```
+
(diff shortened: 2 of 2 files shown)


xenfovn:
it can get addon model ?

Lucas7yoshi:
> it can get addon model ?

this gets all vehicles that are registered, including addon vehicles

xenfovn:
it will be necessary, if it is accepted please mention me , thanks !

code/components/extra-natives-five/src/ModelListNatives.cpp
```diff
@@ -0,0 +1,63 @@
...
+		g_vehicles.insert({ hash, name });
+	}
+	return initVehicleArchetype(name, a2, a3);
+}
+
+static void ArchetypeDtorHook(fwArchetype* at)
+{
+	if (at->assetType == 1)
+	{
+		if (g_vehicles.find(at->hash) != g_vehicles.end())
```
blattersturm, line 29: `.erase()` doesn't mind if the key was or wasn't present.

code/components/extra-natives-five/src/ModelListNatives.cpp
```diff
@@ -0,0 +1,63 @@
...
+		hook::set_call(&initVehicleArchetype, location);
+		hook::call(location, initVehicleArchetype_stub);
+	}
+	// hook to catch deregistration of vehicles
+	{
+		auto location = hook::get_pattern("E8 ? ? ? ? 80 7B 60 01 74 39");
+		hook::set_call(&g_origArchetypeDtor, location);
+		hook::call(location, ArchetypeDtorHook);
+	}
+	{
```
blattersturm, line 51: No variable is being defined here so the `{` scope isn't needed. A little nitpicky, yes.

code/components/extra-natives-five/src/ModelListNatives.cpp
```diff
@@ -0,0 +1,63 @@
...
+		fx::ScriptEngine::RegisterNativeHandler("GET_ALL_VEHICLE_MODELS", [](fx::ScriptContext& context)
+		{
+			// its redundant and possibly confusing to give both hashes and names, so just give the names.
+			std::vector<std::string> models;
```
blattersturm, line 55: I believe you can use `std::string_view` here to reduce some excessive copying. Would need testing to see if the msgpack serializer will handle it though.

code/components/extra-natives-five/src/ModelListNatives.cpp
```diff
@@ -0,0 +1,63 @@
...
+			for (auto& pair : g_vehicles)
```
blattersturm, line 56: `const auto&` may be preferred here but again a nitpick.

blattersturm (commented):
A few nitpicks about language style here.

Lucas7yoshi:
> A few nitpicks about language style here.

I got no issues with nitpicks particularly when I'm working with a language im not familiar with. will address these when Im on a computer 👍

code/components/extra-natives-five/src/ModelListNatives.cpp
```diff
@@ -0,0 +1,63 @@
...
+	if (g_vehicles.find(hash) == g_vehicles.end())
+	{
+		g_vehicles.insert({ hash, name });
+	}
+	return initVehicleArchetype(name, a2, a3);
+}
+
+static void ArchetypeDtorHook(fwArchetype* at)
+{
+	if (at->assetType == 1)
```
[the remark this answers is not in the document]
Lucas7yoshi: This one was done just as a measure to optimize to avoid running on ALL unregistered archetypes (which seems to happen during runtime a fair bit iirc?). Theres no specific one for vehicles so this seemed preferable still just for some minor optimization.

code/components/extra-natives-five/src/ModelListNatives.cpp
```diff
@@ -0,0 +1,63 @@
...
+// Use a map of hashes and names for optimization purposes
+// We just hash it on registration instead of needing to rehash it when checking for it every deregistration
+static std::unordered_map<uint32_t, std::string> g_vehicles;
+
+static uint32_t initVehicleArchetype_stub(const char* name, bool a2, unsigned int a3)
+{
+	uint32_t hash = HashString(name);
+	if (g_vehicles.find(hash) == g_vehicles.end())
+	{
+		g_vehicles.insert({ hash, name });
```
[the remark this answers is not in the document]
Lucas7yoshi: I was using a vector prior before i added hashes for the sake of avoiding hashing everything every run, in which case it does let you add duplicates. So just code that carried over. will address

Lucas7yoshi:
Pretty sure this is requesting the updating of https://github.com/citizenfx/fivem/blob/master/code/client/clrcore/External/VehicleHash.cs

ali-exacute:
This is an awesome addition , thank you @Lucas7yoshi for this

Is there anyway to sort or check if vehicle model is streamed or not ?

Lucas7yoshi:
> This is an awesome addition , thank you @Lucas7yoshi for this
> 
> Is there anyway to sort or check if vehicle model is streamed or not ?

I think you mean if theres a way to check if its a "custom" streamed vehicle, and to my knowledge no unfortunately, theres natives for getting DLC vehicles, but no way to check more beyond that to my knowledge.

Better question for general discord support channels, it might be possible.

PR #1482, opened 2022-06-08
Update locale-fr.json

Florian-1801:


blattersturm:
Localizations should be updated via the translations portal.

PR #1492, opened 2022-06-17
refactor(locale-pl): Polish locale update

michur55:


aiden-z-source:
Please use https://translations.cfx.re/ for client interface translations.

PR #1493, opened 2022-06-18
feat(extra-natives/rdr3): Door system

nikez:
Pretty much the same as V.


Disquse (approved):

Disquse:
Tested on all builds, looks good to me.

nikez:
> Maybe this is the correct PR:
> 
> Is the script GUID extraction correct? For instance, 0xF7424890E4A094C0 (RDR) includes additional dummy object handling. Coincidentally, this is also slightly broken in V (missing object-type check; see 0x701FDA1E82076BA4 in GTA)

I've updated the PR to reflect this. Thanks!

nikez:
Tested with .31 - works there as well!

PR #1498, opened 2022-06-21
tweak(conhost-v2): performance overlay menu tweaks and rdr3 support

Ktos93:
These changes allow the user to choose what they want to display in the performance overlay.Additionally, they allow you to use an overlay for RedM (without GPU information).I am not happy with the implementation of drawOverlay. I have the impression that it is terrible
image
image
image


Ktos93:
Yeah... After publishing the PR I saw that after starting the normal version of the game in the menu options appear that should not be there. So it would be required to clear saved client options.
Thanks for response

PR #1500, opened 2022-06-22
Increase max sync packet length

Disquse:
This change should resolve https://forum.cfx.re/t/ghosting-player-poisoning-due-to-many-components-weapons-applied-to-player-peds/4868864

Clone creation packets in RedM can be huge (>1200 bytes, more than the current limit), while increasing limit isn't a perfect solution, it should work for most cases. Tested locally with 2 clients, haven't found any issues, however, added a warning message on client informing player that there could be sync issues because of receiving a large packet, just in case.

### DIFF
--- a/code/components/gta-net-five/src/CloneManager.cpp
+++ b/code/components/gta-net-five/src/CloneManager.cpp
@@ -43,10 +43,12 @@ rage::netObject* g_curNetObject;
 static std::set<uint16_t> g_dontParrotDeletionAcks;
 
 #ifdef GTA_FIVE
-static constexpr int g_netObjectTypeBitLength = 4;
+static constexpr int kNetObjectTypeBitLength = 4;
 #elif IS_RDR3
-static constexpr int g_netObjectTypeBitLength = 5;
+static constexpr int kNetObjectTypeBitLength = 5;
 #endif
+
+static constexpr int kSyncPacketMaxLength = 2400;
 
 void ObjectIds_AddObjectId(int objectId);
 void ObjectIds_StealObjectId(int objectId);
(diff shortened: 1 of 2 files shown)


t3chman:
FYI, we have tested the server-side changes with numerous users and it is working as expected. It does indeed fix the issue noted in the post above. No new crashes or strange behavior detected.

code/components/gta-net-five/src/CloneManager.cpp
```diff
@@ -1510,6 +1513,11 @@ void CloneManagerLocal::HandleCloneSync(const char* data, size_t len)
 	net::Buffer netBuffer(reinterpret_cast<const uint8_t*>(data), len);
 	g_cloneMsgPacket = netBuffer.Clone();
 
+	if (len >= kSyncPacketWarnLength)
```
blattersturm, line 1516: This check should rather be warning on _send_ on the other side, since _receiving_ a large packet is the thing that's likely to fail.

blattersturm (commented):
Fine, but the warning is on the wrong side.

code/components/gta-net-five/src/CloneManager.cpp
t3chman: Could we keep the receive warning in addition to a send warning? I feel like it can't hurt to have information on both sides, especially once it's out in the wild and devs may want to get a warning when receiving large packets while they're investigating prior to any reproduction/isolation. For the issue this resolves, we weren't able to replicate it for some time and a receive warning would have done wonders.

blattersturm: A receive warning wouldn't even be possible as an oversized packet (not even of this type - it's related to issues that may occur when fixing this) would simply fail to send/not arrive.

blattersturm (approved):

rmrsnpdr:
cant ride horse when someone riding it. you will automatically go down.

rmrsnpdr:
still encountering the ghosting issue when we reach 70+

t3chman:
I believe you were the one I discussed this with. It's impossible for this "ghosting" issues to occur anymore and what you are experiencing is very likely culling at anything over 31 players in scope.

PR #1503, opened 2022-06-23
tweak(conhost) automatic disable overlay when nothing to draw 

Ktos93:
I don't know if such a thing is needed, but personally I would prefer it if it disappeared automatically and at the same time made it impossible to enable without selected metrics


Ktos93:
You can also implement it this way, I don't know which way you prefer [Click There](https://github.com/Ktos93/fivem/commit/fd7ef9ad63d8ffb97a475dc08ca32ecb5ff4ab3e)

PR #1507, opened 2022-06-25
tweak(scripting/five): fix incorrect mappings in CrossMapping_Universal.h

fingaweg:
This commit addresses incorrect mappings in CrossMapping_Universal.h.
The general issue seems to be matching natives at/to the wrong place (swapped).


Disquse:
`CrossMapping_Universal.h` is still based on `[2545.0](https://github.com/citizenfx/fivem/commit/b2ed61e69fc4dd2b1d6ccd586a9c20b5497ef8d1)`, there's no 2612.1 changes

Issue #1490, opened 2022-06-16
Bullets not registering with Automatic Guns


Walker1390:
When shooting another player with a fully automatic gun some, seemingly randomly, bullets do not deal any damage. I don't have the expertise or knowledge to speculate as to why this happens or if it is even something that can be fixed but wanted to make this issue to make someone aware of it if cfx isn't already.

Steps to Reproduce:

1. Tap fire at a player X amount of times slowly until they die.
2. Empty the magazine until there are only enough bullets to kill the player
3. Hold Left Click (Or whatever shooting is bound to) and whats left of the magazine into the player
4. The player won't have died despite you firing enough bullets to kill them.

Clip: https://streamable.com/3m6v0g

This can also be seen just spraying at players, you'll notice not every bullet deals damage however this was the easier way I could replicate it with steps to do it every time. I recreated this on various servers and also a completly [fresh cfx server](https://github.com/citizenfx/cfx-server-data) with only vMenu to give myself a weapon. The fresh server being the one used for the clip. 

If the issue is ping related, I was in the UK and so was the server for the clip, the player I was shooting at was in the US.

Server Artifact: 5652

blattersturm:
The repro steps are still a bit vague and the video isn't very clear on these either (same with how the player is being respawned in place), for example what are 'X amount of times' and how to tell what amount of bullets is enough?

For things like this the actual accuracy of the repro steps is very important, and currently there's just too many unknowns instead of 'have player 1 at coordinate A and player 2 at coordinate B and use weapon Y with Z bullets left and this script to respawn the dying player'.

Korioz:
Hi, i will try to explain a bit better

When shooting with an automatic gun like `GUSENBERG` multiple bullets in a row by holding the left click mouse button, some bullets will not register.

The first part of the repro is made to know how many bullets are needed to kill a player with the chosen automatic weapon.
The second part is made to only have the supposed number of bullets needed to kill a player in our gun's magazine.
The third part we try to kill the player by holding left click with only the number of bullets we are supposed to need to kill a player.

When bullets are not registered the event `weaponDamageEvent` is not fired too, also we observed that by using `ASSAULTRIFLE` instead of `GUSENBERG` for the repro the repro was less reliable and the issue was happening only some times, but with the `GUSENBERG` weapon the repro is very reliable.

For a more clear repro:
1. Get in front of a player we will call player X.
[...]

PR #1521, opened 2022-07-05
feat(onesync): make low-distance sync delay customizable

0xLuca:
This change allows server owners to adjust the maximum distance in which entities are synced with less delay (which is currently hardcoded at a maximum squared distance of `35.0f * 35.0f`) and to adjust how much more frequently the entity is synced (currently hardcoded to one fourth of the default delay which is 50ms).

It adds two new convars: 

- `onesync_lowDelayDistance`, which controls the maximum distance of an entity to be synced with a lower delay, defaults to the current hardcoded value of `35.0f`
- `onesync_lowDelayDivisor`, which sets the integer by which the default delay (50ms) is divided, defaults to the current hardcoded value of `4`.


blattersturm:
I'm not sure what this is meant to accomplish, or if this has been tested to fix whatever issue - providing configuration options is not usually a good idea, since everything should work by default and users will often cargo-cult config values.

Can you provide more info about the context behind this change?

0xLuca:
> I'm not sure what this is meant to accomplish, or if this has been tested to fix whatever issue - providing configuration options is not usually a good idea, since everything should work by default and users will often cargo-cult config values (see how various settings that never even did anything still spread like a wildfire).
> 
> Can you provide more info about the context behind this change? What is this trying to fix? How did you test this? Why a configuration option and not any other solution?

At first: 
It has been tested on a server with about ~200 players each evening and they gave good feedback.

Background:
My users often complained about the "sync" when having gunfights with each other. They said that headshots often dont connect (for the shooting player it is a headshot, for the shot player it's not) when fighting over a long distance.

I went through this server-codebase and found some interesting things, at first I turned on `onesync_automaticResend` and i got good feedback from the players, that shots are now connecting better (but the hitmarker is showing up a bit later than before when we used the default sync style). But they were still experiencing the issue stated above from time to time.

I then implemented this into our private build, increasing the distance and lowering the delay a bit more than in the default codebase. The players gave very good feedback that every "headshot is now connecting" and that they like the "new sync".

Of course, when lowering the sync delay or increasing the maximum distance too much (in combination with a high player culling radius) the FPS can get pretty low, so i made it configurable so that every owner can decide for himself what he values ​​more.

0xLuca:
> ... are you sure the damage issue here isn't about, say, #1490 or the other similar issue with damage event handling that I noticed there?

I'm pretty sure, because the players reported this for non-automatic guns too (for example the `WEAPON_PISTOL50`). 

But we have some issues with players no receiving correct damage. Both players have a "headshot effects" mod, where particles spawn at a players head when it is hit by a bullet. We have situations where the particles spawn on both sides, but the player does not receive the correct damage (for example he is shot in the head with a `PISTOL50` but only loses 1/4 of his health). 

This especially applies to players being in our "Paintball" routing buckets, in the default bucket (0) we have a very low amount of sync complaints (but i cant explain myself why the sync in, for example, bucket 20001 would be not as good as in the default bucket, maybe you know).

> Similarly, wouldn't it make more sense to have lower rate delays _that could be enabled by default_, say, only for peds that are currently having some combat task?

This could help for active combat situations, but for example, when a players is doing a driveby where the target is just running around without any combat task or weapon in his hand, this would not really help (or am i wrong?).

PR #1524, opened 2022-07-12
tweak(clr): Make Player constructor public

vxlurk:
This would add an ability to create derived classes from Player


blattersturm:
What would the use of 'derived classes' be at all?!

PR #1531, opened 2022-07-16
Allow scripts to access Game pause menu page

VisuallyAppealing:
This lets 0x593feae1f73392d4 return 5 per frame when the Game page is navigated to. I feel like it could be useful for servers, which would like to add items such as Rules, Information about the server etc. Tested it and everything seems to work, but if you find anything, feel free to let me know.


ahcenezdh:
@VisuallyAppealing can you provide a code to test your PR please?

prikolium-cfx:
Thanks for your contribution, but can you please add some examples for us to understand how can it be used by server developers?

VisuallyAppealing:
hi, i'm going to close this because i no longer feel like it would have any use. The concept was that It would be checked with the  CodeWantsScriptToTakeControl native.

PR #1539, opened 2022-07-23
Update graphics extra natives pattern for 1436.31

Disquse:
Should resolve recent regression introduced in #1528.


t3chman:
This does patch the regression

PR #1540, opened 2022-07-24
Improvements to building docs

deepdev237:
I tried building Cfx.re and I wasn't successful at first.
So I thought I would help people who would like to do this :)


hjklwasntavailable:
i mean the current instructions are pretty straightforward and cover everything properly, no?

t3chman:
My only issue that wasn't totally straightforward at first was with using anything greater than Python 3.8. 3.8 works fine though.

deepdev237:
> Never had to used the embedded Python.

Well i didn't understand why would I have a tools folder one the c drive and why would I have python in it?
So I searched the python downloads site and I thought this would work, and it did.
running 'fxd gen -game five' crashed for me, and this fixed it for me.

> i mean the current instructions are pretty straightforward and cover everything properly, no? seems like the only part that needs some updating is the python one

Well I agree and disagree at the same time because I wasn't successful for the first time.
Yes, the python part was major.
The visual studio part i added because that was my issue with building.

hjklwasntavailable:
> Well i didn't understand why would I have a tools folder on the c drive and why would I have python in it?
you can tweak it for your needs, you can also just have python in your global PATH variable

everything you stated seems pretty obvious to me, but maybe that's just me

komokvip:
Your instructions really helped a lot. Unfortunately I get an error after using the command:
"fxd vs -game server"

And getting various errors after compiling in Visual Studio 2022.
If anyone has the patience to try to help me, I know little about this area I would be very grateful! Leave the contact discord :)

PR #1548, opened 2022-07-30
tweak(clr): dont remove eventhandler on exception

vxlurk:
For now, if an exception is thrown inside an event handler callback, this callback is removed the event handlers list.
Im not sure if its a bug or feature, but it forces me (im pretty sure not just me) to create some custom workarounds like this

https://gist.github.com/lurkmorr228/d4c4a2d1cef428adeb9dd9c9597ce2cc


lurkmorr228:
Okay, there is another workaround
https://gist.github.com/lurkmorr228/cd63ad7284e28dbf7931a5926c398de0

PR #1550, opened 2022-07-30
Add b2699 Pause Menu tabs

freedy69:
Going through the new pausemenu.xml, these are the only "interesting" changes I have found.
This commit adds Legal and Credit tabs in the GAME header (You need to toggle contexts for one of these to show, can't have both at the same time.)
Also added the 'Display Text Chat' setting item, however, this would not work in older builds and would just appear as blank (https://i.imgur.com/vIpd6q2.png), so It can only be enabled if the 'b2699' context is enabled (With `PauseMenuActivateContext("b2699")`).


blattersturm:
Interesting idea to use a context here. I believe (if this wasn't the case already) this could actually be added via native code as well just like is done already for gameconfig.xml. 

I'll try to look into this soon.

freedy69:
Yeah i think contexts would be the best option for build-limited pause menu features since the game already uses them for stuff like this, for example checking if the player is in multiplayer, using mission creator etc etc

blattersturm:
The pause menu context change I'll add when pushing the merge commit seems to work:

### 2612
image

### 2699
image

PR #1557, opened 2022-08-04
feat(native-decls): Update GetGamePool.md (add usage example)

4mmonium:
Saw a user on Discord asking how to use `GetGamePool`, so provide a basic usage example.

### DIFF
--- a/ext/native-decls/GetGamePool.md
+++ b/ext/native-decls/GetGamePool.md
@@ -21,6 +21,16 @@ follows:
 * `CVehicle`: Vehicles.
 * `CPickup`: Pickups.
 
+## Examples
+```lua
+local vehiclePool = GetGamePool('CVehicle') -- Get the list of vehicles (entities) from the pool
+for i = 1, #vehiclePool do -- loop through each vehicle (entity)
+    if GetPedInVehicleSeat(vehiclePool[i], -1) == 0 then
+        DeleteEntity(vehiclePool[i]) -- Delete vehicles (entities) that don't have a driver
+    end
+end
+```
+
 ## Parameters
 * **poolName**: The pool name to get a list of entities from.
 


ext/native-decls/GetGamePool.md
```diff
@@ -21,6 +21,16 @@ follows:
 * `CVehicle`: Vehicles.
 * `CPickup`: Pickups.
 
+## Examples
+```lua
+local vehiclePool = GetGamePool('CVehicle') -- Get the list of vehicles (entities) from the pool
+for _, veh in ipairs(vehiclePool) do -- loop through each vehicle (entity)
```
BerkieBb, line 27:
```suggestion
for i = 1, #vehiclePool do -- loop for how many vehicles are found in vehiclePool
```
Isn't this more performant?

thorium-cfx: performance should always be measured :)

AvarianKnight: You can see a bunch of performance tests done [here](https://springrts.com/wiki/Lua_Performance#TEST_9:_for-loops) from forever ago

4mmonium: I overlooked this, thanks for noticing. According to the following sources doing this without `pairs` or `ipairs` is faster yes.

https://medium.com/@sleitnick/battle-of-the-loops-c001bcb4961c
https://stackoverflow.com/a/8955122 (this answer sources information from @AvarianKnight's link it seems)

I will be correcting this PR in a bit.

4mmonium: > Isn't this more performant?

Updated the example code to include your suggestion and squashed my commits, thanks :-)

blattersturm:
Noting this down in case I forget if pointed out in the future but the example of course *doesn't* take ownership into account and that's (kind of) intentional as it's just illustrative.

PR #1559, opened 2022-08-06
feat(gamestate/server): GET_PLAYER_CAMERA_COORDS native

niekschoemaker:


blattersturm:
> _No description provided._

> instant emoji reactions

.. uh, am I missing something here?

code/components/citizen-server-impl/src/state/ServerGameState_Scripting.cpp
```diff
@@ -1372,6 +1372,46 @@ static void Init()
 		return pn ? pn->isInvincible : false;
 	}));
 
+	fx::ScriptEngine::RegisterNativeHandler("GET_PLAYER_CAMERA_COORDS", MakePlayerEntityFunction([](fx::ScriptContext& context, const fx::sync::SyncEntityPtr& entity)
+	{
+		auto camData = entity->syncTree->GetPlayerCamera();
+		scrVector resultVector = { 0 };
+
+		float position[3];
+		entity->syncTree->GetPosition(position);
```
blattersturm, line 1381: This looks like a fair bit of code duplication vs. `GetPlayerFocusPos`.

blattersturm (changes requested):

code/components/citizen-server-impl/src/state/ServerGameState_Scripting.cpp
niekschoemaker: Yes I was aware of this, as I copied the code from there and made the changes needed to have it return a single coord.

I don't exactly know how I could make this function work with that same exact function without breaking any of the other code as the code doesn't work exactly the same, ie. in camMode default or 0 you have one focus position but in the other modes it gets added to the list of player positions.

I'm probably missing something here since I'm not too experienced with C++ at the moment.

niekschoemaker:
I could come up with this, please let me know if this is the correct place to put a function like this, code does work, tested it with FiveM.

Mari037:
Como q entra nos servidores estou perdida

EinS4ckZwiebeln:
Anyone have any idea if this will ever be merged? This seems like a useful addition, imo.

PR #1580, opened 2022-08-17
URI handler implementation for RedM

Bobertsawesome:
This essentially clones the implementation of URI handling for FiveM by creating a new handler `redm://connect/<IP>:<port>`

I have tested this as working by building the cfx.re solution and RedM client.


blattersturm:
Nope, sorry.

PR #1592, opened 2022-08-24
feat(nui): media key controls

TheBanHammer:
Allow iframes to access media keys and make use of extended keyboard features to control existing audio context.

An example of how this would be used is a music player could be controlled via keyboard play/pause/etc buttons. I.e. in-game radio, phone, menu music, etc.

### DIFF
--- a/ext/ui-build/data/root.html
+++ b/ext/ui-build/data/root.html
@@ -80,7 +80,7 @@ registerFrameFunction(function(msg, frameName, frameUrl)
 		const frame = document.createElement('iframe');
 		frame.name = frameName;
 		frame.style.visibility = 'hidden';
-		frame.allow = 'microphone *; camera *;';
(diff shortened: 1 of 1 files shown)


blattersturm:
No. People were actually upset media keys *were* enabled.

Also, `encrypted-media` doesn't actually affect media keys. It doesn't do _anything_ as we're not a Widevine certified app. Given your commit apparently was made via GitHub Web, did you even _test this_?

TheBanHammer:
Main reason for myself is Spotify integration. Their web playback SDK requires it to be enabled in order to be run within the iframe.

blattersturm:
> Main reason for myself is Spotify integration. Their web playback SDK requires it to be enabled in order to be run within the iframe.

Again:

1. Did you even test this?
2. This has nothing to do with 'media keys'.
3. Even if it were enabled it'd not work because we are not a Widevine certified app.

PR #1613, opened 2022-09-08
Update scheduler.lua

Elipse458:
When passing nil as eventData to RemoveEventHandler, scheduler.lua errors out with a nil value error, this should prevent that and trigger the "Invalid event data" error from inside the function instead.


FabianTerhorst:
We modify the commit on merge. Changes itself are looking good.

FabianTerhorst (approved):

FabianTerhorst:
Merged https://github.com/citizenfx/fivem/commit/d1bedcca00fd9ea3eaacd4290455f0f7fb9cfd18

PR #1617, opened 2022-09-08
Add setter and getter for the recoil amplitude of weapons

PichotM:
These natives will be useful to set a dynamic recoil for each weapon whether you need it or not
Tested on 1604 and 2699, it resets to the default recoil value if you reconnect/connect to another server

### DIFF
--- a/code/components/extra-natives-five/src/WeaponExtraNatives.cpp
+++ b/code/components/extra-natives-five/src/WeaponExtraNatives.cpp
@@ -11,6 +11,7 @@
 
 static int WeaponDamageModifierOffset;
 static int WeaponAnimationOverrideOffset;
+static int WeaponRecoilShakeAmplitudeOffset;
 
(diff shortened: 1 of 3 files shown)


PichotM:
RecoilShakeAmplitude is a weapon parameter you can set in the weapon metadata file for all weapons
It is a way to balance the recoil of the weapon
Right now the only way to set the recoil of a weapon is to stream a weapon metadata file.

For my use case, I want a smaller recoil in a gamemode while being able to have a higher one in another gamemode. The streaming solution does not make it possible to have dynamic value so this native is a good solution.

It is not relevant for "anti-cheat" stuff as no recoil does not impact this value, you cannot detect or prevent no recoil by using these natives

--
Both natives work fine in 1604 and 2699

blattersturm:
Just to note - I wouldn't assume this to be related to any cheating stuff, as any cheats would probably implement this in different ways than changing values.

PR #634, opened 2021-02-01
Feat: Enable mumble bans + disable text + admin pass convar

manzarek123:
Add: "mumble_adminPass" convar to be able to become mumble admin and kick/ban users
Disable text channels
Enable bans (until server restart)

PR #1621, opened 2022-09-10
fix(mumble/server): incorrect ban detection

Korioz:
- Fixed false positive ban mumble issue introduced in #634 (`Ban_isBanned` use client->hash which is disabled so our only way is ip banning which supports mask)
- Fixed incorrect ban hash being sent to clients
- Remove junk code forgotten a long time ago
- Added fallback for `MUMBLE_IS_PLAYER_MUTED`


code/components/voip-server-mumble/src/ban.cpp
```diff
@@ -201,8 +201,9 @@ message_t *Ban_getBanList(void)
 	msg = Msg_banList_create(bancount);
 	list_iterate(itr, &banlist) {
 		ban = list_get_entry(itr, ban_t, node);
-		//gmtime_r(&ban->time, &timespec);
+		gmtime_s(&timespec, &ban->time);
```
blattersturm, line 204: Does this exist on Linux as well?

Korioz: Good catch, `gmtime_r` is a POSIX thread safe alternative to `gmtime` and `gmtime_s` is for MSVC, thread-safety doesn't matter that much in this context right?

PR #1623, opened 2022-09-10
fix(native-decls): fix declaration for NETWORK_GET_VOICE_PROXIMITY_OV…

AvarianKnight:
…ERRIDE_FOR_PLAYER

This has been broken since its implementation https://github.com/citizenfx/fivem/commit/8d7ac701d32e31eaca3ef6c3f39cbaa8f1739d35

Using the native as is errors with `WARNING: NON-EXISTENT NATIVE 000000007a6462f4`

I guess no one has tried to use this since it was implemented?


AvarianKnight:
Going to bump this as it was likely lost

4mmonium (approved):
Looks good to me, old name doesn't point to its [native handler](https://github.com/citizenfx/fivem/blob/b5c87bf731c721400b336637b0a886f5be726657/code/components/citizen-server-impl/src/state/ServerGameState_Scripting.cpp#L1185).

Great job spotting this, thanks!

thorium-cfx (approved):

PR #1624, opened 2022-09-11
Allow some FX edits in pure level 1

f25a2fc72690b780b2a14e140ef6a9e0:
My logic might not be the best, please co-author if this is the case. Please also note, I was unable to test due to GitHub failing to clone (so please co-author).

Some players might use these to make their game look better (to them), these changes provide minimal-to-no competitive advantages (if they choose to PvP(vE))

### DIFF
--- a/code/components/glue/src/PurePackfileGTA5.cpp
+++ b/code/components/glue/src/PurePackfileGTA5.cpp
@@ -5799,7 +5799,8 @@ static InitFunction initFunction([]
 		if (fx::client::GetPureLevel() == 1)
 		{
 			std::string_view pathTest{ path };
-			if (pathTest == "citizen/common/data/effects" || // BloodFX, "KillFX" etc.
+			if (pathTest == "common.rpf/data/effects" || // BloodFX etc.
+			    pathTest == "common.rpf/data/timecycle" || // KillFX and weather etc.
 			    pathTest == "x64/audio/sfx/RESIDENT.rpf" ||
 				pathTest == "x64/audio/sfx/WEAPONS_PLAYER.rpf" ||
 				pathTest.find("x64/audio/sfx/STREAMED_VEHICLES") == 0 ||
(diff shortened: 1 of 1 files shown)


Scullyy:
Blood effects and time-cycle modifiers (a.k.a rat packs) is one of the reasons people wanted pure mode because too many people abuse them to give them an advantage in PvP situations. If this ends up being allowed on pure mode servers would be forced to use the strict version making it so people can’t use popular verified graphics mods on the server. I wouldn’t like to see this implemented and if it does I think it should have it’s own option to enable them.

f25a2fc72690b780b2a14e140ef6a9e0:
> Blood effects and time-cycle modifiers (a.k.a rat packs) is one of the reasons people wanted pure mode because too many people abuse them to give them an advantage in PvP situations. If this ends up being allowed on pure mode servers would be forced to use the strict version making it so people can’t use popular verified graphics mods on the server. I wouldn’t like to see this implemented and if it does I think it should have it’s own option to enable them.

I think the main issue was mostly these "see people udnerwater" or "mod packs". People used to be abuse a ton, here's a few:

- Shooting Through Walls (materials.dat - I believe)
- Boosting Vehicles (handling.meta)
- Parachute on Cars (vehicles.meta)
- Fast Ladders (ladders.meta)
- No Fall Damage
- Water Bounce
- Explosive Ammo
+ loads more I could name

These are the **main** reason pure mode exist. I don't understand how BloodFX and Timecycle editing could lead to a competitive advantage (even if so, it'd be very minimal??)

Scullyy:
Regardless if it’s minimal it’s still an advantage which I’ve seen people use and say it makes it so much easier for them to kill people in shootouts and see much clearer at night which is honestly a huge advantage.

userMacieG:
**timecycle_mods_1.xml**: removing tinting from glasses
**timecycle_mods_4.xml**: [xray/glow](https://streamable.com/0rlyof)
**water.xml**: removing water
**time.xml**: only day/night
**weather.xml**: removing snow/rain, only `w_extrasunny`
**vehiclepopulation.xml**: removing traffic/peds
**bloodfx.dat**: [custom body hits](https://www.youtube.com/watch?v=bW9ty68EI60)
**wheelfx.dat**: removing smoke
**tcinstbox_types.meta** + **visualsettings.dat**: removing lights
**explosion.ymt**: removing explosions

I could try to find another 10 more files that can be easily modified to get unfair advantage.
So this commit is a big no, no

PR #1636, opened 2022-09-16
tweak(mumble/server): security improvements

Korioz:
This PR features a bunch of improvements addressing mumble server abuses (unwanted users connecting from the official mumble client trolling legit users).
- Admin password can now be provided without being connected this allows us to disconnect no-admin users not responding to some criteria.
- Disconnect users connecting from the official mumble client without an admin password set.
- Disconnect admins who are providing a server id in username field, this help prevent unwanted behaviour with natives or future implementations.
- Disconnect non-admin users who are providing no server id or a server id already used in username field.

### DIFF
--- a/code/components/voip-server-mumble/src/ServerMumbleVoice.cpp
+++ b/code/components/voip-server-mumble/src/ServerMumbleVoice.cpp
@@ -50,7 +50,9 @@ static InitFunction initFunction([]()
 	{
 		int serverId = context.GetArgument<int>(0);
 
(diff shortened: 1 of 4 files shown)


wdesgardin:
Thanks for working on this. I don't think that will completely prevent someone from logging in, but we are having more and more problems with users who log in on mumble server to troll other people so this kind of improvements are needed quickly

thorium-cfx:
Closing this as of blatterstrum's reply, issue has been backlogged to address it later.

PR #1651, opened 2022-10-02
Open m_bagName to allow change handlers to use it

lfshr:
As it stands there's no friendly way of creating a change handler for a `StateBag` as there's no way to access the bag name.

### DIFF
--- a/code/client/clrcore/StateBag.cs
+++ b/code/client/clrcore/StateBag.cs
@@ -7,11 +7,11 @@ namespace CitizenFX.Core
 
 	public class StateBag
 	{
-		private string m_bagName;
+		public string BagName { get; }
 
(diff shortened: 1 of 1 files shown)


lurkmorr228:
maybe better make the ctor public instead?

lurkmorr228:
maybe better add a method for creating handlers?

blattersturm:
Unsure about this one.

lfshr:
> Unsure about this one.

Any ideas about how this would be resolved then? I haven't done any GTA stuff in a while so I'm vaguely remembering the issue. I couldn't create a change handler for a player's state bag because I couldn't access the player's bag name. You can of course rebuild it manually but that requires the backend knowledge about those keys. Another method to create a change handler could be the solution?

Issue #1270, opened 2022-02-02
GET_GAME_POOL parity on server


roentgenium-cfx:
Having the same native used on both client and server would be neat.

AvarianKnight:
This was added in https://github.com/citizenfx/fivem/commit/1f215f37d9687e94787ca700abca65b8c168ae52

PR #1654, opened 2022-10-07
feat(citizen-server-impl): GET_GAME_POOL parity

Ehbw2:
Adds GET_GAME_POOL native to FXServer as mentioned in https://github.com/citizenfx/fivem/issues/1270. 
I have only been able to test it on FiveM as I do not have RedM currently installed.

### DIFF
--- a/code/components/citizen-server-impl/src/state/ServerGameState_Scripting.cpp
+++ b/code/components/citizen-server-impl/src/state/ServerGameState_Scripting.cpp
@@ -1678,6 +1678,66 @@ static void Init()
 
 		return steeringData ? steeringData->steeringAngle * (180.0f / pi) : 0.0f;
 	}));
+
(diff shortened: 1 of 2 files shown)


code/components/citizen-server-impl/src/state/ServerGameState_Scripting.cpp
```diff
@@ -1678,6 +1678,66 @@ static void Init()
...
+		{
+			switch (entity->type)
+			{
+#if STATE_RDR3
+				case fx::sync::NetObjEntityType::Animal:
+				case fx::sync::NetObjEntityType::Horse:
+#endif
+				case fx::sync::NetObjEntityType::Ped:
+				case fx::sync::NetObjEntityType::Player:
+					if (pool.compare("CPed") == 0)
```
AvarianKnight, line 1705: Wouldn't it be better to put these outside the switch so you're not constantly doing the compare?

AvarianKnight (commented):
You have merge conflicts you need to fix too

PR #1662, opened 2022-10-16
Major bug - Temporary removal request

Scullyy:
When the First Person Vehicle Hood setting is enabled other players can't see or hear your gunshots when shooting from inside a vehicle. This is a major bug and can be exploited heavily in RP and PvP servers. I myself am unsure how to fix such an issue but I am creating this to bring it to the attention of someone who possibly can and to provide a temporary removal of the setting while it is awaiting a fix. I know the removal of settings isn't the best option but this setting isn't really used by anyone other than those who are aware of this exploit and disabling it until a solution is provided I believe is much better than keeping such an exploitable setting in place.

Video from evelyn#1571 in the cfx.re discord: https://drive.google.com/file/d/1tNRpT46-BqIODbEJbaN-OUYbGSQFpeeT/view
evelyn#1571 source: https://discord.com/channels/192358910387159041/561197034938564608/979686764262326322


blattersturm:
... this makes no sense even as a mitigation for this supposed 'major' bug. The same setting can be set from the console and any other place, and how is it a 'major' bug that some sounds don't play when the same happens in base GTA Online as well presumably?

blattersturm:
User posted a followup issue at #1663, see also https://forum.cfx.re/t/confused-on-recent-bug-closed/4929799.

PR #1665, opened 2022-10-24
Update ServerWrappers.cs

cashtwizzy:


PappaNiels:
Euh??? I dont know if you are serious right now since this pr only removes one blank line. A pr for this is a bit optimistic in my book.

PR #1678, opened 2022-10-29
Fix curl command parameters

roboriba:
There is badly formatted curl code in code/vendor/v8-base.lua.

### DIFF
--- a/code/vendor/v8-base.lua
+++ b/code/vendor/v8-base.lua
@@ -55,7 +55,7 @@ return {
 				}
 
 				buildcommands {
-					('curl.exe "-z%s/%s" -L "-o%s/%s" "%s/%s"'):format(v8bindir, v8dll, v8bindir, v8dll, v8url, v8dll),
+					('curl.exe -z "%s/%s" -Lo "%s/%s" "%s/%s"'):format(v8bindir, v8dll, v8bindir, v8dll, v8url, v8dll),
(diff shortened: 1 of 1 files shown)


blattersturm:
... huh? This works fine for basically *every* official build machine, my local workstation, and the few other places that use this command too.

Where did you get your 'curl.exe' and is it a different one from the one bundled with Windows?

thorium-cfx:
From looking at the [manual](https://curl.se/docs/manpage.html) it looks like all options have whitespace surrounding it. I don't see any any example that implies our setup is even supported, i.e.: missing a space on the right and enclosing it within the quotes, if at all it seems using `-` at the start of a path has some different meanings to other options.

blattersturm:
Misclicked the 'reopen' button while trying to select the input field. 

> I don't see any example that implies our setup is even supported

That's indeed a potential reason for the confusion, yet this works with the Windows-bundled curl executable. Similarly, I believe there were some issues with either premake or MSBuild translating quoted paths _that lead to actual files_ incorrectly which likely led to this outcome.

Drwhocake:
@blattersturm I had this same issue as @roboriba, and as you suggested in your first comment to check what curl was being used, I discovered (not with were curl but with the output of fxd natives) that there was a curl included with MSYS2 that doesn't work with the script. No idea if its the same issue roboriba had, but I thought it was worth the share

PR #1680, opened 2022-10-30
fix(formats) five->rdr3 bound material assign

Ktos93:
Fixes the problem of materials not assigning after converting five to rdr3.

### DIFF
--- a/code/components/rage-formats-x/include/convert/phBound_five_rdr3.h
+++ b/code/components/rage-formats-x/include/convert/phBound_five_rdr3.h
@@ -93,17 +93,19 @@ rdr3::phBoundComposite* convert(five::phBoundComposite* bound)
 	out->SetChildMatrices(childCount, &childMatrices[0]);
 
 	// add bound flags
-	std::vector<rdr3::phBoundFlagEntry> boundFlags(childCount);
+	five::phBoundFlagEntry* inBoundFlags = bound->GetBoundFlags();
+
+	std::vector<rdr3::phBoundFlagEntry> outBoundFlags(childCount);
 
 	for (uint16_t i = 0; i < childCount; i++)
 	{
-		boundFlags[i].m_0 = 0x3E;
-		boundFlags[i].pad = 0;
-		boundFlags[i].m_4 = 0x7F3BEC0;
-		boundFlags[i].pad2 = 0;
+		outBoundFlags[i].m_0 = inBoundFlags[i].m_0;
+		outBoundFlags[i].pad = 0;
+		outBoundFlags[i].m_4 = inBoundFlags[i].m_4;
+		outBoundFlags[i].pad2 = 0;
 	}
 
-	out->SetBoundFlags(childCount, &boundFlags[0]);
+	out->SetBoundFlags(childCount, &outBoundFlags[0]);
 
 	return out;
 }
@@ -190,7 +192,9 @@ static inline void fillPolyhedronBound(rdr3::phBoundPolyhedron* out, five::phBou
 		}
 		else if (inPolys[i].type == 4)
 		{
-			memcpy(&outPolys[i].box.indices, &inPolys[i].box.indices, sizeof(inPolys[i].box.indices));
+			outPolys[i].cylinder.index =  inPolys[i].cylinder.index;
+			outPolys[i].cylinder.length = inPolys[i].cylinder.length;
+			outPolys[i].cylinder.indexB = inPolys[i].cylinder.indexB;
 			outPolys[i].type = 4;
 		}
 	}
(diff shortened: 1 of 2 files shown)


code/components/rage-formats-x/include/convert/phBound_five_rdr3.h
```diff
@@ -213,9 +213,200 @@ static inline void fillPolyhedronBound(rdr3::phBoundPolyhedron* out, five::phBou
...
-static inline uint8_t ConvertMaterialIndexrdr3(uint8_t index)
+static inline uint16_t ConvertMaterialIndexrdr3(uint8_t index)
 {
-	return index;
+	static std::map<uint8_t, uint16_t> conversionMap;
+
+	#pragma region material conversion mappings
+	if (conversionMap.empty())
+	{
+		conversionMap[0] = 12667; // DEFAULT
```
Disquse, line 223: Why hardcoded indexes when you can generate those from material name? :(
Especially given that you left material names as comments.

Ktos93: It's not faster than doing this every time when conversion function it's called?

Ktos93: Ahh i see how to do this (:

Disquse: `conversionMap` is static so it would be filled only once.

thorium-cfx (changes requested):

code/components/rage-formats-x/include/convert/phBound_five_rdr3.h
```diff
@@ -213,11 +213,225 @@ static inline void fillPolyhedronBound(rdr3::phBoundPolyhedron* out, five::phBou
 	}
 }
 
-static inline uint8_t ConvertMaterialIndexrdr3(uint8_t index)
+static inline uint16_t GenerateMaterialIndex(char* materialName)
```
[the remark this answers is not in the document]
Ktos93: Pushed some changes. I don't know if that's what you meant. i'm really not a c++ programmer ):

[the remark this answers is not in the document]
thorium-cfx: That's looking good. Well you are on your way and now you got some more powerful knowledge :)

You can remove the `inline` in front of it, nowadays compilers don't really adhere to it, becoming more of a hint to other programmers of your intent. If this wasn't evaluated at runtime and would actually be inlined then it would more likely introduce bloated code then any performance gain, I presume.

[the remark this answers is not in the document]
Ktos93: inline Removed Thanks (:

code/components/rage-formats-x/include/convert/phBound_five_rdr3.h
```diff
@@ -213,11 +213,223 @@ static inline void fillPolyhedronBound(rdr3::phBoundPolyhedron* out, five::phBou
 	}
 }
 
-static inline uint8_t ConvertMaterialIndexrdr3(uint8_t index)
+static constexpr uint16_t GenerateMaterialIndex(const char* materialName)
 {
+	if (!materialName)
```
Disquse, line 218: Add brackets here
```cpp
if (!materialName)
{
  return 12667; // DEFAULT
}
```

code/components/rage-formats-x/include/convert/phBound_five_rdr3.h
```diff
@@ -213,11 +213,223 @@ static inline void fillPolyhedronBound(rdr3::phBoundPolyhedron* out, five::phBou
...
+
+	uint32_t hash = 0;
+	const char* next = materialName;
+
+	for (char i = *materialName; i; i = *next)
+	{
+		++next;
+		hash = i + 16 * hash;
+		if ((hash & 0xF0000000) != 0)
+			hash ^= hash & 0xF0000000 ^ ((hash & 0xF0000000) >> 24);
```
Disquse, line 229: Same here

Disquse (changes requested):

Disquse:
@thorium-cfx 's idea about conversion map sounds good to me.

Disquse (approved):
Tested with few models (with embed collision), LGTM.

thorium-cfx (approved):
Looks good to me too, the magic static looks like a good choice :)
Note that I didn't test the code.

cpmodding:
why is it still waiting to be merged? :P

Ktos93:
I scroll down on phone and clicked close... There should be a closing confirmation git...

blattersturm:
Needs a manual review pass (@Disquse?) and perhaps a commit message tweak.

PR #1690, opened 2022-11-10
fix(extra-natives/five): Crosshair convars init

4mmonium:
Resolve an issue with crosshair convar init state:

Currently, crosshair convars are inited from DoCrosshairDraw's method. This method gets called when the game is done loading, thus overriding any replicated cvars the server is sending via 'setr' unless the client were to rejoin said server without fully closing the client via 'quit'.

To resolve this, we move these convars out of the method.

Looks a bit weird to have all these console variables init'd like this in the global scope, but it resolves the issue. I guess a better solution would be to have these variables declared as a shared pointer and then initializing them in an init method like other things do i.e. `g_use3dAudio` or `g_handbrakeCamConvar`.

If that's a better way to do it, let me know and I'll modify the PR, thanks!

### DIFF
--- a/code/components/extra-natives-five/src/DrawImNatives.cpp
+++ b/code/components/extra-natives-five/src/DrawImNatives.cpp
@@ -3236,29 +3236,29 @@ static InitFunction initFunctionSceneExperiment([]()
 });
 #endif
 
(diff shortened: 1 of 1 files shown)


blattersturm:
I'd rather consider it problematic that the convar system 'eats' the replicated flag and removes the _initial_ value in this case.

PR #1699, opened 2022-11-13
fix(client/console): Handle replicated ConVars

4mmonium:
This is a proposal at resolving an issue with replicated ConVars coming from the server. Right now the `ConsoleVariableManager::Register` method takes whatever is fed by the client (via ConVar definitions) and doesn't check if the ConVar was already present in `m_entries` prior to defining it, therefore 'eating' the value (by deleting any old entry) as well as the flags.

This so called 'fix' currently finds any old entries (those that were sent by the server via msgConVars) and checks if they exist.

If they do exist and are replicated we store the old entry flag and value. We then proceed to remove the old entry (mainly so we can use `SetValue`, which calls `SetRawValue`, which happens to call `GetEntryFlags`, which finds the old replicated ConVar and ends up telling us we can't modify replicated variables) and transfer over the data (flags and value) to the new ConVar that is being created.

I'm not sure if this is the best way to solve this, hence why I'm PR'ing it as a proposal to see if there's perhaps a better way.

Thanks.

### DIFF
--- a/code/client/citicore/console/Console.Variables.cpp
+++ b/code/client/citicore/console/Console.Variables.cpp
@@ -148,6 +148,38 @@ ConsoleVariableManager::THandlerPtr ConsoleVariableManager::FindEntryRaw(const s
 
 int ConsoleVariableManager::Register(const std::string& name, int flags, const THandlerPtr& variable)
 {
+	Entry* oldEntry = nullptr;
+
+	// Preserve the actual variable value from this point, just in case
+	std::string oldVariableValue = variable->GetValue();
+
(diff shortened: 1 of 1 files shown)


blattersturm:
I'd say the value-retaining behavior should _not_ be exclusive to replicated convars, as this same concern can and will apply with e.g. `+set` variables, or `seta` variables from the fivem.cfg file.

prikolium-cfx:
Closing it as stale. Feel free to reopen if you have more details on the issue itself and think that's it's critical.

PR #1715, opened 2022-11-22
fix(steam): Xbox controller usable outside Steam again

thorium-cfx:
As of Steam client update (17 Nov) users were no longer able to start FiveM/RedM and use their Xbox controller without using a Steam shortcut or Big Picture Mode. This will restore it.
* Unhooking Steam XInput hooks.
* Wrapper to easily add additional unhooks in the future.

### DIFF
--- a/code/components/steam/src/SteamLoader.cpp
+++ b/code/components/steam/src/SteamLoader.cpp
@@ -8,6 +8,58 @@
 #include "StdInc.h"
 #include "SteamLoader.h"
 
+#include <CoreConsole.h>
+
+template<int Size>
+struct ScopedHookRestoration
+{
+	FARPROC procAddress;
+	uint8_t asmData[Size];
+
+	ScopedHookRestoration()
+		: procAddress(), asmData()
+	{
+	}
+
+	ScopedHookRestoration(const char* moduleName, const char* method)
+		: procAddress(), asmData()
+	{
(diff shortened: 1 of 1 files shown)


code/components/steam/src/SteamLoader.cpp
```diff
@@ -8,6 +8,52 @@
...
+	}
+
+	ScopedHookRestoration(const char* moduleName, const char* method)
+		: procAddress(), asmData()
+	{
+		HMODULE moduleHandle = GetModuleHandleA(moduleName);
+		if (moduleHandle)
+		{
+			procAddress = ::GetProcAddress(moduleHandle, method);
+			if (procAddress)
```
blattersturm, line 29: braces :(

code/components/steam/src/SteamLoader.cpp
```diff
@@ -8,6 +8,52 @@
...
+				memcpy(asmData, procAddress, Size);
+			else
+				trace("Warning: Cfx/SteamLoader could not find method %s in module %s to restore hooks for.\n", method, moduleName);
```
blattersturm, line 32: maybe console::DPrintf? though that does add a CoreConsole.h include so eh

blattersturm (changes requested):

thorium-cfx:
Made unhook errors use `DPrintf` and added a warning for users in case it fails, so we know where to start looking.

PR #1716, opened 2022-11-22
tweak(clrcore): preparing v1 C# files for v2

thorium-cfx:
* Update game related C# files,
   * v2 changes to use namespace `CitizenFX.FiveM`,
   * Other updates to make these files work for both runtimes without changing V1's API,
   * Moved `PlayerList` to `External` as it was disabled for any non-FiveM product,
   * Removed some unused using directives.
* Some math clean up, increasing performance slightly,
* Added `Game.GenerateHashASCII(string)` which is around 6.5 times faster than `Game.GenerateHash(string)`.

Prerequisite to mono_rt2

### DIFF
--- a/code/client/clrcore/MonoScriptRuntime.cs
+++ b/code/client/clrcore/MonoScriptRuntime.cs
@@ -140,7 +140,9 @@ namespace CitizenFX.Core
 		[SecuritySafeCritical]
 		public int HandlesFile(string filename, IScriptHostWithResourceData metadata)
 		{
-			return (filename.EndsWith(".net.dll") ? 1 : 0);
+			metadata.GetNumResourceMetaData("mono_rt2", out IntPtr enableV2);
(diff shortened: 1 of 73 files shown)


manups4e:
Finally

code/client/clrcore/MonoScriptRuntime.cs
```diff
@@ -140,7 +140,9 @@ public int GetInstanceId()
 		[SecuritySafeCritical]
 		public int HandlesFile(string filename, IScriptHostWithResourceData metadata)
 		{
-			return (filename.EndsWith(".net.dll") ? 1 : 0);
+			metadata.GetNumResourceMetaData("mono_rt2", out IntPtr enableV2);
```
blattersturm, line 143: FWIW eventually I'd like to see this be detected using assembly metadata, somehow.

blattersturm (approved):

PR #1724, opened 2022-11-27
feat(clrcore/external): prop hashes

BerkieBb:
All prop hashes as of 2802


thorium-cfx:
I think it's unwise to add these as-is, as it'll bloat our assemblies without much benefit to our user base. Thank you for for this PR though.

PR #1725, opened 2022-11-27
tweak(clrcore/external): update native names

BerkieBb:
New native names for clrcore


thorium-cfx:
Looks good code-wise, think it'll only need some validation checks of the natives on our side.

PR #1728, opened 2022-11-30
feat(nativeAudio/rdr3): Initial implementation for RedM

nikez:
Tested with two clients on 1311, 1355, 1436, 1491. Tested submixes, environment and proximity. Also tested FiveM and had no issues (due to shared file).

**NB**: We had to add padding to the audDspEffect to avoid buffer corruption. See https://i.imgur.com/C4OXlba.png for details.

Co-authored-by: z3t4s <26795431+z3t4s@users.noreply.github.com>
Co-authored-by: okqut <38298546+okqut@users.noreply.github.com>


blattersturm:
Mainly so I won't forget to mention this later, but the amount of code duplication in NuiAudioSink.cpp is extremely worrying.

The game-specific stuff and generic glue here ought to be split to some extent.

PR #1731, opened 2022-12-02
tweak(server/launcher): POPCNT missing on VMs.

MrSentex:
The detection of the POPCNT instruction may give a false negative on virtualized systems even though the processor has the instruction.


blattersturm:
This would be a misconfiguration of said VM, and is not a scenario we aim to support here.

PR #1736, opened 2022-12-05
feat: add GetEntityFromStateBagName helper

AvarianKnight:
Add a helper function as discussed on the [forums](https://forum.cfx.re/t/allow-statebagchangehandlers-to-be-a-partial-name/4963302/4).

### DIFF
--- a/code/components/citizen-server-impl/src/state/ServerGameState_Scripting.cpp
+++ b/code/components/citizen-server-impl/src/state/ServerGameState_Scripting.cpp
@@ -1678,6 +1678,38 @@ static void Init()
 
 		return steeringData ? steeringData->steeringAngle * (180.0f / pi) : 0.0f;
 	}));
+
+	fx::ScriptEngine::RegisterNativeHandler("GET_ENTITY_FROM_STATE_BAG_NAME", [](fx::ScriptContext& context)
+	{
+		// get the current resource manager
+		auto resourceManager = fx::ResourceManager::GetCurrent();
+
+		// get the owning server instance
+		auto instance = resourceManager->GetComponent<fx::ServerInstanceBaseRef>()->Get();
+
+		// get the server's game state
+		auto gameState = instance->GetComponent<fx::ServerGameState>();
+
+		auto entity = 0;
+
+		// get the state bag name
+		std::string bagName = context.CheckArgument<const char*>(0);
+
+		// we only want to handle conversion for entity
+		if (bagName.find("entity:") == 0)
+		{
+			int parsedEntityId = atoi(bagName.substr(7).c_str());
+
+			auto entityPtr = gameState->GetEntity(0, parsedEntityId);
+
+			if (entityPtr)
+			{
+				entity = gameState->MakeScriptHandle(entityPtr);
+			}
+		}
+
+		context.SetResult(entity);
+	});
 }
 
 static InitFunction initFunction([]()
--- a/code/components/gta-net-five/src/CloneExperiments.cpp
+++ b/code/components/gta-net-five/src/CloneExperiments.cpp
@@ -5041,5 +5041,42 @@ static HookFunction hookFunctionDiag([]
 		PedPoolDiagError();
 	});
 #endif
+
+	fx::ScriptEngine::RegisterNativeHandler("GET_ENTITY_FROM_STATE_BAG_NAME", [](fx::ScriptContext& context)
+	{
+		int entityId = 0;
+		std::string bagName = context.CheckArgument<const char*>(0);
+
+		if (bagName.find("entity:") == 0)
+		{
+			int parsedEntityId = atoi(bagName.substr(7).c_str());
+			rage::netObjectMgr* netObjectMgr = rage::netObjectMgr::GetInstance();
+
+			if (netObjectMgr)
+			{
+				rage::netObject* obj = netObjectMgr->GetNetworkObject(parsedEntityId, true);
+				if (obj)
+				{
+					int guid = getScriptGuidForEntity((fwEntity*)obj->GetGameObject());
+					if (guid)
+					{
+						entityId = guid;
+					}
+				}
+			}
+		}
+		else if (bagName.find("localEntity:") == 0)
+		{
+			int parsedEntityId = atoi(bagName.substr(12).c_str());
+			fwEntity* entity = rage::fwScriptGuid::GetBaseFromGuid(parsedEntityId);
+			// Verify the entity exists before returning the entity id
+			if (entity)
+			{
+				entityId = parsedEntityId;
+			}
+		}
+
+		context.SetResult<int>(entityId);
+	});
 });
 #endif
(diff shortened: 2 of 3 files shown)


code/components/citizen-server-impl/src/state/ServerGameState_Scripting.cpp
```diff
@@ -1677,7 +1677,39 @@ static void Init()
...
+		// get the server's game state
+		auto gameState = instance->GetComponent<fx::ServerGameState>();
+
+		auto entity = 0;
+
+		// get the state bag name
+		std::string bagName = context.CheckArgument<const char*>(0);
+
+		// we only want to handle conversion for entity
+		if (bagName.find("entity:") != std::string::npos)
```
blattersturm, line 1699: `== 0`, one would want to check if it's at the start.

code/components/gta-net-five/src/CloneExperiments.cpp
```diff
@@ -5040,6 +5040,40 @@ static HookFunction hookFunctionDiag([]
 	{
 		PedPoolDiagError();
 	});
-#endif
+#endif
```
blattersturm, line 5043: This part again makes me slightly concerned about Git/editor configuration (autocrlf is true, right?)

code/components/gta-net-five/src/CloneExperiments.cpp
```diff
@@ -5040,6 +5040,40 @@ static HookFunction hookFunctionDiag([]
...
+
+	fx::ScriptEngine::RegisterNativeHandler("GET_ENTITY_FROM_STATE_BAG_NAME", [](fx::ScriptContext& context)
+	{
+		int entityId = 0;
+		std::string bagName = context.CheckArgument<const char*>(0);
+
+
+		if (bagName.find("entity:") != std::string::npos)
```
blattersturm, line 5051: Same `== 0` instead of `!= std::string::npos` as above.

code/components/gta-net-five/src/CloneExperiments.cpp
```diff
@@ -5040,6 +5040,40 @@ static HookFunction hookFunctionDiag([]
...
+		{
+			bagName.replace(0, 7, "");
```
blattersturm, line 5053: Same replace stuff as well.

code/components/gta-net-five/src/CloneExperiments.cpp
```diff
@@ -5040,6 +5040,40 @@ static HookFunction hookFunctionDiag([]
...
+			rage::netObject* obj = rage::netObjectMgr::GetInstance()->GetNetworkObject(atoi(bagName.c_str()), true);
+			if (obj)
+			{
+				int guid = getScriptGuidForEntity((fwEntity*)obj->GetGameObject());
```
blattersturm, line 5057: GetGameObject can be null and this'd potentially crash if that's the case. Needs a null check.

code/components/gta-net-five/src/CloneExperiments.cpp
```diff
@@ -5040,6 +5040,40 @@ static HookFunction hookFunctionDiag([]
...
+				if (guid)
+				{
+					entityId = guid;
+				}
+			}
+		}
+		else if (bagName.find("localEntity:") != std::string::npos)
```
blattersturm, line 5064: See above.

code/components/gta-net-five/src/CloneExperiments.cpp
```diff
@@ -5040,6 +5040,40 @@ static HookFunction hookFunctionDiag([]
...
+		{
+			bagName.replace(0, 12, "");
```
blattersturm, line 5066: Same substr() idea here.

code/components/gta-net-five/src/CloneExperiments.cpp
```diff
@@ -5040,6 +5040,40 @@ static HookFunction hookFunctionDiag([]
...
+			int entityIdTransformed = atoi(bagName.c_str());
```
blattersturm, line 5067: `Transformed` implies it's a variant of the outer scope's `entityId`, maybe `parsedId`?

ext/native-decls/GetEntityFromStateBagName.md
```diff
@@ -0,0 +1,15 @@
+---
+ns: CFX
+apiset: shared 
+---
+## GET_ENTITY_FROM_STATE_BAG_NAME
+
+```c
+Entity GET_ENTITY_FROM_STATE_BAG_NAME(char* bagName);
+```
+
```
blattersturm, line 10: A description could help, perhaps as follows:

> `Returns the entity handle for the specified state bag name. For use with [ADD\_STATE\_BAG\_CHANGE\_HANDLER](?_0x5BA35AAF).`

Adding an example, perhaps, too?

ext/native-decls/GetEntityFromStateBagName.md
```diff
@@ -0,0 +1,15 @@
...
+## Parameters
+* **bagName**: The bag name
```
blattersturm, line 12: `A state bag name.`

ext/native-decls/GetEntityFromStateBagName.md
```diff
@@ -0,0 +1,15 @@
...
+
+## Return value
+The entity handle or 0 if it didn't exist
```
blattersturm, line 15: `An entity handle, or 0 if the state bag name did not refer to an entity, or the entity does not exist.`

code/components/gta-net-five/src/CloneExperiments.cpp
```diff
@@ -5040,6 +5040,40 @@ static HookFunction hookFunctionDiag([]
...
 	});
-#endif
+#endif
+
+	fx::ScriptEngine::RegisterNativeHandler("GET_ENTITY_FROM_STATE_BAG_NAME", [](fx::ScriptContext& context)
+	{
+		int entityId = 0;
+		std::string bagName = context.CheckArgument<const char*>(0);
+
+
```
blattersturm, line 5050: Double newline?

blattersturm (changes requested):
Some stylistic remarks, in part for consistency with code elsewhere, in part because this'd handle invalid input a bit weirdly.

code/components/citizen-server-impl/src/state/ServerGameState_Scripting.cpp
```diff
@@ -124,7 +124,7 @@ static void Init()
 		}
 
 		context.SetResult(gameState->MakeScriptHandle(entity));
-	});
+	});
```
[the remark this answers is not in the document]
AvarianKnight: VS code was complaining about clrf in both of these files I kept ignoring it but somehow this still happened

ext/native-decls/GetEntityFromStateBagName.md
AvarianKnight: Maybe it would make more sense to add an example on AddStateBagChangeHandler and reference it here?

AvarianKnight:
It might also be worth having a `GetPlayerSourceFromStateBagName` or something similar

blattersturm:
> It might also be worth having a `GetPlayerFromStateBagName` or something similar

Perhaps, but that's out of scope here.

blattersturm (approved):

PR #1739, opened 2022-12-06
tweak(ext/cfx-ui): Enable NUI in-process GPU option for RedM

Ktos93:
From my tests, it seems safe to add this option to the redm menu (previously only available as a command). Tested on dx12 and Vulkan on AMD GPU
[nui_useInProcessGpu  true](https://streamable.com/jve44a)
[nui_useInProcessGpu  false](https://streamable.com/tofwp0)


Disquse (approved):

blattersturm:
Commit message mostly.

PR #1744, opened 2022-12-13
Only register public BaseScripts

SteveDee98:
It would be much easier to manually register Scripts when they wouldn't be auto registered if they aren't public types.
Otherwise a custom attribute could be added to exclude classes from auto registration.


blattersturm:
This change would very likely be breaking compatibility with existing scripts. Similarly, I believe making the constructor private already works as a suitable solution for this, though one would have to expose a somewhat unnatural factory function as a replacement.

PR #1748, opened 2022-12-16
feat(devtools): add a server version of neteventlog

AvarianKnight:
This was long a long requested feature and might be useful for optimization, or finding exploits in resources

Tips on how this code might be cleaner would be appreciated, but I'm not sure if there is a way :\

### DIFF
--- a/code/components/citizen-devtools/src/NetEventLog.cpp
+++ b/code/components/citizen-devtools/src/NetEventLog.cpp
@@ -1,11 +1,11 @@
 #include <StdInc.h>
 
-#ifndef IS_FXSERVER
 #include <ResourceEventComponent.h>
 #include <ResourceManager.h>
 
+#ifndef IS_FXSERVER
 #include <NetLibrary.h>
-
+#endif
 #include <boost/circular_buffer.hpp>
 
 #include <ConsoleHost.h>
@@ -18,17 +18,27 @@ struct EventLogEntry
 	std::string eventName;
 	size_t payloadSize;
 	bool out;
+	std::optional<std::string> eventSource;
 };
 
-static boost::circular_buffer<EventLogEntry> g_eventLog(150);
+static boost::circular_buffer<EventLogEntry> g_eventLog(
+	// We want a bigger buffer on the server because it will send/receive *a lot* more events, and 150 will quickly get filled
+#if IS_FXSERVER
+	600
+#else
+	150
+#endif
+);
 static std::mutex g_eventLogMutex;
 
-static void EventLog_Add(const std::string& eventName, size_t payloadSize, bool out)
+static void EventLog_Add(const std::string& eventName, size_t payloadSize, bool out, std::optional<std::string> eventSource = std::nullopt)
 {
 	EventLogEntry eve;
 	eve.eventName = eventName;
 	eve.payloadSize = payloadSize;
 	eve.out = out;
+	// this should always be nullopt on client
+	eve.eventSource = eventSource;
 
 	std::unique_lock<std::mutex> lock(g_eventLogMutex);
 	g_eventLog.push_front(std::move(eve));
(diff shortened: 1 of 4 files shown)


code/components/citizen-resources-core/include/ResourceEventComponent.h
```diff
@@ -130,6 +130,9 @@ class RESOURCES_CORE_EXPORT ResourceEventManagerComponent : public fwRefCountabl
 	//
 	fwEvent<const std::string&, const std::string&, const std::string&> OnQueueEvent;
 
+	// Arguments: eventName, eventPayloadSize, eventSource
+	fwEvent<const std::string_view&, size_t&, const std::optional<std::string_view>&> OnClientEventTriggered;
```
blattersturm, line 134: A little odd as the concept of server->client events only exists on the server and is part of ServerResources - I wonder if there'd be a better place for this callout across components, though.

code/components/citizen-devtools/src/NetEventLog.cpp
```diff
@@ -92,17 +124,30 @@ static InitFunction initFunction([]()
...
+#else
 				EventLog_Add(eventName, eventPayload.size(), false);
+#endif
+				
 			}
 		});
+
+		eventManager->OnClientEventTriggered.Connect([](const std::string_view& eventName, size_t& eventSize, const std::optional<std::string_view>& eventTarget)
+		{
+			EventLog_Add(std::string{ eventName }, eventSize, true, std::string{ eventTarget.value_or("-1") });
```
blattersturm, line 138: Instead of showing the target as a number, maybe showing their name (plus a shortcut to the `citizen:server:gui` player list entry for them?) could work.

(similarly, instead of enabling this for the server in `citizen:devtools`, one could just `files` it in `citizen:server:gui` using the COMPILING_CITIZEN_SERVER_IMPL define, solving the callout issue as well)

code/components/citizen-devtools/src/NetEventLog.cpp
```diff
@@ -1,11 +1,11 @@
 #include <StdInc.h>
 
-#ifndef IS_FXSERVER
 #include <ResourceEventComponent.h>
 #include <ResourceManager.h>
 
+#ifndef IS_FXSERVER
```
blattersturm, line 6: For moving this to citizen:server:gui, instead of this, the global define guard would instead be like `#if !defined(IS_FXSERVER) || defined(COMPILING_CITIZEN_SERVER_GUI)`.

code/components/citizen-devtools/src/NetEventLog.cpp
```diff
@@ -55,24 +65,46 @@ static InitFunction initFunction([]()
...
 
 				std::unique_lock<std::mutex> lock(g_eventLogMutex);
 
 				for (auto& entry : g_eventLog)
 				{
-					ImGui::Text((entry.out) ? "C->S" : "S->C");
+
+					ImGui::Text(
+#ifndef IS_FXSERVER
+						(entry.out) ? "C->S" : "S->C"
```
blattersturm, line 96: One could rename `out` to `toServer` and flip it elsewhere.

blattersturm (commented):
Some suggestions.

code/components/citizen-resources-core/include/ResourceEventComponent.h
AvarianKnight: [ServerEventComponent](https://github.com/citizenfx/fivem/blob/65bf224097b1107c10f84f5dfc25ee8e4bddc95d/code/components/citizen-server-impl/include/ServerEventComponent.h#L14) might be a better place to put something like this and just change the initialization to `OnServerCreate`

AvarianKnight:
So after some messing around, I'm not sure if this is really a viable solution at all, on higher population servers this will be relatively useless as there's anywhere from 20-100 events a second.

I wonder if a file based solution would work better and have some external tool to go through the data, I'm not sure where would be a good place to discuss possible implementations though.

AvarianKnight:
Closing as I have no intentions of continuing this PR, if someone wants to continue it feel free to.

PR #1751, opened 2022-12-17
Fix runtime streaming native issue with referenceHash

PichotM:
Fix issue reported here https://forum.cfx.re/t/registerstreamingfilefromurl-not-caching-entries-properly/2697294


blattersturm:
Is this closed because this implementation didn't work, or?

PichotM:
> Is this closed because this implementation didn't work, or?

I'm not familiar with the streaming system and I've discovered that it does not work if two files are the same (for example +hi models/textures)

PichotM:
Any idea how to fix it?

PichotM:
is there anything blocking this PR?

blattersturm:
Mainly, checking if this logic makes sense - using the _header_ as hash seems incorrect as two different RSC files with the same sizes/type would hash as the same in this case. The old ResourceCacheDevice (v1) I believe used the URL as seed, not the pre-read RSC header.

PR #1752, opened 2022-12-18
feat(server/impl): GET_PLAYER_MATCHED_IDENTIFIER

BerkieBb:
A native to get a player identifier by matching a substring to the identifier. This is useful for getting a specific license when you don't know the index of it to use GET_PLAYER_IDENTIFIER and don't want to iterate over GET_PLAYER_IDENTIFIERS to find it.

I made this because I got tired of having to write this over and over again everywhere.


blattersturm:
Instead of some vague concept like 'substrings', why not have some sort of GET_PLAYER_IDENTIFIER_BY_TYPE like the C# wrapper for this use case?

BerkieBb:
> Instead of some vague concept like 'substrings', why not have some sort of GET_PLAYER_IDENTIFIER_BY_TYPE like the C# wrapper for this use case?

Would you say I hardcode the types in it then or just keep the find function to match the type?

BerkieBb:
Also, would you know the reason for why I need to index it directly from the vector instead of doing like `identifiers[i].c_str()`, as that just returns loads of question marks (which VS pointed out is a pointer to the memory).

neptunium-cfx:
Try getting reference to that vector instead of copying it - likely gets deallocated as it's a stack variable, i might be wrong though as extra tired.
```
const std::vector<std::string>& identifiers = client->GetIdentifiers();
```
Not sure why you decided to copy it anyway?

BerkieBb:
> Try getting a reference to that vector instead of copying it - likely gets deallocated as it's a stack variable, i might be wrong though as extra tired.
> 
> ```
> const std::vector<std::string>& identifiers = client->GetIdentifiers();
> ```
> 
> Not sure why you decided to copy it anyway?

I don't code in C++ daily, I code in lua mostly, I sometimes forget references. But I'll try it when I'm back home then

BerkieBb:
I didn't think that'd happen when I renamed it lol

BerkieBb:
I'll restore the branch instead of creating a new PR

BerkieBb:
But changes have been applied, native has been renamed to GET_PLAYER_IDENTIFIER_BY_TYPE and I used a reference instead of a copy successfully and it works by indexing the reference now

BerkieBb:
Forgot to change the docs' parameters

PR #1753, opened 2022-12-19
feat(scripting-mono-v2): all source files for v2

thorium-cfx:
Includes all files required for the mono v2 runtime, without build tools/steps and/or enabling it

* All *.h and *.cpp files
* All *.cs files
* Direct native invocation for server


thorium-cfx:
some txAdmin folder change got in there somehow, removing that one

lurkmorr228:
any eta on that one?

thorium-cfx:
Another prerequisite project first needs to be completed and implemented, which is almost ready to go.

PR #1766, opened 2022-12-30
feat(scripting/state): add GET_PLAYER_SERVER_ID_FROM_PED

AvarianKnight:
This was a native I remember complaining about not existing a while ago and makes working with certain checkable events easier to work with (i.e. weaponDamageEvent and the sort)

This is meant to function somewhat like `NetworkGetPlayerIndexFromPed` does on the client

### DIFF
--- a/code/components/citizen-server-impl/src/state/ServerGameState_Scripting.cpp
+++ b/code/components/citizen-server-impl/src/state/ServerGameState_Scripting.cpp
@@ -1748,6 +1748,18 @@ static void Init()
 
 		context.SetResult(player);
 	});
+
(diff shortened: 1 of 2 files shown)


PichotM:
Technically it is what `NETWORK_GET_ENTITY_OWNER` does
maybe you could group both natives?

AvarianKnight:
For such a simple thing I don't really feel like its really worth doing, you save a negligible amount of space

blattersturm:
> This is meant to function somewhat like NetworkGetPlayerIndexFromPed does on the client

What is the difference - why only 'somewhat'? Why does it need to have its own name instead of using the same name as the game function?

AvarianKnight:
I thought that the native referring to `PlayerIndex` instead of `Player` would be confusing for a server side native, but if wanted I can rename it.

PR #1768, opened 2022-12-31
tweak(onesync): Change f164 to weaponHash for explosionEvent

niekschoemaker:
Not sure if this is wanted for backwards compatibility sake, but does make code a lot more readable.

Had a log of a lot of explosionEvents happening, and for the ones I could find (mostly fire hydrants, vehicles and gas canisters) this value was always `WEAPON_EXPLOSION`

### DIFF
--- a/code/components/citizen-server-impl/src/state/ServerGameState.cpp
+++ b/code/components/citizen-server-impl/src/state/ServerGameState.cpp
@@ -4356,7 +4356,7 @@ struct CExplosionEvent
 	bool f190;
 	bool f191;
 
-	uint32_t f164;
+	uint32_t weaponHash;
 	
 	float posX224;
 	float posY224;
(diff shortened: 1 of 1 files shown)


AvarianKnight:
You should keep the old parameter for backwards compatibility but add new ones with the proper names, you can see [an old pr](https://github.com/citizenfx/fivem/pull/785/files) for reference

code/components/citizen-server-impl/src/state/ServerGameState.cpp
```diff
@@ -4356,7 +4356,7 @@ struct CExplosionEvent
 	bool f190;
 	bool f191;
 
-	uint32_t f164;
+	uint32_t weaponHash;
```
blattersturm, line 4359: See @AvarianKnight's comment above.

blattersturm (changes requested):

prikolium-cfx:
Closed as stale.

PR #1769, opened 2022-12-31
Fixed vector length calculation in MathUtils.DirectionToRotation

lfshr:
Slight error in the calculation. This changed in November from newing up a Vector3 and using Vector.Length, which used to give the correct results.


thorium-cfx (approved):
Requested changes are correct

thorium-cfx:
Can you change your commit title to `fix(clrcore): <your fix message>` then it'll be ready to be pulled

lfshr:
> Can you change your commit title to `fix(clrcore): <your fix message>` then it'll be ready to be pulled

Done

thorium-cfx:
Nice, thank you 👍

PR #1773, opened 2023-01-02
(gta-core): add fwAssetStores pools to pool Management

0x-FADED:
this PR tries to add fwAssetStore pools(ScriptStore, MetaDataStore, AnimStore, TextStore, TxdStore etc) to the pool management this is untested in games.


blattersturm:
> this is untested in games.

uh, what do you mean by that?

PR #1777, opened 2023-01-04
Block to rewrite natives

wmade:
The cheats register a second time the natives to be able to falsify the results So I made sure that if it's declared once you can't declare it a second time


neptunium-cfx:
It's pointless, as cheats can 'bypass' this easily anyway.

wmade:
Okay, but how block the rewrite native?

neptunium-cfx:
Ideally send the cheat to abuse@fivem.net and we will take care of that :)

wmade:
ok, so I close the request? The most famous cheats are still functional. Like Eulen or redEngine.

blattersturm:
Sigh, another weird misguided user...

PR #1782, opened 2023-01-05
feat(rdr3): enable PRI_OPTIONAL_LOW props

0x-FADED:
in RDR3 props with priority level PRI_OPTIONAL_LOW props are not loaded by default, just like it was in GTA5 this PR fixes that.

in both GTA5 and RDR2 `rage::fwMapData::ms_entityLevelCap` is `PRI_OPTIONAL_LOW ` by default. but in `CGameWorld::Init` `rage::fwMapData::ms_entityLevelCap` is set to `PRI_OPTIONAL_MEDIUM ` you guys have set it to `PRI_OPTIONAL_LOW ` in GTA5. I did the same in RDR2. here is a comparison:
b5pEb1672829306

### DIFF
--- a/code/components/gta-streaming-rdr3/src/MiscStuff.cpp
+++ b/code/components/gta-streaming-rdr3/src/MiscStuff.cpp
@@ -1,0 +1,25 @@
+#include <StdInc.h>
+#include <Hooking.h>
+#include <MinHook.h>
+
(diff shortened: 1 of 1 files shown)


blattersturm:
Is this change no longer desired, or what is the reason for closing it?

PR #1786, opened 2023-01-07
fix(server/impl): Use rfind for GET_PLAYER_IDENTIFIER_BY_TYPE

niekschoemaker:
This solution seems a bit more correct, this searches only the start of the string, this shouldn't actually be a problem for now, but could avoid future issues in the case an identifier name would occur in the identifier itself.

This is also a bit more performant than the other solution.


blattersturm:
... oh no, the `== 0` suggestion apparently also hadn't shown up. Heh.

SuperCoolNinja:
finally added ..

PR #1788, opened 2023-01-07
fix(ext/cfx-ui): adaptive card inline action style fix

1A3Dev:
This fixes inline actions on adaptive card inputs showing as a white box.

Before:

After:


blattersturm:
@nihonium-cfx is this an OK change, and if so, can you do this with a proper commit message?

1A3Dev:
Corrected the commit message.

nihonium-cfx (approved):

1A3Dev:
This seems to've already been put in so I guess I'll close this 🤔 https://github.com/citizenfx/fivem/commit/610b56bb450d010af6ab2d350db859d0919e3bf5

PR #1792, opened 2023-01-10
Update SetWeaponsNoAutoSwap.md

orel56000:
Added informtaion about a r* behavior bug

### DIFF
--- a/ext/native-decls/SetWeaponsNoAutoSwap.md
+++ b/ext/native-decls/SetWeaponsNoAutoSwap.md
@@ -12,3 +12,7 @@ Disables autoswapping to another weapon when the current weapon runs out of ammo
 
 ## Parameters
 * **state**: On/Off
+
+```
+This native won't have any affect if setting flag 48 to true, using SetPedConfigFlag native.
+```


ext/native-decls/SetWeaponsNoAutoSwap.md
```diff
@@ -11,4 +11,8 @@ void SET_WEAPONS_NO_AUTOSWAP(BOOL state);
 Disables autoswapping to another weapon when the current weapon runs out of ammo.
 
 ## Parameters
-* **state**: On/Off
\ No newline at end of file
+* **state**: On/Off
+
+```
+This native won't have any affect if setting flag 48 to true, using SetPedConfigFlag native.
```
blattersturm, line 17: Adding descriptions below 'Parameters' won't show at all. Also, why is this in code blocks?

blattersturm (changes requested):
Style issue, also the commit message is ~wrongly formatted.

ahcenezdh:
"This native will not have any effect if the flag 48 is set to true when using the SetPedConfigFlag native."

4mmonium (changes requested):
Hi, @orel56000 check the comments by blattersturm. Things such as descriptions cannot be under the `# Parameters` section, they should be above it, native description shouldn't be in code blocks as well, that's reserved for native db import descriptions.

I will un-assign myself since I don't have much else to add here.

gottfriedleibniz:
Closing as stale. Please reopen if you would continue this work.

PR #1841, opened 2023-01-30
fix(streaming/five): Prevent drawable texture quality degradation due to skin tone blending

z3t4s:
### This fixes an issue mentioned in the forums before:
https://forum.cfx.re/t/disable-r-drawables-texture-compression/4852361
https://forum.cfx.re/t/mipmap-reduction/4893456

### Result after applying the fix:
https://imgur.com/a/X38gRgn

### Tested on:
Every version from 1604-2802
Used the resource linked in the first forum post
AI Peds + CL1 & CL2
Test script: https://pastebin.com/05TmzWqK

### Impact:
Creates 320 2048x2048 empty BC1 textures as a replacement for mp_headtargets/XXXX_diff_nnn_a_whi.ytd
Could not measure an impact on FPS when rendering 70 freemode peds.
Runtime should not be an issue at all because mip levels are still created starting from the increased scratch texture size and the engine does very efficient caching of the final ped texture composition.

### How to test:
Streaming custom clothing without svadhesive/adhesive doesn't seem to be an option.
I tested it by replacing base game drawables and textures with the example from the forums and re-encrypting the RPF

### DIFF
--- a/code/components/gta-streaming-five/src/PatchMeshBlending.cpp
+++ b/code/components/gta-streaming-five/src/PatchMeshBlending.cpp
@@ -1,0 +1,520 @@
+#include <StdInc.h>
+#include <Hooking.h>
+#include <MinHook.h>
+#include <jitasm.h>
+#include <CoreConsole.h>
+
+#define RAGE_FORMATS_GAME_FIVE
+#define RAGE_FORMATS_IN_GAME
+#include <gtaDrawable.h>
+#include <grcTexture.h>
+#include <Streaming.h>
+
+uint32_t g_textureDimensionOverride = -1;
+constexpr uint32_t g_textureDimensionDefault = 512;
+constexpr uint32_t g_meshBlendingSlots = 80; // All known game version have 80 scratch textures
+
+struct MeshBlendReplacements
+{
+	int32_t originalId;
+	int32_t replacementId;
+};
+std::array<MeshBlendReplacements, g_meshBlendingSlots * 3> g_meshBlendReplacements;
+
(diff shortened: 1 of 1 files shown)


blattersturm:
Code looks OK at first glance, will review/merge later. CI build failure is as apparently the VM orchestration logic is odd again.

One extra note for future reference:

> Streaming custom clothing without svadhesive/adhesive doesn't seem to be an option.

Debug builds don't need subdir_file_mapping, but you can also manually comment out the check:

https://github.com/citizenfx/fivem/blob/5fac854ad78c01b13fdcf240554c85b529f7adf4/code/components/gta-streaming-five/src/LoadStreamingFile.cpp#L1716-L1717

Also, using `RPF_FILE`-type data entries in a resource should work fine.

Disquse:
> Debug builds don't need subdir_file_mapping, but you can also manually comment out the check:

I guess `Instance<ICoreGameInit>::Get()->GetData("policy", &policyVal)` is fails on non-adhesive builds. I also had this issue before, moving `#ifndef` before this check makes it work as it was intended.

blattersturm:
A quick additional note:

> Creates 320 2048x2048 empty BC1 textures

Doing this unconditionally would lead to a 640 MB GPU memory allocation vs. 40 MB at 512x512. 600 MB of extra memory usage can hit hard given 2/4 GB GPUs still exist and even 8-12 GB GPUs are often already starved for memory.

Perhaps these could only be resized the first time something 'big' is placed in such a slot?

Issue #1697, opened 2022-11-13
 RegisterPedheadshotTransparent opacity issue


cqmpact:
This seems like a issue [that was ignored on the forums](https://forum.cfx.re/t/registerpedheadshottransparent-opacity-build-2189) and nobody reported it here.

**Issue:** 
RegisterPedheadshotTransparent returns a headshot on which the ped is transparent too. 
See attached code and images.

```
local function getPlayerHeadshot()
    local playerPed = PlayerPedId()
    local handle = RegisterPedheadshotTransparent(playerPed)
    if handle == nil or handle == 0 then handle = RegisterPedheadshot(playerPed) print(handle, 'normal') end
    while not IsPedheadshotReady(handle) or not IsPedheadshotValid(handle) do
        Wait(0)
    end
    local txd = GetPedheadshotTxdString(handle)
    UnregisterPedheadshot(handle)
    return txd
end
```

![FiveM_b2699_GTAProcess_6B7VELoPSN](https://user-images.githubusercontent.com/35305953/201528195-97ffe85d-f372-405f-be46-b502b6967414.png)
![FiveM_b2699_GTAProcess_brsHTxagwe](https://user-images.githubusercontent.com/35305953/201528297-1293b5d0-5e3e-4541-a5d6-9f3d1ae4e8c4.png)

thorium-cfx:
I've been investigating this one for a bit and it seems GTA is storing them differently now. I do want to understand this a bit more before I work on a proper solution.

My findings until now:
1. Texture compressions are still the same, i.e.: BC3,
2. Alpha channel is scaled to a different range [0..1] to [0.333..0.666] (rough estimates), no clue why,
3. RGBs look premultiplied by the (original?) alpha.

gottfriedleibniz:
Unsure if you've made progress on this, so I took a glance at things.

Comparing the same alpha block of 1604 and 2189 generated textures (everything is scaled off min/max when compressed): 1604 produced `62-255` and 2189 produced `37-153`. Given that nothing has changed in the related `PedHeadshotTextureExporter` code-paths, I suspect there is either some change in a target blend description (I guess? unlikely) or a blit shader used.

Looking through some of the intermediate render targets, there seems to be a discrepancy in `Headshot Color 1L` (background colouring). This target uses `PS_BlitAAAlpha` (0x14064C771) which changed in 2189 (`mad + add` vs `mul + mad + mad`). Maybe this is causing the issue?

[...]

PR #1876, opened 2023-02-26
fix(extra-natives-five): translucency issue with ped headshots

thorium-cfx:
Compensate for changes that were made to a blit shader since b2189, which was mapping colors into the [0 .. 0.6] range. This rendered textures made by `RegisterPedheadshotTransparent` half-ish translucent and unwanted.

Tested on builds: 2189, 2372, 2545, 2612, 2699, 2802.

resolves: #1697

### DIFF
--- a/code/components/extra-natives-five/src/NuiTxdSource.cpp
+++ b/code/components/extra-natives-five/src/NuiTxdSource.cpp
@@ -21,6 +21,10 @@
 
 #include <DirectXTex/DirectXTex.h>
 
+#include <CrossBuildRuntime.h>
+#include <Hooking.h>
+#include <Hooking.Stubs.h>
+
 namespace WRL = Microsoft::WRL;
 
 class NuiTxdResourceHandler;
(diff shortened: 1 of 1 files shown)


blattersturm:
> This rendered textures .. unwanted

unwanted textures :(

code/components/extra-natives-five/src/NuiTxdSource.cpp
```diff
@@ -312,11 +316,28 @@ static InitFunction initFunction([]()
...
 	OnGrcCreateDevice.Connect([]()
 	{
 		nui::RegisterSchemeHandlerFactory("http", "nui-img", Instance<NUISchemeHandlerFactory>::Get());
 		nui::RegisterSchemeHandlerFactory("https", "nui-img", Instance<NUISchemeHandlerFactory>::Get());
 	});
+
+	if (xbr::IsGameBuildOrGreater<2189>())
+	{
+		void* ptr = hook::get_pattern("48 8D 4D F7 0F 29 4D F7 0F 29 45 E7", 0xC);
+		MH_CreateHook(ptr, AlterPedHeadshotTransparentPSConstants_Hook, (void**)&g_origAlterPedHeadshotTransparentPSConstants_Hook);
```
blattersturm, line 340: Instead of directly using MinHook I think new code could be better off using 'hook::trampoline' that was added recently, also to avoid some of the initialization order stuff.

code/components/extra-natives-five/src/NuiTxdSource.cpp
```diff
@@ -312,11 +316,28 @@ static InitFunction initFunction([]()
 	}, -500);
 });
 
+__int64(__fastcall* g_origAlterPedHeadshotTransparentPSConstants_Hook)(__m128*, __m128*);
+
+// Since b2189's shader changes `RegisterPedheadshotTransparent`'s texture colors are in the [0 .. 0.6] range, let's inverse it
+// Sets	PS constant buffer 3rd and 4th? (unchecked) float4/vec4/__m128 values, where the 3rd is referred to as `GeneralParams0`
+__int64 __fastcall AlterPedHeadshotTransparentPSConstants_Hook(__m128* a3, __m128* a4)
```
blattersturm, line 323: __fastcall is a little redundant as that's the default on amd64 abi

thorium-cfx: I figured I follow suit with the other examples in the repo, want it removed?

blattersturm: It could stay in order to mark it as 'yeah this definition is from IDA' as the `__int64` looks equally weird like that.

PR #1888, opened 2023-03-04
tweak(native-decls): order entity types

AvarianKnight:
2, 1, 3 instead of 1, 2, 3 is slightly confusing.


AvarianKnight:
I guess GitHubs own web editor doesn't respect line endings 🤣

PR #1892, opened 2023-03-08
Enable multiple instances for featured server (V1)

rky5mart:
This PR is an alternative to #1893.

The focus of this PR is Mega's Games featured server present on the bottom left of the front page.
However, it is made in a generic way which could easily work for any other server.

It enables players to choose which instance they would like to connect to by replacing the connect button with a dropdown.
It's backwards compatible and, if there is just one instance, it shows the usual connect button.

The remaining information (boost power, player count, max players) is summed between all the instances.

The dropdown shows the value of a new variable set by the server using `sets instance_name "US"`, if this variable does not exist it shows the server's locale country.
**The first server read from _https://servers-frontend.fivem.net/api/servers/streamRedir_ needs to return the new variable - is this possible?**

The file https://runtime.fivem.net/pins.json indicates the featured server using the property "noAdServerId".
The new implementation tries to split this value using a comma - the first value is the main server and the remaining are the instances. It's backwards compatible and it can contain only one value just like it does currently.
**The file needs to be updated to include both instances of Mega's Games. So the value should be "45yo89,moyejq".**

Here's a preview of the changes:
preview gif


Keyinator:
I think this is a really valuable feature (also for non-featured servers).

I personally still continue to have struggles with dual-demographic server setup. My national server is at 120+ players with international players and the international server is at constant 0 (occasionally 3 people who want to play alone).

Maybe this could also be expanded upon (with ``instance_name``) to allow other segregations like just multiple instances. Benefits of this would include shared load and better performance since lots of work is exponential and all of the natives have to run on the main-thread leading to some restrictions for nowadays cpu capabilities.

blattersturm:
A proper solution for this (for non-featured servers) is planned for when the server listing will get managed by a reworked version of the current 'keymaster' frontend, sometime later this year.

blattersturm:
This is also implemented to some extent by https://github.com/citizenfx/fivem/commit/014db1ea31b852f27050fee3506d61a0d42e5572.

PR #1893, opened 2023-03-08
Enable multiple instances for featured server (V2)

rky5mart:
This PR is an alternative to #1892.

The focus of this PR is Mega's Games featured server present on the bottom left of the front page.

It enables players to choose which instance they would like to connect to by duplicating the information/connect button for each instance.
If there is just one instance, it shows the usual information and connect button.

If there are multiple instances the connect button shows a suffix which is the value of a new variable set by the server using `sets instance_name "US"`, if this variable does not exist it shows the server's locale country.
**The first server read from _https://servers-frontend.fivem.net/api/servers/streamRedir_ needs to return the new variable - is this possible?**

The file https://runtime.fivem.net/pins.json indicates the featured server using the property "noAdServerId".
The new implementation tries to read a new property "noAdServer2Id" which indicates the new server.
**The file needs to be updated to include the new property "noAdServer2Id" with the value "moyejq".**

Here's a preview of the changes:
preview gif


blattersturm:
A similar solution is applied in https://github.com/citizenfx/fivem/commit/014db1ea31b852f27050fee3506d61a0d42e5572.

PR #1895, opened 2023-03-10
tweak(scripting/v8): improve exports typings

pedr0fontoura:
This change allows us to use the advantage of TypeScript's interface extending to add better typings for known exports in a project.

As the interface will be available globally, I called it `CitizenExports` to prevent conflict with other packages.

For example, you could do the following in a TypeScript module file to extend the exports interface:

```ts
declare global {
  interface CitizenExports {
    coolResource: {
      doSomething(parameter: string): void;
    }
  }
}
```

Then in another file, you would get the auto complete:

```ts
global.exports.coolResource.doSomething('Snaily');
```

### DIFF
--- a/data/shared/citizen/scripting/v8/index.d.ts
+++ b/data/shared/citizen/scripting/v8/index.d.ts
@@ -128,7 +128,14 @@ declare var GlobalState : StateBagInterface
 declare function Player(entity: number|string): EntityInterface
 declare var LocalPlayer : EntityInterface
 
-declare var exports: any;
+interface CitizenExports {
+    (exportKey: string | number, exportFunction: Function): void;
+    [resourceName: string] : {
+        [exportKey: string | number]: Function
+    };
+}
+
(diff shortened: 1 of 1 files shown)


lurkmorr228:
the typings seem a bit weird 
why not 

interface CitizenExports {
    (exportName: string, exportedFunction: Function): void;
    [resource: string] : Function;
}

AvarianKnight:
> the typings seem a bit weird why not
> 
> ```
> interface CitizenExports {
>     (exportName: string, exportedFunction: Function): void;
>     [resource: string] : Function;
> }
> ```

This would make more sense, probably better to use `exportFunction` instead though.

pedr0fontoura:
Numbers are also valid export keys/names. I wouldn't recommend using it, but making it string only would be opinionating the code and I don't know if that's intended. I'll update the names of the parameters and fix a mistake I spotted.

PR #1896, opened 2023-03-11
feat(gameState): Add manual players culling native

Identity-labs:


blattersturm:
We generally do not accept big PRs like this without discussing them beforehand. Especially given the 'No description provided', I'm not sure what you are expecting.

Identity-labs:
@prikolium-cfx can we talk about this POC, if it's used we can do manual culling to improve what players we can see.
On redm it will be game changer because actual culling display random player in scope if we are above 32 players.

PR #1903, opened 2023-03-16
feat(extra-natives/five): runtime water manipulation

nikez:
Thanks to @Disquse for the initial implementation.

Tested on:
1604-2802
Debug & Release
DX10, DX11


nikez:
Added a missing lock in the latest force push (`RESET_WATER`)

PR #1906, opened 2023-03-17
GNU Affero LICENSE for FiveM

RyoTabuki:


PappaNiels:
You know that there is a license already, right? [License](https://github.com/citizenfx/fivem/blob/master/code/LICENSE)

blattersturm:
... what?

PR #1909, opened 2023-03-17
Implement BREAK_OFF_VEHICLE_WHEEL for FiveM

Disquse:


ext/native-decls/BreakOffVehicleWheel.md
```diff
@@ -0,0 +1,31 @@
...
+---
+## BREAK_OFF_VEHICLE_WHEEL
+
+```c
+void BREAK_OFF_VEHICLE_WHEEL(Vehicle vehicle, int wheelIndex, BOOL leaveDebrisTrail, BOOL deleteWheel, BOOL unknownFlag, BOOL putOnFire);
+```
+
+Break off vehicle wheel by index. The `leaveDebrisTrail` flag requires `putOnFire` to be true.
+
+## Examples
```
AvarianKnight, line 13: Does it matter if the example is before or after paramaters or does hugo automatically order this

Disquse: Hmmm, good note. I used my old `GET_INTERIOR_ROOM_NAME` native as a reference, but it seems on native reference the `Parameters` block is not visible. I guess the order is important indeed. Would also need to fix old natives then.

PR #1915, opened 2023-03-20
Various network check bypass variables for FiveM

Disquse:
Added `game_enablePlayerRagdollOnCollision` allowing to activate native ragdoll on collision game logic.
Added `game_enablePlayerJumpRagdollControl` allowing to activate ragdoll on jump like in SP (Jump+R).
Moved `game_enableFlyThroughWindscreen` into a separate file.

https://user-images.githubusercontent.com/10367215/226429970-1574d853-4208-4b66-879a-3a1027776b01.mp4

### DIFF
--- a/code/components/extra-natives-five/src/PatchNetworkChecks.cpp
+++ b/code/components/extra-natives-five/src/PatchNetworkChecks.cpp
@@ -1,0 +1,66 @@
+/*
+* This file is part of the CitizenFX project - http://citizen.re/
+*
+* See LICENSE and MENTIONS in the root of the source tree for information
+* regarding licensing.
+*/
+
+#include "StdInc.h"
+#include <Hooking.h>
+#include <nutsnbolts.h>
+#include <GameInit.h>
+#include <CoreConsole.h>
+
+static bool* g_flyThroughWindscreenDisabled;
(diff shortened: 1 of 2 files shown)


gottfriedleibniz:
Declaring a ConVar as replicated from the clients end may run into issues after [disconnect](https://github.com/citizenfx/fivem/blob/master/code/components/glue/src/BindNetLibrary.cpp#L115).

Disquse:
Not sure if this is the best way to clear variables, but works as expected in-game.

gottfriedleibniz:
Join your local server with all three of those variables set, disconnect to clear `ConVar_Replicated`, and now those variables are freely usable on servers that do not use/recognize these options. 

While the current uses are fairly innocuous, that interaction is busted. Presuming the intent is that these variables are server authoritative, why not just have them be controlled via native?

Unrelated...

In the recent BreakOffVehicleWheel native, `unknownFlag` looks like `tyreDeformSwitchOn` (requires verification).

blattersturm:
> Presuming the intent is that these variables are server authoritative, why not just have them be controlled via native?

I believe this was mainly for values that have to 'match' on all clients, and that do not necessarily involve script.

PR #1918, opened 2023-03-22
Added note for .net framework 4.6 to the build instructions

ktmitton:
Without this, you are unable to load the Citimono and Citimonoref projects in VS.


thorium-cfx (approved):

FabianTerhorst:
@jakub-cfx another change that is maybe relevant for the building.md improvements.

prikolium-cfx:
Thanks for the PR but we already added note about .NET 4.6 requirement and an instruction how to add it.

PR #1930, opened 2023-04-05
tweak(scripting/csharp): add blip cone function

PappaNiels:
This function will add a cone to the blip (see picture).  This can be used for for example camera's or whatever you like!

image

The following code snippet is an example of how you can use the 'new' function.

```cs
using System;
using CitizenFX.Core;
using CitizenFX.Core.Native;
using static CitizenFX.Core.Native.API;

namespace TEST
{
    public class Class1 : BaseScript
    {
        public Class1() 
        {
            RegisterCommand("testblip", new Action<int, List<object>, string>(async (src, args, raw) =>
            {
                int hash = (uint)GetHashKey("ch_prop_ch_cctv_cam_02a");

                RequestModel(hash);

                while (!HasModelLoaded(hash)) await Delay(10);

                Vector3 v = GetEntityCoords(Game.PlayerPed.Handle, true);
                int obj = CreateObject(GetHashKey("ch_prop_ch_cctv_cam_02a"), v.X, v.Y, v.Z, false, false, false);

                while (!DoesEntityExist(obj)) await Delay(10);

                Blip blip = new Blip(AddBlipForEntity(obj));

                blip.AddCone(1.2f, 8.2f, 3.0f, 1);

                await Delay(3000);

                blip.RemoveCone();

                DeleteObject(ref obj);
            }), false);
        }
    }
}
```

### DIFF
--- a/code/client/clrcore/External/Blip.cs
+++ b/code/client/clrcore/External/Blip.cs
@@ -3,7 +3,9 @@ using System.Security;
 
 #if MONO_V2
 using CitizenFX.Core;
+using CitizenFX.FiveM.Native;
 using API = CitizenFX.FiveM.Native.Natives;
+using Function = CitizenFX.FiveM.Native.Natives;
 
 namespace CitizenFX.FiveM
 #else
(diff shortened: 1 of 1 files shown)


code/client/clrcore/External/Blip.cs
```diff
@@ -482,6 +482,34 @@ public void RemoveNumberLabel()
...
+		/// </summary>
+		/// <param name="width">Width of the cone (0.0f - 3.0f)</param>
+		/// <param name="length">Length of the cone</param>
+		/// <param name="heading">Heading of the cone in radians (degrees * pi / 180)</param>
+		/// <param name="color">Color of the cone (see <see href="https://docs.fivem.net/docs/game-references/hud-colors/">hud colors</see>/></param>
+		/// <param name="unk0">Unknown at this time</param>
+		/// <param name="unk1">Unknown at this time</param>
+		/// <param name="unk2">Unknown at this time</param>
+		/// <param name="unk3">Unknown at this time</param>
+		public void AddCone(float width, float length, float heading, int color, float unk0 = -1.0f, float unk1 = 1.0f, float unk2 = 1.0f, int unk3 = 1)
```
thorium-cfx, line 496: 1. `ShowCone` and `HideCone` would be better, "add" implies we can have multiple.
2. It seems like `width` is the angle in radians, but used as: `angle * 2`, i.e.: ~3.14 is a full circle (oddly...).
3. `distance` would be more descriptive than `length`.

thorium-cfx (changes requested):
Some changes requested. Example code worked right away 👍

Apart from the inline comment; it's missing some `using`s for mono rt2.

thorium-cfx (approved):

Issue #1942, opened 2023-04-12
Performance issue on getting a state bag value (server side)


joelwurtz:
We are adding metrics on our server to detect and tackle performance on some of our scripts, and we where having a strange performance lag on one part of our code where the code is only doing something like : 

```js
setTick(() => {
     const data = GlobalState.data;
     
     // do some stuff
     
     GlobalState.data = newData;
     
     await wait(1000);
});
```

Where there is a "low" player count (between 1 to 80 players, this code run normally between 5ms, however when there is more player it can go up to more than 100ms (even more than 250ms in one or two tick), 

See the attached graph, (first one is the tick duration, second one is our player count)

![image](https://user-images.githubusercontent.com/90466/231560010-8a4cce63-87e8-4d14-8e11-98ee2de479cb.png)

At first we were believing there was some problems when setting the value to the callback since it will lock the data and need to update other players.

So we add a metrics around this call to mesure it

![image](https://user-images.githubusercontent.com/90466/231560699-ed1b4d78-7fc4-4dba-9bc0-8dcbb41e5d86.png)

As you can see, the set in this call was always under 5 ms so this was not it, after looking more at the code and others events, we where having another tick that have the same problem, however this tick was only getting a value from an Entity state bag (and do nothing else except calculating some data and keeping it in memory)

![image](https://user-images.githubusercontent.com/90466/231561748-65c35fb3-db3d-4eb0-afbe-8b18e848a80f.png)

As you can see both tick have (at some details) the exact same time in duration, after adding some more metrics around the `GET_STATE_BAG_VALUE` native we see that it was this native that was inducing the lag : 

![image](https://user-images.githubusercontent.com/90466/231562493-3a217b06-08f5-4470-a510-a79dae9c729c.png)

When looking at the code in fivem, there is a Mutex when getting a state bag see https://github.com/citizenfx/fivem/blob/master/code/components/citizen-resources-core/src/StateBagComponent.cpp#L473

Since this problemes appear on the first time that we actually need a state bag on and different, we strongly believe that there is a lot of write pressure on this lock so it's is slow to release (maybe from another thread ?) when having a large player count.

At one point we were thinking that, since we have more people we may have more entity created (which create a state bag every time if i'm not wrong) but this was not the case as you can see here : 

![image](https://user-images.githubusercontent.com/90466/231564180-29047cb7-7f45-4f4f-8e46-d65ecd554b7e.png)

There seems to be no correlation between both events

[...]

PR #1943, opened 2023-04-13
Decrease state bag lock pressure

joelwurtz:
Should help https://github.com/citizenfx/fivem/issues/1942

I add a new lock that handle target updates only
I also use a try lock for removal so it should release some pressure when there is a lot of entity despawning / state bag changes in the same time

### DIFF
--- a/code/components/citizen-resources-core/src/StateBagComponent.cpp
+++ b/code/components/citizen-resources-core/src/StateBagComponent.cpp
@@ -581,17 +581,22 @@ void StateBagComponentImpl::AttachToObject(fx::ResourceManager* object)
 		decltype(m_erasureList) erasureList;
 
 		{
-			std::unique_lock _(m_erasureMutex);
-			erasureList = std::move(m_erasureList);
-		}
-
-		if (!erasureList.empty())
-		{
-			std::unique_lock lock(m_mapMutex);
-
-			for (const auto& id : erasureList)
+			if (!m_mapMutex.try_lock())
 			{
-				m_stateBags.erase(id);
+				return;
+			}
+
+			{
+				std::unique_lock _(m_erasureMutex);
+				erasureList = std::move(m_erasureList);
+			}
+
+			if (!erasureList.empty())
+			{
+				for (const auto& id : erasureList)
+				{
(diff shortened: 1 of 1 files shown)


joelwurtz:
I also noted that the state bag map contains weak reference to the precreated state bag map items, i'm wondering if this is really necessary ?

Since the state bag is reacreated into the precreate state bag we could, instead of remove the weak reference, removing the actual reference in the precreated state bag, and check expiration of the weak reference when getting the state bag ? Then there is no more lock on the state bag map itself when removing (here is only a lock on the precreated state bag map)

I can add this change to the PR if you find this better

blattersturm:
I'm unsure how to exactly review this PR and without being sure if this fixes your performance issue (did you test it?) I'm also a bit hesitant on merging it.

(also, a bit of a mess with three commits?)

joelwurtz:
We try to stay on recommended version so i'm not sure on which state the current server release is, is the last version stable enough ? If it's stable enough we can try it with this fix on our server, otherwise we can wait for the recommended version and compile our own server with those changes to confirm that it fixes our problem.

I can squash commit into one, or at least the last one into the second one, since both commit do not change the same thing, tell me what you prefer.

blattersturm:
Latest should be fine (and if it isn't, there's more fuel for regression bug reports :) ) to run to test such.

blattersturm:
Did this get checked on?

joelwurtz:
Sorry for the late reply, i have tested it a while ago but it doesn't seem to fix our latency issue with state bag, i try to look at the ETW trace : nothing was out of ordinary but clearly not my field of predilection. 

I did notice on the ETW there was a lot of switch context but i guess it's "normal" given the server architecture ?

We end up removing all our code that was relying on state bags and we don't have any latence anymore.

During this rewrite we also try to update less often some states : certainly the current implementation have some hot paths issues when setting values to aggressively. We were using it mostly to sync some vehicles states (like oil / fuel / open / close / radio configuration / etc ...), some player data and other global states.

thorium-cfx:
As this solution seems not to resolve the issue I'll be closing it, made a backlog item to tackle this in the future.

PR #1944, opened 2023-04-17
Implement additional hud component natives for FiveM

Disquse:
Plus few minor tweaks in `InterfaceExtraNatives.cpp`

```cpp
char* GET_HUD_COMPONENT_NAME(int id);
void GET_HUD_COMPONENT_ALIGN(int id, int* horizontalAlign, int* verticalAlign);
void SET_HUD_COMPONENT_ALIGN(int id, int horizontalAlign, int verticalAlign);
Vector3 GET_HUD_COMPONENT_SIZE(int id);
void SET_HUD_COMPONENT_SIZE(int id, float x, float y);
```

### DIFF
--- a/code/components/extra-natives-five/src/InterfaceExtraNatives.cpp
+++ b/code/components/extra-natives-five/src/InterfaceExtraNatives.cpp
@@ -46,10 +46,10 @@ static char LoadZoomMapDataMeta()
 static HookFunction initFunction([]()
 {
 	{
-		auto location = hook::get_pattern("33 C0 0F 57 C0 ? 0D", 0);
+		auto location = hook::get_pattern<char>("33 C0 0F 57 C0 ? 0D", 0);
 
-		expandedRadar = hook::get_address<uint64_t*>((char*)location + 7);
-		revealFullMap = hook::get_address<uint64_t*>((char*)location + 37);
+		expandedRadar = hook::get_address<bool*>(location + 7);
+		revealFullMap = hook::get_address<bool*>(location + 37);
 	}
 
 	{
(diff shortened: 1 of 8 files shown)


code/components/extra-natives-five/src/InterfaceExtraNatives.cpp
```diff
@@ -21,12 +27,115 @@ struct MapZoomData
...
+
+static void EnsureHudComponentBackup(int index)
+{
+	if (index < hudComponentsCount)
+	{
+		auto& component = hudComponentsArray[index];
+
+		if (auto& it = hudComponentOldEntries.find(index); it == hudComponentOldEntries.end())
+		{
+			HudComponentData origData = component;
```
blattersturm, line 95: Is there a reason for the second copy here, outside of `.insert({})` potentially being annoyed?

(if it's just the latter, would `.emplace(index, component)` work?)

code/components/extra-natives-five/src/InterfaceExtraNatives.cpp
```diff
@@ -43,13 +152,42 @@ static char LoadZoomMapDataMeta()
...
+	}
+
+	fx::ScriptEngine::RegisterNativeHandler(nativeHash, [=](fx::ScriptContext& ctx)
+	{
+		auto index = ctx.GetArgument<uint32_t>(0);
+
+		if (auto& it = hudComponentOldEntries.find(index); it != hudComponentOldEntries.end())
+		{
+			auto component = &hudComponentsArray[index];
+			component->alignX[0] = it->second.alignX[0];
```
blattersturm, line 174: This partial reset code seems to be in *two* places, maybe it could be unified (also in case any other editable values get added in the future?).

code/components/extra-natives-five/src/InterfaceExtraNatives.cpp
```diff
@@ -43,13 +152,42 @@ static char LoadZoomMapDataMeta()
...
+}
+
 static HookFunction initFunction([]()
 {
 	{
 		auto location = hook::get_pattern("33 C0 0F 57 C0 ? 0D", 0);
 
-		expandedRadar = hook::get_address<uint64_t*>((char*)location + 7);
-		revealFullMap = hook::get_address<uint64_t*>((char*)location + 37);
+		expandedRadar = hook::get_address<bool*>((char*)location + 7);
```
blattersturm, line 189: For easier change tracking, can this type fix be moved to a _separate_ commit?

(it can be in the same PR still, I think)

PR #1945, opened 2023-04-17
fix(build-tools): creation of out/v2

thorium-cfx:
make's target check failed when the `out` folder was already present, ignoring the creation of `out/v2`


blattersturm:
unsure if this handles out/ itself not existing in order but i don't know make

PR #1948, opened 2023-04-18
feat(extra-natives/five): dynamic door creation, network bypass

nikez:


Ekinoxx0:
This seems nice, could you please explain what is this for (trying to solve an issue, anticheat, other) for those of us that doesn't know about dynamic door creation.
This could end up being useful for others, but as it's not documented in code or in commit description :/

Also thanks for being active, it's nice to see servers developers implicated in PRs for FiveM 👍

nikez:
Basically, this allow you to create doors with the last parameter in `CreateObject` to be set to false instead of true in MP.

The current instruction for CreateObject is that the last parameter has to be set to true in "network mode", but when doing that it creates it as a CObject and you have to freeze it and it no longer behaves like a door.

Example:
```
CreateObjectNoOffset(`xs_prop_x18_garagedoor02`, -1314, -2784.54, 13.94, false, false, false)
```

Disquse:
Eh, GitHub ignored my merge commit after rebasing. Merged here: https://github.com/citizenfx/fivem/commit/13ecc5494bcbd65da3ffe937a8894f1f675ba471

PR #1949, opened 2023-04-18
feat(scripting/csharp): Ai blip class

PappaNiels:
This pull request will add the so called 'Ai blips' to the object oriented scripting environment. These blips are run by the game itself, so no updates needed scriptwise. 

Use for the most important 'functions' and construction of the objects:
```cs
// Ai blip without color
AiBlip ai = new AiBlip(ped.Handle /*any ped*/, 270 /*sprite*/);

// Ai blip with color
AiBlip ai = new AiBlip(ped.Handle /*any ped*/, 270 /*sprite*/, 5 /*color (yellow in this case)*/);
```
Enable or disable a vision cone => `ai.Cone`

Force or don't force the blip on the map => `ai.Forced`

Get or set the 'gang id' => `ai.GangId`

Get or set the sprite => `ai.Sprite`

Get or set the notice range => `ai.Range`

This code has been tested on build 2802

### DIFF
--- a/code/client/clrcore/External/AiBlip.cs
+++ b/code/client/clrcore/External/AiBlip.cs
@@ -1,0 +1,158 @@
+using System;
+
+#if MONO_V2
+using CitizenFX.Core;
+using API = CitizenFX.FiveM.Native.Natives;
+
+namespace CitizenFX.FiveM
+#else
+using CitizenFX.Core.Native;
+
+namespace CitizenFX.Core
+#endif
+{
+	public class AIBlip : PoolObject, IEquatable<AIBlip>
+	{
+		/// <summary>
+		/// Creates a <see cref="AIBlip"/>. This blip will be handled by the game itself and are used for peds (guards, allies, etc). 
+		/// </summary>
+		/// <param name="handle">Handle of the ped that will have the blip</param>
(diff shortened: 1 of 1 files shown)


code/client/clrcore/External/AiBlip.cs
```diff
@@ -0,0 +1,177 @@
...
+using API = CitizenFX.FiveM.Native.Natives;
+
+namespace CitizenFX.FiveM
+#else
+using CitizenFX.Core.Native;
+
+namespace CitizenFX.Core
+#endif
+{
+	public sealed class AiBlip : PoolObject, IEquatable<AiBlip>
```
thorium-cfx, line 14: `AIBlip`, convention is to turn acronyms into Pascal case when we have more than 2 characters. Natives are like that purely because the tools indiscriminately apply PascalCase.

Also why seal it?

thorium-cfx (changes requested):
Some concerns as of design:
Currently properties are storing and returning the blip's states. Though this will be an issue when external script/code changes the value, we'll then retrieve wrong values. Think for now it's best to leave the getters out (them being unreliable) and if/when we find the getters reimplement them.

thorium-cfx:
Closing this as alternative #2062 has been approved and merged

PR #1954, opened 2023-04-22
feat(scripting-mono): enable mono v2 runtime

thorium-cfx:
* Will enable the mono v2 runtime on FiveM and Server
  - Resources can use this runtime by defining `mono_rt2 'Prerelease expiring 2023-06-30. See https://aka.cfx.re/mono-rt2-preview for info.'` in their `fxmanifest.lua` file,
  - User's C# assemblies need to be compiled against the `v2/` assemblies.
* RedM needs some internal native changes before it'll be activated,
* Replace mono_rt2 check with an expiration date and informative value.

Final commit for publishing mono_rt2

### DIFF
--- a/code/components/citizen-scripting-mono-v2/src/MonoScriptRuntime.cpp
+++ b/code/components/citizen-scripting-mono-v2/src/MonoScriptRuntime.cpp
@@ -28,6 +28,8 @@
 * We also don't need to do domain != current_domain checks as it's already included in mono_domain_set_internal().
 */
 
+using namespace std::literals; // enable ""sv literals
+
 // backwards compatibility, without this the v1 runtime will crash
 static void back_to_root_domain()
 {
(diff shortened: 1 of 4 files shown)


blattersturm:
Flagged as manual review because:

- needs reviewing and adding of mscorlib change in data/client/
- ideally needs to have some expiration for a temporary flag during the pilot program: we can't guarantee backwards compatibility, so users should not expect it

thorium-cfx:
> we can't guarantee backwards compatibility, so users should not expect it

What did you have in mind? we can print a warning when using this new runtime, would actually not be a bad idea.

>  it should detect based on assembly reference?

Yep, this was discussed and it's in the post release tasks

blattersturm:
> What did you have in mind? we can print a warning when using this new runtime, would actually not be a bad idea.

A warning, and perhaps an expiration date (tied to a specific value in the manifest).

thorium-cfx:
> an expiration date

We'd first need a solution for this one. As with the current setup; the intermediate file system (on client) disallows mono to precheck assemblies and its references, at least before we've fully loaded them and that's done by the runtime itself.

blattersturm:
Expiration date could be as simple as the 'mono_rt2'  value being checked for a (hardcoded) date that we may bump at times.

thorium-cfx:
Interesting idea, think I now understand the target group you're aiming at. I'll get at it

thorium-cfx:
Alright, I made a date checking alternative with a max date we can bump and warnings/errors. As to the link; do you want to put something on the cfx/fivem website or send them to my github repo?

thorium-cfx:
As a reminder, with https://github.com/citizenfx/fivem/pull/1959 the mscorlib replacement is no longer needed.

PR #1956, opened 2023-04-25
A new face for the loading screen

Imjuliio:
Changes to the background image of the screen.


blattersturm:
While we appreciate the effort here, this isn't a change we're likely to accept.

PR #1959, opened 2023-04-26
tweak(clrcore-v2): shared library, reorganization, and fixes

thorium-cfx:
* Fix concurrent loading of scripting-mono v1 and v2,
* Reorganized clrcore-v2 libraries to enable the use of shared libraries,
* Internal clrcore-v2 reorganization and fixes.


thorium-cfx:
Removed redundant server check in `Remote`, thanks to @Keyinator for spotting it.

PR #1961, opened 2023-04-28
tweak(scripting-mono-v2): use core scheduler

thorium-cfx:
Schedule tick calls with the core scheduler, aka Bookmark system or tick-less.
* Removes unnecessary mono-environment-switching costs when there's nothing to execute,
* Renamed parameters and added documentation on the `Scheduler`'s functions.


thorium-cfx:
This has a dependency on https://github.com/citizenfx/fivem/pull/1959, will rerun AppVeyor after that one is merged.

thorium-cfx:
Ready for review/merge; dependency solved and tested the builds.

PR #1968, opened 2023-04-30
Add SET_WATER_AREA_CLIP_RECT & Fix LOAD_WATER_FROM_PATH path handling

ranstar74:
Added **SET_WATER_AREA_CLIP_RECT** - allows to place custom water file outside playable area
Fixed **LOAD_WATER_FROM_PATH** - wasn't handling path separator correctly, creating invalid paths like 'water_handler//water.xml' (two slashes instead of single one)

### DIFF
--- a/code/components/extra-natives-five/src/WaterExtraNatives.cpp
+++ b/code/components/extra-natives-five/src/WaterExtraNatives.cpp
@@ -8,6 +8,9 @@
 #include <ResourceMetaDataComponent.h>
 #include <VFSManager.h>
 
+#include "atArray.h"
+#include "GameInit.h"
+
 struct sDepthBound
 {
 	__m128 min, max;
@@ -68,6 +71,16 @@ struct WaveQuad
 	int8_t directionY; // +11
 }; // 12
 
+struct CWaterArea
+{
+	// Those define maximum allowed coordinates for water quads
+
+	static inline int32_t* sm_ClipMinX;
+	static inline int32_t* sm_ClipMinY;
+	static inline int32_t* sm_ClipMaxX;
+	static inline int32_t* sm_ClipMaxY;
+};
+
 static atArray<WaterQuad>* g_waterQuads;
 static atArray<CalmingQuad>* g_calmingQuads;
 static atArray<WaveQuad>* g_waveQuads;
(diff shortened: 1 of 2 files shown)


ext/native-decls/SetWaterAreaClipRect.md
```diff
@@ -0,0 +1,25 @@
...
+```
+
+Sets world clip boundaries for water quads file (water.xml, water_heistisland.xml)
+Used internally by LOAD_GLOBAL_WATER_FILE
+
+## Examples
+
+```lua
+SetWaterAreaClipRect(-4000, -4000, 4500, 8000)
+end
```
AvarianKnight, line 18: You don't need an `end` here

ranstar74: Good catch. Fixed.

code/components/extra-natives-five/src/WaterExtraNatives.cpp
```diff
@@ -273,6 +285,20 @@ static HookFunction initFunction([]()
 	// Visual calls
 	hook::call(hook::get_pattern("E8 ? ? ? ? 33 F6 84 C0 0F 84 ? ? ? ? F3 0F 10 4F ? F3 0F 10 07 E8"), CustomTestForUnderwaterVisuals);
 
+	// Water level
+	{
+		// This function processes water file set request by LOAD_GLOBAL_WATER_FILE native
+		uintptr_t updateWaterData = (uintptr_t)hook::get_pattern("48 89 5C 24 08 48 89 6C 24 10 48 89 74 24 18 57 48 83 EC 20 8A 05 ? ? ? ? 33");
+
+		// Loads water file & sets hardcoded clip rect for LosSantos / Heist Island
+		uintptr_t setWaterAreaType = hook::get_call<uintptr_t>(updateWaterData + 0x59);
```
Disquse, line 294: This hook is only valid on b2372+ (aka The Cayo Perico Heist), but we also support older builds of the game (1604, 2060, 2189).

ranstar74: I've looked up water extents on 1604 and well its just inlined stack variable there, that would be hell to fix it in all places (50+ refs on 2699). is it fine if I just disable it on versions before Cayo? 
изображение_2023-05-01_224707228

Disquse: Yeah, this is fine. You can put the whole code block inside `if (xbr::IsGameBuildOrGreater<2189>())` check. Would be nice to add a note that your native only work on `b2189`+.

Disquse:
Have you tried to test if those values getting reset during session shutdown (i.e. when you disconnect from a server)? If no, you will have to use `OnKillNetworkDone` event and reset those values to default state yourself.

ook3D:
> Have you tried to test if those values getting reset during session shutdown (i.e. when you disconnect from a server)? If no, you will have to use `OnKillNetworkDone` event and reset those values to default state yourself.

i can confirm it gets reset when disconnecting and reconnecting

ranstar74:
Additionally added water reset branch on reconnecting as @Disquse suggested, to be sure that it works properly on all supported versions.

blattersturm:
Merge was requested in CW chat. I reworded commit messages adding a prefix and squashed the fixes for the new functionality into the initial commit.

PR #1970, opened 2023-05-01
feat(nui): midi API access

manzarek123:
This is to allow navigator.requestMIDIAccess() and be able to connect a real keyboard for a piano resource.

navigator.requestMIDIAccess({sysex: true}) will still be blocked so it should be safe :)

### DIFF
--- a/ext/ui-build/data/root.html
+++ b/ext/ui-build/data/root.html
@@ -80,7 +80,7 @@ registerFrameFunction(function(msg, frameName, frameUrl)
 		const frame = document.createElement('iframe');
 		frame.name = frameName;
 		frame.style.visibility = 'hidden';
-		frame.allow = 'microphone *; camera *;';
+		frame.allow = 'microphone *; camera *; midi *;';
 		frame.src = frameUrl;
 		frame.tabIndex = -1;
 


blattersturm:
Ugh. I don't particularly feel okay with enabling more 'NUI for things that aren't UI' abuse.

manzarek123:
Do you think it can be abused in a bad way?
If there is only good use cases then it would be nice to allow it and will allow the creation of nice resources.

blattersturm:
The same could be done in a cleaner way *without* requiring to abuse NUI and the complex inter-process communication jank from Chromium which can and will lead to issues on 'weird' end-user systems or with 500 other NUI frames running side by side.

'it would be nice to allow it and it will allow the creation of nice resources' is not a sufficient argument here, sorry.

manzarek123:
I am curious, how a cleaner way would work.
In any case we would need an UI and so have an NUI frame and send every midi message to chromium to display pressed keys.
So the WEB MIDI API seemed like a good choice, its there, its proven and I don't see how it will harm more than let's say the Gamepad API.

Its a new feature so yes the argument is just as simple as "it is adding a new feature" that is the point of it.

You are probably right there are probably better ways to do it but no one will ever implement it, so yeah it is a bit disappointing but I understand

blattersturm:
> and I don't see how it will harm more than let's say the Gamepad API.

'the gamepad API' is already a thing that is being misused in this way, the difference is that this didn't need any explicit enabling so people already started doing so without being able to tell them not to.

PR #1973, opened 2023-05-02
fix(scripting-lua): Wrap InvokeNative in trycatch

cybooo:
This could potentionally fix the problem reported at https://forum.cfx.re/t/server-crash-on-version-6415/5086045/14


blattersturm:
no it wouldn't.

PR #1985, opened 2023-05-12
feat(gamestate/server): add filter vehicles mode to sv_filterRequestControl

0xLuca:
This adds a new filter mode to sv_filterRequestControl to route all requests except for entities of type vehicle.


blattersturm:
Supposedly is obsoleted by #1987.

PR #1987, opened 2023-05-12
feat(gamestate/server): add SetEntityIgnoreRequestControlFilter native

manzarek123:
This will allow to use _sv_filterRequestControl_ but still allow to cover some edge cases that would discourage people from using it.
I believe that this will help with the adoption of _sv_filterRequestControl_ and especially _sv_filterRequestControl 4_.


blattersturm:
Right, I do think this would be a reasonable workaround to allow adapting existing scripts.

MatinHQ:
> This will allow to use _sv_filterRequestControl_ but still allow to cover some edge cases that would discourage people from using it. I believe that this will help with the adoption of _sv_filterRequestControl_ and especially _sv_filterRequestControl 4_.

Does this still work when using mode 2 or 3?

PR #1989, opened 2023-05-13
Update Simplified Chinese Translations.

vosscat:


DaniGP17:
You can't make a pull request to update translations, you should update here: https://translations.cfx.re/projects/citizenfx/cfx-ui/

PR #1991, opened 2023-05-13
tweak(cfx-ui): 'Already have account' -> 'Already have an account' 

MortalD3V:


blattersturm:
1. Can you actually make the commit message have the correct title? Doing so in the pull request only won't magically be sufficient, and the commit message is still 'Update locale-en.json'.
2. I'm a little hesitant to merge this without approval from @nihonium-cfx.

PR #1992, opened 2023-05-13
feat(streaming-five): Scaleform debug logging

thorium-cfx:
Print `DEBUG_LOG` messages from scaleforms onto the console
* Togglable with convar `game_enableScaleformDebugLog`

Tested on: 1604, 2189, 2545, 2699, and 2802.

### DIFF
--- a/code/components/gta-streaming-five/src/ScaleformHacks.cpp
+++ b/code/components/gta-streaming-five/src/ScaleformHacks.cpp
@@ -4,6 +4,7 @@
 #include <nutsnbolts.h>
 #include <sfFontStuff.h>
 
+#include <CoreConsole.h>
 #include <Streaming.h>
 #include <CrossBuildRuntime.h>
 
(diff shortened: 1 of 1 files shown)


code/components/gta-streaming-five/src/ScaleformHacks.cpp
```diff
@@ -389,6 +401,46 @@ namespace sf
...
+		"PAUSE_TYPE",
+		"STORE",
+		"GAMESTREAM",
+		"SF_BASE_CLASS_VIDEO_EDITOR",
+		"SF_BASE_CLASS_MOUSE",
+		"SF_BASE_CLASS_TEXT_INPUT",
+	};
+
+	static bool g_scaleformDebugLog = 0;
+	static ConVar<bool> g_scaleformDebugLogVar("cl_scaleform_debug_log", ConVar_Archive | ConVar_UserPref, false, &g_scaleformDebugLog);
```
blattersturm, line 422: I think for game behavior we generally use `game_` prefix, so something like `game_enableScaleformDebugLog`?

thorium-cfx: aight, looks better yeah.

PR #2012, opened 2023-05-25
WIP: Bundle ReShade with FiveM

blattersturm:
Thought of during the aftermath of the recent ReShade concerns, this patch set will build a variant of ReShade as part of `rage:graphics:five`.

There's a few things to note:
* It *should* only proxy the base game device. This hasn't been tested completely, and there might be some vtable patches left.
* It will read shaders and config files from `plugins/` as working directory.
* It can be enabled using the `reshade` convar, with the overlay toggled via the `reshade_gui` convar. We might consider exposing the former as a setting in `cfx-ui`, and the latter as a `devgui` button.
* It's only received minimal testing on my main workstation, with the configuration from NVE, primarily. NVE.asi from the current release seems to work fine.
* ReShade has single-instance support, which means that since `rage-graphics-five.dll` has the ReShade exports, other ReShade DLLs will skip loading. This is a mild compat break, as even if `reshade` is set to `false`, it will silently ignore existing ReShade installs. **This also might mean that if another ReShade is loaded before rage-graphics-five.dll is, unexpected things may occur.**
* The fork hasn't been tested to be fully functional, there might be some logic errors.
* There's a lot of code cleanup that'll be needed if/when this does ship, mainly for some of the 'selective' ReShade ignoring.


blattersturm:
Some additional concerns seen during testing:

- The forkawesome icon font isn't loaded and glyphs from it show as '?'. Issue is the TTF fails to be loaded by FreeType and might only work with stbz
- The input system might have to be implemented anyway as effect authors expect the 'use effects' and overlay toggles at least. Alternately, these could be mapped to game binds.
- FPS and time overlays go out of the game window, these need viewport position to be added.
- Steam overlay sizing is weird but this might be related to the beta client or not be affected by ReShade mode.

blattersturm:
Pushed a commit fixing server/release builds failing on AppVeyor.

PR #2025, opened 2023-05-30
Corrected Date in MonoScriptRuntime.cpp

NiftyR:


blattersturm:
the fuck?

NiftyR:
If the message is not incorrect that means the logic for expiring the runtime is incorrect.

AvarianKnight:
Maybe you're misunderstanding?

The values set for the mono runtime are set to expire on [2023-06-30](https://github.com/citizenfx/fivem/blob/1649e2a2afa47d505c078ac472a7a88543c57e12/code/components/citizen-scripting-mono-v2/src/MonoScriptRuntime.cpp#L266)

jcf54:
Right, but the prerelease has expired now. We're in GMT timezone, the date local to us is 2023-05-31.

image

NiftyR:
Looking at the code it's likely using || is being the issue since the day could be triggering this if
```cpp
if (time->tm_year > maxYear || time->tm_mon >= maxMonth || time->tm_mday > maxDay)
```

AvarianKnight:
Should probably be
```c++
if (time->tm_year > maxYear || (time->tm_mon >= maxMonth && time->tm_mday > maxDay))
```

AvarianKnight:
> That looks about right, you can see where my confusion came from though surely.

Yes, but it would've been better to report this as an issue instead of making a PR to change the time.

PR #2036, opened 2023-06-04
feat(gamestate/server): disable networked sounds convar

wdesgardin:
This is a continuation of an old pull request I made regarding networked sounds #1316 

As discussed, I simply added a convar to disable the routing of these events.


code/components/citizen-server-impl/src/state/ServerGameState.cpp
```diff
@@ -6756,6 +6758,14 @@ std::function<bool()> fx::ServerGameState::GetGameEventHandler(const fx::ClientS
 		return GetRequestControlEventHandler(client, std::move(buffer));
 	}
 
+	if(eventType == NETWORK_PLAY_SOUND_EVENT)
+	{
+		return []()
+		{
+			return g_networkedSoundsEnabledVar->GetValue();
```
AvarianKnight, line 6765: Should this not be a static bool (see other convars here for reference) instead of using `GetValue()`

PR #2038, opened 2023-06-06
feat(onesync): add RPC natives

manzarek123:
2 new RPC to turn a ped, both tested


ahcenezdh:
any update on this?

FabianTerhorst:
Needs discussion if new rpc natives should be added while they have reliability issues, because they don't guarantee execution when the client is the owner or when the entity exists on the client.

prikolium-cfx:
Due to reliability issues with RPC natives we prefer to avoid adding new ones.

PR #2041, opened 2023-06-07
fix(extra-natives/five): entity fire id not released

nikez:
Test script: https://pastebin.com/xv4jrcFf


gottfriedleibniz:
This is a bit broken.

Any call to `StopEntityFire` or `StartEntityFire` will stop/free any active fires previously created by `StartScriptFire` (where "flag" is set while the "entity" field is null).

PR #2062, opened 2023-06-19
tweak(scripting/csharp): Add ai blip functions

PappaNiels:
This is an alternative pull request to #1949 . After some discussing with thorium and de C# scripters, we decided this would be the best option out of the two pull requests. This pull request puts the ai blip features in the ped class instead of making a separate class. You can decide which one to use.

### DIFF
--- a/code/client/clrcore/External/Ped.cs
+++ b/code/client/clrcore/External/Ped.cs
@@ -1572,6 +1572,120 @@ namespace CitizenFX.Core
 			}
 		}
 
+		/// <summary>
+		/// Sets the status of the vision cone of the AI Blip
+		/// </summary>
+		public bool AIBlipConeEnabled
+		{
+			set
+			{
+				API.SetPedAiBlipHasCone(Handle, value);
+			}
+		}
+
+		/// <summary>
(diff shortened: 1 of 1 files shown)


code/client/clrcore/External/Ped.cs
```diff
@@ -2,16 +2,16 @@
...
 using API = CitizenFX.FiveM.Native.Natives;
-using Function = CitizenFX.FiveM.Native.Natives;
-namespace CitizenFX.FiveM
+using Function = CitizenFX.FiveM.Native.Natives;
+namespace CitizenFX.FiveM
 #else
-using CitizenFX.Core.NaturalMotion;
-namespace CitizenFX.Core
+using CitizenFX.Core.NaturalMotion;
+namespace CitizenFX.Core
```
thorium-cfx, line 14: Only new line changes? better discard it

thorium-cfx (changes requested):

PappaNiels:
forgot to add a 'check' method and a method to disable the ai blip. its there now :)

thorium-cfx (approved):

PR #2081, opened 2023-07-01
fix(clrcore-v2): native long/ulong types, Coroutine.Run, OutString cast, and transparency issue

thorium-cfx:
fix(clrcore-v2): use (u)long types on natives when requested
* Adds overrides for (u)long types for v2 native generation to compensate for `type.nativeType` being just `int` on our 64 bit integer friends.

fix(clrcore-v2): Coroutine.Run, OutString cast, and transparency issue
* Only run Coroutine.Run(...) functions directly when we're on the main thread, otherwise we run it next frame.
* Fix return type of Coroutine.Run<T>(...);
* Move OutString -> CString conversion operator.
* Fix CAS transparency issue with EventHandlerSet's + operator for `Delegate` types (don't know why SecAnnotate didn't pick up on this).

resolves thorium-cfx/mono_v2_get_started#22

### DIFF
--- a/code/client/clrcore-v2/Coroutine/CoroutineMethods.cs
+++ b/code/client/clrcore-v2/Coroutine/CoroutineMethods.cs
@@ -1,6 +1,7 @@
 using System.Collections.Generic;
 using System;
 using System.Runtime.InteropServices;
+using System.Threading;
 
 namespace CitizenFX.Core
 {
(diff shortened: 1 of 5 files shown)


Sasino97:
Thank you ❤ I look forward to a stable mono v2 and I'll keep reporting issues as I encounter them. My codebase is very big and diverse, and once the main crashes are solved I hope to test with 50+ players on my secondary server

PR #2089, opened 2023-07-08
Clamp attribute index in `SET_PED_COMBAT_ATTRIBUTES` for FiveM and RedM

Disquse:
See https://github.com/citizenfx/natives/pull/892

Additional changes related to FiveM:
- Reworked rage parser code to use the modern component exporting approach.
- Added `enumElemCount` to `parMemberDefinition` to avoid manual iterations.

In order to make it also work in RedM, these additional changes was made:
- Implemented side-by-side rage parser support for RDR3.
- Fixed a bug that was breaking native handlers overriding.

### DIFF
--- a/code/components/extra-natives-five/src/NativeFixes.cpp
+++ b/code/components/extra-natives-five/src/NativeFixes.cpp
@@ -12,6 +12,9 @@
 #include <Hooking.h>
 #include <scrEngine.h>
 #include <CrossBuildRuntime.h>
+#include "RageParser.h"
+#include "Resource.h"
+#include "ScriptWarnings.h"
 
 static void FixVehicleWindowNatives()
 {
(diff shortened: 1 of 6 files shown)


gottfriedleibniz:
A few semi-related issues that can be addressed later:
1. RDR2 suffers from this as well.
1. `parMemberEnumData` already contains the count of `enumData` (`+ 0x30`) so you wouldn't have to manually count.

Disquse:
Going to be reworked soon to include RDR3 and use precalculated length. Thank you, @gottfriedleibniz

gottfriedleibniz:
Depending on how defensive you want to be, in RDR2 both `0xF4A5C4509BF923B1` and `0x06565032897BA861` seem to suffer from a similar issue (presumably R* is using asserts in game-code). Besides proper documentation, I'm not sure how nit-picky you need to be with those though.

PR #2090, opened 2023-07-08
fix(extra-natives/five): relaxed entity fire id cleanup

jakub-cfx:
changes the cleanup condition to one that doesn't interfere with StartScriptFire initiated fires but still effectively prevents orphaned fire IDs

tested on b1604 w/ scripts provided by the original fix dev (PR https://github.com/citizenfx/fivem/pull/2041), their tests still pass & script fires now function as expected again

### DIFF
--- a/code/components/extra-natives-five/src/NativeFixes.cpp
+++ b/code/components/extra-natives-five/src/NativeFixes.cpp
@@ -239,7 +239,7 @@ void FreeOrphanFireEntries()
 {
 	for (auto& i : *g_fireInstances)
 	{
-		if (!i.entity() && i.flags() & 0x4)
+		if (!i.entity() && i.flags() == 0x4)
 		{
 			i.flags() &= ~0x4;
 		}


nikez:
LGTM: was going to fix this today, but timing :)

jakub-cfx:
> LGTM: was going to fix this today, but timing :)

sorry for the snipe 😅

gottfriedleibniz:
Is the reason why this is failing/not-releasing  known?

It's not something I've looked into, but was wondering because `0x7F0DD2EBBB651AFF` (on PC anyway) has always had its own (somewhat-different?) obfuscation (See: `8B 15 ? ? ? ? 33 15 ? ? ? ? 48 01 D5 E9`).

jakub-cfx:
> Is the reason why this is failing/not-releasing known?
> 
> It's not something I've looked into, but was wondering because `0x7F0DD2EBBB651AFF` (on PC anyway) has always had its own integrity check (See: `8B 15 ? ? ? ? 33 15 ? ? ? ? 48 01 D5 E9`), and it's not like this native is doing anything complicated.
> 
> (After second thought, that is probably an unrelated coincidence).

Haven't looked into the root cause myself yet, will check out the pattern in a bit

PR #2108, opened 2023-07-13
feat(state-bag): initial implementation of state bag filters

AvarianKnight:
This is an initial implementation of state bag rejections

This was tested and works, this will reject the state bag change and properly sync the original value back to the calling client. 

```js
AddStateBagFilter("key", "", (bagName, key, value) => {
    console.log("requested change", key, value);
    setTimeout(() => {
        console.log("actual value", Player(1).state.key)
    }, 0)
	// false is used to reject changes
	return false;
})
```

There are a few things that I still need to do

## Todos
- [x] Figure out why returning nothing in AddStateBagChangeHandler causes the server to crash (maybe its just on debug)
- [x] Reset value properly if `originalValue` is none
- [x] Document state bag rejection

### DIFF
--- a/code/components/citizen-resources-core/include/StateBagComponent.h
+++ b/code/components/citizen-resources-core/include/StateBagComponent.h
@@ -72,7 +72,7 @@ public:
 	//
 	// Sets data for a key.
 	//
-	virtual void SetKey(int source, std::string_view key, std::string_view data, bool replicated = true) = 0;
+	virtual void SetKey(int slotId, std::string_view key, std::string_view data, bool replicated = true) = 0;
 
 	//
 	// Sets the owning peer ID.
@@ -112,7 +112,7 @@ public:
 	// Should be called when receiving a state bag control packet.
 	// arg: outBagNameName; if given (!= nullptr) and if the state bag wasn't found then this string will contain the bag name, otherwise outBagNameName is unchanged.
 	//
-	virtual void HandlePacket(int source, std::string_view data, std::string* outBagNameName = nullptr) = 0;
+	virtual void HandlePacket(int slotId, std::string_view data, std::string* outBagNameName = nullptr) = 0;
 
 	//
 	// Gets a state bag by an identifier. Returns an empty shared_ptr if not found.
@@ -145,6 +145,10 @@ public:
 	//
 	virtual void AddSafePreCreatePrefix(std::string_view idPrefix, bool useParentTargets) = 0;
 
+	//
+	// A rejection handler for state bags, this is ran before OnStateBagChange
+	//
+	fwEvent<std::string_view, std::string_view, const msgpack::object&> ShouldAllowStateBagChange;
 	//
 	// An event handling a state bag value change.
 	//
--- a/code/components/citizen-resources-core/src/StateBagComponent.cpp
+++ b/code/components/citizen-resources-core/src/StateBagComponent.cpp
@@ -27,7 +27,7 @@ public:
 
 	virtual void Reset() override;
 
-	virtual void HandlePacket(int source, std::string_view data, std::string* outBagNameName = nullptr) override;
+	virtual void HandlePacket(int slotId, std::string_view data, std::string* outBagNameName = nullptr) override;
 
 	virtual std::shared_ptr<StateBag> GetStateBag(std::string_view id) override;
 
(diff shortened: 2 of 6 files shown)


AvarianKnight:
This should be good to review now but I have a few questions about the final implementation.

1. Should the client be able to reject server changes
2. Should the server be able to reject server changes

Yum1x:
*not related to this topic but would be nice have a feature to delete/clear a state bag.*

blattersturm:
> This should be good to review now but I have a few questions about the final implementation.
> 
> 1. Should the client be able to reject server changes
> 2. Should the server be able to reject server changes

Both should be 'no' to reduce unintended side effects.

AvarianKnight:
I'm wondering if it wouldn't be better to have a `ADD_STATE_BAG_REJECT_HANDLER` or something similar, which would be a server-only handler to run before `ADD_STATE_BAG_CHANGE_HANDLER`.

The main issue with how I have implemented is that its order dependent and still triggers other change handlers even if its rejected, and there isn't any way to really go back and re-run those changes.

blattersturm:
`_FILTER` might be more sensible than `_REJECT_HANDLER`, though even then a separate function is also a bit ugly. Is there a practical use case where this'd be needed?

AvarianKnight:
> Though even then a separate function is also a bit ugly. Is there a practical use case where this'd be needed?

I'm afraid I'm not sure what you mean here

thelindat:
> Is there a practical use case where this'd be needed?

Rejecting on the change handler seems like it could have some unintended side-effects. Pretending these handlers were added to separate scripts..

```lua
local somevar

AddStateBagChangeHandler('key', '', function(bag, key, value)
    -- value is assigned to somevar, state change is assumed to succeed
    somevar = value
    return false
end)

AddStateBagChangeHandler('key', '', function(bag, key, value)
    -- reject the state change
    return true
end)

local somevar2

AddStateBagChangeHandler('key', '', function(bag, key, value)
    -- state was already rejected, but this is triggered anyway
    -- value is assigned to somevar2
    somevar2 = value
    return false
end)
```

code/components/citizen-resources-core/include/StateBagComponent.h
```diff
@@ -72,7 +72,7 @@ class CRC_EXPORT StateBag : public std::enable_shared_from_this<StateBag>
 	//
 	// Sets data for a key.
 	//
-	virtual void SetKey(int source, std::string_view key, std::string_view data, bool replicated = true) = 0;
+	virtual void SetKey(int slotId, int source, std::string_view key, std::string_view data, bool replicated = true) = 0;
```
AvarianKnight, line 75: Just a note for me later, if people need this they can just get the entity owner, this adds a lot of mess

AvarianKnight:
>Is there a practical use case where this'd be needed?

The reason to add another native is because as Linden pointed out, it might lead to unexpected behavior by people who use `AddStateBagChangeHandler`, explicitly having a function that runs before it makes more sense and won't mess with expected behavior.

If the question was more towards "why is this needed", being able to deny state bags is generally a pretty requested feature and would overall be nice, as you can server-authoritative data. This is particularly useful for when working with garage resources as you can have the vehicles table index as a state bag and use that to save later and reject any clients change to that value.

ahcenezdh:
Any update on this?

AvarianKnight:
Going through and bumping a bunch of random prs is generally frowned upon. They'll get to it when they get to it

thorium-cfx:
Related to #2257

nihonium-cfx:
Doesn't this has a potential to cause server's main-thread lags?

Each SBag update would be calling into ScRT [blocking main-thread](https://github.com/citizenfx/fivem/blob/66d052957b28a5102eba86fad5947e40b22597dc/code/components/citizen-server-impl/src/state/ServerGameState.cpp#L4208) and filtering code can be of arbitrary complexity. Main-thread ticks every 50ms ([1000 / 20](https://github.com/citizenfx/fivem/blob/66d052957b28a5102eba86fad5947e40b22597dc/code/components/citizen-server-impl/src/GameServer.cpp#L194)), calling into ScRT is not cheap, I'd say it's safe to eyeball that even for the most simple filter function and just one of it, it'd take ~0.5ms, now let's say you've got 1000 players and your normal game logic sends updates each frame, now you will have ~500ms worth of filtering checks for main-thread, this will cause tick delays and main-thread hitch warnings.

Of course, this is exaggerated example, but even SBag value update on each frame isn't that unrealistic (say you store car's fuel level there), but the more checks you have, the more updates you have, the more sophisticated the filter is - the more delays it will potentially create for main-thread.

thelindat:
https://github.com/citizenfx/fivem/pull/2257 should cover most cases where filtering would be unnecessarily slow or complex, especially with different sync flags (i.e. owner only, no sync, read only). Not sure how the team intends or intended on making statebags rejectable in the first place, and doing it in the change handler itself might lead to side effects (and doesn't solve performance concerns).

PR #2117, opened 2023-07-18
Update gameconfig.xml

normalday123:
admin @blattersturm @Disquse please accept this so we can stream more ytd, you have to.


blattersturm:
nah we don't 'have to'

also not 'admin'.

something like #1419 would be way better here.

also, FragmentStore isn't even TxdStore, this isn't even related to 'ytd'.

PR #2118, opened 2023-07-19
Update gameconfig.xml

normalday123:
how about try this number admin @Disquse and @blattersturm 

i get this number from here https://www.gta5-mods.com/misc/gta-5-gameconfig-300-cars gameconfig i want to increase my server to 2944 without remove clothing pack pleaseeee

### DIFF
--- a/data/client/citizen/common/data/gameconfig.xml
+++ b/data/client/citizen/common/data/gameconfig.xml
@@ -458,7 +458,7 @@
 						</Item>
 						<Item>
 							<PoolName>TxdStore</PoolName>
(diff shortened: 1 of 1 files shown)


AvarianKnight:
This doesn't change what blattersturm said in the last PR, having dynamically sized pools would be a better option.

Also note that constantly opening PR's and then @'ng people is a good way to annoy people and possibly get blocked from the repo.

normalday123:
nice info, i hope admin increase they skill to solve this problem, thanks

PR #2121, opened 2023-07-20
fix(mumble): fix possible data race

### LINKED ISSUE #2115: MumbleSetPlayerMuted / MumbleIsPlayerMuted leads to server crash
Hi, 

Yesterday i was coding a script where an admin can mute a player from mumble from these natives: [https://docs.fivem.net/natives/?_0x1D5D50C2](url) and [https://docs.fivem.net/natives/?_0xCC6C2EB1](url) but when you try to mute a player it lead to a huge main stack root loop and the server crash

AvarianKnight:
Possibly fixes #2115 this doesn't fix the underlying implementation though.

### DIFF
--- a/code/components/voip-server-mumble/src/client.cpp
+++ b/code/components/voip-server-mumble/src/client.cpp
@@ -112,6 +112,7 @@ int Client_getfds(struct pollfd *pollfds)
 
 bool Client_is_player_muted(int serverId)
 {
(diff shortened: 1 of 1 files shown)


AvarianKnight:
Fixed in https://github.com/citizenfx/fivem/commit/9b1f6ee2497a7c6502715dcedf8c662642c05494

PR #2122, opened 2023-07-20
feat(cfx-ui/sorting): add a new server list sorting for relevant ^server name search

PichotM:
### Goal of this commit

When you are searching for a specific server in the server list, sometimes the name of your community might be used in other project descriptions, making a user struggle to find the server he is looking for.

For example, if you type in the search bar names like Echo, RSM, GLife, Impulse, or many others, you will end up with non-relevant servers putting random names into their project description and showing up at the top of the list because they've more boost power (Except if the server is pinned).

The goal of this change is to sort in a better way when the user inputs a server name (or project name). If the project's name starts with was the user is inputting, it will be sorted above the top boosted servers.

### Result

I have added this gif to show you how it will impact your search.

### Potential issues

One issue with my edit is that we do not rely on the cached result of the server list if we have a relevant name to sort. We also don't uniquely cache every new sorting, so it is fine, I guess.

Feel free to share any ideas/comments you may have

### DIFF
--- a/ext/cfx-ui/src/cfx/common/services/servers/source/WorkerSource.worker.ts
+++ b/ext/cfx-ui/src/cfx/common/services/servers/source/WorkerSource.worker.ts
@@ -3,7 +3,7 @@ import { filterList } from "./listFilter";
 import { getAllMasterListServers } from "./utils/fetchers";
 import { IAutocompleteIndex, IListableServerView } from "./types";
 import { IServerListConfig } from "../lists/types";
-import { sortList } from "./listSorter";
+import { sortFilteredList, sortList } from "./listSorter";
 import { serverView2ListableServerView } from "../transformers";
 import { AutocompleteIndexer } from "./autocomplete";
 import { shouldPrioritizePinnedServers } from "cfx/base/serverUtils";
@@ -152,9 +152,15 @@ export class ServersWorker {
       return;
     }
 
-    this.sendList(list.config, list.id, filterList(
+    const filteredList = filterList(
       this.listableServersMap,
       this.getSortedList(list.config),
+      list.config,
+    );
+
+    this.sendList(list.config, list.id, sortFilteredList(
+      this.listableServersMap,
+      filteredList,
       list.config,
     ));
   }
--- a/ext/cfx-ui/src/cfx/common/services/servers/source/listSorter.ts
+++ b/ext/cfx-ui/src/cfx/common/services/servers/source/listSorter.ts
@@ -1,4 +1,4 @@
-import { shouldPrioritizePinnedServers } from "cfx/base/serverUtils";
+import { shouldPrioritizePinnedServers, getPrioritizedServerName } from "cfx/base/serverUtils";
 import { IServerListConfig, ServerListSortDir, ServersListSortBy } from "../lists/types";
 import { IPinnedServersConfig } from "../types";
 import { IListableServerView } from "./types";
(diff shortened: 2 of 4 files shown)


nihonium-cfx (changes requested):
First of all, thank you for contribution! Conceptually I think this is a good change.

### Performance issue
Resorting the whole servers list is not that slow normally, however, this can and should be done better by sorting the list with server name "relevance" after the list has been filtered, as it is guaranteed to be a much smaller list and thus much faster to sort.

### Improving relevance factor
I think using the index of where the `searchName` starts in the `.sortableName` can yield even better results in comparison to only relying on whether `.sortableName` starts with the `searchName` or not.

In other words: the closer the `searchName` is to the beginning of the `.sortableName` - the higher in the list it should be.

Need not to forget about the cases when `.sortableName` doesn't include `searchName` at all, these should be pushed to the end of the list respecting all other sorting factors so the sorting is still stable.

#### Code style issues
1. Use of semicolon in the end of expressions is required.
2. Arguments of arrow functions are always wrapped with parens, even if it's just one argument.

#### Commit naming convention
The name of the changed component must be `ext/cfx-ui` instead of `cfx-ui/sorting`.

PichotM:
The new changes might be better

ext/cfx-ui/src/cfx/common/services/servers/source/WorkerSource.worker.ts
```diff
@@ -152,18 +152,24 @@ export class ServersWorker {
...
+      list.config,
     ));
   }
 
   private getSortedList(config: IServerListConfig): string[] {
     const listHash = getSortedListHash(config);
 
     let sortedList = this.sortedLists[listHash];
-    if (!sortedList) {
+    if (!sortedList || listHash.indexOf(':serverNameOnTop') !== -1) {
```
nihonium-cfx, line 172: This change is not needed anymore

ext/cfx-ui/src/cfx/common/services/servers/source/listSorter.ts
```diff
@@ -102,3 +102,47 @@ function sortByProperty(servers: Record<string, IListableServerView>, property:
 
   return -dir;
 }
+
+export const sortFilteredList = (servers: Record<string, IListableServerView>, sortedList: string[], config: IServerListConfig): string[] => {
+  const sorters: Array<(a: string, b: string) => number> = [];
+
+  if (shouldPrioritizeServerName(config)) {
```
nihonium-cfx, line 109: This function can be made into a `getPrioritizedServerName` that'd return the `serverName` if found so we don't have to fetch it from search terms again

nihonium-cfx (changes requested):
Looks good to me, apart from these two comments and commit name!

PichotM:
done

nihonium-cfx (approved):
LGTM

Georgie15:
I think this change has made it harder for players to find the servers they are attempting to find. For example, our server is called "Paradise State Roleplay" but we are also known as "PSRP," abbreviated. If a player types "PSRP," they are given a bunch of smaller servers called "PSRP" in the title, yet the player is likely looking for the bigger server. With the old search, if a player finds a bigger server, the server is typically big enough that boosts will display them at the top. I understand that an easy fix is to add "PSRP" in our server title, but I feel that this change has made the default search worse. With the old search, adding more boosts allowed us to be above the other servers since they had few players. Now, servers have little control over where they rank within the search regardless of size or boosts unless filters are selected manually.

A better alternative for this change would be to set it back to what it was before **but have a filter option** that allows players to see more accurate results, such as how the search is now. As a server owner for two years, this is my honest opinion on the new change. I hope to see this considered and later added since the old search was controllable for owners on where they wanted to place.

Issue #2018, opened 2023-05-26
StateBags issue


Marky-S:
The problem is when i set replicated statebag from `AddStateBagChangeHandler` handler on client, this value is partial not replicating, i mean, all okay until entity restream, after restream this statebag value just disappearing on client and value is null. 

But if in  `AddStateBagChangeHandler` handler before setting statebag i make `await Delay(1000)` all works fine even after restream.

Theres is some repro code:

- Client
```c#
    public class Entry : BaseScript
    {
        public Entry()
        {
            AddStateBagChangeHandler("tunning", null, new Action<string, string, object, int, bool>(StateBagChanged));
            AddStateBagChangeHandler("tunningDelayed", null, new Action<string, string, object, int, bool>(StateBagChangedDelayed));
        }

        private void StateBagChanged(string bagName, string key, object value, int reserved, bool replicated)
        {
            var vehicleHandle = GetEntityFromStateBagName(bagName);
            var vehicle = new Vehicle(vehicleHandle);

            if (vehicle.State.Get("tunningApplied") == true)
            {
                Debug.WriteLine("Tunning was applied earlier");
                return;
            }

            vehicle.State.Set("tunningApplied", true, true);

            Debug.WriteLine("First tunning apply");
        }

        private async void StateBagChangedDelayed(string bagName, string key, object value, int reserved, bool replicated)
        {
            var vehicleHandle = GetEntityFromStateBagName(bagName);
            var vehicle = new Vehicle(vehicleHandle);

            if (vehicle.State.Get("tunningApplied") == true)
            {
                Debug.WriteLine("Tunning was applied earlier");
                return;
            }

            await Delay(1000);

            vehicle.State.Set("tunningApplied", true, true);

            Debug.WriteLine("First tunning apply");
        }

        [Command("restream")]
        private async void OnRestreamCommand()
        {
            Game.PlayerPed.Position = new Vector3(-10_000f, -10_000f, 70f);

            await Delay(1000);

            Game.PlayerPed.Position = new Vector3(-14.03844f, 19.02692f, 70.88116f);
        }

        [Command("createVeh")]
        private async void OnCreateVehCommand()
        {
            var modelHash = GetHashKey("t20");

            var vehicleModel = new Model(modelHash);
            await vehicleModel.Request(3_000);

            TriggerServerEvent("createVehFromClient");
        }

        [Command("createVehDelayed")]
        private async void OnCreateVehDelayedCommand()
        {
[...]

PR #2133, opened 2023-07-29
Update AddStateBagChangeHandler.md

Marky-S:
Updating documentation due to https://github.com/citizenfx/fivem/issues/2018

### DIFF
--- a/ext/native-decls/AddStateBagChangeHandler.md
+++ b/ext/native-decls/AddStateBagChangeHandler.md
@@ -36,6 +36,9 @@ If bagName refers to a player, use [GET_PLAYER_FROM_STATE_BAG_NAME](?_0xA56135E0
 ## Return value
 A cookie to remove the change handler.
 
+## Notes
+In its current form, `StateBagChangeHandler`'s are also invoked as the state bag is being reconstructed/restreamed. Due to the built-in implementation, the `keyFilter` parameter is always sorted in dictionary of all state bags. In reconstructed/restreamed state bag scenarios, keep in mind that in the `handler` function other state bags with specified `keyFilter` may not be processed yet at this moment, depending on their order.
+
 ## Examples
 ```js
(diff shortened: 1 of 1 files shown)


AvarianKnight:
I feel like this would probably be better off in the [docs](https://docs.fivem.net/docs/scripting-manual/networking/state-bags/) vs native-decls as then you can provide examples, the wording here is will also confuse the common FiveM dev.

### Everything below here should be considered "docs", and hopefully easy to understand

`StateBagChangeHandler`'s are invoked as the state bag is replicated, or set. This means that the `StateBagChangeHandler` can be set twice on the client if they where the ones to set the state bag as replicated.`

### Example
```js
AddStateBagChangeHandler("someKey", null, (bagName, key, value, _, willReplicate) => {
	if (willReplicate) {
        console.log(`We (the client) set the key: ${key} to be replicated to the server with value ${value}`)
    } else {
		// We will get back the value that we set to be replicated to the server
        console.log(`We (the client) got the key: ${key} back from the client it replicated the value: ${value}`)
    }
})
```

There are also some caveats with replication from Server to Client:

* When the player is in scope of the entity/player, `StateBagChangeHandler`'s will run in order from which they are replicated.
* If the player enters scope of another entity/player, the `StateBagChangeHandler`'s will get ran in alphabetic order as the underlying implementation is a dictionary.

gottfriedleibniz:
It'd also be nice to get clarification on whether the lexicographic ordering is an implementation detail or not. For instance, https://github.com/citizenfx/fivem/issues/1942 is still an open issue and it wouldn't surprise me if the data map (and its mutex) are a little hot.

AvarianKnight:
> It'd also be nice to get clarification on whether the lexicographic ordering is an implementation detail or not. For instance, #1942 is still an open issue and it wouldn't surprise me if the data map (and its mutex) are a little hot.

It would be nice to know, maybe bubble or thorium could shed some light on that.

PR #2141, opened 2023-08-04
feat(net): attribute proper packet names for timeout

AvarianKnight:
This makes it so if a player times out, instead of getting packet hash as hex they'll get the actual packet name

### DIFF
--- a/code/components/debug-net/src/NetDebug.cpp
+++ b/code/components/debug-net/src/NetDebug.cpp
@@ -16,6 +16,8 @@
 
 #include <ConsoleHost.h>
 
+#include <net/PacketNames.h>
+
 #include <imgui.h>
 #include <imguivariouscontrols.h>
 
@@ -32,12 +34,11 @@ const int g_netOverlaySampleCount = 150;
 class RageHashList
 {
 public:
-	template<int Size>
-	RageHashList(const char* (&list)[Size])
+	RageHashList()
 	{
-		for (int i = 0; i < Size; i++)
+		for (const auto packet : net::PacketNames)
 		{
-			m_lookupList.insert({ HashRageString(list[i]), list[i] });
+			m_lookupList.emplace(packet);
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


LWSS:
Looks good, but while ur at it, maybe add the hash as a // comment so we can ctrl-f for it sometime later

code/shared/KnownPackets.h
```diff
@@ -0,0 +1,59 @@
+#pragma once
+
+#include <utility>
+#include "Utils.h"
+
+constexpr std::pair<uint32_t, const char*> g_knownPackets[]
```
blattersturm, line 6: `inline` might still be needed in case this ends up used twice in the same module.

code/shared/KnownPackets.h
```diff
@@ -0,0 +1,59 @@
+#pragma once
+
+#include <utility>
+#include "Utils.h"
```
blattersturm, line 4: I think this include is already assumed to exist fine.

code/shared/KnownPackets.h
```diff
@@ -0,0 +1,59 @@
+#pragma once
+
+#include <utility>
```
blattersturm, line 3: Same goes for this one.

blattersturm (changes requested):

AvarianKnight:
Applied the requested changes, also replaced g_knownPackets in NetDebug to use g_knownPackets

thorium-cfx (changes requested):

code/shared/KnownPackets.h
```diff
@@ -0,0 +1,70 @@
+#pragma once
+
+inline constexpr std::pair<uint32_t, const char*> g_knownPackets[]
```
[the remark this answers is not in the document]
thorium-cfx: Good call on the other code, had a proper look and seems like they're better optimized as-is for those use cases.

code/components/debug-net/src/NetDebug.cpp
```diff
@@ -32,12 +33,11 @@ const int g_netOverlaySampleCount = 150;
...
-	RageHashList(const char* (&list)[Size])
+	RageHashList()
 	{
-		for (int i = 0; i < Size; i++)
+		for (const auto packet : g_knownPackets)
 		{
-			m_lookupList.insert({ HashRageString(list[i]), list[i] });
+			m_lookupList.insert({ packet.first,  packet.second });
 		}
 	}
```
thorium-cfx, line 42: Seeing our lists are the same now, we can just refer to the original and not make a copy

thorium-cfx (changes requested):
2 more things that stand out
1. Can you move `KnownPackets.h` in a subfolder called `net` and put it in namespace `fx::net`?
2. See comment below

AvarianKnight:
This pretty simple PR has been sitting for 9 months, so I'm going to assume the changes are unwanted.

FabianTerhorst (approved):

PR #2144, opened 2023-08-05
tweak(gta-net-five): ignore automobile types for now

LWSS:


Keyinator:
I recommend a push to prod since this fixes https://github.com/citizenfx/fivem/issues/2145

PR #2148, opened 2023-08-05
Removed blocking of ReShade above 5.9

Jorbann:
Set original block of 5.9+ to just 5.9 so that users can update to 5.9.1 and have no crashing.

### DIFF
--- a/code/components/rage-graphics-five/src/ReShadeFixups.cpp
+++ b/code/components/rage-graphics-five/src/ReShadeFixups.cpp
@@ -67,9 +67,9 @@ bool IsValidGraphicsLibrary(const std::wstring& path)
 						}
 						// ReShade = 5.9 causes heap corruption basically *all the time* so we hard-block it
 						// ReShade 5.9.1 fixed this issue and blocking above 5.9 is no longer needed
-						else if (fixedInfo->dwProductVersionMS = 0x50009)
+						else if (fixedInfo->dwProductVersionMS == 0x50009)
 						{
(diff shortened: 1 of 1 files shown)


blattersturm:
This would still block 5.9.1, since it's not checking the least-significant-bit. Similarly, these commits are a mess.

We'll do this ourselves somewhen.

Jorbann:
🤷‍♀️ Looking at the code there, it's not a hard fix - especially considering that the issue in ReShade does not exist any longer.

The least you can do is fix the issue relativly soon, as the crashing is no longer present, and your user base is just about as illiterate as a small kangaroo - and the harassment towards our community when something is blocked in FiveM is beyond attrocious.

Telling individuals to contact the ReShade developer, rather than getting in contact with them yourselves is beyond petty and it puts every single one of us within harassment's way.

Site Contact Fourm - https://reshade.me/sitenotice
Discord Server - https://discord.gg/PrwndfH
ReShade Fourms - https://reshade.me/forum

blattersturm:
It's 'not a hard fix' sure but currently the build infrastructure is offline and it's much more difficult to fix up a blatantly wrong PR than to just implement stuff ourselves. We'd already been told about 5.9.1 fixing it, there's no need for you to harass with a pushy issue, PR and a lot of other weirdness either, _especially not at 1 AM on a Sunday when there's more critical issues we have to deal with like the platform being offline entirely_.

Your behavior here (are you even a 'developer' on their end, or just a CM?) also shows why we don't reach out to you guys either, too: not going to sign up to a forum that looks like it's questionable security-wise, nor join a Discord server to just get ganged up on by non-technical folks, and similarly, the user base harassment works both ways: we'd gotten a lot of weird hell from the broken release you guys shipped for over a week too.

Jorbann:
Good luck with that infrastructure being offline! Hope for the best :)
image

PR #2151, opened 2023-08-08
feat(resources/chat): Enable command toggleChat for RedM

outsider31000:
allow command toggleChat for RedM users to be able to controll the chat window

an addSuggestion was also added for users to know what states does it have as well.

cfx forum thread of the request 
https://forum.cfx.re/t/allow-togglechat-for-redm-users/5158193

### DIFF
--- a/ext/system-resources/resources/chat/cl_chat.lua
+++ b/ext/system-resources/resources/chat/cl_chat.lua
@@ -225,27 +225,21 @@ local kvpEntry = GetResourceKvpString('hideState')
 local chatHideState = kvpEntry and tonumber(kvpEntry) or CHAT_HIDE_STATES.SHOW_WHEN_ACTIVE
 local isFirstHide = true
 
-if not isRDR then
-  -- Allow only for Fivem until RedM supports it
-  if RegisterKeyMapping then
+if RegisterKeyMapping then
     RegisterKeyMapping('toggleChat', 'Toggle chat', 'keyboard', 'l')
-  end
 end
 
-  RegisterCommand('toggleChat', function(source,args,rawCommand)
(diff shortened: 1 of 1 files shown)


Ekinoxx0 (commented):
You should probably either : 
 - Wait for "chatLoaded" var to be true (repeat Wait(0) until chatLoaded)
 - Move this code into RegisterNUICallback('loaded', function() end)
 
 Random wait is not really useful to know what we are waiting for 👍

outsider31000:
i added a thread for that reason so it doesnt mess with the rest  it only waits for the function to be caeld, but you also right it can be done differently yeah

thorium-cfx:
This code needs to be cleaned up, before we can consider it. @Disquse could you tell me why `toggleChat` is disabled for RedM?

Disquse:
I guess because RedM doesn't support `RegisterKeyMapping` and other stuff, no idea.

outsider31000:
Yeah RegisterKeymapping is not used in RedM. This is an issue for redm users i have to constantly tell them what they need to do to remove chat from always visible  and since the command is non existente  they have no other choice than to modify the script.  @thorium-cfx

thorium-cfx:
This feature is dependent on KeyMapping to be enabled on RedM, will consider this PR once that's added

outsider31000:
@thorium-cfx  That is why is outside of RDR scope it will only apply to fivem
image

outsider31000:
Closing this pr in reference to #2152 
They are both for the same thing which is to allow RedM to use toggleChat command but #2152 has an extra feature  a small optimisation and chat suggestion for that command.
@Disquse

PR #2152, opened 2023-08-08
feat(resources/chat): Add `toggleChat` for RedM and extra parameter to set chat state

outsider31000:
This pr aims to allow the use of toggleChat command for RedM users.
It also allows an extra argument to use for chat states, for example when players have left session with state as `visible` since it uses ResourceKvp when they re enter it shows visble, this extra parameter allows to manipulate chat state in other resources
example:
`ExecuteCommand("toggleChat hidden")`

In summary:
- use of toggleChat command for RedM users
- extra parameter to allow resources to manipulate chat states
- adds a chat suggestion so players are aware of what states are supported
- smal optimisation on the look up for input for rdr gtav that was done inside the while loop
- removed isRDR check in the future if RegisterKeymapping is implemented in RedM  the check for `if RegisterKeymapping` should  and is enough, and a future pr wont be needed here.

### DIFF
--- a/ext/system-resources/resources/chat/cl_chat.lua
+++ b/ext/system-resources/resources/chat/cl_chat.lua
@@ -232,25 +232,25 @@ if RegisterKeyMapping then
 end
 
 RegisterCommand('toggleChat', function(source, args, rawCommand)
-	if not args[1] then
-		if chatHideState == CHAT_HIDE_STATES.SHOW_WHEN_ACTIVE then
-			chatHideState = CHAT_HIDE_STATES.ALWAYS_SHOW
-		elseif chatHideState == CHAT_HIDE_STATES.ALWAYS_SHOW then
-			chatHideState = CHAT_HIDE_STATES.ALWAYS_HIDE
-		elseif chatHideState == CHAT_HIDE_STATES.ALWAYS_HIDE then
-			chatHideState = CHAT_HIDE_STATES.SHOW_WHEN_ACTIVE
-		end
-	else
-		if args[1] == "visible" then
-			chatHideState = CHAT_HIDE_STATES.ALWAYS_SHOW
-		elseif args[1] == "hidden" then
-			chatHideState = CHAT_HIDE_STATES.ALWAYS_HIDE
-		elseif args[1] == "whenactive" then
-			chatHideState = CHAT_HIDE_STATES.SHOW_WHEN_ACTIVE
-		end
-	end
-	SetResourceKvp('hideState', tostring(chatHideState))
-	isFirstHide = false
+  if not args[1] then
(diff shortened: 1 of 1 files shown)


ext/system-resources/resources/chat/cl_chat.lua
```diff
@@ -227,34 +227,30 @@ local CHAT_HIDE_STATES = {
 local kvpEntry = GetResourceKvpString('hideState')
 local chatHideState = kvpEntry and tonumber(kvpEntry) or CHAT_HIDE_STATES.SHOW_WHEN_ACTIVE
 local isFirstHide = true
-
-if not isRDR then
```
outsider31000, line 229: this check is redundante  since we are already checking for if RegisterKeyMapping exists, so best to remove it.

outsider31000:
Closing this pr, as insignificant as it is, to allow RedM users to use `toggleChat` been here for months, makes no sense and basically due to the recente events of firing the only person that gave some hope for RedM users being fired the way he was fired.

outsider31000:
reopening this pr upon request.

outsider31000:
#2618  this also needs to be merged, upon testing the messages dont update unless you press pageup or pagedown due to the latest commits.
In other words you cant see new messages.

ext/system-resources/resources/chat/cl_chat.lua
```diff
@@ -225,39 +227,45 @@ local CHAT_HIDE_STATES = {
...
-      chatHideState = CHAT_HIDE_STATES.ALWAYS_SHOW
-    elseif chatHideState == CHAT_HIDE_STATES.ALWAYS_SHOW then
-      chatHideState = CHAT_HIDE_STATES.ALWAYS_HIDE
-    elseif chatHideState == CHAT_HIDE_STATES.ALWAYS_HIDE then
-      chatHideState = CHAT_HIDE_STATES.SHOW_WHEN_ACTIVE
-    end
-
-    isFirstHide = false
-
-    SetResourceKvp('hideState', tostring(chatHideState))
```
jakub-cfx, line 245: Why is the order of these flipped compared to the original function?

jakub-cfx (commented):
Looks reasonable, one small request I'd have is to fix the indents to match the rest of the file. The new code seems to unintentionally be indented twice.

ext/system-resources/resources/chat/cl_chat.lua
outsider31000: something I overlooked probably. i have put it where it was.

prikolium-cfx:
https://github.com/citizenfx/fivem/commit/8094e1e8b81c04bc7d35197a17404dc472397087
Squashed and merged.

PR #2156, opened 2023-08-13
tweak(build): nuke vcpkg for all projects

### LINKED ISSUE #2155: Vcpkg breaks build because of OpenSSL dependency
Title essentially. Causes linker errors.
image

duk-37:
Vcpkg breaks stuff because of OpenSSL. Fun.

Closes #2155


prikolium-cfx:
Can you add more details about this issue? When it happens and in what environment. Currently we don't have such issue in compilation process.

duk-37:
> Can you add more details about this issue? When it happens and in what environment. Currently we don't have such issue in compilation process.

The summary is exactly that simple as described in #2155: if you install pre-built openssl packages through vcpkg (`vcpkg install openssl:x64-windows`) you will get linker errors because vcpkg links them in by default.

PR #2165, opened 2023-08-20
feat(extra-natives/five): CScenarioConditionIsMultiplayerGame network check patch

z3t4s:
This was talked about in the CW discord and also formally requested in the [forums](https://forum.cfx.re/t/add-some-way-to-enable-special-car-jacking-animations/5158220).

It patches a virtual function inside the 'CScenarioConditionIsMultiplayerGame' vtable to allow a few CConditionalClipset's to execute in networked environments.
This seems to be exclusively used for alternative car entry/exit/jacking animations (a different condition called 'CScenarioConditionIsPlayerInMultiplayerGame' is used in other places for conditional animations and scenarios)

Note: This has been tested with two clients on 1604-2802 Debug and Release. This was the best pattern possible due to the generic layout of the check function, the ctor of the vft and the 'parser_Register' function

### DIFF
--- a/code/components/extra-natives-five/src/PatchNetworkChecks.cpp
+++ b/code/components/extra-natives-five/src/PatchNetworkChecks.cpp
@@ -15,11 +15,13 @@ static bool* g_flyThroughWindscreenDisabled;
 static bool* g_playerRagdollOnCollisionDisabled;
 static bool* g_playerJumpRagdollControlDisabled;
 static bool* g_dynamicDoorCreationDisabled;
+static bool* g_SingleplayerScenarioConditionsDisabled;
 
 static bool isFlyThroughWindscreenEnabledConVar = false;
 static bool isPlayerRagdollOnCollisionEnabledConVar = false;
 static bool isPlayerJumpRagdollControlEnabledConVar = false;
 static bool isDynamicDoorCreationEnabledConVar = false;
+static bool isSingleplayerScenarioConditonsEnabledConVar = false;
 
 static bool* isNetworkGame;
 
@@ -31,7 +33,8 @@ static HookFunction hookFunction([]()
 	static ConVar<bool> enablePlayerRagdollOnCollision("game_enablePlayerRagdollOnCollision", ConVar_Replicated, false, &isPlayerRagdollOnCollisionEnabledConVar);
 	static ConVar<bool> enablePlayerJumpRagdollControl("game_enablePlayerJumpRagdollControl", ConVar_Replicated, false, &isPlayerJumpRagdollControlEnabledConVar);
 	static ConVar<bool> enableDynamicDoorCreation("game_enableDynamicDoorCreation", ConVar_Replicated, false, &isDynamicDoorCreationEnabledConVar);
-	
+	static ConVar<bool> enableSingleplayerScenarioConditions("game_enableSingleplayerScenarioConditions", ConVar_Replicated, false, &isSingleplayerScenarioConditonsEnabledConVar);
+
 	// replace netgame check for fly through windscreen with our variable
 	{
 		g_flyThroughWindscreenDisabled = (bool*)hook::AllocateStubMemory(1);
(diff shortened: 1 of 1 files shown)


code/components/extra-natives-five/src/PatchNetworkChecks.cpp
```diff
@@ -60,13 +63,50 @@ static HookFunction hookFunction([]()
 		hook::put<int32_t>(location, (intptr_t)g_dynamicDoorCreationDisabled - (intptr_t)location - 4);
 	}
 
+	// Replace netgame check in CScenarioConditionIsMultiplayerGame::"check"
+	// This patch was designed to make SP car jacking animations work in MP (CConditionalClipSet)
+	//
+	{
+		uint64_t* result = 0;
+		auto parCguBuildStructureFromStaticData = hook::pattern("48 8D 15 ? ? ? ? 45 33 C9 44 8B C0 48 8B CB E8");
```
blattersturm, line 71: Patterns without a defined count or count_hint might make initial scan a little slow. Does the result count differ on different game builds?

code/components/extra-natives-five/src/PatchNetworkChecks.cpp
```diff
@@ -60,13 +63,50 @@ static HookFunction hookFunction([]()
...
+	}
+
 	OnKillNetworkDone.Connect([]()
 	{
 		se::ScopedPrincipal principalScopeInternal(se::Principal{ "system.internal" });
 		enableFlyThroughWindscreen.GetHelper()->SetValue("false");
 		enablePlayerRagdollOnCollision.GetHelper()->SetValue("false");
 		enablePlayerJumpRagdollControl.GetHelper()->SetValue("false");
 		enableDynamicDoorCreation.GetHelper()->SetValue("false");
+		enableSingleplayerScenarioConditions.GetHelper()->SetValue("false");
```
blattersturm, line 109: Unrelated to the review here, but I wonder if this could be simplified so that variables that should 'reset' per session don't have to be added in multiple different places.

code/components/extra-natives-five/src/PatchNetworkChecks.cpp
blattersturm: Right, I see what you're doing here now- you're iterating through structure registration routines in order to find `CScenarioConditionIsMultiplayerGame`, and go from _there_ to patch stuff beyond.

I don't have IDA or the game at hand here, but is there any more direct route to get there that won't need _this_?

Alternately, how many matches in does this target usually end up? One could use `.count()` to limit until a bit _after_ that point in time, as order usually matches across executable versions.

gottfriedleibniz: You could check if `49 2E 5B FA + 0x23c` works (or `04 00 01 00 49 2E 5B FA + 0x240`). Little bit of a hack, but I don't think (verification required) this has changed since 323.

LWSS: hmm something like this would work for 2372/2699, but would need different offset on 1604 I think. It is a pretty hard exact spot to sig.
```
		char* clonedScenarioInfoCtor = hook::get_pattern<char>("48 8D 05 ? ? ? ? 49 89 06 49 8B C6 48 83 C4 30 41 5E C3");
		char* vtable = hook::get_address<char*>(clonedScenarioInfoCtor, 3, 7);
		vtable -= 0xB90;
```

my only nitpick is that several of your asserts dont do anything - it's more likely that if something goes wrong, it's going to just be random junk

Rm1998:
So I assume this is going to allow MP characters do specific animations that they can't do in FiveM but can do in story? If so then that is pretty sweet and a much needed feature!

gottfriedleibniz:
I have noticed strange behaviour when trying to hijack remote vehicles (and drivers). While I have not fully debugged this, it does seem dependent on *when* the migration to the client doing the hijacking happens. 

Likely some task data being reset or not synchronized. This leads to a desynchronization between as one client will have the Ped in the driver seat and another client will not (executing a flee or motion task, iirc).

PR #2178, opened 2023-08-27
fix(clrcore): actually use sound ID when playing audio

DatGuy1:
The current implementation is incorrect and fetches a random sound ID at the end.

### DIFF
--- a/code/client/clrcore/External/Audio.cs
+++ b/code/client/clrcore/External/Audio.cs
@@ -90,35 +90,52 @@ namespace CitizenFX.Core
 		};
 		#endregion
 
+		/// <remarks>This method allocates a sound ID. It is imperative you release the sound ID when you are done using it through <see cref="ReleaseSound"/></remarks>
 		public static int PlaySoundAt(Vector3 position, string sound, string set)
 		{
-			API.PlaySoundFromCoord(-1, sound, position.X, position.Y, position.Z, set, false, 0, false);
-			return API.GetSoundId();
+			var soundId = API.GetSoundId();
+			API.PlaySoundFromCoord(soundId, sound, position.X, position.Y, position.Z, set, false, 0, false);
+			return soundId;
 		}
+
+		/// <inheritdoc cref="PlaySoundAt(Vector3, string, string)"/>
 		public static int PlaySoundAt(Vector3 position, string sound)
 		{
-			API.PlaySoundFromCoord(-1, sound, position.X, position.Y, position.Z, null, false, 0, false);
-			return API.GetSoundId();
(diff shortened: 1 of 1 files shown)


thorium-cfx:
Thank you for pointing this out.

A few concerns on the native usage:
It seems that `GetSoundId` is allocating a new sound id on each call, which can have a max of 100 at a time if my research results are correct. This requires users to call `ReleaseSoundId` afterwards to properly release the sound id as to not hit the max (-1 being returned if we hit the max). This would need to be communicated to the user, so adding docs to these methods with this mentioned in the `<remarks> is required. Could you make sure of this? You can use `<inheritdoc />` to prevent duplicates.

DatGuy1:
You mean to put the <remarks> on the Audio class and then inheritdoc on all the methods?

thorium-cfx:
Uhm no, document 1 method and then inheritdoc the others from the first method

code/client/clrcore/External/Audio.cs
```diff
@@ -90,36 +90,42 @@ public static class Audio
...
 		#endregion
 
+		/// <remarks>This method allocates a sound ID. It is imperative you release the sound ID when you are done using it through <see cref="ReleaseSound"/></remarks>
 		public static int PlaySoundAt(Vector3 position, string sound, string set)
 		{
 			var soundId = API.GetSoundId();
 			API.PlaySoundFromCoord(soundId, sound, position.X, position.Y, position.Z, set, false, 0, false);
 			return soundId;
 		}
+        /// <inheritdoc cref="PlaySoundAt(Vector3, string, string)"/>
```
thorium-cfx, line 100: Can you swap these with tab `\t` indentation like the rest of the file? While you're at it can you also add a new line between the methods as well? Will clean up things nicely.

thorium-cfx (changes requested):
Documentation looks great, thanks! Only left are some layout concerns

DatGuy1:
Done. I'll let you squash and merge it through GitHub.

A possibility would be making a disposable Sound class that allocates the ID on initialisation and releases the ID with a destructor, but that's an architectural question more up to you.

thorium-cfx:
> A possibility would be making a disposable Sound class that allocates the ID on initialisation and releases the ID with a destructor, but that's an architectural question more up to you.

Doesn't sound bad to me, though that'll be a separate task ;)

For now could you squash this yourself? We're still exploring options for squashing and keep proper track of the history.

DatGuy1:
Squashed.

thorium-cfx (approved):

nihonium-cfx (approved):

PR #2179, opened 2023-08-27
fix(streaming/native): fix issue with empty referenceHash from stream…

PichotM:
…ing natives

Updated version for this PR: https://github.com/citizenfx/fivem/pull/1751


code/components/citizen-resources-client/src/ResourceCacheDeviceV2.cpp
```diff
@@ -342,8 +342,8 @@ concurrency::task<RcdFetchResult> ResourceCacheDeviceV2::FetchEntry(const std::s
 	std::unique_lock<std::mutex> lock(ms_lock);
 
 	const auto& e = entry->get();
-	const auto& referenceHash = e.referenceHash;
-	auto it = ms_entries.find(referenceHash);
+	const auto& remoteHash = (e.referenceHash.empty()) ? e.remoteUrl.c_str() : e.referenceHash.c_str();
```
blattersturm, line 345: `.c_str()`?

code/components/citizen-resources-client/src/ResourceCacheDeviceV2.cpp
```diff
@@ -371,8 +371,8 @@ void ResourceCacheDeviceV2::UnfetchEntry(const std::string& fileName)
 		std::unique_lock<std::mutex> lock(ms_lock);
 
 		const auto& e = entry->get();
-		const auto& referenceHash = e.referenceHash;
-		auto it = ms_entries.find(referenceHash);
+		const auto& remoteHash = (e.referenceHash.empty()) ? e.remoteUrl.c_str() : e.referenceHash.c_str();
```
blattersturm, line 374: Same here.

blattersturm (changes requested):

PichotM:
i removed that

ktx-mega:
up

Ekinoxx0:
I just tried using the native without knowing the underlying issue (based a system of ugc ytds on it), then i noticed this pr and i'm now hopeful i won't have to trash my code if it's merged.
After testing this locally with a custom build, it seems to fix the issue without causing any side effects 👍

PichotM:
up

Lucas7yoshi:
I am a person who has ideas that would be most practically served through a native like this. (namely, storing copious .yvr files in a blob store like S3)
which is to say
bump

prikolium-cfx:
Thanks for the Pull Request and your interest in it, but as you know this native command is experimental and we don't want to extend this experiment now.

PR #2180, opened 2023-08-28
tweak(ros/accountid): EGS command-line blob string behavior

blattersturm:
This came up during an automated security scan. It's likely of no ill effect, but fixing rather than ignoring.

Launching RDR with an EGS copy was tested and worked fine.

### DIFF
--- a/code/components/ros-patches-five/src/AccountID.cpp
+++ b/code/components/ros-patches-five/src/AccountID.cpp
@@ -92,13 +92,18 @@ static std::string MapMessage(const nlohmann::json& j)
 	return message;
 }
 
+// this structure is passed via HostSharedData so should have a linear layout
 struct ExternalROSBlob
 {
 	uint8_t data[16384];
 	uint8_t steamData[64 * 1024];
 	size_t steamSize;
-	uint8_t epicData[64 * 1024];
+
+	// epicData is a copy of GetCommandLine result, so can't exceed the maximum length of a command line
+	// (https://devblogs.microsoft.com/oldnewthing/20031210-00/?p=41553, https://archive.today/nab64)
+	wchar_t epicData[UNICODE_STRING_MAX_CHARS];
 	size_t epicSize;
+
 	uint32_t steamAppId;
 	bool valid;
 	bool tried;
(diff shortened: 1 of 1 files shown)


code/components/ros-patches-five/src/AccountID.cpp
```diff
@@ -97,7 +98,7 @@ struct ExternalROSBlob
 	uint8_t data[16384];
 	uint8_t steamData[64 * 1024];
 	size_t steamSize;
-	uint8_t epicData[64 * 1024];
+	char epicData[64 * 1024];
```
thorium-cfx, line 101: Why not go for `std::string` instead? Looking at the code below, special functions are now put in place to keep everything working and null-terminated, truncation on `StringCbCopyA` and `StringCbCatA` also doesn't seem desirable.

thorium-cfx (changes requested):
Seems like it's better to go for dynamic allocated strings at this point.

code/components/ros-patches-five/src/AccountID.cpp
blattersturm: This structure is shared cross-process via HostSharedData, so has to be fixed-size and not contain any external pointers. `std::string` would be awfully unsuited for this.

The case where the command-line string will be truncated in and of itself is extremely unlikely since we launch this subprocess ourselves, the entire request itself was solely to meet the 'solve the automated security scan' bullet point.

Additionally, the 65535 length is because the process command line [can't exceed this length](https://devblogs.microsoft.com/oldnewthing/20031210-00/?p=41553). It can't possibly get truncated.

thorium-cfx: I see, thank you. Making comments in code on the last points should be wise then, don't need to look up this PR again :)

gottfriedleibniz: > extremely unlikely

Or impossible? If the former, some window-dressing/defensiveness (informal: log spam) could be added to `GetCommandLineAStub` to at least handle those silently routines failing. It would also provide feedback on:

> truncation on StringCbCopyA and StringCbCatA also doesn't seem desirable.

blattersturm:
Pushed a second pass reducing the static buffer use and ensuring that even if EGL launches with a very long command line we'll still have space.

PR #2185, opened 2023-09-01
tweak(core/sbac): cleanup code, return on read failing, add ReadBitsSafe

AvarianKnight:
This is a cleanup of PR #2161

This makes `HandlePacket` properly fail when a buffer read fails, and adds `ReadBitsSafe` which will check to make sure the length is greater then 0 & we won't read over m_maxBits before allocating the vector.


AvarianKnight:
Closing this as I'm not quite sure if this is necessary (or even safe)

PR #2186, opened 2023-09-03
Create locale-nn_NO.json

johanrong:
Adding support for Norwegian nynorsk.


blattersturm:
Translations aren't supposed to be submitted here.

PR #2189, opened 2023-09-06
tweak(scripting-mono-v2): bump pilot date

Nova-px:
Bump pilot date to 2023-11-31 because it's currently outdated

### DIFF
--- a/code/components/citizen-scripting-mono-v2/src/MonoScriptRuntime.cpp
+++ b/code/components/citizen-scripting-mono-v2/src/MonoScriptRuntime.cpp
@@ -263,11 +263,12 @@ int MonoScriptRuntime::HandlesFile(char* filename, IScriptHostWithResourceData* 
 	}
 
 	// last supported date for this pilot of mono_rt2, in UTC
-	constexpr int maxYear = 2023, maxMonth = 8, maxDay = 31;
+	constexpr int maxYear = 2023, maxMonth = 12, maxDay = 31;
 
 	// Allowed values for mono_rt2
 	constexpr std::string_view allowedValues[] = {
 		// put latest on top, right here ↓
+		"Prerelease expiring 2023-12-31. See https://aka.cfx.re/mono-rt2-preview for info."sv,
 		"Prerelease expiring 2023-08-31. See https://aka.cfx.re/mono-rt2-preview for info."sv,
 		"Prerelease expiring 2023-06-30. See https://aka.cfx.re/mono-rt2-preview for info."sv,
 	};


code/components/citizen-scripting-mono-v2/src/MonoScriptRuntime.cpp
```diff
@@ -263,11 +263,12 @@ int MonoScriptRuntime::HandlesFile(char* filename, IScriptHostWithResourceData*
 	}
 
 	// last supported date for this pilot of mono_rt2, in UTC
-	constexpr int maxYear = 2023, maxMonth = 8, maxDay = 31;
+	constexpr int maxYear = 2023, maxMonth = 11, maxDay = 31;
```
thorium-cfx, line 266: November has 30 days 😏 Let's put this to December (12) instead, too much going on right now.

thorium-cfx (changes requested):
Some date/day issues

blattersturm:
Replacing the first commit in this PR might be better - this is now a bit tricky to merge as it'll contain both commits.

AvarianKnight:
Is it a good idea to put the pilot day during the middle of the holidays?

Also, what is the point of date locking this vs just having it under an experimental flag (kind of like rdr), as while breakages are expected, it would be a better experience for the end user to  fix their resource for any breaking changes vs the entire runtime being disabled randomly. 

Possible flag:

`use_experimental_mono_v2_runtime 'I acknowledge that the mono v2 runtime will experience API & ABI breakages, and that I will be responsible to fix my resource(s) when these breakages do happen. ' `

blattersturm:
> Also, what is the point of date locking this

To prevent the situation from RDR occurring again, where this becomes a compatibility breaking point and the temporary situation is permanent.

AvarianKnight:
> > Also, what is the point of date locking this
> 
> To prevent the situation from RDR occurring again, where this becomes a compatibility breaking point and the temporary situation turns permanent - indeed, the RDR situation isn't changeable now without breaking everything made there since 2019.
> 
> It's not only resource _authors_ being affected, it's also people buying or downloading resources written by others. Having a hard date lock means there's at least a hard deadline for resources using the preview feature to 'break', discouraging longer-term use by people who can't maintain such.

Thanks for the explanation, this makes a lot more sense when explained like that.

DaniGP17:
Any news about this?

thorium-cfx:
> Any news about this?

We've had some technical difficulties that had to be solved first. This one is scheduled next.

Sorry for the inconvenience.

thorium-cfx (approved):

DaniGP17:
> > Any news about this?
> 
> We've had some technical difficulties that had to be solved first. This one is scheduled next.
> 
> Sorry for the inconvenience.

Oh no problem, sorry for bothering, good luck!

PR #2194, opened 2023-09-08
fix(scripting/v8): raw event listeners

william-des:
### Goal of this PR

Fixes issue where raw event listeners weren't triggered without standard subscribers.
Skips payload deserialization when no standard event subscribers are present.

Fixing these issues in js scripting event function

ScRT: JS

### Successfully tested on

**Game builds:** 3095

**Platforms:** Windows, Linux

### Checklist

### DIFF
--- a/data/shared/citizen/scripting/v8/main.js
+++ b/data/shared/citizen/scripting/v8/main.js
@@ -463,10 +463,21 @@ const EXT_LOCALFUNCREF = 11;
 				global.source = parseInt(source.substr(13));
 			}
 
-			const payload = unpack(payloadSerialized) || [];
-			const listeners = emitter.listeners(name);
+			// Running raw event listeners
+			try {
+				rawEmitter.emit(name, payloadSerialized, source);
+			} catch (e) {
+				console.error('Unhandled error during running raw event listeners', e);
+			}
 
-			if (listeners.length === 0 || !Array.isArray(payload)) {
+			const listeners = emitter.listeners(name);
+			if (listeners.length == 0) {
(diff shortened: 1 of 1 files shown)


data/shared/citizen/scripting/v8/main.js
```diff
@@ -463,30 +463,31 @@ const EXT_LOCALFUNCREF = 11;
...
-
-					if (retval instanceof Promise) {
-						(async () => {
-							try {
-								await retval;
-							} catch (e) {
-								console.error('Unhandled promise failure:', e);
-							}
-						})();
+			if (listeners.length > 0) {
```
thorium-cfx, line 467: Prefer to see this being applied as an early out, rather than more scoping. Especially to reduce the amount of lines being changed.

thorium-cfx (changes requested):

data/shared/citizen/scripting/v8/main.js
```diff
@@ -463,10 +463,21 @@ const EXT_LOCALFUNCREF = 11;
...
-			const payload = unpack(payloadSerialized) || [];
+			// Running raw event listeners
+			try {
+				rawEmitter.emit(name, payloadSerialized, source);
+			} catch (e) {
+				console.error('Unhandled error during running raw event listeners', e);
+			}
+
 			const listeners = emitter.listeners(name);
+			if (listeners.length == 0) {
```
c-wide, line 474: strict equality comparison?

FabianTerhorst (approved):
Tested and works fine 👍

PR #2195, opened 2023-09-10
feat(nui/resources): SEND_DUI_KEY_DOWN & SEND_DUI_KEY_UP

Mycroft-Studios:
This PR adds 2 natives:
SEND_DUI_KEY_DOWN
SEND_DUI_KEY_UP

These new natives allow you to inject keyboard events directly into the browser. which was previously:
A. impossible on any external sites
B. very janky and too specific (having to apply the code to each individual input) on internal UI

The need for this was also explained [on the forums](https://forum.cfx.re/t/dui-nui-request-for-send-dui-key-down-send-dui-key-up/5166146) and I have made a [preview video](https://youtu.be/oDJYPvRLcWM) to showcase the type of amazing things these natives can enable.

### DIFF
--- a/code/components/nui-core/include/CefOverlay.h
+++ b/code/components/nui-core/include/CefOverlay.h
@@ -371,7 +371,7 @@ namespace nui
 	bool OVERLAY_DECL HasMainUI();
 	void OVERLAY_DECL SetMainUI(bool enable);
 	void OVERLAY_DECL SetHideCursor(bool hide);
-
+	int  OVERLAY_DECL GetCefKeyboardModifiers(WPARAM wparam, LPARAM lparam);
 	void ProcessInput();
 
 	void OVERLAY_DECL ExecuteRootScript(const std::string& scriptBit);
--- a/code/components/nui-core/src/CefInput.cpp
+++ b/code/components/nui-core/src/CefInput.cpp
@@ -191,89 +191,89 @@ namespace nui
 	void ProcessInput()
 	{
 	}
-}
 
-int GetCefKeyboardModifiers(WPARAM wparam, LPARAM lparam)
-{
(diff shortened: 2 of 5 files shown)


blattersturm:
In addition to (or instead of?) this I think it'd also make sense to have some functionality to assign NUI focus *directly* to a specific DUI page, instead of using this to simulate user input in a potentially tacky way.

Mycroft-Studios:
> I think it'd also make sense to have some functionality to assign NUI focus _directly_ to a specific DUI page,

I agree with this, and this would be a better way of implementing it *but*, with the current way NUI focus works, that is not possible. I attempted multiple different ways of adding a "DUI focus", but since NUI focus currently applies focus to only the main Nui tab, and then to the iframe that belongs to that resource, it simply doesn't work with DUI, since each DUI is separate and not contained to a set frame.

> (and inconsistent! as different scripts would do this differently)

My implementation specifically focuses on ASCII (which is the same set of characters the FXServer supports anyway), and specifically the JavaScript keys, meaning u can send them either by capturing with NUI, or do manually with IsControlJustPressed.
 
 > Similarly, this doesn't handle things like text input across keyboard layouts well either
 Why would it not? it generates the scanCode based upon the users layout and "e" is still 101 in ascii, regardless of user layout?

And while I agree with the security concerns, the same can apply for normal NUI, and therefore would be a separate topic in its own right.

blattersturm:
> it simply doesn't work with DUI, since each DUI is separate and not contained to a set frame.

This would entail routing input events to that browser instead of the main browser, of course.

thorium-cfx:
Closing this as I agree with blattersturm's assessment. I've put the request in the backlog and we'll first evaluate the options and use cases.

PR #2198, opened 2023-09-11
increase FragmentStore pool

CDocc:
by increasing the FragmoreStore pool it will allow servers to be able to increase the limit on cars with mods, we have recently run into an issue after updating to the latest DLC build from GTA Online that having so many custom cars is now cause the client to crash while on loading screen & that there FragmentStore pool is full.


prikolium-cfx:
Better way already implemented in Canary branch. More information currently can be found there https://github.com/citizenfx/fivem-docs/pull/481

PR #2205, opened 2023-09-16
feat(extra-natives/five): LOAD_ROPE_DATA_FROM_PATH

packfile:
Adds a way to load a custom rope data file, useful for modifying/creating new ropes with additional textures, radii and colours as seen in https://i.imgur.com/1EpFIyP.png
Additionally fixes the game crashing if calling `ADD_ROPE` with an invalid index (8 onwards with default rope data loaded)

### DIFF
--- a/code/components/extra-natives-five/include/ropeManager.h
+++ b/code/components/extra-natives-five/include/ropeManager.h
@@ -46,6 +46,10 @@ public:
 	atArray<ropeData*> typeData; // 0x8
 
 	static ropeDataManager* GetInstance();
+
+	static void Init();
+
+	static void Shutdown();
 };
 static_assert(sizeof(ropeDataManager) == 0x18);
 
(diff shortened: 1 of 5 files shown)


Disquse (approved):
Looks good, played around these changes, work as expected, tested on all available game builds

thorium-cfx:
What would be reasons to add this? And why wouldn't we just load this prior to gameplay?

packfile:
> What would be reasons to add this? 

Someone I know wanted to make a rope that was thicker and used different textures for a script. Currently there is not a trivial way (afaik) to replace the default ropedata file as the path is hardcoded

> And why wouldn't we just load this prior to gameplay?

No particular reason, did this this way to match a [similar](https://github.com/citizenfx/fivem/blob/3e793d3b2b5ec400436d192533f9749b6d4a66ca/code/components/extra-natives-five/src/WaterExtraNatives.cpp#L734) native that was implemented

blattersturm:
> And why wouldn't we just load this prior to gameplay?

Resources can be started and stopped during gameplay as well.

PR #2206, opened 2023-09-17
Population ped creation event for RedM

Disquse:
Plus some code cleanup.
The event work in the same way as in FiveM (`populationPedCreating`), see [docs](https://docs.fivem.net/docs/scripting-reference/events/list/populationPedCreating/).

Example from docs with `CS_dutch` model:
image

### DIFF
--- a/code/components/extra-natives-five/src/RuntimeAssetNatives.cpp
+++ b/code/components/extra-natives-five/src/RuntimeAssetNatives.cpp
@@ -1072,23 +1072,6 @@ static hook::cdecl_stub<void*(fwEntityDef*, int fileIdx, fwArchetype* archetype,
 	return hook::get_call(hook::pattern("4C 8D 4C 24 40 4D 8B C6 41 8B D7 48 8B CF").count(1).get(0).get<void>(14));
 });
 
-static hook::cdecl_stub<fwArchetype*(uint32_t nameHash, uint64_t* archetypeUnk)> getArchetype([]()
-{
-	return hook::get_call(hook::pattern("89 44 24 40 8B 4F 08 80 E3 01 E8").count(1).get(0).get<void>(10));
-});
-
-fwArchetype* GetArchetypeSafe(uint32_t archetypeHash, uint64_t* archetypeUnk)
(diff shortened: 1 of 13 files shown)


code/components/extra-natives-five/src/RuntimeAssetNatives.cpp
```diff
@@ -1231,8 +1214,8 @@ static InitFunction initFunction([]()
 					fwEntityDef* entityDef = (fwEntityDef*)MakeStructFromMsgPack("CEntityDef", entityData);
 					mapData->entities.Set(i, entityDef);
 
-					uint64_t archetypeUnk = 0xFFFFFFF;
-					fwArchetype* archetype = GetArchetypeSafe(entityDef->archetypeName, &archetypeUnk);
+					rage::fwModelId modelId{ 0xFFFFFFF };
```
thorium-cfx, line 1217: Some more context on this `0xFFFFFFF` (7 Fs, magic number) and others down the line is necessary. Some ideas: a constant (in) or a static method in `rage::fwModelId` could also be an option.

thorium-cfx (changes requested):
Love the cleanup. Only one concern on the magic number/hex.

Would like to see @blattersturm's review on this as well.

code/components/extra-natives-five/src/RuntimeAssetNatives.cpp
blattersturm: `fwModelId` probably is a bitfield (28 bit handle, 4 bits for flags/type) and the handle being `-1` feels like a default for 'invalid'.

This could likely be a constructor, either default or with a magic type (like how nullptr_t/nullopt_t work).

blattersturm: Checking game code (fwMapData::ConstructEntity in companion), it appears the lower 16 bits are the model index, the 12 bits above are the index of a map types file, and that leaves 4 bits of flags.

thorium-cfx (approved):

PR #2207, opened 2023-09-18
docs(ext/natives): Change documentation SET_PED_CONFIG_FLAGS

BugCode1:
1. Adding a complete list of flags
2. Changing the maximum number of flags in b944


AvarianKnight:
I don't think these files are meant to be touched they're hashed for the specific fx_version (or whatever the legacy version of it is for _resource.lua)

You should use [the native docs](https://github.com/citizenfx/natives/blob/839863d349f90f3ab47c350dfd199ddc777ba702/PED/SetPedConfigFlag.md?plain=1#L4) to update SET_PED_CONFIG_FLAG

BugCode1:
mb, I didn't know that

PR #2217, opened 2023-09-25
feat(onesync): ped and player appearance data node research

packfile:
Adds research for `CPlayerAppearanceDataNode` and `CPedAppearanceDataNode` with most names documented. Code is behind a `#if 0` whilst the parsed data is unused

### DIFF
--- a/code/components/citizen-server-impl/include/state/SyncTrees_Five.h
+++ b/code/components/citizen-server-impl/include/state/SyncTrees_Five.h
@@ -2298,7 +2298,110 @@ struct CPedMovementGroupDataNode
 };
 
 struct CPedAIDataNode { };
-struct CPedAppearanceDataNode { };
+
+struct CPedAppearanceDataNode
+{
+	bool Parse(SyncParseState& state)
+	{
+#if 0
+		DeserializePedProps(state);
+
+		DeserializePedComponents(state);
+
+		int phoneMode = state.buffer.Read<int>(2);
+
+		bool isTakingOffVehicleHelmet = state.buffer.ReadBit();
+		bool isPuttingOnVehicleHelmet = state.buffer.ReadBit();
+		bool hasVehicleHelmet = state.buffer.ReadBit();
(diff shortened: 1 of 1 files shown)


Disquse (approved):
LGTM. For future researches it's better to use new generic serializer, however since this PR is mostly for documenting, I see no problems in using old way of serializing data nodes (assuming it will be reworked in the future anyway).

code/components/citizen-server-impl/include/state/SyncTrees_Five.h
```diff
@@ -3005,9 +3108,267 @@ struct CPlayerAppearanceDataNode
...
+		if (hasDamagePack)
+		{
+			uint32_t damagePackName = state.buffer.Read<uint32_t>(32);
+		}
+#endif
 
 		return true;
+	}
+
+	void SerialiseHeadBlend(SyncParseState& state)
```
thorium-cfx, line 3205: static?

code/components/citizen-server-impl/include/state/SyncTrees_Five.h
```diff
@@ -3005,9 +3108,267 @@ struct CPlayerAppearanceDataNode
...
+		if (hasParents)
+		{
+			int firstParentHead = state.buffer.Read<int>(8);
+			int secondParentHead = state.buffer.Read<int>(8);
+			float firstParentBlend = state.buffer.ReadFloat(8, 1.0f);
+			float secondParentBlend = state.buffer.ReadFloat(8, 1.0f);
+		}
+	}
+
+	void SerialisePedSecondaryTask(SyncParseState& state)
```
thorium-cfx, line 3305: static?

thorium-cfx (changes requested):
Thank you for this research, looks great. While reading it seems like we are deserializing here instead of serializing, but the function names say otherwise.

thorium-cfx (approved):

PR #2229, opened 2023-10-07
Improve player talking hook for FiveM

Disquse:
Patterns should future-proof this code from getting broken again, hopefully. Plus added a bit of explanation in the comments. Reworking this code to use `rage::netInterface::GetPlayerFromGamerId` might be considered in the future, however this function was heavily obfuscated in latest game builds, plus using it will add performance overhead. Simple memory layout hacks may work for some time.

Plus added a shared function to get net object from entity, it should decrease hard-to-read code duplication and ensure consistency.


blattersturm (approved):
LGTM.

adweex:
Oh boi, we are waiting for this update.

PR #2230, opened 2023-10-07
Ignore certain ped model sync blocking tunable in RedM

Disquse:
Yet another attempt to fix spawn of peds with "bad models" in RDO. Fortunately, we don't need such measures. But unfortunately, we don't have an obvious way to opt-out from these beforehand as for now, so it's always a build-by-build investigation.

Original report: https://forum.cfx.re/t/broken-human-peds-for-specific-models-and-other-ambient-ped-outfits-not-syncing/5111409


blattersturm (approved):
Changeset seems OK.

PR #2231, opened 2023-10-08
Enhanced Error Handling in 'MapDataContents_Remove' Function

Thunderwolfx:
More robust error handling to the MapDataContents_Remove function to address intermittent crashes observed. By adding a try-catch block around the function call that's been identified as problematic, we aim to capture and log any exceptions that arise. This will provide insights into the root causes of these exceptions and prevent abrupt crashes.

Changes:
1. Wrapped the g_origMapDataContents_Remove(contents, a2) function call inside a try-catch block.

2. Exceptions derived from std::exception will log their message to the console.

3. Other unknown exceptions will also be captured and logged, indicating an unidentified error.

4. In both cases, after logging the exception, the program will continue executing the subsequent lines instead of crashing, providing a more graceful failure mode.

### DIFF
--- a/code/client/launcher/GameCache.cpp
+++ b/code/client/launcher/GameCache.cpp
@@ -1067,15 +1067,15 @@ static bool PerformUpdate(const std::vector<GameCacheEntry>& entries)
 							return false;
 						}
 
-						fseek(f, 0, SEEK_END);
+						_fseeki64(f, 0, SEEK_END);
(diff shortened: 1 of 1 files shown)


gottfriedleibniz:
The code was not merged. `extra-natives-five` is not configured to catch asynchronous exceptions w/ `catch(...)` anyway so the PR as-is does not seem particularly useful.

PR #2237, opened 2023-10-12
feat(onesyn/events): add ragdoll request event

goncalobsccosta:
Since some cheats have now an option to ragdoll any player server owners should be able to cancel that event.

My knowledge is limited, and I don't know how to figure out the event data. I put a placeholder so that someone with that knowledge should be able to replace that.

### DIFF
--- a/code/components/citizen-server-impl/src/state/ServerGameState.cpp
+++ b/code/components/citizen-server-impl/src/state/ServerGameState.cpp
@@ -5163,6 +5163,22 @@ struct CRemoveAllWeaponsEvent
 	MSGPACK_DEFINE_MAP(pedId);
 };
 
+struct CRagdollRequestEvent
+{
+	void Parse(rl::MessageBuffer& buffer)
+	{
+	}
+
+	inline std::string GetName()
+	{
+		return "ragdollRequestEvent";
+	}
+
+	int placeHolder;
+
+	MSGPACK_DEFINE_MAP(placeHolder);
+};
+
 /*NETEV startProjectileEvent SERVER
 /#*
  * Triggered when a projectile is created.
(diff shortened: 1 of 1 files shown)


AvarianKnight:
Do you have reproduction for actually causing other players to ragdoll?

goncalobsccosta:
Hi @AvarianKnight, I didn't figure out how to do it with natives, but I uploaded a clip so you can see https://streamable.com/f58m64

Yum1x:
> Hi @AvarianKnight, I didn't figure out how to do it with natives, but I uploaded a clip so you can see https://streamable.com/usr93i

It isn't a client-side effect? Did u confirmed if the target players really takes a ragdoll?

goncalobsccosta:
It does affect the target, because in that clip I'm the target and I asked the cheater to take the clip!

william-des:
This event contains only one value: the network id of the target entity. It can be parsed in this manner:

```
void Parse(rl::MessageBuffer& buffer)
{
	entityNetId = buffer.Read<uint16_t>(13);
}

```
However, given the recent feedback on PRs concerning game events, I believe the recommended approach here would be to use a convar to disable the routing of these events, similar to how it's done for the RequestControlEvent and NetworkPlaySoundEvent. Optionally, offer a mode that blocks the events only if the target is a player, and another one blocking for both players and NPCs.

MoskalykA:
> Do you have reproduction for actually causing other players to ragdoll?

I can confirm this, I can give you a replay of a stream where the streamer is ragdoll many times.

goncalobsccosta:
Ok, I added the correct parameter. Can someone give me feedback if this method is ok? Or do I have to make this work like @william-des said?

WattSkill:
wouldnt it be the same as https://github.com/citizenfx/fivem/commit/0fa984209cedbebe5e962ea4fb2a36794f1fce16 with REQUEST_CONTROL_EVENT?

goncalobsccosta:
This was fixed with [468a8fa](https://github.com/citizenfx/fivem/commit/468a8fa50d2073cbe23b0b5a063b53130d26d5d9)

PR #2238, opened 2023-10-12
Train config index crash workaround for RedM

Disquse:
Fixes a base game crash caused by an invalid train config index on remote train net objects. Sentry crash identifier: `REDM-18W4` (b1491).

### DIFF
--- a/code/components/gta-core-rdr3/src/PatchTrainConfigCrash.cpp
+++ b/code/components/gta-core-rdr3/src/PatchTrainConfigCrash.cpp
@@ -1,0 +1,63 @@
+#include <StdInc.h>
+
+#include "atArray.h"
+#include "Hooking.h"
+#include "Hooking.Stubs.h"
+
+struct CTrainToken
+{
+	void* vtable;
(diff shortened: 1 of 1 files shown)


code/components/gta-core-rdr3/src/PatchTrainConfigCrash.cpp
```diff
@@ -0,0 +1,63 @@
+#include <StdInc.h>
+
+#include "atArray.h"
+#include "Hooking.h"
+#include "Hooking.Stubs.h"
```
blattersturm, line 5: Does this file need `Hooking.Stubs.h`?

blattersturm (changes requested):

code/components/gta-core-rdr3/src/PatchTrainConfigCrash.cpp
Disquse: Yeah, because I'm using `hook::trampoline`.

thorium-cfx (approved):

PR #2240, opened 2023-10-14
feat(devtools/five): archetype list

packfile:
Adds a tool to view statistics related to loaded archetypes, this can be used by server developers to help diagnose one of the causes of LOD/SLODs displaying (aka ‘playdough’), as mentioned in https://forum.cfx.re/t/too-many-archetypes-leading-to-lod-slods-displaying.

The UI displays the distribution of archetypes across different model types (base, vehicle, weapon, ped, etc) to guide where a developer should look to resolving the issue. Also, the names of loaded YTYPs and their contribution towards the overall count are included, so one can see what content is problematic or unnecessarily loaded (looking at those that glob everything with a `DLC_ITYP_REQUEST`).

Example from an empty vanilla server:
FiveM_b2545_GameProcess_H2PU2BIe4j

Example from a heavily modded server with addon maps, vehicles and weapons:
image


BugCode1:
So, is this problem related to the limitation of the number of archetypes? And if their amount is full 100%, how can this problem be solved?

packfile:
> So, is this problem related to the limitation of the number of archetypes? And if their amount is full 100%, how can this problem be solved?

Not necessarily, there are other ways to cause similar behaviour (notably, issues arising from duplicate archetype definitions); the linked forum post and related topics have more in depth discussion on this

BugCode1:
It is true. But there is no exact solution to solve this problem anywhere, and the question is, is this a general limit for archetypes or can this problem be solved by separating ytyps into several ytyps or creating several different resources for each ytyp?

blattersturm:
> is this a general limit for archetypes 

It's a limit for archetypes _loaded at the same time_.

> It is true. But there is no exact solution to solve this problem anywhere, and the question is, is this a general limit for archetypes or can this problem be solved by separating ytyps into several ytyps or creating several different resources for each ytyp?

If you load the type files at the same time anyway, this won't be helped if split.

BugCode1:
> > is this a general limit for archetypes
> 
> It's a limit for archetypes _loaded at the same time_.
> 
> > can this problem be solved by separating ytyps into several ytyps or creating several different resources for each ytyp?
> 
> If you load the type files at the same time anyway, this won't be helped if split.

Is it possible to increase this limit?

code/components/devtools-five/src/ArchetypeDebug.cpp
```diff
@@ -0,0 +1,292 @@
+#include <StdInc.h>
+#ifdef GTA_FIVE
```
Disquse, line 2: I wonder why you decided to guard this code? This is a `-five` component :P

code/components/devtools-five/src/ArchetypeDebug.cpp
```diff
@@ -0,0 +1,292 @@
...
+								delta = SortCompare(left.name, right.name);
+								break;
+							case 1:
+								delta = SortCompare(left.count, right.count);
+								break;
+							case 2:
+								delta = SortCompare(left.usage, right.usage);
+								break;
+							}
+							if (delta > 0)
```
Disquse, line 200: We really need brackets here :(

code/components/devtools-five/src/ArchetypeDebug.cpp
```diff
@@ -0,0 +1,292 @@
...
+							switch (sortSpec->ColumnIndex)
+							{
+							case 0:
+								delta = SortCompare(left.name, right.name);
+								break;
+							case 1:
+								delta = SortCompare(left.count, right.count);
+								break;
+							}
+							if (delta > 0)
```
Disquse, line 261: And here!

code/components/devtools-five/src/ArchetypeDebug.cpp
```diff
@@ -0,0 +1,292 @@
...
+	});
+
+	ConHost::OnDrawGui.Connect([]()
+	{
+		if (!archetypeListEnabled)
+		{
+			return;
+		}
+
+		if (g_archetypeFactories->GetSize() == 0 || g_archetypeFactories->GetCount() < 7)
```
Disquse, line 126: Why there's a check for `< 7` ? Would be nice to have this explained or documented in the code.

code/components/devtools-five/src/ArchetypeDebug.cpp
```diff
@@ -0,0 +1,292 @@
...
+			"CWeaponModelInfo",
+			"CVehicleModelInfo",
+			"CPedModelInfo",
+			"CCompEntityModelInfo",
+		};
+
+		std::unordered_map<uint32_t, FactoryBucket> factoryBucketsLookup;
+
+		size_t archetypePoolSize = *(uint32_t*)(((char*)rage__fwArchetypeManager__ms_ArchetypePool) + 0x8);
+		size_t archetypePoolFree = *(uint32_t*)(((char*)rage__fwArchetypeManager__ms_ArchetypePool) + 0x10);
```
Disquse, line 146: So our `atPoolBase` code completely lack support for raw values in "flags" (aka "free count" or so) field, we definitely need to improve this in the future.

Disquse (changes requested):

code/components/devtools-five/src/ArchetypeDebug.cpp
packfile: > There's multiple factories and we're waiting for everything to be initialized?

Yeah, I've amended the code to be more clear on this

code/components/devtools-five/src/ArchetypeDebug.cpp
packfile: Removed 😅

Disquse (approved):
Looks good, tested a bit, work as intended.

code/components/devtools-five/src/ArchetypeDebug.cpp
thorium-cfx: @Disquse can you make a task on this?

PR #2243, opened 2023-10-17
tweak(server): scriptless startup notices

jakub-cfx:
This PR changes the server startup notice handler from Lua to a tailored logic processor written in C++. The needs of the notice system were reviewed before the change and the new processor's feature set should cover all that's needed.

Please make sure to update the dynamically served JSON to adhere to the new schema before shipping an FXServer update, otherwise there will be no data for the new notice internals to process. I can provide the updated JSON if necessary.

### DIFF
--- a/code/components/citizen-server-impl/include/NoticeLogicProcessor.h
+++ b/code/components/citizen-server-impl/include/NoticeLogicProcessor.h
@@ -1,0 +1,60 @@
+#pragma once
+
+#include <StdInc.h>
+#include <ServerInstanceBase.h>
+#include <ResourceManager.h>
+
+#include <json.hpp>
+
+namespace fx
+{
+class NoticeLogicProcessor
+{
+private:
+	enum RuleOp
+	{
+		And,
+		Or,
+		Not,
+		NullOrEmpty,
+		Contains,
+		Equals
+	};
+
+	enum RuleType
+	{
+		ConVar,
+		StartedResourceList
+	};
+
+public:
+	NoticeLogicProcessor(fx::ServerInstanceBase* server);
+	bool ProcessNoticeRule(const nlohmann::json& ruleRef, uint32_t nestingLevel) const;
+
+	static void BeginProcessingNotices(fx::ServerInstanceBase* server, const nlohmann::json& noticesBlob);
+
+private:
+	ConsoleVariableManager* m_cvManager;
+	fx::ResourceManager* m_resManager;
+
+	// Macro'd map entries to prevent accidental typos or bad mapping
+#define StrEnumMapEntry(EnumName, ValueName) \
+	{                                        \
+		#ValueName, EnumName::ValueName      \
+	}
+
+	const std::map<std::string, RuleOp> m_ruleOpStringToEnum{
+		StrEnumMapEntry(RuleOp, And),
+		StrEnumMapEntry(RuleOp, Or),
+		StrEnumMapEntry(RuleOp, Not),
+		StrEnumMapEntry(RuleOp, NullOrEmpty),
+		StrEnumMapEntry(RuleOp, Contains),
+		StrEnumMapEntry(RuleOp, Equals)
+	};
+
+	const std::map<std::string, RuleType> m_ruleTypeStringToEnum{
+		StrEnumMapEntry(RuleType, ConVar),
+		StrEnumMapEntry(RuleType, StartedResourceList),
+	};
+};
+}
--- a/code/components/citizen-server-impl/src/NoticeLogicProcessor.cpp
+++ b/code/components/citizen-server-impl/src/NoticeLogicProcessor.cpp
@@ -1,0 +1,158 @@
+#include <StdInc.h>
+#include "NoticeLogicProcessor.h"
+
+fx::NoticeLogicProcessor::NoticeLogicProcessor(fx::ServerInstanceBase* server)
+{
+	m_cvManager = server->GetComponent<console::Context>()->GetVariableManager();
+	m_resManager = server->GetComponent<fx::ResourceManager>()->GetCurrent();
+}
+
+bool fx::NoticeLogicProcessor::ProcessNoticeRule(const nlohmann::json& ruleRef, uint32_t nestingLevel) const
+{
+	if (nestingLevel >= 10)
+	{
+		throw std::invalid_argument("Maximum nesting level for notice rules was exceeded");
+	}
+	if (!ruleRef.is_object())
+	{
+		return false;
+	}
+	auto ruleObject = ruleRef.get<nlohmann::json::object_t>();
+
+	auto& opObject = ruleObject["op"];
+	if (!opObject.is_string())
+	{
+		return false;
+	}
+
+	auto mappedOp = m_ruleOpStringToEnum.find(opObject.get<std::string>());
+	if (mappedOp == m_ruleOpStringToEnum.end())
+	{
+		return false;
+	}
+
+	auto opNum = mappedOp->second;
+	switch (opNum)
+	{
+		case RuleOp::And:
+		case RuleOp::Or:
+		{
+			auto& nestedRules = ruleObject["rules"];
+			if (!nestedRules.is_array())
+			{
+				return false;
+			}
+
+			for (auto& subRule : nestedRules)
+			{
+				auto isSubRuleTrue = ProcessNoticeRule(subRule, nestingLevel + 1);
+				if (opNum == RuleOp::And)
+				{
+					if (!isSubRuleTrue)
+					{
+						return false;
+					}
+				}
+				else // ::Or
+				{
+					if (isSubRuleTrue)
+					{
+						return true;
+					}
+				}
+			}
+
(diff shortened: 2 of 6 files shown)


code/components/citizen-server-impl/src/NoticeLogicProcessor.cpp
```diff
@@ -0,0 +1,149 @@
...
+{
+	m_cvManager = server->GetComponent<console::Context>()->GetVariableManager();
+
+	server->GetComponent<fx::ResourceManager>()->ForAllResources([this](const fwRefContainer<fx::Resource>& resource)
+	{
+		if (resource->GetState() == fx::ResourceState::Started)
+		{
+			m_resourceList.push_back(resource->GetName());
+		}
+	});
```
thorium-cfx, line 16: Can you put this into `fx::ResourceManager` like discussed?

thorium-cfx (changes requested):
Missing something, will properly review when that's done!

code/components/citizen-server-impl/src/NoticeLogicProcessor.cpp
jakub-cfx: Toyed around with it a bit and ended up ditching the local cache & instead chose to tap into ResourceManager's GetResource within the logic processor loop.

Implementing a limited use resource identity/name cache into the manager itself didn't really sit well with me. If we chose to go that route we'd need to outline when such cache should be refreshed or what its lifetime is supposed to be in general, which feels a bit overkill at the moment.

Even without this cache vector evaluation is still very fast - the latest commit includes this 😄

code/components/citizen-server-impl/src/NoticeLogicProcessor.cpp
```diff
@@ -0,0 +1,140 @@
...
+NoticeLogicProcessor::NoticeLogicProcessor(fx::ServerInstanceBase* server)
+{
+	m_cvManager = server->GetComponent<console::Context>()->GetVariableManager();
+	m_resManager = server->GetComponent<fx::ResourceManager>()->GetCurrent();
+}
+
+bool NoticeLogicProcessor::ProcessNoticeRule(json& ruleRef, uint32_t nestingLevel)
+{
+	if (nestingLevel >= 10)
+		throw std::invalid_argument("Maximum nesting level for notice rules was exceeded");
```
thorium-cfx, line 13: We need to catch this somewhere

code/components/citizen-server-impl/src/NoticeLogicProcessor.cpp
```diff
@@ -0,0 +1,140 @@
...
+	if (!ruleRef.is_object())
+		return false;
```
thorium-cfx, line 15: Style guide: use braces everywhere and group them by functionality :)
For some instances down below; you can use ternary operators or return the condition check straight away

code/components/citizen-server-impl/include/NoticeLogicProcessor.h
```diff
@@ -0,0 +1,59 @@
...
+	ConsoleVariableManager* m_cvManager;
+	fx::ResourceManager* m_resManager;
+
+	// Macro'd map entries to prevent accidental typos or bad mapping
+#define StrEnumMapEntry(EnumName, ValueName) \
+	{                                        \
+		#ValueName, EnumName::ValueName      \
+	}
+
+	std::map<std::string, RuleOp> m_ruleOpStringToEnum{
```
thorium-cfx, line 46: better `const` these

code/components/citizen-server-impl/include/NoticeLogicProcessor.h
```diff
@@ -0,0 +1,59 @@
...
+
+	enum RuleType
+	{
+		ConVar,
+		StartedResourceList
+	};
+
+public:
+	NoticeLogicProcessor(fx::ServerInstanceBase* server);
+	bool ProcessNoticeRule(json& ruleRef, uint32_t nestingLevel);
```
thorium-cfx, line 32: It looks like this one is not making changes to either the processor instance nor the json file, we could then declare it as: `bool ProcessNoticeRule(const json& ruleRef, uint32_t nestingLevel) const;`.

The static function can then also be changed to a `const json&`  parameter. I don't think `ServerInstanceBase` is correctly set up to be const-qualified yet, otherwise I'd like to see that one `const` as well.

code/components/citizen-server-impl/include/NoticeLogicProcessor.h
```diff
@@ -0,0 +1,59 @@
...
+
+#include <StdInc.h>
+#include <ServerInstanceBase.h>
+#include <ResourceManager.h>
+
+#include <json.hpp>
+
+using json = nlohmann::json;
+
+class NoticeLogicProcessor
```
thorium-cfx, line 11: We'll be moving to namespacing everything, so best to apply it here as well.

code/components/citizen-server-impl/include/NoticeLogicProcessor.h
```diff
@@ -0,0 +1,59 @@
+#pragma once
+
+#include <StdInc.h>
+#include <ServerInstanceBase.h>
+#include <ResourceManager.h>
+
+#include <json.hpp>
+
+using json = nlohmann::json;
```
thorium-cfx, line 9: better put this one in the class itself, also better for consumers of this processor, hence: `NoticeLogicProcessor::json`

thorium-cfx (changes requested):

code/components/citizen-server-impl/src/NoticeLogicProcessor.cpp
thorium-cfx: Your solution is what I had more or less in mind. My idea was to just let it find the resource and return its status or another status if it didn't exist. But making this processor agnostic to the resource system was my main goal, which looks a lot better now. Getting rid of these caches was my second goal, the use of those are generally not needed.

So good job!

code/components/citizen-server-impl/src/NoticeLogicProcessor.cpp
jakub-cfx: NoticeLogicProcessor exceptions are caught by its consumer, [DownloadAndProcessNotices in ServerNucleus.cpp](https://github.com/jakub-cfx/fivem/blob/tweak/server-notice-logic/code/components/citizen-server-impl/src/ServerNucleus.cpp#L35)
I could move the try-catch from there to BeginProcessingNotices so its contained within NoticeLogicProcessor but I feel like the handling of exceptions should be left to the highest level consumer

code/components/citizen-server-impl/include/NoticeLogicProcessor.h
jakub-cfx: I'm not entirely sure I understand, if I move the `#include <json.hpp>` from the processor's .h to the .cpp I lose the ability to declare function prototypes with JSON types in the argument list. There's no separate `json` class that NoticeLogicProcessor declares or implements, I just stick to `using json = nlohmann:json;` because it makes referencing the 3rd party (nlohmann) types less lengthy and it's what [ServerNucleus.cpp did too before](https://github.com/citizenfx/fivem/blob/master/code/components/citizen-server-impl/src/ServerNucleus.cpp#L25) this change.

code/components/citizen-server-impl/include/NoticeLogicProcessor.h
jakub-cfx: Would `fx` be the namespace of choice for this?

code/components/citizen-server-impl/src/NoticeLogicProcessor.cpp
jakub-cfx: Added the braces and made use of ternary operators where applicable 👍

code/components/citizen-server-impl/include/NoticeLogicProcessor.h
jakub-cfx: Added const where applicable, no part of the current codebase uses ServerInstanceBase in conjunction with const so I left that one as is.

code/components/citizen-server-impl/include/NoticeLogicProcessor.h
thorium-cfx: yes that's fine, we'll get to others later when things are better designed

code/components/citizen-server-impl/src/NoticeLogicProcessor.cpp
thorium-cfx: Looks good, as this is no game or performance heavy code I think it's fine to keep it like this.

code/components/citizen-server-impl/include/NoticeLogicProcessor.h
thorium-cfx: Alright, next time you can just put the `using ...` or `typedef ...` in the class/struct itself. Look at `std` types for a reference

thorium-cfx (approved):

PR #2244, opened 2023-10-17
feat(gamestate/fivem): add a ConVar to disable REQUEST_RAGDOLL_EVENT

AvarianKnight:
This is a follow up to some of the discussion from #2237

This adds a ConVar for server owners to disable ragdoll requests `set sv_enableRagdollRequests false`.

### DIFF
--- a/code/components/citizen-server-impl/src/state/ServerGameState.cpp
+++ b/code/components/citizen-server-impl/src/state/ServerGameState.cpp
@@ -99,6 +99,9 @@ static std::shared_ptr<ConVar<int>> g_requestControlSettleVar;
 static RequestControlFilterMode g_requestControlFilterState;
 static int g_requestControlSettleDelay;
 
+static std::shared_ptr<ConVar<bool>> g_enableRagdollRequestsVar;
+static bool g_enableRagdollRequests;
+
 static uint32_t MakeHandleUniqifierPair(uint16_t objectId, uint16_t uniqifier)
 {
 	return ((uint32_t)objectId << 16) | (uint32_t)uniqifier;
@@ -6773,7 +6776,15 @@ std::function<bool()> fx::ServerGameState::GetGameEventHandler(const fx::ClientS
 		return GetRequestControlEventHandler(client, std::move(buffer));
 	}
 
-	if(eventType == NETWORK_PLAY_SOUND_EVENT)
+	if (eventType == RAGDOLL_REQUEST_EVENT)
+	{
+		return []
+		{
+			return g_enableRagdollRequests;
+		};
+	}
+
+	if (eventType == NETWORK_PLAY_SOUND_EVENT)
 	{
 		return []()
 		{
@@ -6869,6 +6880,7 @@ static InitFunction initFunction([]()
 		}
 
 		g_networkedSoundsEnabledVar = instance->AddVariable<bool>("sv_enableNetworkedSounds", ConVar_None, true, &g_networkedSoundsEnabled);
+		g_enableRagdollRequestsVar = instance->AddVariable<bool>("sv_enableRagdollRequests", ConVar_None, true, &g_enableRagdollRequests);
 
 		g_networkedPhoneExplosionsEnabledVar = instance->AddVariable<bool>("sv_enableNetworkedPhoneExplosions", ConVar_None, false, &g_networkedPhoneExplosionsEnabled);
 


code/components/citizen-server-impl/src/state/ServerGameState.cpp
```diff
@@ -6869,6 +6914,7 @@ static InitFunction initFunction([]()
 		}
 
 		g_networkedSoundsEnabledVar = instance->AddVariable<bool>("sv_enableNetworkedSounds", ConVar_None, true, &g_networkedSoundsEnabled);
+		g_enableRagdollRequestsVar = instance->AddVariable<bool>("sv_enableRagdollRequests", ConVar_None, true, &g_enableRagdollRequests);
```
4tl4nt4, line 6918: hank yourself

AvarianKnight:
The initial implementation which filtered to not allow this for player peds specifically would break game functionality (i.e. stun guns, and possibly other stuff).

Players who disable `sv_enableRagdollRequests` are expected to handle syncing stun gun shots themselves via `weaponDamageEvent`

goncalobsccosta:
Hi, @AvarianKnight I think this should not break game functionality, I've tested with stun guns and everything is working fine with my code https://github.com/citizenfx/fivem/pull/2237 (and canceling the event)

AvarianKnight:
> Hi, @AvarianKnight I think this should not break game functionality, I've tested with stun guns and everything is working fine with my code #2237 (and canceling the event)

Stun gun was not working when testing this, are you you had it working?

packfile:
The video below shows a few cases where this event is triggered in gameplay (a [trace](https://gist.github.com/packfile/374ef2fb0b1b9fe7257ca47300361c37) has been added added for clarity): https://streamable.com/g6z3nr. Clip contains ramming peds with a vehicle, using the up-n-atomiser, punching peds, falling over in rapids, shooting peds, and setting them alight to mention a few.

Unsure if disabling this event would have any negative impact on the above, so it's worth the PR author(s) verifying this?

blattersturm:
Is it possible for us to somehow validate expected use cases here? A toggle like this seems risky given the side effects.

AvarianKnight:
> Is it possible for us to somehow validate expected use cases here? A toggle like this seems risky given the side effects.

Expected use case is to block players from exploiting the game event

AvarianKnight:
Going to post some previous discussion from the engineering discord:

Tasing a ped you don't own does a `WEAPON_DAMAGE_EVENT` (Player or Regular)

If you tase a regular ped you go through a weird chain:
1. Player 2 tases Ped
2. Player 2 Sends out a `WEAPON_DAMAGE_EVENT`
3. For Player 1 the ped is tased instantly, Player 2 has to wait for Player 1 to do a `REQUEST_RAGDOLL_EVENT` even though they no longer own it
4. Player 2 sees Ped being tased half a second or more later

If you tase a Player everything happens pretty much instantly.

https://github.com/citizenfx/fivem/assets/13681939/a4ffc8b5-563f-4dd7-92c9-d6fd5584931b

A weird quirk is that it doesn't matter if it gets canceled, at least for `WEAPON_TASER`, you need to actually handle the weaponDamageEvent to cancel it

CiastekbatakPro:
Messing with ragdoll request event isn't required at all to block the cheat functionalities in most cases you can use following code

```lua
CreateThread(function()
while true do
        if IsPedInAnyVehicle(PlayerPedId(), false) or IsPedFalling(PlayerPedId()) then
            SetPedCanRagdoll(PlayerPedId(), true)
        Wait(10000)
        else
            SetPedCanRagdoll(PlayerPedId(), false)
        end
        Wait(100)
    end
 end)
```
 eventually adjust it to your scripts
 I hope I was helpful

AvarianKnight:
I would like the opinion of @blattersturm here, as there's a bit of a cross roads.

1.Users will complain that there is no way to block this abuse by third-party code.
2. Users will complain that certain parts of expected base game functionality don't work

The end-user will complain either way, I think that having a specific way to block this abuse is probably the better idea even with the draw backs.

That being said, I think if the user opts-in to this possibly breaking change they should be greeted with a warning message that cannot be disabled so they know (even if they copy paste from somewhere else) that this can possibly break functionality.

goncalobsccosta:
Hi @AvarianKnight, I think with the PR that I made users have the ability to customize the blocked cases. It would'nt be just an on/off switch!

Tell me what you think.

AvarianKnight:
> Hi @AvarianKnight, I think with the PR that I made users have the ability to customize the blocked cases. It wouldn't be just an on/off switch!
> 
> Tell me what you think.

Newer event handlers aren't supposed to expose the event.

blattersturm:
In the worst case scenario since this is very game logic dependent, it might be viable to expose an event anyway. 

Alternately, filtering this in client-side code to only act in legitimate use cases (or filter out blatantly illegitimate use cases) might also help.

AvarianKnight:
Closing this as this PR didn't resolve the initial issue, likely cases of people abusing this are via `weapon_stungun` and `weapon_stungun_mp` which users can block via [weaponDamageEvent](https://docs.fivem.net/docs/scripting-reference/events/server-events).

AvarianKnight:
Reopening as @tens0rfl0w mentioned this *did* fix some cases of abuse

AvarianKnight:
This was fixed with https://github.com/citizenfx/fivem/commit/468a8fa50d2073cbe23b0b5a063b53130d26d5d9

PR #2248, opened 2023-10-18
Update IsPedStrafing.md

Mani2421:


ext/native-decls/IsPedStrafing.md
```diff
@@ -12,4 +12,6 @@ bool IS_PED_STRAFING(Ped ped);
 * **ped**: The target ped.
 
 ## Return value
-Whether or not the ped is strafing.
\ No newline at end of file
+Whether or not the ped is strafing.
+
+Returns 1 if you're in first person for some reason.
```
4mmonium, line 17: When you go into first person camera, the ped goes into strafing mode, hence why it might be returning 1 😊

Mani2421: I didn't know that's what strafing meant, thanks ☺️

PR #2250, opened 2023-10-21
Add GET_OBJECT_TYPE in serverside

CiastekbatakPro:
its, useful because models of PICKUP on server-side are 0 so its hard to define which object is pickup and which is not


AvarianKnight:
You shouldn't change line endings :(

PR #2253, opened 2023-10-24
Add GET_OBJECT_TYPE

CiastekbatakPro:
its, useful because models of PICKUP on server-side are 0 so its hard to define which object is pickup and which is not


LWSS:
You should minify your commit so it doesn't have a bunch of whitespace changes

Issue #1854, opened 2023-02-12
GlobalState in clientside return nil if Table is large.


renzuzu:
as instructed by bubble to produce repros here.

this issue cannot be reproduce on any existing Globalstate that already have a value. ( as trying to use the command in client side its only returned the old value , this is probably an issue too)

i can only reproduce with New GlobalState

Server: 
```
RegisterCommand('fillgs', function()
	local stat = GlobalState.TEST or {} -- TEST is new
	for i = 1, 10000 do -- 10k to replicate a large table
		table.insert(stat,{
			data = 100000,
			data2 = 100000,
			data3 = 10000,
			data4 = 1000000,
			data5 = 1000000,
			data6 = 10000000,
			data7 = 10000000,
			data8 = 1000000,
		})
	end
	GlobalState.TEST = stat
end)
```
Server:  to see the globalstate haves a data in server
```
RegisterCommand('vehiclestat', function() -- to see the globalstate have a data in server
	local stat = GlobalState.TEST
	for k,v in pairs(GlobalState.TEST) do
		print(k,v)
		--stat[k] = nil
	end
	GlobalState.TEST = stat
end)
```
Client: 
```
RegisterCommand('vehiclestat2', function()
	for k,v in pairs(GlobalState.TEST) do
		print(k,v)
	end
end)
```

The Client Returned Nil value and produce and error.

i tought this is a default behavior of Oneysnc Globalstate. to prevent client receiving a large data sent from server.
but as i asked bubble, he said this could be a bug and need a repro.

im currently using Canary version of client and server artifact version 6183

renzuzu:
today i did the same repro steps.

![image](https://user-images.githubusercontent.com/82306584/218673566-f203fdc0-681e-4ada-a7ae-37d9ed1115f0.png)
this seems fixed, im not sure which commits from bubble fix this.
i even tried 20k loops and its still available in client.
Great as always dbub.

blattersturm:
... silly thing is nothing even changed in this regard.

renzuzu:
what? i tought you update something, this issue is existing even last year  probably 1.2 years ago when i start implementing global statebags, i always tought that this is a default behavior of this feature thats why i didnt bother asking, when i message you just few days ago, is just to confirm if its really a default behavior or a bug.

i remember doing this fix to not save vehicle properties data to avoid large datas.
https://github.com/renzuzu/renzu_garage/commit/9f247804828d36dc400dd08fa9a8a10c851ba1ec
[...]

PR #2257, opened 2023-11-01
fix(onesync/server): basic state bag write policies

tens0rfl0w:
This introduces a basic rejection logic for replicated player/entity state bags.
"Use-case" should be any type of state bag usage where "truthful" data matters, for example syncing economy-related data.

Implementation:
- ScRTs can use `SET_STATE_BAG_KEY_POLICY` to set a write policy for a certain state bag key.
- If write-protected, any client changes will be rejected by the server and the server value will be resent to the requesting client.
- A write policy affects all state bags. If set this will reject all value changes on the protected key.

---
I originally implemented this on a per state-bag/entity/player -> key logic, but reconsidering the actual use cases for this type of feature this didn't make much sense.


thorium-cfx:
I'm changing this into a draft, as this feature requires more design work and consider more use-cases before it's a feature we can keep supporting in the future.

Some internal points to consider I have on this,  to give context:
- Allowing/disallowing clients to read/write global, player, and entity values, checked on both server and client.
- Allowing/disallowing server to read/write global, player, and entity values.
- Different policies.
- Be performant.
- Updates (or not) statebag changes according to their rules.
- Allows the resource developer to set their desired policy on a per key basis.
- Delays between server and client (correcting changed values mid-policy change).
- Several use-cases (per-key, per-bag, etc.)

AvarianKnight:
I think there should be a way to go back to the default way how state bags worked before, where only the server could create state bags, and also have a way to default everything to be server write, client read unless set otherwise for the state bag.

This would be the most sane security default but I don't think it can be swapped easily without breaking compatibility so it would be good to just have a "sane default" ConVar.

> Allowing/disallowing clients to read/write global, player, and entity values, checked on both server and client.

I think @tens0rfl0w hit this on the head, clients should never be allowed to write to global state. As for the disallowing clients to read player/entity/global state the simplest solution would to probably be to have an enum.

```c
enum eStateSyncFlags {
	NoFlags = 0, // default flag
	OwnerOnly = 1, // only sync the state to the owning client
	NoSync = 2, // only the server should ever know this value, and it shouldn't ever be sync'd to clients
	ReadOnly = 4, // this would reject any change made by the client before ever being sent to the server
	// more if needed later
}
```

> Allowing/disallowing server to read/write global, player, and entity values.

I don't think this should be under consideration, the server is meant to be authoritative, nothing should be able to tell it its not able to touch anything.

 > Different policies.
 
 I think more context is required here by what you mean @thorium-cfx (I'm assuming this might be related to other stuff i posted here already)

> Be performant.

On this regard it might be worth swapping away from std::map as iirc @gottfriedleibniz mentioned this was causing some bad access times (though I can't find the comment right now).

It might be work making our own way of handling data to maintain consistency, like maintaining ordering that state bags are set in or allow state bags to be assigned priority to be able to be used consistently to [avoid some confusion](https://github.com/citizenfx/fivem/issues/2018)

> Updates (or not) statebag changes according to their rules.

I think this is somewhat explained how I would think to do it elsewhere

Another other issue that should probably be addressed https://github.com/citizenfx/fivem/issues/1854

thelindat:
> consider more use-cases
> Be performant.

So long as policies can be implemented in way future updates don't break backwards compatibility, I believe it's worth implementing what can easily be done right now. Right now there is nothing in place to prevent bad actors from creating a limitless number of statebag keys with enormous amounts of data.

- There is nothing to prevent instantiating arbitrary statebags.
- There is no rate limit on how frequently the client can update statebags.
- There is no limit on the size of the data being set.
- There is no limit on the number of states that can be stored on a statebag.

I ran a simple function to spam statebags and managed to increase fxserver's memory use by 1gb, which cleared the moment I dropped and statebags were wiped. Considering the 100,000 states I added would then get synced to any player who entered my scope.. bad time for all.

AvarianKnight:
> I ran a simple function to spam statebags and managed to increase fxserver's memory use by 1gb, which cleared the moment I dropped and statebags were wiped. Considering the 100,000 states I added would then get synced to any player who entered my scope.. bad time for all.

In this situation I'm not sure why the default was changed away from a server has to initialize the state bag to client can initialize any state bag it wants (which is the main reason this becomes an issue).

PR #2260, opened 2023-11-01
feat(discord): user toggle for rich presence

tens0rfl0w:
Requested in https://forum.cfx.re/t/5018677

Allow users to either completely disable Discords Rich Presence status or only show the current game version without displaying any game status.

### DIFF
--- a/code/components/citizen-resources-client/src/ConsoleScriptFunctions.cpp
+++ b/code/components/citizen-resources-client/src/ConsoleScriptFunctions.cpp
@@ -12,8 +12,10 @@ static InitFunction initFunction([]()
 
 		// this should not be exposed to script
 		// #TODO: 'not exposed to script' convar flags
-		if (HashString(varName.c_str()) == HashString("cl_ownershipTicket") || // ownership ticket is *very* bad to expose as users can pretend to be other users using this
-			HashString(varName.c_str()) == HashString("ui_updateChannel"))     // update channel is often misused as a marker for other things
+		uint32_t hashedVarName = HashString(varName.c_str());
+		if (hashedVarName == HashString("cl_ownershipTicket") ||   // ownership ticket is *very* bad to expose as users can pretend to be other users using this
+			hashedVarName == HashString("ui_updateChannel") ||     // update channel is often misused as a marker for other things
+			hashedVarName == HashString("cl_discordRichPresence")) // users probably set this for privacy reasons, no need for peeking eyes
 		{
 			context.SetResult(context.CheckArgument<const char*>(1));
 			return;
--- a/code/components/discord/src/DiscordPresence.cpp
+++ b/code/components/discord/src/DiscordPresence.cpp
@@ -9,6 +9,8 @@
 
 #include <ScriptEngine.h>
 
+#include <CoreConsole.h>
+
 #ifdef GTA_FIVE
 #define DEFAULT_APP_ID "382624125287399424"
 #define DEFAULT_APP_ASSET "fivem_large"
@@ -53,6 +55,12 @@ static std::string g_discordAppAssetText;
 static std::string g_discordAppAssetSmallText;
 
 static std::optional<std::tuple<std::pair<std::string, std::string>, std::pair<std::string, std::string>>> g_buttons;
+
+static std::shared_ptr<ConVar<std::string>> g_richPresenceState;
+
+static std::string g_lastRichPresenceState = "disabled";
+
+static size_t g_onGameFrameCookie = -1;
 
 static void UpdatePresence()
 {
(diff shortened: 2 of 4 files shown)


AvarianKnight:
Shouldn't the flag for the ConVar be `ConVar_UserPref`

tens0rfl0w:
> Shouldn't the flag for the ConVar be `ConVar_UserPref`

Yes, correct! I missed this one, thank you.

code/components/net/src/NetLibrary.cpp
```diff
@@ -404,11 +406,17 @@ void NetLibrary::ProcessOOB(const NetAddress& from, const char* oob, size_t leng
...
-				));
+				if (g_richPresenceEnabled->GetValue())
+				{
+					richPresenceSetValue(0, fmt::sprintf(
+						"%s%s",
+						std::string(cleaned).substr(0, 110),
+						(strlen(cleaned) > 110) ? "..." : ""
+					));
+				}
+				else {
```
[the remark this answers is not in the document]
tens0rfl0w: I also thought about this way, consideration was how most other games (I know of) handle this type of thing. I would say if you don’t want to display the game you're currently playing, use the global toggle in Discord itself and use the integrated toggle if you just want to hide the current server info. Another option would be to remove the game status completely and just display "FiveM" and the default app logo.

Unsure which is the best option here.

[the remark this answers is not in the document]
AvarianKnight: IMO the same argument could be said for what this PR does, just disable discord rich presence in your settings if you don't want games to show.

The [original platform suggestion](https://forum.cfx.re/t/disable-discord-rich-presence-toggle-option/5018677) wanted to be able to completely disable this, someone else suggested only hiding but that was not what was originally requested.

AvarianKnight:
You could probably handle all of this in OnGameFrame and not have to modify the natives to have a bunch of `g_richPresenceEnabled->GetValue()` calls everywhere

```cpp
static std::shared_ptr<ConVar<bool>> g_richPresenceEnabled = std::make_shared<ConVar<bool>>("game_enableDiscordRichPresence", ConVar_Archive | ConVar_UserPref, true);

static bool g_discordRichPresenceEnabled;

static bool g_wasRichPresencePreviouslyEnabled = true;

OnGameFrame.Connect([&handlers]()
{
	if (!g_discordRichPresenceEnabled)
	{
		if (g_wasRichPresencePreviouslyEnabled)
		{
			g_wasRichPresencePreviouslyEnabled = false;
			Discord_Shutdown();
		}
		return;
	}
	
	if (!g_wasRichPresencePreviouslyEnabled)
	{
		g_wasRichPresencePreviouslyEnabled = true;
		Discord_Initialize(g_discordAppId.c_str(), &handlers, 1, nullptr);
	}
	
	Discord_RunCallbacks();

	UpdatePresence();
});
```

tens0rfl0w:
> You could probably handle all of this in OnGameFrame and not have to modify the natives to have a bunch of `g_richPresenceEnabled->GetValue()` calls everywhere

Decided to go with your approach: If disabled, Rich Presence should be completely killed.
Implemented like you proposed.

code/components/discord/src/DiscordPresence.cpp
```diff
@@ -193,6 +202,24 @@ static InitFunction initFunction([]()
...
+			}
+			return;
+		}
+
+		if (!g_wasRichPresencePreviouslyEnabled)
+		{
+			g_wasRichPresencePreviouslyEnabled = true;
+			DiscordEventHandlers handlers;
+			memset(&handlers, 0, sizeof(handlers));
+			Discord_Initialize(g_discordAppId.c_str(), &handlers, 1, nullptr);
```
AvarianKnight, line 220: Something I didn't think about before, there should be a `g_richPresenceChanged = true;` after all of this so `UpdatePresence()` gets ran.

tens0rfl0w: Makes sense to get an instant update/re-activation. Resolved.

code/components/discord/src/DiscordPresence.cpp
```diff
@@ -162,6 +170,7 @@ static InitFunction initFunction([]()
 	g_discordAppAssetSmall = DEFAULT_APP_ASSET_SMALL;
 	g_discordAppAssetText = DEFAULT_APP_ASSET_TEXT;
 	g_discordAppAssetSmallText = DEFAULT_APP_ASSET_SMALL_TEXT;
+	g_richPresenceEnabledVar = std::make_shared<ConVar<bool>>("game_enableDiscordRichPresence", ConVar_Archive | ConVar_UserPref, g_richPresenceEnabled, &g_richPresenceEnabled);
```
thorium-cfx, line 173: There's an overload on `ConVar` to add a callback for changes, there you can do (un)initialization and (un)register the update function (see `fwEvent<T>::Connect` and `fwEvent<T>::Disconnect(cookie)` on `OnGameFrame`).

Will remove the per frame overhead.

thorium-cfx (changes requested):

code/components/discord/src/DiscordPresence.cpp
AvarianKnight: This would still have overhead per-frame anyways because you would have to do relatively the same check  inside of `UpdatePresence` to make sure it doesn't try to update rich presence.

thorium-cfx: I don't see how that adds up. You check for `g_richPresenceEnabled` and return if it's not true, which would not call `UpdatePresence()` at all. Unregistering the this update call will result in the same behavior, just with less calls and checks

AvarianKnight: You're right, I completely glossed over you mentioning disconnecting the event.

tens0rfl0w: This may be resolved with the most recent push, hopefully, implemented like expected.

thorium-cfx: Looking nice, I'll need to confirm that this is being called at startup as well, I'll probably get to that next week or maybe someone else will get to that sooner.

Another thing I'm considering is if `game_` should be changed into something like `cl_` as `game_` is used for game (GTAV/RDR2) related features.

tens0rfl0w: Tested startup behavior, should work. Let me know if anything else should be changed :) Thanks you!

ahcenezdh:
It would be cool to have a getter to check if the player has enabled Rich Presence (for example, to offer rewards if the player enables this option).

AvarianKnight:
Servers have notoriously abused stuff like that

tens0rfl0w:
You can already get the current ConVar value with `GetConvar("game_enableDiscordRichPresence", "true")`.
I already wanted to block this ConVar from ScRTs reads to avoid a situation like [this](https://github.com/citizenfx/fivem/commit/555579d2db7244c58cc5ce8ab5c1af8a02880e02) one, seems like I forgot about this while implementing.
Any objections to blocking this?

thorium-cfx:
I've discussed the mentioned improvements in the forum post and discussed this feature internally, we'd like to see it offer at least 3 options:
1. Off
2. On - only mentioning FiveM/RedM
3. On - with server information

I'm more drawn to naming it `cl_discordRichPresence` for these options.

tens0rfl0w:
Will implement like proposed, thank you for the feedback.

tens0rfl0w:
This should now include all desired changes requested by @thorium-cfx.

The current commit has two implementation aspects I'm not sure about, but wasn't able to accomplish without changing other things first:
-ConVar change handlers won't get called if the ConVar value is unset and the default value is returned, so on a "fresh" install, Discord Rich Presence will not work till the user restarts their game at least once to get the ConVar set.
-cfx-ui doesn't expose any ConVar setters for int values, this is why I went with string here.

Also, regarding possible "unwanted" side-effects:
Should this ConVar be readable by ScRTs to avoid servers kicking users if they decide to disable their rich presence?

Moozdzn:
> Should this ConVar be readable by ScRTs to avoid servers kicking users if they decide to disable their rich presence?
Probably out of scope, but can't this be achieved with a discord bot? Most servers already force you to join their discord server.

thorium-cfx:
Thank you @tens0rfl0w, I've been quite busy and will review the changes soon, but wanted to address this question:
> Also, regarding possible "unwanted" side-effects: Should this ConVar be readable by ScRTs to avoid servers kicking users if they decide to disable their rich presence?

Seeing that the player wants to hide this for personal or privacy reasons, there's reason enough not to expose this to anyone, so no it shouldn't be readable.

tens0rfl0w:
Returning `default_` now if read by `GET_CONVAR`.

code/components/citizen-resources-client/src/ConsoleScriptFunctions.cpp
```diff
@@ -12,8 +12,9 @@ static InitFunction initFunction([]()
 
 		// this should not be exposed to script
 		// #TODO: 'not exposed to script' convar flags
-		if (HashString(varName.c_str()) == HashString("cl_ownershipTicket") || // ownership ticket is *very* bad to expose as users can pretend to be other users using this
-			HashString(varName.c_str()) == HashString("ui_updateChannel"))     // update channel is often misused as a marker for other things
+		if (HashString(varName.c_str()) == HashString("cl_ownershipTicket") ||   // ownership ticket is *very* bad to expose as users can pretend to be other users using this
+			HashString(varName.c_str()) == HashString("ui_updateChannel") ||     // update channel is often misused as a marker for other things
+			HashString(varName.c_str()) == HashString("cl_discordRichPresence")) // avoid scripts denying user-choice
```
thorium-cfx, line 17: Weird all these line changes, though while you're at it, can you cache `HashString(varName.c_str())` and check that variable against all these others? And I think we should go with a more constructive comment like "Users probably set this for privacy reasons, no need for peeking eyes"

code/components/discord/src/DiscordPresence.cpp
```diff
@@ -54,6 +56,12 @@ static std::string g_discordAppAssetSmallText;
 
 static std::optional<std::tuple<std::pair<std::string, std::string>, std::pair<std::string, std::string>>> g_buttons;
 
+static std::shared_ptr<ConVar<std::string>> cl_richPresenceState;
+
+static std::string cl_lastRichPresenceState = "disabled";
```
thorium-cfx, line 61: Globals need prefix `g_`.

code/components/citizen-resources-client/src/ConsoleScriptFunctions.cpp
tens0rfl0w: Line changes were due to comment indentation.

code/components/discord/src/DiscordPresence.cpp
```diff
@@ -153,17 +161,81 @@ static void UpdatePresence()
 	}
 }
 
+static void OnRichPresenceStateChange(internal::ConsoleVariableEntry<std::string>* richPresenceState)
+{
+	std::string state = richPresenceState->GetValue();
+
+	if (state != "disabled" && state != "partial" && state != "enabled")
```
prikolium-cfx, line 168: What do you think if we replace "partial" with "restricted"? It sounds better if we talk about privacy

prikolium-cfx (approved):
Looks good to me. Great feature for privacy protection.

fivem20:
KLKLKL KLK

PR #2261, opened 2023-11-02
feat(client/clr): ability to send byte[] as InputArgument

KubasGC:
There is no way to pass byte[] to native now using C#. This PR makes it possible - I tested it [RedM] with native `0x26E87218390E6729` (notification toast showing on the left side), which needs two byte arrays passed as arguments.


thorium-cfx:
Hello KubasGC, thank you for your PR.

This feature comes with more things to consider that we're currently working on already. As such we can't accept this PR, but we appreciate the work and request.

PR #2262, opened 2023-11-02
tweak(docs/building): add note when using python 3.12+

AvarianKnight:
- Python deprecated distutils in [PEP 632](https://peps.python.org/pep-0632/) and removed it in 3.12

### DIFF
--- a/docs/building.md
+++ b/docs/building.md
@@ -21,10 +21,12 @@ Then, execute the following commands in a `cmd.exe` shell to set up the build en
 
 ```bat
 set BOOST_ROOT=C:\libraries\boost_1_71_0
-set PATH=C:\tools\python27amd64;%path%
 git clone https://github.com/citizenfx/fivem.git -c core.symlinks=true
 cd fivem
 git submodule update --jobs=16 --init
+
+:: If you're using Python 3.12 or higher you may need to install setuptools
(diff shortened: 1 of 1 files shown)


thorium-cfx:
Could you put this one back to [2912c9c](https://github.com/citizenfx/fivem/commit/2912c9c93386df977580eb87f42ff1f51522dbbb)?

AvarianKnight:
> If you're using Python 3.12 or higher you may need to install setuptools when it isn't installed already

The wording here seems a bit redundant does it not?

> If you're using Python 3.12 or higher you may need to install setuptools

thorium-cfx:
True, your proposal looks good 👍

docs/building.md
```diff
@@ -8,9 +8,9 @@ To build FiveM, RedM or FXServer on Windows you need the following dependencies:
   - .NET desktop environment
   - Desktop development with C++
   - Universal Windows Platform development
-
```
thorium-cfx, line 11: Please no unnecessary changes like these, this'll make git blame look wild

thorium-cfx (changes requested):

AvarianKnight:
Every editor I have force removes redundant white spaces 😢

thorium-cfx (approved):

PR #2266, opened 2023-11-06
Glow sphere drawing system in FiveM

Disquse:
Original report: https://forum.cfx.re/t/graphics-drawsphere-renders-only-one-sphere-per-frame/2500594?u=disquse

```c
void DRAW_GLOW_SPHERE(float posX, float posY, float posZ, float radius, int colorR, int colorG, int colorB, float intensity, BOOL invert, BOOL marker);
```

This native allowing to draw a "glow sphere" as a marker or overlay. Limited to 256 spheres per single frame.
Example script: https://pastebin.com/05wnuWCD

FiveM_JT6PJ0oAeD
FiveM_NQtjNzCSHd

### DIFF
--- a/code/components/extra-natives-five/src/GraphicsExtraNatives.cpp
+++ b/code/components/extra-natives-five/src/GraphicsExtraNatives.cpp
@@ -1,10 +1,10 @@
 #include <StdInc.h>
-
+#include <array>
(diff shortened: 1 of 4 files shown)


ook3D:
Nice

thorium-cfx (approved):

PR #2272, opened 2023-11-11
Fix (clrcore/fivem) Ped class using wrong handle

xZToxicZx:
The Ped class for _GetHeadBlendData was using the local player Ped rather then the Ped its representing, this update fixes that.

### DIFF
--- a/code/client/clrcore/External/Ped.cs
+++ b/code/client/clrcore/External/Ped.cs
@@ -202,7 +202,7 @@ namespace CitizenFX.Core
 			UnsafePedHeadBlendData data;
 			unsafe
 			{
-				Function.Call(Hash._GET_PED_HEAD_BLEND_DATA, API.PlayerPedId(), &data);
+				Function.Call(Hash._GET_PED_HEAD_BLEND_DATA, Handle, &data);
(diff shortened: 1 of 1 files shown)


code/client/clrcore/External/Ped.cs
```diff
@@ -2,16 +2,16 @@
 using CitizenFX.Core.Native;
 using System.Security;
 
-#if MONO_V2
-using CitizenFX.Core;
+#if MONO_V2
```
thorium-cfx, line 5: What happened here, can you undo this?

thorium-cfx (changes requested):
Good find. Just some unnecessary changes at the top, can you undo those?

code/client/clrcore/External/Ped.cs
xZToxicZx: Oh no idea what happened there... can do

xZToxicZx: All of these are updates to line endings, VS is being a bit stupid and not letting me just stage a single line...

thorium-cfx (approved):

PR #2278, opened 2023-11-16
Update AuthController.cs

cundyis:
Constants for Repeated Strings: Strings that are reused multiple times (like "/webadmin/" and error messages) are moved to constants. This helps avoid typos and makes future changes easier.

Simplified Return in Login Method: The Login method is simplified to a single line.

Combined Error Checks: In LoginRcon, the two separate checks for an invalid RCON password are combined into one for simplicity and to avoid code duplication.

Removed Unused Imports: Unused using directives like System.Dynamic and System.Expando are removed for clarity.

Improved Readability: Some lines are broken into multiple lines for better readability, especially where method calls with multiple parameters are involved.

### DIFF
--- a/ext/webadmin/server/Controllers/AuthController.cs
+++ b/ext/webadmin/server/Controllers/AuthController.cs
@@ -1,24 +1,24 @@
 using System.Collections.Generic;
-using System.Dynamic;
 using System.Security.Claims;
 using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


AvarianKnight:
Is webadmin even put into builds anymore? I thought it got no-oped a while ago.

FabianTerhorst:
Thanks for your contribution, but webadmin will be removed soon.

PR #2282, opened 2023-11-23
fix(ext/cfx-ui): adaptive card disabled button style fix

1A3Dev:
This fixes disabled actions on adaptive cards not having proper styling.

Before:
old
After:
new

### DIFF
--- a/ext/cfx-ui/src/cfx/apps/mpMenu/parts/LegacyConnectingModal/AdaptiveCardPresenter/AdaptiveCardPresenter.module.scss
+++ b/ext/cfx-ui/src/cfx/apps/mpMenu/parts/LegacyConnectingModal/AdaptiveCardPresenter/AdaptiveCardPresenter.module.scss
@@ -126,6 +126,7 @@
     }
 
     :global(.ac-pushButton),
+    :global(.ac-pushButton-disabled),
     :global(.ac-inlineActionButton) {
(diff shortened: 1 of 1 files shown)


ext/cfx-ui/src/cfx/apps/mpMenu/parts/LegacyConnectingModal/AdaptiveCardPresenter/AdaptiveCardPresenter.module.scss
```diff
@@ -185,6 +186,13 @@
           border-color: ui.color('error');
         }
       }
+
+      &:global(.ac-pushButton-disabled) {
+        color: ui.color-token('button-disabled-text');
+        border-color: ui.color-token('button-disabled-border');
+        background-color: ui.color-token('button-disabled-background');
+        cursor: default;
```
nihonium-cfx, line 194: Please use the same style here as we already have for disabled buttons: https://github.com/citizenfx/fivem/blob/master/ext/cfx-ui/src/cfx/ui/Button/Button.module.scss#L67-L70

nihonium-cfx (changes requested):
Thank you for the contribution!

nihonium-cfx (changes requested):

PR #2283, opened 2023-11-23
added playerDeathEvent

gottaster0id:
Has many different use cases such as fully server side player death handling.

Example:
```
AddEventHandler('playerDeathEvent', function(source, causeOfDeath, sourceOfDamage)
    print('playerDeathEvent', source, causeOfDeath, sourceOfDamage)
end)
```

### DIFF
--- a/code/components/citizen-server-impl/src/state/ServerGameState.cpp
+++ b/code/components/citizen-server-impl/src/state/ServerGameState.cpp
@@ -961,6 +961,24 @@ void ServerGameState::Tick(fx::ServerInstanceBase* instance)
 				}
 			}
 
+			if (entity->type == sync::NetObjEntityType::Player)
+			{
+				auto entityHealth = entity->syncTree->GetPedHealth();
+
+				if (entityHealth->health <= 0 && !entityHealth->isDead)
+				{
+					auto ownerRef = entity->GetClient();
+
+					entity->syncTree->SetPedDead(true);
+
+					evMan->TriggerEvent2("playerDeathEvent", {}, fmt::sprintf("%d", ownerRef->GetNetId()), entityHealth->causeOfDeath, entityHealth->sourceOfDamage);
+				}
+				else if (entityHealth->isDead && entityHealth->health >= 1)
+				{
+					entity->syncTree->SetPedDead(false);
+				}
+			}
+
 			float position[3];
 			entity->syncTree->GetPosition(position);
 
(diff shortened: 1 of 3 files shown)


gottaster0id:
> A similar event is already part of the [stock fxserver [baseevents] resource](https://github.com/citizenfx/cfx-server-data/blob/master/resources/%5Bsystem%5D/baseevents/deathevents.lua), though this will likely be more efficient and work without any resources.

Your right, but since a event like that should be fully server side instead of client->server i created this requests.

code/components/citizen-server-impl/src/state/ServerGameState.cpp
```diff
@@ -960,6 +960,23 @@ void ServerGameState::Tick(fx::ServerInstanceBase* instance)
 					}
 				}
 			}
+			
+			if (entity->type == sync::NetObjEntityType::Player)
+			{
+				auto entityHealth = entity->syncTree->GetPedHealth();
+				auto ownerRef = entity->GetClient();
+
+				if (entityHealth->health <= 0 && !entityHealth->isDead)
```
Gogsi, line 969: Peds have a [fatal injury threshold](https://docs.fivem.net/natives/?_0xD839450756ED5A80) under which they are considered dead. This doesn't have to be 0.

gottaster0id: Yes, but this native is only client side and since when a player is dead the life is at 0 I decided to check the health. But are they any scenarios that the health is not 0 and your dead?

Gogsi: Just tested this script on a blank txadmin server. Setting the player's health to anything below 100 kills them

```lua
  Wait(6000)
  SetEntityHealth(GetPlayerPed(-1), 99);
end)
```

Gogsi: Nevermind, it seems that something sets your health to 0 as part of the dying process. In this image I set it to 99 and it went to 0. This is probably not an issue then.
image

gottaster0id: Yes, but it will work anyway since if you die your health will be automatically set to 0.

ahcenezdh:
@gottfriedleibniz  sorry to ping you but any news on this PR? would be awesome to have those server side events

FabianTerhorst:
Merge of this is going to be delayed for now until the event system received improvements to reduce the overhead it creates.

n3xuuu:
is there any eta for when the event system will receive those improvements? it's a really needed feature and would help everyone

evr76:
ya entro

PR #2285, opened 2023-11-24
feat(extra-natives/rdr3): GET_GAME_POOL

Ktos93:
This PR adds ability to get `EntityBatch` pool using `GET_GAME_POOL` native. Which can be usefull in a lot of cases like get position for trees

### DIFF
--- a/code/components/extra-natives-five/src/PoolTraversalNatives.cpp
+++ b/code/components/extra-natives-five/src/PoolTraversalNatives.cpp
@@ -294,16 +294,18 @@ static InitFunction initFunction([]()
 	fx::ScriptEngine::RegisterNativeHandler("GET_GAME_POOL", [](fx::ScriptContext& context)
 	{
 		std::string_view pool = context.CheckArgument<const char*>(0);
-		if (pool.compare("CPed") == 0)
+		uint32_t poolHash = HashString(pool);
+
+		if (poolHash == HashString("CPed"))
 			SerializePool<PedPoolTraits>(context);
-		else if (pool.compare("CVehicle") == 0)
+		else if (poolHash == HashString("CVehicle"))
 			SerializePool<VehiclePoolTraits>(context);
-		else if (pool.compare("CObject") == 0)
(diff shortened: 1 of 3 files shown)


AvarianKnight:
You should add this to the native description too

Ktos93:
I've never been good at doing documentation. I have no idea how to properly mark that this is only for RedM

gottfriedleibniz:
Since you're looking at this native, you can also convert the unnecessary `string` construction to `string_view` (compiler optimizes this away anyway). From usage rates, a handful of cycles could also be saved by moving the `CVehicle` above `CObject`. 

Much of this file can be simplified anyway given how much more things are understood since the time of writing.

Ktos93:
you mean something like this?
image

gottfriedleibniz:
It should also be worth it to take another easy win and just compare case-insensitive hashes (e.g., `HashString("CPed")`). Case sensitivity is not that important in this context. Just make sure the `pool` hash is only computed once.

Ktos93:
Is that what you meant?

gottfriedleibniz:
Sure. The next issue is whether creating so many ScriptGuid extensions makes sense. That value is often in the thousands and it is not entirely obvious it will only contain "trees", or whatever is desired, that are actually visible/useful to the client/script.

It may be possible to `GET_SHAPE_TEST_RESULT_INCLUDING_MATERIAL` in a clever way to achieve what you are trying to desire. However, if you continue this path, it may be worth reversing `CEntityBatch` further to see if you can reduce this (e.g., ownership and visibility flags).

Ktos93:
Using shape test you cant get model of this tree just return 0. `GET_SHAPE_TEST_RESULT_INCLUDING_MATERIAL` when i last time tested this native on RedM its doesnt work. Also you still cant make sure its a tree or maybe a wooden building.
Using `GET_GAME_POOL` give ability to call this native one time for a long while and save all trees with data you need
(image GetEntityModel , hitEntity)
image

Issue #2196, opened 2023-09-10
Async export returns incorrect value when second export is called after a value is awaited


mkafrin:
This seems to be an issue in the JS runtime that occurs whenever any value is awaited in an async export, and then any export is called after that point. If those two requirements are met, then the return of the async export will be replaced by the return of the second export that was called. I looked at other issues but none seemed to be specific to this issue (one was about exports that _sometimes_ return a promise, but that isn't the case here).

Repro:
```js
exports('test2', () => 'wrong');
exports('test1', async () => {
    await Promise.resolve(); // Can await anything here and it will still be broken (unresolved promise, non-promise, etc.)
    exports[GetCurrentResourceName()].test2(); // return value from this will replace the explicit return on the next line
    return 'right';
});
RegisterCommand('test', async () => {
  console.log(await exports[GetCurrentResourceName()].test1()); // 'wrong' is logged
});
```
Expected Behavior: Run the "test" command and "right" will be logged.
Actual Behavior: Run the "test" command and "wrong" is logged.

gallejens:
https://github.com/citizenfx/fivem/issues/1611

mkafrin:
Is it? You mentioned in your issue that the problem only appeared when the secondary function _sometimes_ returns a promise, and that is only happens when that function returns `undefined`.

Is that not the case still for you? If not the original issue should be updated. Because this issue happens for me regardless of what the `test2` export returns, and it doesn't have to be awaited either (though the issue also happens if it's awaited).

gallejens:
You are correct that its not the same behaviour but is probably the same underlying issue.
Just wanted to link the other issue as it might contain relevant information.

blattersturm:
This is an interesting one, likely caused by function reference invocation technically not being reentrant _enough_.

The issue here stems from V8's microtask queue executing right after the initial invocation, overwriting the stored return value buffer (containing `test1`'s promise) with that of the later return - in this case, `test2` completing.

I'm still investigating this but that explanation would make sense.

TeKGameR950:
Still happening, i'd love this to be fixed, its very annoying when you are using multiple exports that are awaited.

Mathu-lmn:
This can be reopened for the time being I think !

PR #2288, opened 2023-11-28
DRAFT: WIP/PoC: funcref correctness (#2196)

blattersturm:
This is a fix to #2196 I was working on mid-October, which ended up reworking function references entirely. I dug this up in my local working tree while preparing for an alternative implementation of https://github.com/citizenfx/fivem/commit/c9ee6587ad10c9ce7b02831b43c103a5a6d7b145, and am sharing it for completeness, and so it doesn't get lost in case this VM gets marked as obsolete at some point.

(Among likely other things not including it not applying on the current tree anymore, the Mono changes here are incomplete)

There's a few other concerns about function references going on as well, mainly performance overhead (as seen in e.g. Warxander's [Vein menu system](https://forum.cfx.re/t/vein-an-easy-to-use-gui-library-for-fivem/4788714)), and solving that could be seen as an extension to this refactor.


TeKGameR950:
Any plans for this to be merged ? Would solve an issue i'm having :/

PR #2312, opened 2023-12-22
feat(net): Add NetworkSetEntityOwner server native

ivanzaida:
### Goal of this PR

This pull request introduces the `NETWORK_SET_ENTITY_OWNER` native to the server side of the application, facilitating the manual migration of the owning player for entities.
This may be extremely useful alongside the [sv_filterRequestControl](https://docs.fivem.net/docs/server-manual/server-commands/#sv_filterrequestcontrol-mode) being set to 4.

The created native uses FiveM's inner method that migrates entities owners, I believe the `ServerGameState->ReassignEntity ` handles all the edge cases regarding entity owner migration.

Server

### Successfully tested on
**Game builds:** 2944
**Platforms:** Windows, Linux

### Checklist


ivanzaida:
Thanks for your reply.
The ownership reassignment is indeed a side-effect, but from what I've tested it only happens to population entities.
It seems like the population ownership system is designed in some way that a player owns some kind of a territory around him and the game does not let to reasign an entity to a different player (it actually allows, but after a few seconds the ownership gets reset).
From what I've tested, server-created entities do not get automatically reasigned to the territory-owner, and the assigned user can manipulate the entity's state in more efficient way than doing RPC requests.
Also, I got tired defining custom events for managing an entity state / running some client-specific natives on the owning client's side, who is not the actual target of the script, e.g. :
`
// client
onNet('PlayPedAmbientSpeechNative', (pedNetId: number, speechName: string, speechParam: string) => {
  const entity = NetworkGetEntityFromNetworkId(pedNetId);
  if (!entity) {
    return;
  }
  PlayPedAmbientSpeechNative(entity, speechName, speechParam);
})
`
I believe, I'm not the only one who is tired of doing this kind of boilerplate job, and the changes present in this PR would save lots of developer time

AvarianKnight:
I don't think this is a proper solution to the problem.

This PR would expect everyone using to have a general understanding on how everything *should* work and also expect them to not break "rules" (i.e. don't set the owner to someone that it *can't* be relevant to).

A proper solution would be to either have server-side setters (which seems like a pipe dream if we're being honest), or fixing the RPC system so instead of sending a RPC once and hoping it gets executed, add RPC calls to a vector/map and send the RPC execution to the client until it gets acked.

Something similar to this could be added *internally* and not exposed to the end user, so if there are RPCs pending for an entity, the server should refuse to mitigate the entity unless it is no longer relevant to the currently locked owner (to hopefully not have the owner ping-pong around players while trying to sync) , after which it should find another relevant player and try to execute the RPC calls again.

ivanzaida:
would be nice if you guys could take care of the proper solution, but it seems like a pipe dream as well

FabianTerhorst:
In my opinion there is a missing a way to prevent network owner change after using this setter for this specific entity. Having that possibility would make this setter more useful.

code/components/citizen-server-impl/src/state/ServerGameState_Scripting.cpp
```diff
@@ -2049,7 +2049,41 @@ static void Init()
...
+		if (!client)
+		{
+			return false;
+		}
+
+		auto gameState = instance->GetComponent<fx::ServerGameState>();
+
+		gameState->ReassignEntity(entity->handle, client);
+
+		auto newOwner = entity->GetLastOwner();
```
iridium-cfx, line 2078: I'm not too familiar with ReassignEntityInner, but it looks like you would want to check GetClient instead of GetLastOwner, to see if it was updated.

idan-github:
This native implementation is very good. Will they ever add it?

This native feature would help me a lot.

Does anyone know why they haven't added this native yet? What's been holding it up for the past 2.5 years?

AvarianKnight:
> This native implementation is very good. Will they ever add it?
> 
> This native feature would help me a lot.
> 
> Does anyone know why they haven't added this native yet? What's been holding it up for the past 2.5 years?

If you read the previous discussion you would understand why this native hasn't been accepted. 

https://github.com/citizenfx/fivem/pull/4007 is meant to solve a part of this problem

It still fails in a certain regards though, to quote myself from a few years ago

> This PR would expect everyone using to have a general understanding on how everything should work and also expect them to not break "rules" (i.e. don't set the owner to someone that it can't be relevant to), and we also shouldn't be spilling internals to the end user.

This would have to be solved before the other PR should be considered merge ready, currently its just meant for testing.

PR #2314, opened 2023-12-23
feat(net/client): Add state bag methods

ivanzaida:
The modifications in this pull request introduce two new natives along with supporting methods for state bags.

### Goal of this PR
`STATE_BAG_HAS_VALUE`: This native is designed to eliminate unnecessary deserialization of state bag values. It serves the purpose of determining whether a value exists for a specific key within the state bag without the need for full deserialization.

`GET_STATE_BAG_KEYS`: This native expands the utility of state bags by enabling the construction of custom models around the state bag without the necessity to manually check each key for existence. This enhancement streamlines the process of working with state bags and facilitates the creation of more efficient and organized models.

FiveM, RedM, Server,

### Successfully tested on
**Game builds:**  2944

**Platforms:** Windows, Linux

### Checklist

### DIFF
--- a/code/components/citizen-resources-core/include/StateBagComponent.h
+++ b/code/components/citizen-resources-core/include/StateBagComponent.h
@@ -96,6 +96,16 @@ public:
 	// Gets data for a key.
 	//
 	virtual std::optional<std::string> GetKey(std::string_view key) = 0;
+
+	//
+	// Returns whether the state bag has data for this key
+	//
+	virtual bool HasKey(std::string_view key) = 0;
+
+	//
+	// Returns all the keys that the state bag has data for
+	//
+	virtual std::vector<std::string> GetKeys() = 0;
 };
 
 class CRC_EXPORT StateBagComponent : public fwRefCountable, public IAttached<ResourceManager>
--- a/code/components/citizen-resources-core/src/StateBagComponent.cpp
+++ b/code/components/citizen-resources-core/src/StateBagComponent.cpp
@@ -101,6 +101,8 @@ public:
 	virtual ~StateBagImpl() override;
 
 	virtual std::optional<std::string> GetKey(std::string_view key) override;
+	virtual bool HasKey(std::string_view key) override;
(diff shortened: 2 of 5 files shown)


ivanzaida:
Thanks for the feedback.
I actually wanted to expose the added methods thru all the ScRTs, but it would be breaking the compatibility, since I'm prety sure some of the script makers already used `has` as their state bag key and it would be a breaking change for them, so I decided not to add this method to JS/Lua ScRTs.

The ideas behind every feature:
1. Setting a state bag value if it's not present
```ts
const entity = Entity(someHandle)
if (!entity.metadataKey) { // why need to deserialize here, if I only want to know if the value is set?
  entity.metadataKey = {...}
}
```

2. The idea around the GET_STATE_BAG_KEYS was that a state bag is a key/value storage, and It would be nice to expose this kind of method to iterate over all the keys that are present in the statebag. Im not pushing this feature, it's just an improvement of the existing API

ivanzaida:
There is no problem for me to separate the features into two different PR's, if is required.
Need to consult the docs to see if there is a way to integrate the features into the existing ScRTs `Entity `API

FabianTerhorst:
Agree with the concern that it doesn't match the current state bag api, but there is no easy way to adjust the current api to support this without breaking changes or without changing it complete. So the change will be accepted and added until we have a new state bag api iteration.

FabianTerhorst (approved):
Tested and works fine 👍

slashkeyvalue:
Wouldn't hasvalue also return true for keys whose values were set to null? that'd be weird.

FabianTerhorst:
> Wouldn't hasvalue also return true for keys whose values were set to null? that'd be weird.

Yeah, because null values are also replicated. But there is no way to remove state bag values as well. So filtering nulls seems to make sense.

FabianTerhorst:
With https://github.com/citizenfx/fivem/pull/2717 this change makes sense again, because null values will no longer exists inside the data.

Issue #962, opened 2021-09-06
Allow server-spawned entities to be protected from client-side deletion


blattersturm:
Currently, the server will unconditionally accept a client-side clone-delete request for a server-spawned entity. It could be worthwhile to add/abuse some flag (potentially stored in state so the client can be aware as well? though it'd need to be ensured a client can't set these) that will restrict this for such an entity, and also refuse client-side script/etc. deletion where possible to prevent 'flickering' scenarios.

Server-side DELETE_ENTITY should be unaffected by this.

WhatTheNull:
is there any problem by not allowing client-side delete entity's no matter what its created by server or client ? by allowing deletion only by server we can have a event that can be called by a delete request like `entitycreating` or such then we can let the programmers decide/check if     the owner of the entity = the person's server id who is requesting the delete       or using a ace for admins stuff etc and much more ...

blattersturm:
> is there any problem by not allowing client-side delete entity's no matter what its created by server or client ? by allowing deletion only by server we can have a event that can be called by a delete request like `entitycreating` or such then we can let the programmers decide/check if the owner of the entity = the person's server id who is requesting the delete or using a ace for admins stuff etc and much more ...

What?

WhatTheNull:
i mean . dont allow the client-side delete any entity at all , client-side ask the server for deleting the entity , and then server-side decide whether that entity should be deleted or not

blattersturm:
I'm still not sure what you're talking about. This is what already happens.

WhatTheNull:
client's can delete the server-side created entity's right ? what if we block the deleting from client totally and handle the deletion entity's to server-side for all entity's

blattersturm:
What? Deletion is already handled server side. 

Can you please stay out of this comment thread? I don't want to have to block you from interacting with this repo.

WhatTheNull:
ok , i just talked based on your comment : `Currently, the server will unconditionally accept a client-side clone-delete request from the owner of a server-spawned entity.` i thought its not handled by server side , my bad then , i will delete my comments to keep this clean

blattersturm:
Why does GitHub have self deletion nonsense now? It looks like I'm talking to myself here. 🙄

blattersturm:
Original series of comments from hooks:

@WhatTheNull:
[...]

PR #2322, opened 2023-12-31
feat(gamestate/server): CREATE_TRAIN(_CARRIAGE) natives

### LINKED ISSUE #2197: General train issues
I've been digging around to implement trains in the fiveM ecosystem and I faced few issues.

## 1st approach

My first approach was to create the trains server side via the CreateVehicleServerSetter method, here is some pseudo code
> server
```lua
RegisterCommand("train", function(source)
    local train = CreateVehicleServerSetter(`metrotrain`, 'train', -888.1231, -2310.7957, -12.6248, 0.0)
    print(GetVehicleType(train)) -- 'train'
    Wait(50)

    TriggerClientEvent("test", source, NetworkGetNetworkIdFromEntity(train))
end, false)
```

> client

The train renders as derailed
```lua
RegisterNetEvent("test", function(trainNetId)
    local fwTrain = NetworkGetEntityFromNetworkId(train)
    NetworkRequestControlOfEntity(fwTrain)
    Wait(500)

    SetMissionTrainCoords(fwTrain, -891.9408, -2321.6558, -12.6177) -- works
    SetRenderTrainAsDerailed(fwTrain, false) -- no effect
    SetTrainSpeed(fwTrain, 1.0) -- fails to execute the native
end)
```

## 2nd approach

My other idea was then to implement a serverside train streamer to create my own sync layer with clientside entities, and simulate the trains movement by interpolating the trains sector nodes when no player stream that virtual entity, this way the system would work even in relaxed lockdown mode, by the way, having the NetworkTime exposed server side would make such time-based sync really easy

> client
```lua
    loadTrainModels()

    local fwTrain = CreateMissionTrain(26, pos.x, pos.y, pos.z, direction, false, true) -- the 6th argument supposed to be 'isNetwork'
    print(NetworkGetEntityIsNetworked(fwTrain)) -- true

    freeTrainModels()
```

This makes a networked train, that is blocked by the relaxed lockdown mode, which is really meh

## Solutions

Well, I see few solutions

- Make the server-side created trains entities work with the client-side mission train natives
and / or
- Make the client-side native CreateMissionEntity 6th 'isNetwork' param work
and / or
- Implement a server side event regarding the lockdown modes, where we can accept or deny an entity created by a client, therefore we would have the full control, and it's nice / secured

Ehbw:
### Goal of this PR

Allows for server-side creation of functioning trains, to be used in strict and relaxed entity lockdown modes.

By adding the ``CREATE_TRAIN`` and ``CREATE_TRAIN_CARRIAGE`` to allow for the creation of trains and its carriages through server-setters. The following ``LOAD_TRAIN_CONFIG_FROM_PATH`` and ``LOAD_TRAIN_TRACKS_FROM_PATH`` are also added to address the issue of the server lacking knowledge of train configurations and train index's. Currently the information is only used to prevent invalid track ID's and train config indexes which cause unintended client behaviour. 

Modifications were required for the client, to have server-setter trains behave as intended when re-entering a the scope of a client or changing ownership. By updating the trains current track node to its relative position and then force blending on ownership change (if the server was the last owner). This has no effect on client created trains or trains actively owned by other players.

While creating this PR I also completed the ``CTrainGameStateDataNode`` based on the research provided by [gottfriedleibniz](https://forum.cfx.re/t/train-carriages-becoming-detached/3430416/8?u=ehbw), and some of my own.

Examples of Invalid and Valid uses of these natives can be found here: https://gist.github.com/Ehbw/c40e88f0686275f4a6fe417573babb64

A functioning script snippet can be found here: https://gist.github.com/Ehbw/ef9d2302c4833bab5b23160588ff3a6f

FiveM, Server, Natives

### Successfully tested on

Builds prior to the tuners update (2372) results in all trains being deleted by the owning client when leaving the owner's scope. This behaviour is identical to client created trains on these same builds. This would be resolved with blocking deletion of any server-owned entity as suggested here https://github.com/citizenfx/fivem/issues/962

**Game builds:** 1604-3095

**Platforms:** Windows

### Checklist

### Fixes issues
resolves #2197

### DIFF
--- a/code/components/citizen-server-impl/src/state/ServerSetters.cpp
+++ b/code/components/citizen-server-impl/src/state/ServerSetters.cpp
@@ -5,13 +5,16 @@
 #include <state/ServerGameState.h>
 #include <ScriptEngine.h>
 
+#include <parser/TrainParser.h>
 #include <GameServer.h>
 
+#include <ResourceManager.h>
 #include <Resource.h>
 #include <fxScripting.h>
 
 #include <boost/mpl/at.hpp>
 #include <boost/function_types/parameter_types.hpp>
+#include <ServerInstanceBaseRef.h>
 
 namespace fx
 {
@@ -156,6 +159,64 @@ std::shared_ptr<sync::SyncTreeBase> MakeVehicle(uint32_t model, float posX, floa
 	return tree;
 }
 
+template<typename TTree>
+std::shared_ptr<sync::CTrainSyncTree> MakeTrain(uint32_t model, float posX, float posY, float posZ, float heading, uint32_t resourceHash, bool isEngine, bool direction, bool stopAtStation, float speed, int trackid, int trainConfigIndex)
+{
+	auto tree = std::make_shared<sync::CTrainSyncTree>();
+
+	SetupNode(tree, [model](sync::CVehicleCreationDataNode& cdn)
+	{
+		cdn.m_model = model;
+		cdn.m_creationToken = msec().count();
+		cdn.m_needsToBeHotwired = false;
+		cdn.m_maxHealth = 1000;
+		cdn.m_popType = sync::POPTYPE_MISSION;
+		cdn.m_randomSeed = rand();
+		cdn.m_tyresDontBurst = false;
+		cdn.m_vehicleStatus = 2;
+		cdn.m_unk5 = false;
+	});
+
+	SetupPosition<sync::CSectorDataNode, sync::CSectorPositionDataNode>(tree, posX, posY, posZ);
+	SetupHeading(tree, heading);
+
+	SetupNode(tree, [resourceHash](sync::CEntityScriptInfoDataNode& cdn)
+	{
+		cdn.m_scriptHash = resourceHash;
+		cdn.m_timestamp = msec().count();
+	});
+
+	SetupNode(tree, [isEngine, speed, direction, trackid, trainConfigIndex, stopAtStation](sync::CTrainGameStateDataNode& cdn)
+	{
+		cdn.data.linkedToBackwardId = 0;
+		cdn.data.linkedToForwardId = 0;
+		cdn.data.distanceFromEngine = 0.0f;
+
+		cdn.data.trainState = 3;
+
+		cdn.data.isEngine = isEngine;
+		cdn.data.cruiseSpeed = speed;
+		cdn.data.direction = direction;
+		cdn.data.trackId = trackid;
+		cdn.data.trainConfigIndex = trainConfigIndex;
+
+		cdn.data.isCaboose = true;
+
+		cdn.data.isMissionTrain = true;
+		cdn.data.hasPassengerCarriages = true;
+		cdn.data.renderDerailed = false;
+
+		if (Is2372())
+		{
+			cdn.data.allowRemovalByPopulation = false;
+			cdn.data.shouldStopAtStations = stopAtStation;
+			cdn.data.highPrecisionBlending = false;
+		}
+	});
+
+	return tree;
+}
+
 std::shared_ptr<sync::SyncTreeBase> MakePed(uint32_t model, float posX, float posY, float posZ, uint32_t resourceHash, float heading = 0.0f)
 {
 	auto tree = std::make_shared<sync::CPedSyncTree>();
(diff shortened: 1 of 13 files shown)


Jamie9192:
#Justice4FjamZoo

ext/native-decls/server/CreateTrain.md
```diff
@@ -0,0 +1,32 @@
...
+Creates a functioning train using 'server setter' logic (like CREATE_VEHICLE_SERVER_SETTER and CREATE_AUTOMOBILE).
+
+## Parameters
+* **modelHash**: The model of train to spawn.
+* **x**: Spawn coordinate X component.
+* **y**: Spawn coordinate Y component.
+* **z**: Spawn coordinate Z component.
+* **direction**: The direction in which the train will go (true or false)
+* **stopsAtStations**: Should the train stop at stations on the track (>b2372)
+* **trackId**: The track the train should follow (0 = Main Track, 3 = Metro track)
```
Korioz, line 20: `cruiseSpeed` is missing here.

Ehbw: Thanks, added it now.

code/components/citizen-server-impl/src/state/ServerSetters.cpp
```diff
@@ -447,6 +539,163 @@ static InitFunction initFunction([]()
...
+			float y = ctx.GetArgument<float>(2);
+			float z = ctx.GetArgument<float>(3);
+			bool direction = ctx.GetArgument<bool>(4);
+			bool stopAtStations = ctx.GetArgument<bool>(5);
+			float speed = ctx.GetArgument<float>(6);
+			uint32_t trackId = ctx.GetArgument<uint32_t>(7);
+			uint32_t trainConfigIndex = ctx.CheckArgument<uint32_t>(8);
+
+			//Don't create train if an invalid trackid has been provided.
+			if (trackId != 0 && trackId != 3)
```
glitchdetector, line 569: While it's understandable to check for track IDs, it's possible for servers to load new tracks into the game. This information is not available to the server though, so the check (afaik) can't be changed to include those. Plus, the default tracks outside of ID 0 and 3 are perfectly valid and I see no reason why you would not be allowed to use them.

code/components/citizen-server-impl/src/state/ServerSetters.cpp
```diff
@@ -447,6 +539,163 @@ static InitFunction initFunction([]()
...
+			{
+				throw std::runtime_error(va("Tried to spawn train on invalid track id: %u", trackId));
+
+				ctx.SetResult(0);
+				return;
+			}
+
+			//Don't create train if an invalid trainConfigIndex has been provided, otherwises the client crashes.
+			if (trainConfigIndex > (uint32_t)(Is2802() ? 26 : (Is2372() ? 25 : 24)))
```
glitchdetector, line 578: The same applies for train config IDs, as a server can add new ones to the game. This check stops those from being usable.

glitchdetector (commented):
Validations do not take into account potential content additions by the server.

code/components/citizen-server-impl/src/state/ServerSetters.cpp
glitchdetector: Since it's not currently feasable to have the server be aware of valid track and train config IDs, then a way to bypass these checks would be useful, at least for servers that specifically have additional content. 
Possibly ConVar check to bypass, or some way to "add valid IDs"(?)

Mycroft-Studios:
imo, it would be nice to get the feature out, and then add support for custom tracks later, a working but limited feature is better than no feature

FjamZoo:
Any update on this???

CoZmoTheGod:
Bump, please merge this PR. A lot of work was done in testing this by @Ehbw and this would be an amazing thing to have!

ahcenezdh:
> Bump, please merge this PR. A lot of work was done in testing this by @Ehbw and this would be an amazing thing to have!

Yes, would be amazing

CeebDev:
Any update on this???

Legacy-TacticalGamingInteractive:
Is this essentially something we can use, if merged ofc, in order to rewrite a script such as https://github.com/LexTheGreat/TrainSportation in order to stop so much desync between our client sided trains etc?

Heisenberg691:
please merge this, we want trains

Legacy-TacticalGamingInteractive:
> please merge this, we want trains

one can only hope heh

ahcenezdh:
:eyes:

MasterXDev:
Any update on this???

guhblag:
Any update on this???

SSnowly:
MERGE IT, MERGE IT WE WANT TRAINS ON SERVER SIDE GOOOO

Ehbw:
Closing this for now. It needs some improvements (especially around handling custom train and track configs). Generally trains could do with a few more fixes (taken the only client side train fix in this pr and made a new PR for that #2911)

Legacy-TacticalGamingInteractive:
</3

PR #2325, opened 2024-01-02
Wildcard problem with REGISTER_RESOURCE_AS_EVENT_HANDLER

Alchemy-000:
### Goal of this PR
Is to fix wildcard problem with native REGISTER_RESOURCE_AS_EVENT_HANDLER that is used in scheduler.lua and main.js when adding event handlers and if you use the native in your own code and disable filter with wildcard, events will be called multiple times....

Changes will make sure that if wildcard is presence it will not add event names to that filter list and if theres some when adding wildcard it will remove those....

Atleast FiveM Client and Server

### Checklist

### DIFF
--- a/code/components/citizen-scripting-core/include/ResourceScriptingComponent.h
+++ b/code/components/citizen-scripting-core/include/ResourceScriptingComponent.h
@@ -86,7 +86,15 @@ public:
 
 	inline void AddHandledEvent(const std::string& eventName)
 	{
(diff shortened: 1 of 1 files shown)


Alchemy-000:
Changes were made to wrong function and didn't change anything basically....

AvarianKnight:
This doesn't change the fact that you really shouldn't be using natives marked internal.

PR #2359, opened 2024-01-22
update scheduler.lua to handle if error is nil

MWRPLLC:
Current behavior produces an error in scheduler.lua and this commit fixes that.

### Goal of this PR

wasabi_bridge + ox_target produced an error -that was nil somehow causing scheduler.lua to throw an error.

Resolves a nil value not being handled.

FiveM

### Successfully tested on

**Game builds:** .. 

**Platforms:** Windows, Linux

### Checklist

### DIFF
--- a/data/shared/citizen/scripting/lua/scheduler.lua
+++ b/data/shared/citizen/scripting/lua/scheduler.lua
@@ -472,7 +472,7 @@ local function doStackFormat(err)
 	local fst = FormatStackTrace()
 	
 	-- already recovering from an error
-	if not fst then
+	if not fst or not err then
 		return nil
 	end
 
--- a/.gitmodules
+++ b/.gitmodules
@@ -179,9 +179,6 @@
 [submodule "vendor/libnyquist"]
 	path = vendor/libnyquist
 	url = https://github.com/ddiakopoulos/libnyquist.git
-[submodule "ext/txAdmin"]
-	path = ext/txAdmin
-	url = https://github.com/citizenfx/txAdmin.git
 [submodule "vendor/prometheus-cpp"]
 	path = vendor/prometheus-cpp
 	url = https://github.com/jupp0r/prometheus-cpp.git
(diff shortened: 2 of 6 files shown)


nihonium-cfx (approved):

data/shared/citizen/scripting/lua/scheduler.lua
```diff
@@ -472,7 +472,7 @@ local function doStackFormat(err)
 	local fst = FormatStackTrace()
 	
 	-- already recovering from an error
-	if not fst then
+	if not fst or not err then
```
blattersturm, line 475: This will silently suppress a `nil` error, and won't fix the case where an error object is e.g. a `number` or `boolean`. Doing something like the Lua codebase itself does might be better:

https://github.com/lua/lua/blob/e288c5a91883793d14ed9e9d93464f6ee0b08915/lua.c#L135-L142

blattersturm (commented):
I'm not sure how this got approved. First off, the commit message doesn't match the naming scheme so it'd look odd in changeloggen. Secondly, there's a random merge commit tossed in of the author merging their own merge request.

Finally, there's no real reproduction case, it doesn't fix whatever underlying issue (if any, like why would the combination of resources "return a nil error"), doesn't reference any report, nor does it really fix the behavior it's meant to fix, instead introducing _new_ issues like silently-dropped errors.

data/shared/citizen/scripting/lua/scheduler.lua
thelindat: At the very least `string.format` would resolve the issue with concatenating non-string values.
```lua
return string.format('^1SCRIPT ERROR: %s^7\n%s', err or '', fst)
```

thorium-cfx:
@nihonium-cfx want to put this back into triage or want to assign it to someone?

thorium-cfx:
Marked as stale and invalid, please look into the feedback offered by blattersturm and thelindat

FabianTerhorst:
Closing this in favor of #2631

PR #2360, opened 2024-01-24
Rework camera shake toggle convar in FiveM

Disquse:
### Goal of this PR
The way how this functionality was implemented before is very fragile and got broken in b3095. The one-way nopping was also breaking the ability to disable this option after disconnecting from a server. So it was decided to rework it a bit.

By reworking the way how this functionality was implemented.

FiveM

### Successfully tested on

**Game builds:** 1604, 2060, 2189, 2372, 2545, 2612, 2699, 2802, 2944, 3095

**Platforms:** Windows

### Checklist

### Fixes issues
Addresses the problem reported in this forum topic (`#3`):
https://forum.cfx.re/t/a-couple-of-problems-since-the-last-updates/5200554

### DIFF
--- a/code/components/gta-core-five/src/DisableCameraShake.cpp
+++ b/code/components/gta-core-five/src/DisableCameraShake.cpp
@@ -6,40 +6,45 @@
  */
 
 #include <StdInc.h>
-#include <GameInit.h>
 #include <Hooking.h>
 #include <CoreConsole.h>
-#include <nutsnbolts.h>
 
-static std::shared_ptr<ConVar<bool>> g_cameraShakeConvar;
+#include "Hooking.Stubs.h"
 
-static void CameraShakeOverride()
+static bool g_disableCameraShake;
+
+static void (*g_origUpdateCameraVehicleHighSpeedShake)(void*);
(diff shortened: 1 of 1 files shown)


code/components/gta-core-five/src/DisableCameraShake.cpp
```diff
@@ -6,40 +6,45 @@
...
+#include "Hooking.Stubs.h"
 
 static std::shared_ptr<ConVar<bool>> g_cameraShakeConvar;
 
-static void CameraShakeOverride()
+static void (*g_origUpdateCameraVehicleHighSpeedShake)(void*);
+static void UpdateCameraVehicleHighSpeedShake(void* a1)
 {
-	if (g_cameraShakeConvar->GetValue())
+	if (!g_cameraShakeConvar->GetValue())
```
nihonium-cfx, line 19: Feels like `g_disableCameraShakeConvar` would be a better name to reflect the value of this convar meaning. Since this is a rework - we can improve this as well.

blattersturm: Even better: how about not calling GetValue on the convar in the hook but instead using the tracking variable support so that there's only a single `bool` check?

PR #2362, opened 2024-01-24
fix(state/statebag): filter frequent & large data from client

### LINKED ISSUE #2361: Lack of state bag rate limiting
### What happened?

Somebody set numerous (100+) statebags on an entity with payloads of around a megabyte each on an entity.
This in turn causes the server to crash because this data has to be networked to all players around them.

### Expected result

Instead of a server crash the state bag update should have been rejected

### Reproduction steps

1. Have a lot of players in one area
2. Set a state bag on a entity you have ownership over (so create a object)
3. Add statebag with random keys and random enormous payloads onto this entity
4. Wait for server to crash

### Importancy

Crash

### Area(s)

FXServer

### Specific version(s)

Server 7290 Linux & Server 7339 Win32

### Additional information

I would suggest a simple rate limiter the same way they are implemented on msgServerEvent would probably be enough to fix any issues regarding this, so a rate limiter for the number of events and one for the size of the payload.

Also having the possibility to cancel the AddStateBagChangeHandler somehow would really be appreciated to be able to add some more custom security regarding this, so being able to validate state bags with custom logic dependant on the server.

I'm not to familiair with the internals on how the rate limiters work exactly with the console context so that's the reason I'm creating an issue instead of a pull request.

AvarianKnight:
* if we accept these willy-nilly, clients could possibly OOM the server, or do a one-to-many make-shift DoS on larger a server
* the size limits here are kept high to keep compatibility with people who "misuse" state bags
* sane defaults for servers that don't send large data via state bags would be `set rateLimiter_stateBagSize_rate 1000` and `set rateLimiter_stateBagSize_burst 4000`

### Goal of this PR

Fix the possible abuse cases for state bags, though this doesn't cover all of them [theres already another pr sitting](https://github.com/citizenfx/fivem/pull/2108) that allows better filtering.

The rate and size of state bags similar to `msgServerEvent`

Server

### Successfully tested on

**Game builds:** N/A

**Platforms:** Windows

### Checklist

### Fixes issues
Fixes #2361

### DIFF
--- a/code/components/citizen-server-impl/src/state/ServerGameState.cpp
+++ b/code/components/citizen-server-impl/src/state/ServerGameState.cpp
@@ -30,6 +30,8 @@
 #include <boost/range/adaptors.hpp>
 #include <boost/math/constants/constants.hpp>
 
+#include <KeyedRateLimiter.h>
+
 #include <OneSyncVars.h>
 #include <DebugAlias.h>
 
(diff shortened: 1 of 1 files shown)


blattersturm:
Do we have a good outlook on what common data sizes people use are, even in worst-case scenarios? Imposing a total size limit might also help for other scenarios (memory leak spam, e.g.).

The report, of course, rather is an issue with clients submitting data to *other clients* arbitrarily and it overloading the network layer, for which this fix does appear reasonable.

AvarianKnight:
> Do we have a good outlook on what common data sizes people use are, even in worst-case scenarios? Imposing a total size limit might also help for other scenarios (memory leak spam, e.g.).

The common use case for us with state bags is small data, generally only 100 bytes at maximum, except for some global state usage (which is around 200-500 bytes, which still wouldn't get affected by this). 

Would be interesting to know other servers general usage.

AvarianKnight:
It should also be noted this is more of a mitigation then an actual fix (though this should have these to prevent spammy attacks anyways).

My idea for fixing 
1. Revert the default to server-side creation as a default
2. Add a new native to allow for certain state bag names to be created by the client
A bad name would be something like: `AllowClientCreationOfStateBag("key")`
3. Assign protobuf definitions to keys so the server can automatically reject any malformed requests for state bags.

blattersturm:
The third point would be a fairly interesting project for other serialization bits as well - schema validation so there's no 'send an object where a bool is expected'-style replication issues for events that have such an annotation.

FlawwsX:
> Revert the default to server-side creation as a default

I wouldn't expect this to get approved because it would probably break a lot of scripts out there, but some sort of configuration **option** for server owners to control this would be nice.

AvarianKnight:
> > Revert the default to server-side creation as a default
> 
> I wouldn't expect this to get approved because it would probably break a lot of scripts out there, but some sort of configuration **option** for server owners to control this would be nice.

When it comes to security vs compat, security should probably be of higher priority.

AvarianKnight:
FiveM has currently fallen to having most sane defaults behind some 'configuration option' due to maintain compatibility, which means that most of the time the server *has to run into* the issue before it can actually know that it doesn't have the right default. 

This is also *very hard* to find out for this issue as state bags show as hashes in the timeout log instead of proper packet attribution  which was provided here https://github.com/citizenfx/fivem/pull/2141

Changing from server-first creation to client can create all was a massive mistake that I voiced at the time in Discord, but I the channels that the conversations were done in were wiped so I can't go back and reference them.

blattersturm:
> FiveM has currently fallen to having most sane defaults behind some 'configuration option' due to maintain compatibility

This is what the eventual goal for the _original_ FXv2 effort was, yes - a large watershed point of a few big compat changes that then, if a server runs _only_ FXv2 resources, would put the server in some 'modern' mode which then could be acted upon in e.g. discovery UI in some way to make it a 'desirable' goal for server owners and resoruce authors to work towards.

code/components/citizen-server-impl/src/state/ServerGameState.cpp
```diff
@@ -4245,13 +4247,75 @@ void ServerGameState::AttachToObject(fx::ServerInstanceBase* instance)
...
+		// You can adjust this with `set rateLimiter_stateBagFlood_burst [max burst]`
+		// this allows the rate to temporarily increase without instantly flagging to kick them
+		constexpr double kStateBagRateLimitFloodBurst = 100.f;
+		static auto stateBagRateFloodLimiter = stateBagRateLimiterStore.GetRateLimiter("stateBagFlood", fx::RateLimiterDefaults{ kStateBagRateFloodLimit, kStateBagRateLimitFloodBurst });
+
+		// You can adjust this with `set rateLimiter_stateBagSize_rate [max state bag size]`
+		// Example: `set rateLimiter_stateBagSize_rate 1024` would reduce the maximum state bag size to 1 kilobyte
+		constexpr double kStateBagSizeLimit = 64 * 1024.0;
+		// You shouldn't adjust this value. This value is hard coded to be 128kb internally and should never be larger than this..
+		constexpr double kStateBagSizeLimitBurst = 128 * 1024.0;
```
blattersturm, line 4282: 'internally': perhaps specify *where* in the codebase?

code/components/citizen-server-impl/src/state/ServerGameState.cpp
```diff
@@ -4245,13 +4247,75 @@ void ServerGameState::AttachToObject(fx::ServerInstanceBase* instance)
...
+
+		const uint32_t netId = client->GetNetId();
+
+		if (!stateBagRateLimiter->Consume(netId))
+		{
+			if (!stateBagRateFloodLimiter->Consume(netId))
+			{
+				if (!g_ignoreStateBagWarnings->GetValue() && !client->IsDropping())
+				{
+					console::Printf("state-bag-warning", "Client %s %i got dropped for sending too many state bag value updates.\n", client->GetName(), netId);
```
blattersturm, line 4293: Do we handle channel names with `-` properly in suppression logic?

code/components/citizen-server-impl/src/state/ServerGameState.cpp
```diff
@@ -4245,13 +4247,75 @@ void ServerGameState::AttachToObject(fx::ServerInstanceBase* instance)
...
+					console::Printf("state-bag-warning", "If you believe this to be a mistake please increase your rateLimiter_stateBag_rate and rateLimiter_stateBag_burst. ");
+					console::Printf("state-bag-warning", "You can do this with `set rateLimiter_stateBag_rate [new value]`. The default rate limit is 50 and burst limit is 100\n");
+					console::Printf("state-bag-warning", "You can disable this warning with `set sv_ignoreStateBagWarnings true` if you think you have this properly set up.\n");
```
blattersturm, line 4296: If this has a channel name, couldn't one use channel silencing instead of a dedicated convar for warnings?

(`con_addChannelFilter state-bag-warning drop` or the likes)

code/components/citizen-server-impl/src/state/ServerGameState.cpp
AvarianKnight: They seem to work fine.

code/components/citizen-server-impl/src/state/ServerGameState.cpp
AvarianKnight: I actually forgot about this, swapped to using this instead of the ConVar.

nihonium-cfx (approved):

PR #2368, opened 2024-01-31
tweak(client): remove non-ASLR game executables

blattersturm:
Since b6cc153df04ec066b01f7647ca150a684501844a blocks the client from running on Windows 7 or 8, these builds aren't needed anymore, as they served as a workaround for Windows before 8.1 not supporting high-entropy virtual addresses, leading to the logic in [`hook::get_unadjusted`](https://github.com/citizenfx/fivem/blob/66d052957b28a5102eba86fad5947e40b22597dc/code/client/shared/Hooking.h#L82) not working.

When merging this PR, keep in mind:
1. The change in `sticky` should be mirrored in `adhesive` in `fivem-private` as well.
2. If incremental builds do not remove old binary artifacts, the non-ASLR binaries should be removed manually.

### Goal of this PR
Clean up the build output and remove unneeded executables.

Dropping the unneeded non-ASLR builds.

FiveM, RedM

### Successfully tested on

**Game builds:** 3095
**Platforms:** Windows

### Checklist

### Fixes issues
N/A

### DIFF
--- a/code/client/launcher/premake5.lua
+++ b/code/client/launcher/premake5.lua
@@ -73,11 +73,14 @@ local function isGamePersonality(name)
 	return false
 end
 
-local function launcherpersonality_inner(name, aslr)
+local function launcherpersonality_inner(name)
 	local projectName = name == 'main' and 'CitiLaunch' or ('CitiLaunch_' .. name)
+	local subprocessName = name
 
-	if aslr then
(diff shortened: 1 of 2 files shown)


gottfriedleibniz:
`launcherpersonality_inner` could be renamed to `launcherpersonality` since the indirection is not required anymore. That can be done at another time.

PR #2371, opened 2024-02-01
feat(extra-natives/five): Add `SET_INTERIOR_PROBE_LENGTH` for FiveM

z3t4s:
### Goal of this PR

Allow servers to customize the probe distance used in `CPortalTracker`.

All dynamic entities own an instance of `CPortalTracker`. It is used to detect traversal from and into interiors, through portals and other related actions. 
By default the probing function inside `CPortalTracker`'s tick/update function probes downwards 4.0 units, trying to extract the interior the entity is in from collision data (staticboundsstore).

In many cases this will not work, leading to ~30 hard coded special cases based on interior archetype hash. Rockstar extends the probing distance for these interiors, most notably in `xs_arena_interior` to 150.0f.

By adding a new FiveM native `SET_INTERIOR_PROBE_LENGTH` to give servers individual control of game behaviour depending on their needs.

- The override value is applied to all probes shorter than itself. 
- User input is clamped to `0.0f - 150.0f`
- Session reset is handled, the override is reset on disconnect

Test code:
```lua
RegisterCommand("setInteriorProbeLength", function(src, args, raw)
    local probeLength = (tonumber(args[1]) + 0.0)

    print("Extending interior detection probes to: ", probeLength)
    SetInteriorProbeLength(probeLength)
end)

RegisterCommand("resetInteriorProbeLength", function()
    print("Resetting interior detection probes to default settings")
    SetInteriorProbeLength(0.0)
end)
```

Examples of use cases:
- Specific custom interiors with unusual dimensions
- Housing systems with script created floors
- Improved render behaviour when trying to locate and noclip into underground interiors

Demo:
https://streamable.com/jxm0ux

image

FiveM

### Successfully tested on

**Game builds:** All available FiveM builds
**Configurations:** Debug / Release
**Platforms:** Windows

### Checklist

### Fixes issues
Base game behavior

### DIFF
--- a/code/components/extra-natives-five/src/InteriorExtraNatives.cpp
+++ b/code/components/extra-natives-five/src/InteriorExtraNatives.cpp
@@ -8,6 +8,9 @@
 #include <atPool.h>
 #include <DirectXMath.h>
 #include <CrossBuildRuntime.h>
+#include <GameInit.h>
+
+#include "GameValueStub.h"
 
 #define DECLARE_ACCESSOR(x) \
 	decltype(impl.m2060.x)& x()        \
@@ -303,8 +306,37 @@ static int GetInteriorRoomIdByHash(CMloModelInfo* arch, int searchHash)
 	return -1;
 }
 
+static GameValueStub<float> g_emitterAudioEntityProbeLength;
+static float g_interiorProbeLengthOverride = 0.0;
+
+static bool (*g_CPortalTracker__Probe)(Vector3* pos, CInteriorInst** ppInteriorInstance, int* roomId, Vector3* traceImpactPoint, float traceLength);
+static bool CPortalTracker__Probe(Vector3* pos, CInteriorInst** ppInteriorInstance, int* roomId, Vector3* traceImpactPoint, float traceLength)
+{
+	// game code has a lot of different special case handling in CPortalTracker::vft0x8
+	// joaat('xs_arena_interior') seems to be the case with the longest traceLength override (150.f)
+	//
+	if (g_interiorProbeLengthOverride > 0.0f && traceLength < g_interiorProbeLengthOverride)
+	{
+		traceLength = g_interiorProbeLengthOverride;
+	}
+
+	return g_CPortalTracker__Probe(pos, ppInteriorInstance, roomId, traceImpactPoint, traceLength);
+}
+
 static HookFunction initFunction([]()
 {
+	{
+		auto location = hook::get_pattern<void>("E8 ? ? ? ? 40 8A F8 49 ? ? E8");
+		hook::set_call(&g_CPortalTracker__Probe, location);
+		hook::call(location, CPortalTracker__Probe);
+	}
+
+	{
+		auto location = hook::get_pattern<uint32_t>("33 ED 39 A9 ? ? ? ? 0F 86 ? ? ? ? F3", 18);
+		g_emitterAudioEntityProbeLength.Init(*hook::get_address<float*>(location));
+		g_emitterAudioEntityProbeLength.SetLocation(location);
+	}
+
 	{
 		auto location = hook::get_pattern<char>("BA A1 85 94 52 41 B8 01", 0x34);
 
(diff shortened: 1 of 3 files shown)


qalle-git:
Great work!

blattersturm:
I assume the expectation is the script command will be invoked from an existing spatial check/trigger script on a server, correct? The checks in `CPortalTracker::Update` for probe type overrides look _nasty_ so mimicking those would seem undesirable.

nikez:
Set it once and forget or have individual spatial checks, I guess that's entirely up to whoever implements it. I've tested noclip:ing around with it set to 150f without any issues, but not with full test coverage, of course. I think most people would survive with 16f as length.

And yeah, it's unreasonably expensive to implement this on a per interior basis.. due to the nastiness of the internals of this function.

hamcoremusic:
Really looking forward to this extra native as it will greatly help my team in being able to create the visuals they want in their MLO interiors when using dynamic props. Is there a chance this is in an update coming up?

gottfriedleibniz:
Looking at `CPortalTracker::Update` there is one case unrelated to those `xs_arena_interior` interior hard codes that I'm still trying to figure out to determine potential side-effects. It is the one where `traceLength` is determined by the archetypes `bsRadius` plus another hard coded value. 

This conceptually seems different to the "we are in X; increase `traceLength` to 150.f"; `CInteriorInst` creation perhaps? Would be interested to know if you found anything here.

Another question would be on the naming here, e.g., `SET_` vs `OVERRIDE_` vs. `OVERRIDE_MIN...`  vs. `CLAMP_`. I have no strong opinion, but others may.

gottfriedleibniz:
> Completely indifferent to that too. Happy to accept suggestions.

I trust Disquse's judgment on this.

> I assume this is a mission script fix. Probably a floating or ceiling mounted object somewhere.

This is part of the confusion. From my own testing this path seemed to be hitting quite a few ped and vehicle entities with registered `rage::netObject`'s. This being `CInteriorInst` creation related would make sense, i.e., figuring out what belongs to the interior.

Consequently, one concern is whether this override (esp. large values) for this specific case will cause any weird interactions with any interior-related bits in netcode. This would include our own `NetInteriorLocationHacks` and servers without OneSync enabled (e.g., ownership of far away entities). The limited probe length for mission entities w/ how some custom interiors are designed is certainly in issue in this case. Ugh.

Hard to be definitive here since so *much* of this (dis)assembly is nasty and gross. I do not think this should be prohibitive of throwing this on canary and seeing if users notice any unusual artifacts.

> audEmitterAudioEntity probes a fixed 20.0 units

Overriding this specific value seems like an instant value add.

> Something that is related to ped and vehicle population.

Keeping this value default would probably be best 😅. The others I have no strong opinion on (e.g., modifying native behaviour vs. providing two new ones which offer custom probe lengths.)

Mellowtri:
This is awesome! I have been wanting a solution like this for a long time. Thank you, can't wait to try it out. Hope to see it on canary soon.

Out of curiosity, did you look into why vehicles do not seem to care about floor distance in the same way peds do?
If I am in a vehicle and fly into an interior, it will load without any issues, even when I am not above the floor collision and the distance doesn't seem to matter at all.

gottfriedleibniz:
Would it be possible to split the `UpdatePortalTracker` changes into its own PR? Bit of feature creep and a change that is a bit more dangerous. This will also allow us to get the interior probe length changes onto canary a bit faster.

Other comments:

- The `__fastcall` keyword can be dropped (x64 and all).
- Should `SET_INTERIOR_PROBE_LENGTH` allow values smaller than 1.0f? That seems to be the smallest value used by the game. Would there be any reason to go below that?
- The default value to `game_emitterAudioEntityProbeLength` should match the current in-game value (20.0f; same on reset). Secondly, would there be any value in this being made scriptable? Would the same assumption [above](https://github.com/citizenfx/fivem/pull/2371#issuecomment-1927821954) apply here?
- After further consideration, I question whether changing those natives is something worth risking at this point. In addition to creating a weird dependency between natives, @Disquse in another conversation raised a good point in that we have *generally* avoided altering behaviour here except to maintain backward compatibility or fix in-game bugs.

gottfriedleibniz:
Rebased branch to 1b01a54ea0803c8e06ef7e15bb1ae3a3d64de085 and included a small fix for `g_emitterAudioEntityProbeLength` needing to be 20.0f by default (and on resetg).

gottfriedleibniz (approved):

PR #2383, opened 2024-02-10
feat(extra-natives/five): current amount and maximum pool size

nikez:
### Goal of this PR

To allow you to see the current pool size and it's maximum size.
...

By exposing the current amount and the maximum pool size.
...

FiveM, Natives
...

### Successfully tested on

**Game builds:** 2372

**Platforms:** Windows

### Checklist

### DIFF
--- a/code/components/extra-natives-five/src/PoolTraversalNatives.cpp
+++ b/code/components/extra-natives-five/src/PoolTraversalNatives.cpp
@@ -19,8 +19,9 @@ class RefPool
 {
 private:
 	TEntry** m_baseAddress;
-	uint32_t m_count;
-	char m_pad[36];
+	size_t m_count;
+	size_t m_freeSpace;
+	char m_pad2[24];
 	uint32_t* m_validBits;
 	// ...
 
@@ -44,6 +45,12 @@ public:
 	size_t GetSize()
 	{
 		return m_count;
+	}
+
+	// copying the behaviour of atPoolBase by returning the used space
+	size_t GetCountDirect()
+	{
+		return (m_count - m_freeSpace);
 	}
 };
 #endif
@@ -161,6 +168,15 @@ static void SerializePool(fx::ScriptContext& context)
 	}
 
 	context.SetResult(fx::SerializeObject(guids));
+}
+
+template<typename TTraits>
+static void GetPoolSize(fx::ScriptContext& context)
+{
+	TTraits::PoolType* pool = TTraits::GetPool();
+
+	context.SetResult<int>(pool->GetCountDirect());
+	*context.GetArgument<int*>(1) = pool->GetSize();
 }
 
 struct FindHandle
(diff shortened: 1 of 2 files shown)


thorium-cfx:
Hello Nikez, could you add a bit more context on why you want to be able to see it? Thank you

nikez:
Hi,

Sure, we had an instance where one of our scripts caused the `CObject` pool to overflow, effectively crashing people when a networked cloning event came in, as there's a nullptr check that will assert/crash you. So this is only for people to add a safeguard. So to avoid this in the future, we'll be using this for our "synced objects" (local objects create on every client rather than a networked object).

In general it's also a pretty good metric to track.

code/components/extra-natives-five/src/PoolTraversalNatives.cpp
```diff
@@ -291,6 +307,23 @@ static InitFunction initFunction([]()
 			throw std::runtime_error(va("Invalid pool: %s", pool));
 		}
 	});
+
+	fx::ScriptEngine::RegisterNativeHandler("GET_GAME_POOL_SIZE", [](fx::ScriptContext& context)
+	{
+		std::string pool = context.CheckArgument<const char*>(0);
+		if (pool.compare("CPed") == 0)
```
Disquse, line 314: We're using brackets even for single-line blocks.

```cpp
if (pool.compare("CPed") == 0)
{
  GetPoolSize<PedPoolTraits>(context);
} 
else if (pool.compare("CObject") == 0)
{
// etc...
```

nikez: I knew this would come up - so, do you want us to **deviate** from already existing implementations? Previously this has been frowned upon.

https://github.com/citizenfx/fivem/blob/c73d37249419fa9ef06947652a6a2065d0057e24/code/components/extra-natives-five/src/PoolTraversalNatives.cpp#L294

nikez: If so, why .compare() and not string_view, etc. It opens up for further improvements.. But do we _not_ fix the other (almost identical) native?

Disquse: Iirc there was some PR that meant to rework the related a bit, we also suggested to add brackets. Not sure what happened to that PR...

nikez: I get it, but as active contributors having clear guidelines on how to proceed with this in the future would be much appreciated.

Do we **follow** already existing code and style, even if wrongly implemented?
Do we only follow the guide line for the new code we add and live with the old, non-compliant code?
Do we implement the new code as per the style guide and also **fix** any related natives that are also incorrect?

Disquse: > Do we follow already existing code and style, even if wrongly implemented?

Usually this, yes.

Disquse:
This PR also needs to be tested in RedM. @nikez said he doesn't have anything for the RedM workflow so I will test it myself soon.

gottfriedleibniz:
> local objects create on every client rather than a networked object

There seems to be a common use-case with player generated dynamic interiors, sets, whatever, causing the # of objects to sky-rocket (or players with 3000 weapons attached to their Ped). Being able to capture statistics e.g., like `GET_USED_CREATOR_BUDGET`, seems to be a desired one affected of this PR. 

> This seems to be intentional behaviour to ensure integrity between remote sessions, as failure to create an object would cause desync.

Do you think this failure is possible with the other `*CreationDataNode` code paths?

> So this is only for people to add a safeguard.

Presumably for safety any use of this native would have to subtract whatever the value of the `CNetObjObject` pool is - and this *seems* like to be really fighting against game internals. This part is definitely something we should attempt to handle more gracefully.

prikolium-cfx:
Hi Nikez. Is it still needed including that fact that now you can extend your pools on connect to server?

KillStr3aK:
> Hi Nikez. Is it still needed including that fact that now you can extend your pools on connect to server?

mind elaborating?

nikez:
> Hi Nikez. Is it still needed including that fact that now you can extend your pools on connect to server?

Yes! This would help us collect data for crash analytics.

prikolium-cfx:
I don't really like multipurpose functions like this, but to improve current solution please name it `GET_GAME_POOL_COUNT_AND_SIZE` and make this function return `bool` (false if pool not found instead of runtime error) and return both values by pointer.

PR #2287, opened 2023-11-27
tweak(gameconfig): up TxdStore limit to 105500

manzarek123:
I am currently hitting this limit on my server with a vanilla game. I am streaming a lot of clothing so increasing this limit would be appreciated so I don't have to remove content

Issue #2094, opened 2023-07-10
txd poolsize on 2802


SKITTLE6969:
2802 has little space for extra txd Is it possible bump up the txd poolsize to like around 105500 something?

Disquse:
How many texture dictionaries you're trying to stream? I have a resource with more than 4300 dictionaries (map conversion) and it doesn't hit the limit.

SKITTLE6969:
> 

i have no idea how its happening, my server is currently on pure mode 2 and when i change it to pure mode 1 people starts to get txdpool size crashes non stop

indonesiafreeroam:
![image](https://github.com/citizenfx/fivem/assets/139290857/75da283c-b0eb-47d9-b4d4-0db36964663f)

can you please increase txd to game build 2944 thanks

normalday123:
![image](https://github.com/citizenfx/fivem/assets/138086530/bbd624d9-d56d-43b0-b916-0587fb9ad2aa)
![image](https://github.com/citizenfx/fivem/assets/138086530/bad3729b-0a62-455b-bc61-2a156164bde9)

can you please increase txd pool
may server have 33.000 ytd item

thanks

normalday123:
because we add car house and clothess, can you pliss increase, so we can use dlc 2944

Moozdzn:
We are also hitting this limit but I havent seen how many we have

> Why would you need to stream SO MANY #tds?

Assuming only clothing and with every piece having 26 textures (26 ytds)

33.000 ( the number mentioned above ) equals to 1269 pieces of clothing in total with 115 per component, 11 in total.

Doesn't seem to much, nor does it fill an entire clothing dlc

(Unless my math is wrong and this comment is useless)

normalday123:
![image](https://github.com/citizenfx/fivem/assets/138086530/3ecc8674-c95f-4439-8a6c-b4215040ae29)
here ytd car just 4000++++

![image](https://github.com/citizenfx/fivem/assets/138086530/56293c99-e3c8-4e2e-99be-86694c13b533)
and here the ytd clothes 

if i remove clothes, i can use gamebuild 2944, but what point use fivem key stream clothes, if you cant stream all the clothes you liked. i think other server having this problem too, but they dont complain here...

this no include house, custom map, ped, pleasee increasee ytd fragment limit thank you..

gottfriedleibniz:
[...]

PR #2392, opened 2024-02-14
increase txd store pool size

EnzoDCdev:
### Goal of this PR

with this update, it will be possible to use clothes made by the most famous creators with many variables and not limit yourself to using old gamebuilds...

for exampe: i have a pack of clothes that works with gamebuild 2802, with the latest gamebuild 3095 I get the error, like many other people, of txd store limit to 105500)

the pr archive the goal increasing the value...

fivem

### Successfully tested on
**Game builds:** .. 

**Platforms:** Windows

### Checklist

### DIFF
--- a/data/client/citizen/common/data/gameconfig.xml
+++ b/data/client/citizen/common/data/gameconfig.xml
@@ -458,7 +458,7 @@
 						</Item>
 						<Item>
 							<PoolName>TxdStore</PoolName>
-							<PoolSize value="105500"/>
+							<PoolSize value="205500"/>
 						</Item>
 						<Item>
 							<PoolName>CNetObjVehicle</PoolName>


yannbcf:
My take is, reduce the amount of assets that you are providing to your users, it is content that they download / stream, simply doubling a pool size is not a solution by itself

gottfriedleibniz:
In an ideal world we continue the work started here: https://github.com/citizenfx/fivem/pull/1419, however, getting that code to a completed and well-tested state would take (an unknown amount of) time.

Vanilla builds of 3095 seem to use a number in the vicinity ~76000 while some servers are apparently adding [33000+](https://github.com/citizenfx/fivem/issues/2094#issuecomment-1631640837). So even with the recent [105500](https://github.com/citizenfx/fivem/pull/2287) bump there is a (known) line being crossed.

Unsure what a reasonable value would be here - the comment above is certainly the safest approach to take.

gbalazsfalvi:
I would really appreciate, if you can accept this PL

Pixel-Nomad:
please increase the pool size currently i'm using 400+ cars in my server with 100+ mlos and lots of lots clothing and litrally hit the 105500 pool mark can we increase it to 155500 and not to 205500 since i think i does include default textures aswell but in general it will gonna hit when gta releases new updates if this pool size counts gta default textures

prikolium-cfx:
> please increase the pool size currently i'm using 400+ cars in my server with 100+ mlos and lots of lots clothing and litrally hit the 105500 pool mark can we increase it to 155500 and not to 205500 since i think i does include default textures aswell but in general it will gonna hit when gta releases new updates if this pool size counts gta default textures

We are designing better and more universal solution for this issue. No ETA but we will try to deliver that as soon as possible.

AlphaCarry:
> please increase the pool size currently i'm using 400+ cars in my server with 100+ mlos and lots of lots clothing and litrally hit the 105500 pool mark can we increase it to 155500 and not to 205500 since i think i does include default textures aswell but in general it will gonna hit when gta releases new updates if this pool size counts gta default textures

did you find the way of increase poolsize?

liberty40:
Any update ?

PR #2394, opened 2024-02-16
Earlier connect to server when using "fivem://connect"

slashkeyvalue:
### Goal of this PR

When using ""fivem://connect"", the time between launching the game and getting any "connecting" feedback can about a minute on slower machines, so this PR aims to make use of a very cheap speed up to try and improve user feedback as it seems that the game can run initfunctions while connecting to a server instead of having to wait before connecting.

The game launcher now attempts a ConnectTo when netLibrary is created instead of waiting on INIT_CORE

FiveM, RedM

### Successfully tested on

**Game builds:** .. 

**Platforms:** Windows

### Checklist

### DIFF
--- a/code/components/glue/src/ConnectToNative.cpp
+++ b/code/components/glue/src/ConnectToNative.cpp
@@ -1477,14 +1477,11 @@ void Component_RunPreInit()
 		{
 // #TODOLIBERTY: ?
 #ifndef GTA_NY
-			rage::OnInitFunctionStart.Connect([](rage::InitFunctionType type)
+			NetLibrary::OnNetLibraryCreate.Connect([](NetLibrary*)
 			{
-				if (type == rage::InitFunctionType::INIT_CORE)
-				{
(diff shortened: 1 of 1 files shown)


gottfriedleibniz:
I am not sure this is the correct approach to take at the moment.

In releng we have reports of slow game startup times due to other factors which your users may be experiencing the brunt of. It would be nice to address those problems prior to considering having this instance of "ConnectTo" be moved to so early in the initialization/hooking stage (which is likely to only be a temporary solution).

slashkeyvalue:
Unfortunately i don't have access to releng, so i just assumed that initfunction stuff was *supposed* to be a little slow ( as it'd be loading game stuff and whatnot ). But yeah, if there's some deeper problems related to loading times it'd be wise to focus on that.

That being said, even if though it's a temporary solution, i still think users could benefit of the increased connecting-to-server feedback while research/planning on actually improving startup times is done

PR #2396, opened 2024-02-20
feat(scripting/v8): allow to set the max listeners count

joelwurtz:
### Goal of this PR
By default any event that is listened more than 10 times will trigger a log error saying that there may be a memory leak. 

In some cases users may want to listen more than 10 times to a specific event this patch allow to not display a log message in those cases by setting the correct value to the setMaxEventListeners.

It's not a real problem, but often users will print this error saying it's the cause of a bug (but it's not) and it's misleading.

This PR allow to configure the max listeners count so a developer can set it according to his code behavior.

FiveM, RedM, Server, ScRT: JS

### Successfully tested on

**Platforms:** Windows, Linux

### Checklist

### Fixes issues

N/A

### DIFF
--- a/data/shared/citizen/scripting/v8/index.d.ts
+++ b/data/shared/citizen/scripting/v8/index.d.ts
@@ -83,10 +83,12 @@ type CitizenImmediate = Omit<CitizenTimer, 'refresh'>;
 declare var Citizen: CitizenInterface;
 
 declare function addRawEventListener(eventName: string, callback: Function): void
+declare function setMaxRawEventListeners(max: number): void
 
 declare function addEventListener(eventName: string, callback: Function, netSafe?: boolean): void
 declare function on(eventName: string, callback: Function): void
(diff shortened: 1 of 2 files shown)


nihonium-cfx (changes requested):
Hey there, thanks for the PR!

1. Commit message needs to refer to `scripting/v8` instead of `scripting-v8`, see this commit for example: https://github.com/citizenfx/fivem/commit/01a29ec5bdd3ee4aed99b0af85879424f2ee774e
2. [TypeScript definitions](https://github.com/citizenfx/fivem/blob/master/data/shared/citizen/scripting/v8/index.d.ts) need to be extended to reflect new API added.

Not required, but as a side-quest: would be great to add this functionality for the raw event listener (`rawEmitter`) as well for feature parity of these.

joelwurtz:
Thanks for the review, i have applied the feedback

nihonium-cfx (approved):

PR #2397, opened 2024-02-21
fix(nui): Fixed to be able to change the resolution setting of Fixed-size NUI

wacha-329:
### Goal of this PR

The purpose is to be able to change the resolution setting of Fixed-sizeNUI, which is one of the items in the FiveM settings.
The current implementation is fixed at 1920x1080, and when playing in 4K, the game cannot be set to 4K and the size of the NUI to WQHD, which lacks scalability, so this modification will make it possible.

By defining FixedSizeWindowX and FixedSizeWindowY in CitizenFX.ini, the resolution can be changed to any desired resolution.
If not defined, the resolution will be 1920x1080.
FixedSizeWindowX ranges is 640-7680
FixedSizeWindowY range is 480-4320
If this range is exceeded, the default is 1920x1080.

FiveM, RedM

### Successfully tested on

**Platforms:** Windows

### Checklist

### Fixes issues

N/A

### Supplement
I was thinking of loading ini within NUIWindow::UpdateFrame(), but I thought it was not a good idea because UpdateFrame is a function that is called frequently.
As a result, we implemented it in a way that the file is loaded in Initialized and the value is held in a variable.
If there are any problems with this implementation, please let us know.
I am new to programming and may have implemented it incorrectly. My apologies.

### DIFF
--- a/code/components/nui-core/include/NUIWindow.h
+++ b/code/components/nui-core/include/NUIWindow.h
@@ -212,4 +212,8 @@ public:
 	void HandlePopupShow(bool show);
 
 	bool IsFixedSizeWindow() const;
+
+private:
+	int m_fixedWidth;
+	int m_fixedHeight;
 };
--- a/code/components/nui-core/src/NUIWindow.cpp
+++ b/code/components/nui-core/src/NUIWindow.cpp
@@ -41,7 +41,7 @@ extern void AddSchemeHandlerFactories(CefRefPtr<CefRequestContext> rc);
 NUIWindow::NUIWindow(bool rawBlit, int width, int height, const std::string& windowContext)
 	: m_rawBlit(rawBlit), m_width(width), m_height(height), m_renderBuffer(nullptr), m_dirtyFlag(0), m_onClientCreated(nullptr), m_nuiTexture(nullptr), m_popupTexture(nullptr), m_swapTexture(nullptr),
 	  m_swapRtv(nullptr), m_swapSrv(nullptr), m_dereferencedNuiTexture(false), m_lastFrameTime(0), m_lastMessageTime(0), m_roundedHeight(0), m_roundedWidth(0),
-	  m_syncKey(0), m_paintType(NUIPaintTypeDummy), m_windowContext(windowContext)
+	  m_syncKey(0), m_paintType(NUIPaintTypeDummy), m_windowContext(windowContext), m_fixedWidth(1920), m_fixedHeight(1080)
 {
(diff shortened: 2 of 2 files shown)


AvarianKnight:
I'm confused why this is necessary? The only reason this option exists is because developers *don't* scale their UI properly for higher resolution monitors.

If you're playing at 4k then you should ideally *not* use this on the server you're playing on if the developers scale their UI properly.

wacha-329:
Thank you for presenting the original commit.
I understood that it was added as a setting to address the case where the UI display is broken.

> If you're playing at 4k then you should ideally not have to use `nui_useFixedSize` if the server owner/developers properly scale their UI.


I agree with that opinion.
However, there are some Plugins that do not support scale, and some of them are not supported even if we ask the developer to do so.
Therefore, I believe that if the client can change the resolution to any desired resolution, such problems will no longer occur.
So I suggested this option.

Thank you.

Mycroft-Studios:
> However, there are some Plugins that do not support scale

thats exactly **why** its locked at 1080p, changing the resoltuion it locks to, wouldnt make literally any "better" scaling experience,
and would just be the same as changing your games resolution

wacha-329:
Let me talk about what made me decide to make this correction.
I want to play the game in 4K, in which case the UI will also scale to 4K.
However, that made the UI too small and the text hard to read, so I enabled Fixed-size NUI and the next thing I knew the UI was too large.
Currently, the resolution is fixed at 1920x1080, and I proposed this PR because I would like to be able to set the resolution of the Fixed-size NUI to 2560x1440 (WQHD).
This modification was born from the idea that we would like to play the game in 4K and the UI in WQHD.
I think this is different from the intention of setting Fixed-size NUI, so please close this PR if you think it is unnecessary.
Thank you.

Mycroft-Studios:
> I want to play the game in 4K, in which case the UI will also scale to 4K.
yeah, that, isnt what this is intended for, or what it does

wacha-329:
OK :)
Thank you for taking the time to help me with this PR.

> I want to play the game in 4K, in which case the UI will also scale to 4K.

Hopefully one day this will happen as a different feature.
Thank you very much.

AvarianKnight:
> > I want to play the game in 4K, in which case the UI will also scale to 4K.
> 
> yeah, that, isnt what this is intended for, or what it does

I don't think this is completely right, this mode is meant to fix bad "assumed" scaling, adding an option to switch between 1080 & 1440 might be worthwhile and more intuitive then the implementation here.

PR #2400, opened 2024-02-24
fix(client): Cleanup various static sig patterns and put them in Hook…

osmium-cfx:
…Function's.

We had a few rare cases here where sigs would be resolved on-demand.

There is some small overhead from checking each local static variable, but really this was done so that all the sigs are done before going in-game.

### Goal of this PR

Put Most if not all sigs into HookFunctions or function stubs that run at startup.

This means that sigs are ran less conditionally(without needing an option) and at startup time instead of 1st time called.

By changing some rare cases of sigs not doing this.

Client

### Successfully tested on

Did basic testing without hints file and it seemed OK

**Game builds:** Several

**Platforms:** Windows

### Checklist

- [x ] Code compiles and has been tested successfully.
- [ x] Code explains itself well and/or is documented.
- [ x] My commit message explains what the changes do and what they are for.
- [ x] No extra compilation warnings are added by these changes.

### DIFF
--- a/code/components/extra-natives-five/src/NuiAudioSink.cpp
+++ b/code/components/extra-natives-five/src/NuiAudioSink.cpp
@@ -1074,14 +1074,15 @@ namespace rage
 		static audCategoryControllerManager* GetInstance();
 	};
 
+	static audCategoryControllerManager** g_audCategoryControllerManager = nullptr;
 	audCategoryControllerManager* audCategoryControllerManager::GetInstance()
 	{
-#ifdef GTA_FIVE
-		static auto patternRef = hook::get_address<audCategoryControllerManager**>(hook::get_pattern("45 33 C0 BA 90 1C E2 44 E8", -4));
-#elif IS_RDR3
-		static auto patternRef = hook::get_address<audCategoryControllerManager**>(hook::get_pattern("48 8B 0D ? ? ? ? E8 ? ? ? ? 8B 15 ? ? ? ? 48 8D 0D ? ? ? ? E8", 3));
-#endif
-		return *patternRef;
+		if (!g_audCategoryControllerManager)
+		{
+			return nullptr;
+		}
+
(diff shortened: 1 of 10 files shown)


code/components/extra-natives-five/src/NuiAudioSink.cpp
```diff
@@ -1074,14 +1074,15 @@ namespace rage
...
+	static audCategoryControllerManager** g_audCategoryControllerManager = nullptr;
 	audCategoryControllerManager* audCategoryControllerManager::GetInstance()
 	{
-#ifdef GTA_FIVE
-		static auto patternRef = hook::get_address<audCategoryControllerManager**>(hook::get_pattern("45 33 C0 BA 90 1C E2 44 E8", -4));
-#elif IS_RDR3
-		static auto patternRef = hook::get_address<audCategoryControllerManager**>(hook::get_pattern("48 8B 0D ? ? ? ? E8 ? ? ? ? 8B 15 ? ? ? ? 48 8D 0D ? ? ? ? E8", 3));
-#endif
-		return *patternRef;
+		if (!g_audCategoryControllerManager)
```
gottfriedleibniz, line 1080: `audCategoryControllerManager::GetInstance` being called from another HookFunction prior to this pattern running a bit nasty. Any potential side-effects here (i.e., until the first `Mumble_RunFrame` call)? I don't believe so, but a second opinion would not hurt.

osmium-cfx: I believe this is only called here?

https://github.com/citizenfx/fivem/blob/master/code/components/extra-natives-five/src/NuiAudioSink.cpp#L2878

Should be OK. Fixed the merge conflicts anyway.

gottfriedleibniz (approved):
Moving code around. ItWorksOnMyMachine sticker certified.

PR #2403, opened 2024-02-27
feat(nui-core): add options to switch between fixed size modes

AvarianKnight:
### Goal of this PR
Allow users to choose between `1920x1080` and `2560x1440` for the scaled UI, fixing bad UI scaling by server developers for users using 4k monitors.

See https://github.com/citizenfx/fivem/pull/2397 for more discussion

Adds `nui_fixedSizeMode` and UI elements to allows the user to choose between `0` (1920x1080) and `1` (2560x1440).

FiveM, NUI

### Successfully tested on
@wacha-329 tested the build and said it works as expected

### Checklist

### Fixes issues
https://github.com/citizenfx/fivem/pull/2397


wacha-329:
Thanks for the great suggestions!
This PR is going to solve the problem I had.
I will build and try it out.

Another reason for the fix I suggested in #2397 is that I have seen other users playing on FiveM with misaligned UI and thought it would be useful to be able to freely determine the resolution of the UI.
I don't know what the resolution of that player is, but it is probably a 16:10 display.
I think it would be useful to have some other resolution options as well.
So I implemented it in the form of a .ini file so that the user can freely change it.
GTAV also has a resolution option, and I think it would be even more usable if we could match it.

Reference: Resolution options when launching GTAV on a 4K display
```
3840x2160, 2560x1600, 2560x1440, 2048x1536, 2048x1280, 2048x1152, 1920x1440, 1920x1200, 1920x1080, 1680x1050,
1600x1200, 1600x1024, 1600x900, 1440x1080, 1440x900, 1400x1050, 1366x768, 1360x768, 1280x1024, 1280x960,
1280x800, 1280x768, 1280x720, 1176x664, 1152x864, 1024x768, 800x600
```

wacha-329:
[](https://github.com/citizenfx/fivem)

We tried builds on both the [latest source code](https://github.com/citizenfx/fivem) and on the [PR branch](https://github.com/AvarianKnight/fivem/tree/feat/additional-fixed-modes).
Both builds were successful and the UI resolution changes worked correctly.
I am attaching a screenshot for reference.
Fixed-size NUI-disable
Fixed-size NUI-enable-1920x1080
Fixed-size NUI-enable-2560x1440

PR #2407, opened 2024-02-29
fix(nui/core): set user-agent for imgur URLs

mninc:
### Goal of this PR

Fix imgur images in NUI.

Imgur recently started blocking requests with `CitizenFX` in the user-agent, responding with status code 429. This overrides the user-agent for requests to imgur.

FiveM, RedM

### Successfully tested on

**Game builds:** 2944

**Platforms:** Windows

### Checklist

### DIFF
--- a/code/components/nui-core/src/NUIClient.cpp
+++ b/code/components/nui-core/src/NUIClient.cpp
@@ -410,6 +410,25 @@ auto NUIClient::OnBeforeResourceLoad(CefRefPtr<CefBrowser> browser, CefRefPtr<Ce
 		}
 	}
 
+	// Imgur started blocking CitizenFX in the user-agent
+	{
+		CefURLParts parts;
+		if (CefParseURL(request->GetURL(), parts))
+		{
+			auto hostString = CefString(&parts.host).ToString();
+
+			if (boost::algorithm::ends_with(hostString, "imgur.com"))
+			{
+				CefRequest::HeaderMap headers;
+				request->GetHeaderMap(headers);
+	
+				headers.erase("User-Agent");
+				headers.emplace("User-Agent", "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/112.0.0.0 Safari/537.36");
+				request->SetHeaderMap(headers);
+			}
+		}
+	}
+
 #if !defined(_DEBUG)
 	if (frame->IsMain())
 	{


Yum1x:
Definitely it's not a solution. Use your own cdn or buy a cheap S3/R2 bucket instead cheating services

mninc:
> Definitely it's not a solution. Use your own cdn or buy a cheap S3/R2 bucket instead cheating services. Also, you can do the same on ur end, by intercepting the fetch request

Using a CDN isn't ideal for user-uploaded images. This approach is already used for discord URLs which is why I think it's ok.

AvarianKnight:
> Using a CDN isn't ideal for user-uploaded images. This approach is already used for discord URLs which is why I think it's ok.

I'm not sure what you mean, this would be the *ideal* solution for handling user-uploaded content

I personally don't think FiveM should keep working around people misusing other websites free APIs, if services don't want FiveM to use their API them maybe FiveM *shouldn't*

mninc:
> I'm not sure what you mean, this would be the *ideal* solution for handling user-uploaded content

Rethinking this and doing some tests, yeah it could probably be implemented to be just as easy as submitting a URL.

> I personally don't think FiveM should keep working around people misusing other websites free APIs, if services don't want FiveM to use their API them maybe FiveM *shouldn't*

Agreed, but discord URLs should probably be treated the same way. I'm not sure why NUI needs an updated user-agent - ideally it should be treated by websites just like a normal browser. Using the default user-agent would be a better solution than specifically working around certain websites.

PsychoShock:
Remember that Discord is **NOT** a method to upload a file. People that have been using it, is wrong and shouldn't never have happen. However, I think Avarian is right on 

> I personally don't think FiveM should keep working around people misusing other websites free APIs, if services don't want FiveM to use their API them maybe FiveM shouldn't

Instead of doing this PR, people should have docs saying what recommanded method to use for cloud management for game servers to upload images, media files, and logs. This is not something "new" and should have happen long time ago. Maybe a partnership with someone (like ZapHosting) would be better here, but this will take a lot of time.

Lucas7yoshi:
Misusing hosting files on third party services for content in servers is something that happens a LOT and frankly is understandable for the third parties to object to.

IMO this is not a solution, and could just trigger a game of cat and mouse and cause shaky stability for users "why imgur go away in my nui? oh its back? oh its gone again...". - Like discord did with requiring expiring tokens in urls to view outside of the app.

The solution is to use a proper service for it, which more often than not doesn't exist (because $^7 at scale) so you have to make your own.

edwardexo:
While I completely understand Imgur's reasoning for this, I believe the PR should be accepted (at least temporarily), giving server owners time to transition to other services. The same change has been done for Discord in the past by Cfx.re.

Historically, Imgur has been the primary platform for hosting user-generated media. In RP servers, this content could be custom backgrounds, profile pictures, uploads from phones, logos, icons etc.

With more and more CDNs beginning to restrict this behavior (recent URL changes in Discord), the ultimate safe solution would be for servers to create their own hosting solutions. 

However, transitioning away from Imgur isn't a task that people can accomplish overnight. Especially hard for smaller communities, where setting up an image upload server might not even be doable for them. Therefore a transition period would be needed, where people could explore other options.

PsychoShock:
> While I completely understand Imgur's reasoning for this, I believe the PR should be accepted (at least temporarily), giving server owners time to transition to other services. The same change has been done for Discord in the past by Cfx.re.

Again, this shouldn't have happen in the first place. CFX doesn't need to give people "time" when it's not caused by the platform. If Imgur blocked CFX is for a reason and it will not be appropriate for a "big" compagny like R* to do such a thing.

> With more and more services beginning to restrict this kind of usage (recent URL changes in Discord), the ultimate safe solution would be for servers to create their own hosting solutions.

Like I said, multiples websites give this service for free every month that people could start using!

> However, transitioning away from Imgur isn't a task that people can accomplish overnight. Especially hard for smaller communities, where setting up an image upload server might not even be doable for them. Therefore a transition period would be needed, where people could explore other options.

I mean, it is. People could switch on free hosting server for logs, upload files and more in few minutes. Remember that [CFX discord](discord.gg/cfx) is always available for people that doesn't know how to ask help.


This whole conversation shouldn't even have place. This PR just try to conturn something a compagny purposely blocked and again, I do agree something should be done since it's used by many server like docs with some recommandations. This is a decision that should be handle by @4mmonium and the team (on docs).

jaymo1011:
Bottom line, using Imgur the way it has been used by resource developers (and therefore, forced upon server owners) is not in the spirit of the service, and almost certainly against Imgur ToS. The fact of the matter is that Imgur has previously turned a blind eye to this misuse, as it does for most other non-commercial uses, but now does not. The decision was made by Imgur to block this misuse of their service and that's absolutely expected.

Perhaps this would have been considered in the past when the project was smaller and was not considered *legitimate* in most business to business relations but times have changed. This is not something to be subverted, this is just the current situation. 

A much better way to do fix the issue itself is to suggest the use of different services for the time being and to work on documentation which makes it clear that all images used by NUI should be included in the client resource package or hosted by the FXServer instance, not internet image hosts. This is the standard for embedded web apps anyways.

edwardexo:
I can't agree that no immediate action/guidance is needed, with such a large part of the playerbase is being affected.

Legally, about Cfx.re/Rockstar 'doing the right thing' - it's not Cfx.re who is breaking the ToS (by setting the user agent). It's the uploader's responsibility to make sure his picture is not breaking the ToS. The same way Google isn't responsible if a picture is uploaded from Chrome's user agent.

Is it the _right thing_ to do though.. Ideally not. 

Although, at the same time, all legitimate requests are also being blocked (cases where Imgur isn't being abused as a CDN). Imgur is actually the one 'not appropriate' here, by not notifying Cfx.re that they're making a breaking change like this (which they easily could have and should have). This would have would have allowed Cfx.re to notify server owners in advance.

Taking this into account, I would say Cfx.re allowing this (at least temporarily) would be justified (both legally and morally).

> Like I said, multiples websites give this service for free every month that people could start using!

> A much better way to do fix the issue itself is to suggest the use of different services for the time being

If any considerable amount of users move to a similar service, surely that will get blocked even quicker. As Imgur's one of the largest ones (having the most capacity to handle something like this). Making using a different service one of the worst options, as you'll soon likely have to deal with the issue again (and lose your content.. _again_).

> All images used by NUI should be included in the client resource package or hosted by the FXServer instance

Generally, most resources already include NUI images in the client resource package. That isn't the issue here.

Issue is in the cases where a regular player needs to upload content (not the server owner/resource developer!).
If there were any good available alternatives, surely developers would have used them.

> or hosted by the FXServer instance, not internet image hosts.

That would be the solution for in-game pictures/videos that phones generate (screenshot-basic). 

However, servers would need to set up endpoints in their FXServer needing to handle file types, sizes, comperssion, security, API keys, disk size etc. Doable by a regular FiveM developer? Probably not. Maybe an official resource or documentation by Cfx.re like @PsychoShock suggested could be a solution here.

Does the FiveM default gamehost (ZAP) even support potentially hosting 300GB of user uploaded content on their VPS?
(videos, images could take a lot of space in big servers over years). 

**However..**

That still doesn't solve the issue of users needing to upload content. AFAIK, you can't open the File Explorer in NUI so the player could choose a file to upload. Meaning that for a full solution, you'd need to set up **your own image uploading website**, handling security, illegal uploads etc. Definitely not doable by 99.999% of the developers, making the issue as important as it is.

There isn't an easy alternative. A chime in from Cfx.re would be good here - allowing Imgur/setting up documentation on how to set up your own image website/maybe providing an image upload themselves?

nihonium-cfx:
Hey there,

Understand the desire to fix the core issue here (content not loading), but we will not workaround such restrictions imposed by other companies in terms of use of their CDN anymore.

For now, I'd suggest moving your static content somewhere it is allowed/accepted, e.g. your own web hosting or CDN provider.

PR #2423, opened 2024-03-14
fix(scripting-mono-v2): allow scheduling from non-main threads

thorium-cfx:
### Goal of this PR
Enable scheduling tick calls from non-main threads, this is necessary to call back into the runtime (on the main thread) after external libraries used async/await with `Task/Task<T>` and we had nothing else scheduled. This didn't work before.

By removing its dependency on the Bookmark system (not thread-safe) and replacing it with a simplified version that achieves the same goal, but thread-safe.

ScRT: C#

### Successfully tested on
Server and client

**Game builds:** 2802

**Platforms:** Windows, Linux

### Checklist

### Fixes issues
fixes thorium-cfx/mono_v2_get_started#23

### DIFF
--- a/code/client/clrcore-v2/BaseScript.cs
+++ b/code/client/clrcore-v2/BaseScript.cs
@@ -278,7 +278,6 @@ namespace CitizenFX.Core
 #else
 			if (inputMapper != null && inputParameter != null)
 			{
-				Debug.WriteLine(command);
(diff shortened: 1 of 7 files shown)


jakub-cfx (approved):
Looks good to me 😊 Very nice and clean approach for a change that delivers such a big improvement. With the code already tested I think we can go ahead and merge.

PR #2426, opened 2024-03-18
feat(state): add NET_TO_* and *_TO_NET helpers

AvarianKnight:
### Goal of this PR

Add the same helpers that the client has to convert to/from network ids instead if having to use the extremely verbose `NetworkGetEntityFromNetworkId` and `NetworkGetNetworkIdFromEntity`

Adds the same helpers from the client

Server

### Successfully tested on

**Platforms:** Windows

### Checklist

### DIFF
--- a/code/components/citizen-server-impl/src/state/ServerGameState_Scripting.cpp
+++ b/code/components/citizen-server-impl/src/state/ServerGameState_Scripting.cpp
@@ -95,7 +95,7 @@ static void Init()
 		context.SetResult(true);
 	});
 
-	fx::ScriptEngine::RegisterNativeHandler("NETWORK_GET_ENTITY_FROM_NETWORK_ID", [](fx::ScriptContext& context)
+	auto getEntityFromNetworkId = [](fx::ScriptContext& context)
 	{
 		// get the current resource manager
 		auto resourceManager = fx::ResourceManager::GetCurrent();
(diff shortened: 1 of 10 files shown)


FabianTerhorst:
Hi, Thanks for the PR. 
Can you add a check for the entities `NetObjEntityType` to match the expected entity type of the `NET_TO_*` methods? 
So that e.g. a vehicle is not returned by `NET_TO_PED`.

AvarianKnight:
I don't understand why this is really necessary, this would also not match the same behavior on the client.

Disquse:
The server-side API must be as consistent as possible with the client-side API. However, it would be helpful to add some warnings if the returning entity is not of the expected type.

FabianTerhorst:
> The server-side API must be as consistent as possible with the client-side API. However, it would be helpful to add some warnings if the returning entity is not of the expected type (on client-side too, but it's not the part of this PR obviously).

What do you think about adding the typecheck to the client native? Or would that break to much?

Disquse:
> Or would that break to much?

It will break way too much, yes. Not an option in our case, unfortunately. Warnings at least might be useful for developers without breaking anything, but still hard to predict the outcome.

FabianTerhorst:
> > Or would that break to much?
> 
> It will break way too much, yes. Not an option in our case, unfortunately. Warnings at least might be useful for developers without breaking anything, but still hard to predict the outcome.

Thanks, so a warning on both the server and the client sounds like the best way for now.

gottfriedleibniz:
Is the desire to have @AvarianKnight update the server implementation to sanitize based on entity type or match *current* client behaviour while we make a work item about improving feedback? There seems to be some confusion here.

At the moment we already do provide [some](https://github.com/citizenfx/fivem/blob/5e71f91d7bd846d8eb059c6a7a0db79f16e9d601/code/components/extra-natives-five/src/EntityDeletionNatives.cpp#L121) [feedback](https://github.com/citizenfx/fivem/blob/5e71f91d7bd846d8eb059c6a7a0db79f16e9d601/code/components/extra-natives-five/src/EntityDeletionNatives.cpp#L99-L107) about non-existent net-objects and game entities which is currently not reflected on server side.

AvarianKnight:
> At the moment we already do provide [some](https://github.com/citizenfx/fivem/blob/5e71f91d7bd846d8eb059c6a7a0db79f16e9d601/code/components/extra-natives-five/src/EntityDeletionNatives.cpp#L121) [feedback](https://github.com/citizenfx/fivem/blob/5e71f91d7bd846d8eb059c6a7a0db79f16e9d601/code/components/extra-natives-five/src/EntityDeletionNatives.cpp#L99-L107) about non-existent net-objects and game entities which is currently not reflected on server side.

I think these types of warning would be weird to have on the server, any entity call (i.e. `ENT_TO_NET`) will go through makeEntityFunction, and if this is called with an invalid entity it will cause an exception that gets bubbled into the current runtime.

`NET_TO_ENT` will return 0 if the entity doesn't exist, this is currently an implementation detail but with the added docs it becomes "defined behavior", which is how a lot of stuff returns that "hey, this entity is invalid" and, imo, is better then being spammed with warnings that the entity doesn't exist.

PR #2427, opened 2024-03-18
fix(scripting/lua): Fix not closing to-be-closed variables on error

kajkul:
### Goal of this PR

Call `__close` metamethod on to-be-closed variables ([added in Lua 5.4](https://www.lua.org/manual/5.4/manual.html#3.3.8)) when an uncaught error occurs. Before this change variables would be closed only if the function/event/script execution ended without an error.

This PR calls the `lua_resetthread` method, when handling an error case, which unwinds the stack and closes all to-be-closed variables present there.

ScRT: Lua

### Successfully tested on

**Platforms:** Windows

### Checklist

### DIFF
--- a/code/components/citizen-scripting-lua/src/LuaScriptRuntime.cpp
+++ b/code/components/citizen-scripting-lua/src/LuaScriptRuntime.cpp
@@ -994,15 +994,39 @@ bool LuaScriptRuntime::RunBookmark(uint64_t bookmark)
 			}                        
 
 			static auto formatStackTrace = fx::ScriptEngine::GetNativeHandler(HashString("FORMAT_STACK_TRACE"));
-			auto stack = FxNativeInvoke::Invoke<const char*>(formatStackTrace, nullptr, 0);
 			std::string stackData = "(nil stack trace)";
-			
-			if (stack)
+
+			try
 			{
-				stackData = stack;
+				auto stack = FxNativeInvoke::Invoke<const char*>(formatStackTrace, nullptr, 0);
+				if (stack)
+				{
+					stackData = stack;
+				}
 			}
+			catch (...)
+			{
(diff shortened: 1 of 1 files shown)


kajkul:
Don't know why the checks failed. It doesn't look like anything related to with this PR.

gottfriedleibniz:
LGTM.

gottfriedleibniz:
Amended previous message, just a few concerns that should probably be addressed.

kajkul:
> Actually. Would it be possible to call `lua_resetthread` prior to invoking `FORMAT_STACK_TRACE`?

I guess it might get the wrong stack trace if an error occurs during closing a variable.
> That execution path is not exception-free. ... In fact, it would probably be best if we wrapped that `FxNativeInvoke::Invoke` invocation in a try/catch

Just did that.

code/components/citizen-scripting-lua/src/LuaScriptRuntime.cpp
```diff
@@ -994,15 +994,39 @@ bool LuaScriptRuntime::RunBookmark(uint64_t bookmark)
...
+			if (lua_type(thread, -1) == LUA_TSTRING)
+			{
+				resetErr = lua_tostring(thread, -1);
+			}
+
+			// We can't just check whether the resetStatus is OK because
+			// if there was no error lua_resetthread returns the same status it received
+			if (resetStatus != resumeValue || resetErr != err)
+			{
+				ScriptTrace("^1Error while closing to-be-closed varaibles: %s^7\n", resetErr);
```
AvarianKnight, line 1027: small typo here `varaibles` -> `variables`

kajkul: thanks fixed

PR #2428, opened 2024-03-18
tweak(minidump): don't report crashes if we're in debug mode

AvarianKnight:
### Goal of this PR
Ignore mindumps when in debug mode so I don't upload useless dumps

Disables HTTP request when in debug mode

Server

### Successfully tested on
**Platforms:** Windows

### Checklist


thorium-cfx:
We should look into disabling this for all local/custom builds, including debug. Not sure if we have any directives for this.

PR #2429, opened 2024-03-18
feat(fivem/state): add ConVar & Native to disallow server created entities being deleted by the client

AvarianKnight:
### Goal of this PR
Disallow server-created entities from being deleted by clients when `sv_disallowClientDelete` is set to true.

Adds a new check to `ProcessCloneRemove` to block the deletion of the server spawned created and recreate the entity for the calling client.

The way this is done isn't perfect (the entity will flicker for the person requesting the delete) but this shouldn't have any possible regressions (hopefully)

Server

### Successfully tested on
**Platforms:** Windows

### Checklist

### DIFF
--- a/code/components/citizen-server-impl/include/state/ServerGameState.h
+++ b/code/components/citizen-server-impl/include/state/ServerGameState.h
@@ -833,6 +833,8 @@ struct SyncEntityState
 	bool passedFilter = false;
 	bool wantsReassign = false;
 	bool firstOwnerDropped = false;
+	bool serverCreated = false;
+	bool rejectClientDeletion = false;
 
 	std::list<std::function<void(const fx::ClientSharedPtr& ptr)>> onCreationRPC;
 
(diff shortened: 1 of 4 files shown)


william-des:
It's a great idea to provide a way to prevent the deletion of server-created entities. However, I think it would be beneficial to add native functions to allow or disallow the deletion of entities regardless of the convar value (similar to [SetEntityIgnoreRequestControlFilter](SetEntityIgnoreRequestControlFilter))  A convar is not very flexible; for some servers, it could involve modifying a considerable number of scripts, which can be particularly complicated with the increasing use of escrowed resources.

AvarianKnight:
> It's a great idea to provide a way to prevent the deletion of server-created entities. However, I think it would be beneficial to add native functions to allow or disallow the deletion of entities regardless of the convar value (similar to [SetEntityIgnoreRequestControlFilter](SetEntityIgnoreRequestControlFilter)) A convar is not very flexible; for some servers, it could involve modifying a considerable number of scripts, which can be particularly complicated with the increasing use of escrowed resources.

This is a good idea, I will implement it here in a bit.

AvarianKnight:
This now adds `SET_ENTITY_REJECTS_CLIENT_DELETION` and adds another flag `rejectClientDeletion` that will default to the value of `sv_disallowClientDelete`

AvarianKnight:
There were some possible issues with this PR that could lead to some weird/bad regressions.

PR #2430, opened 2024-03-20
feat(redm/script): Add fxmanifest metadata for overriding train track and trolley cable files

### LINKED ISSUE #2424: [RedM] Cannot override base-game railway data
### What happened?

Thank you for your time.

### Expected result

### Reproduction steps

None.

### Importancy

Slight inconvenience

### Area(s)

RedM, Natives, ScRT: Lua

### Specific version(s)

RedM 7619

### Additional information

Sage-of-Mirrors:
### Goal of this PR

Allow a resource developer to replace the default `traintracks.xml` and `trolleyCableTracks.xml` files, thus allowing for RedM servers to have proper custom train tracks. And trolley cables, if so desired.

The fxmanifest metadata statements `replace_traintrack_file <path>` and `replace_trolley_cable_file <path>` have been added. When present, they store their paths to global variables in a patch, located at `gta-core-rdr3/src/PatchTrainTrackFileOverride.cpp`. The patch intercepts any calls to `LoadTrackXML()`, and if either of the override paths are valid, it will pass *them* into the track loading native rather than the default filepaths that were passed in.

RedM, Server, ScRT: Lua

### Successfully tested on

**Game builds:** b1311

**Platforms:** Windows

### Checklist

### Fixes issues

resolves #2424


Sage-of-Mirrors:
This resource can be used on a server to test overriding `traintracks.xml`: [tracktest.zip](https://github.com/citizenfx/fivem/files/14662336/tracktest.zip). It adds rough track nodes corresponding to the Great Plains and Mexico from RDR1. A resource that allows the player to spawn a train and manipulate junctions is required to see more than the `trace()` statements in the log.

PR #2439, opened 2024-03-27
fix(natives): Vector3 will generate a tuple type instead of an array …

AleksanderEvensen:
### Goal of this PR
The vector3 type has until now been generated as an array type. In my opinion it should be generated as a tuple type instead.
The reason i think the tuple type is a better option is because it can be used properly with the spread operator in functions that takes in the x, y and z component (ref picture below)

image

...

...

### Successfully tested on

**Game builds:** .. 

**Platforms:** Windows, Linux

### Checklist

### DIFF
--- a/ext/natives/codegen_out_dts.lua
+++ b/ext/natives/codegen_out_dts.lua
@@ -103,7 +103,7 @@ local function printReturnType(type)
 	elseif type.nativeType == 'float' then
 		return 'number'
 	elseif type.nativeType == 'vector3' then
-		return 'number[]'
+		return '[number, number, number]'
 	elseif type.nativeType == 'int' then
 		return 'number'
 	elseif type.nativeType == 'object' then


AvarianKnight:
I think this would be better, though it would generate type errors where there were none before, but also not sure what the opinions on "compatibility breaks" are for types.

I think this change is better for the better as it should always return 3 values

AleksanderEvensen:
I didn't event think about the compatibility breaking for the types 😅

FabianTerhorst:
I'm not sure if we can handle this change considering it causes to much compatibility issues.

AleksanderEvensen:
This could be done by a major version bump to the package (if that isn't off the table).

Using a tuple type in a function that takes an array as parameter doesn't yield an error in TypeScript.
The only instance where TypeScript yields an error is when an array is passed to a function where the parameter is a tuple type. I haven't seen any function in the type definitions for the natives where a function takes an array as input, so compatibility with other natives wouldn't break.

Error in user code wouldn't be a problem either, because if they're using a tuple type in their own functions they're also typecasting the array type anyway.

Would therefor argue that this wouldn't break so many things that I might have thought in the beginning.

Maybe some weird eslint rule would catch this case, but I didn't have the time to test this right now. If there is anything else that i might have missed, I would love to hear about it.

image

tsconfig used with example above:
```json
{
    "compilerOptions": {
      "strict": true,
      "allowUnusedLabels": false,
      "allowUnreachableCode": false,
      "exactOptionalPropertyTypes": true,
      "noFallthroughCasesInSwitch": true,
      "noImplicitOverride": true,
      "noImplicitReturns": true,
      "noPropertyAccessFromIndexSignature": true,
      "noUncheckedIndexedAccess": true,
      "noUnusedLocals": true,
      "noUnusedParameters": true,
  
      "isolatedModules": true,
  
      "checkJs": true,
  
      "esModuleInterop": true,
      "skipLibCheck": true
    },
    "$schema": "https://json.schemastore.org/tsconfig",
    "_version": "2.0.0"
  }
```

FabianTerhorst:
Thanks for the explanation. When the array typed function parameter accepts the tuple as well its not really a compatibility issue.

PR #2445, opened 2024-04-02
fix(statebag): check for replication state on SendAll

### LINKED ISSUE #2438: State bags set on the server do not respect replicated flag
### What happened?

As reported originally back in 2021 by @AvarianKnight here https://forum.cfx.re/t/server-side-entity-state-bags-sync-when-replicate-is-false/2313177 and seems it was never opened here.

I ran into this issue and did some research and found the original report.

I would argue that this is a security issue because currently when scripting developers are setting this value to be false there could be an expectation that its working correctly, therefore data stored with the intent to not replicate in fact is replicating which would be a data security issue.

### Expected result

It should respect the replicated flag being set to false and not replicate the state to the clients.

### Reproduction steps

1. Create an entity or player state server side and set the replicated flag to false.
2. Query the state client side and see that the data is replicated to other players when entering their scope.

There is also a reproduction script in @AvarianKnight's original report.

### Importancy

Security issue

### Area(s)

OneSync

### Specific version(s)

FiveM, Server 7752, Linux

### Additional information

_No response_

tens0rfl0w:
### Goal of this PR

Saving the replication state of a sbag key to allow checking if data should be replicated when the server sends all sbag data to the client after the client enters the scope of an entity/a player.

Storing the replication state of a sbag key in an unordered set on every set operation, and checking the saved state when the server wants to send all sbag data to the client.

I opted for the separate object approach to maintain simplicity. If you prefer storing the replication state directly within the data object, please let me know.

Server

### Successfully tested on

**Game builds:** 2699

**Platforms:** Windows

### Checklist

### Fixes issues
fixes #2438

### DIFF
--- a/code/components/citizen-resources-core/src/StateBagComponent.cpp
+++ b/code/components/citizen-resources-core/src/StateBagComponent.cpp
@@ -138,6 +138,7 @@ private:
 
 	std::shared_mutex m_dataMutex;
 	std::map<std::string, std::string, std::less<>> m_data;
+	std::unordered_set<std::string> m_replicatedKeys;
 
 	std::atomic<bool> m_replicationEnabled; //< #TODO: potentially remove this once the throttling system is in place
 	std::mutex m_replicateDataMutex;
@@ -236,6 +237,15 @@ void StateBagImpl::SetKeyInternal(int source, std::string_view key, std::string_
 	{
 		std::unique_lock _(m_dataMutex);
 
+		if (replicated)
+		{
+			m_replicatedKeys.insert(std::string(key));
+		}
+		else
+		{
+			m_replicatedKeys.erase(std::string(key));
+		}
+
 		if (auto it = m_data.find(key); it != m_data.end())
 		{
 			if (data != it->second)
@@ -308,7 +318,10 @@ void StateBagImpl::SendAll(int target)
 
 		for (auto& [ key, value ] : m_data)
 		{
-			SendKeyValue(target, key, value);
+			if (m_replicatedKeys.find(key) != m_replicatedKeys.end())
+			{
+				SendKeyValue(target, key, value);
+			}
 		}
 	}
 }


AvarianKnight:
Would it not be better performance-wise to only keep track of what we *dont* want to replicate? We're far more likely to want to replicate something over not.

tens0rfl0w:
While I agree on the assumption that an enabled replication state is much more likely, other than some memory overhead this shouldn't really have a performance impact as `std::string`'s hashing function should make any insert/delete/find operation on `std::unordered_set` take `O(1)` in this use case. (Hash collision seems unlikely here? Might be wrong..)

Can be changed ofc.

thorium-cfx:
I'm not 100% sure about the goal here, is this to privatize key/value pairs?

tens0rfl0w:
This PR intends to fix a current bug in the statebag implementation that causes non-replicated player and entity statebags to still get synced to clients when a client enters the scope of an entity/a player. See the mentioned issue for the actual issue reproduction.

thorium-cfx:
I understand, but in this case wouldn't you expect the old value – that was replicated – to be sent instead before we set another value that shouldn't be replicated? The current design or implementation seems flawed either way IMO.

So I'm trying to figure out if we still want to consider the `replicate` boolean here. You see there's currently a proposal going through the last stages that will allow you define if key/value pairs are synced at all (among more configurability), overruling this `replicate` state, hence why I asked if it's meant to privatize key/value pairs.

tens0rfl0w:
Thank you for clarifying.

I personally see switching the replication state on an existing key as something that will not happen that often; that's why I didn't really consider this when implementing. For me, disabling the replication of a key means 'ignoring' the sync-state, which implies we don't sync any data and 'forget' this was replicated to clients, which would mean no syncing of the previous (replicated) value.

So this isn't meant to privatize key/value pairs but rather to actually respect replication flags, so we don't overwrite non-replicated client keys or put additional load on the network thread when someone would expect key-data isn't synced.

jxckUK:
Sorry to jump in, but as I raised the GitHub issue thought I'd throw my comments in as well.

As @tens0rfl0w mentions and as I see this is to ensure that replication state is honoured as it was intended or at least documented [here](https://docs.fivem.net/docs/scripting-manual/networking/state-bags/#:~:text=This%20can%20be%20overridden%20on%20a%20case%2Dby%2Dcase%20basis%20by%20using%20set%3A).

The specific use case that caused me to open this (after finding AvarianKnights original repro) did not change the replication flag state on an existing key, instead a new key with replication set to false. I can say from my experience that I've not yet changed the replication state of an existing key once, so would agree that's uncommon.

Future proposals and extensions of the solution shouldn't necessarily change the need to fix the existing behaviour _I guess_?

thorium-cfx:
Interesting, wouldn't you instead want to write non-sync data to a non-synced key? And synced data to a key you do enable to sync? That with a lot less overhead than a `std::unordered_set` lookup? Mind you, `O(1)` being the complexity, not the clock cycles.

At least it's obvious that `replicate` means different things to different people and its function differs per the type of state bag. To me the global state bag works as intended (you keep the old value) but those rules/expectations are no longer applicable to range based state bags, I would then still expect to get the old replicated state, as it was requested to replicate. Though moving away from this design, to me, sounds like the solution.

tens0rfl0w:
Moving away from implementation details and looking at what "replication" state actually means, I have to disagree. Difficult to say what the "majority" of feature-users think the intended behavior should be, but having separate values for the same key depending on the replication state, as in:
```lua
Player(source).state:set("key", "value1", true)
```
and
```lua
Player(source).state:set("key", "value2", false)
```
not targeting the same key seems confusing for me. (If that's what you meant, sorry If I got you wrong here)

If would rather propose sending a delete operation to all routing targets and remove the key/value from the client statebag when replication is switched to off.

thorium-cfx:
Reading this, I wonder if we're thinking about the current set up and not the actual goal we're trying to achieve (see X/Y problem). What is the real goal we want to achieve? Make the replicate parameter do X or Y, or do we want to prevent synchronization of certain keys (maybe whole statebags) to keep data private/local and/or reduce network load and unnecessary storage?

Tell me why would I want to use:

```lua
Player(source).state:set("shared_key", "show everyone", true)
Player(source).state:set("shared_key", "don't sync this data", false)
``` 

and not just:
```lua
Player(source).state:set("shared_key", "show everyone", true)
Player(source).state:set("local_key", "don't sync this data", false)
``` 

@tens0rfl0w correct, though taking the definition, it doesn't define when, how, or for how long it will replicate, it could just be a single time, or two? for all I know. Also the docs don't define the behavior as well, merely "By default, state set from the server is replicated, and state set from clients is not replicated.", where "replicated" is not defined explicitly.

Though what I meant is not having 2 states for a single key, just that the previous value that we do wanted to replicate will be known by clients and still replicate (if server -> client), it'll effectively act the same as the global state bag, but with removal/addition when going out/in scope.

tens0rfl0w:
I think my confusion came from the fact that global state bags are also affected by this flawed implementation. If you're setting a global state bag key like:
```lua
GlobalState:set("key", "value", true)
GlobalState:set("key", "value2", false)
```
this works for already connected clients, but any client that connects after this set operation will still have "value2" synced as the key/value.

So what should be the change here?
We could store the last replicated value and still replicate it when a client requests the full sbag data, or we could nullify all synced clients and stop replication altogether. I'm unsure what the best approach is here.

PR #2463, opened 2024-04-16
feat(mono-rt2): add NuiEvent attribute

AvarianKnight:
- allows the client to use `[NuiEvent("NuiEventName")]` to receive Nui messages

### Goal of this PR
Add similar attributes as used by `EventHandler` for Nui

Adds new `NuiEventAttribute` and registers it via `RegisterNuiCallback`

ScRT: C#

### Successfully tested on
RedM

**Game builds:** N/A

**Platforms:** Windows, Linux

### Checklist

### DIFF
--- a/code/client/clrcore-v2/Attributes.cs
+++ b/code/client/clrcore-v2/Attributes.cs
@@ -116,6 +116,32 @@ namespace CitizenFX.Core
 			Binding = binding;
 		}
 	}
+	
+#if !IS_FXSERVER
+	/// <summary>
+	/// Register this method to listen for the given <see cref="CallbackName"/> when this <see cref="BaseScript"/> is loaded
+	/// if <see cref="IsRawCallback"/> is specified this will use a raw NUI callback instead
+	/// </summary>
+	/// <remarks>Only works on <see cref="BaseScript"/> inherited class methods</remarks>
+#else
+	/// <summary>Does nothing on server side</summary>
+	[EditorBrowsable(EditorBrowsableState.Never)]
+#endif
+	[AttributeUsage(AttributeTargets.Method, AllowMultiple = true)]
+	public class NuiCallbackAttribute : Attribute
+	{
+		public string CallbackName { get; }
+		public bool IsRawCallback { get; }
+		public NuiCallbackAttribute(string callbackName, bool isRawCallback = false)
+		{
+			CallbackName = callbackName;
+			if (isRawCallback)
+			{
+				throw new NotImplementedException("Raw Nui Callbacks are not currently implemented");
+			}
+			IsRawCallback = isRawCallback;
+		}
+	}
 
 	/// <summary>
 	/// Register this method to listen for the given <see cref="Export"/> when this <see cref="BaseScript"/> is loaded
--- a/code/client/clrcore-v2/BaseScript.cs
+++ b/code/client/clrcore-v2/BaseScript.cs
@@ -34,6 +34,8 @@ namespace CitizenFX.Core
 		}
 
 		private readonly List<KeyValuePair<int, DynFunc>> m_commands = new List<KeyValuePair<int, DynFunc>>();
+		
+		private readonly Dictionary<string, DynFunc> m_nuiCallbacks = new Dictionary<string, DynFunc>();
 
 #if REMOTE_FUNCTION_ENABLED
 		private readonly List<RemoteHandler> m_persistentFunctions = new List<RemoteHandler>();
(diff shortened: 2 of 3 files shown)


thorium-cfx:
Nice addition, thank you.

Thinking if we'd instead add this to the `EventHandler` as a `Binding` option, would require some extra changes, but will enable them to be registered during runtime (not startup) as well.

AvarianKnight:
> Nice addition, thank you.
> 
> Thinking if we'd instead add this to the `EventHandler` as a `Binding` option, would require some extra changes, but will enable them to be registered during runtime (not startup) as well.

Hmm, I'll be honest I don't know if this is entirely necessary since unlike event handlers users can already do this themselves.

```c#
Natives.RegisterNuiCallback("TestHandler", new Action<ExpandoObject, Callback>((ExpandoObject obj, Callback cb) => {

}));
```

thorium-cfx:
Like discussed privately, I'm looking at this from a user's perspective, one who doesn't know about the natives system for such functionality. In that case I would also expect methods like `BaseScript.AddNuiCallback(...)` to go with this feature, which is not needed when we combine it with the event handlers. Feel free to shoot holes in that idea.

It'll look like this:
```cs
[EventHandler(Binding.NUI)]
Coroutine Function() {}
```

@manups4e can we poke your mind, maybe ask the C# community as well?

manups4e:
I'll ask in my guild for sure :) i like the idea tho

AvarianKnight:
These changes have not been tested.

After some discussion some changes where made:

The attribute was renamed from `NuiEvent` to `NuiCallback` to avoid confusion that this is an event.

The `Register/Unregister` calls were changed to public on the `BaseScript` class to make it easier for end users to find.

The `Register/Unregister` will add/remove themselves whenever a BaseScript is enabled/disabled.

There was a field added to mark a `NuiCallback` as raw, this is currently unimplemented and will be added to a future PR

AvarianKnight:
This has been tested and works but has a minor bug.

If its called in a constructor it will be enabled even if `Initialize()` and `Enabled()` isn't called. 

I opted to copy the current behavior as fixing this is outside the scope of this PR.

code/client/clrcore-v2/Attributes.cs
```diff
@@ -116,6 +116,32 @@ public EventHandlerAttribute(string name, Binding binding = Binding.All)
...
+#else
+	/// <summary>Does nothing on server side</summary>
+	[EditorBrowsable(EditorBrowsableState.Never)]
+#endif
+	[AttributeUsage(AttributeTargets.Method, AllowMultiple = true)]
+	public class NuiCallbackAttribute : Attribute
+	{
+		public string CallbackEvent { get; }
+		public bool IsRawCallback { get; }
+		public NuiCallbackAttribute(string callbackEvent, bool isRawCallback = false)
```
thorium-cfx, line 135: `CallbackEvent` indicates that this is the event and/or callback. I'd suggest something like `Name` or what we generally call this part:
```
                                   vvvvvvvvvvv
https://${GetParentResourceName()}/getItemInfo
```

Could you also apply the same changes in the other places?

code/client/clrcore-v2/Native/Native.cs
```diff
@@ -15,6 +15,8 @@ internal static class CoreNatives
 		private static UIntPtr s_0x637f4c75;
 		private static UIntPtr s_0x8d50e33a;
 		private static UIntPtr s_0x91310870;
+		private static UIntPtr s_registerNuiCallback;
+		private static UIntPtr s_unregisterRawNuiCallback;
```
thorium-cfx, line 19: Could you add the hashes as comments behind them? Easier to find them :)

thorium-cfx (changes requested):
Overall looks good, nice job, just a naming and documentation concern. To be extra sure: last changes were tested right?

AvarianKnight:
> To be extra sure: last changes were tested right?

Yes the changes were tested as noted with a minor bug (noted here: https://github.com/citizenfx/fivem/issues/2476) the test code I was using is also listed there.

code/client/clrcore-v2/BaseScript.cs
```diff
@@ -283,7 +298,90 @@ internal void RegisterKeyMap(string command, string description, string inputMap
 			m_commands.Add(new KeyValuePair<int, DynFunc>(ReferenceFunctionManager.CreateCommand(command, dynFunc, false), dynFunc));
 #endif
 		}
+		
+		#endregion
+		
+		#region NUI Callback registration
+
+		internal void RegisterNuiCallback(string callbackEvent, DynFunc dynFunc)
```
thorium-cfx, line 306: Still got some `callbackEvent` names lingering ;) also down below, can you change these to the `callbackName` as well, including all the other occurrences?

thorium-cfx (changes requested):

thorium-cfx (approved):
Looks good, only needs a second pair of eyes and then we can merge it 👍

martonp96 (approved):
Looks good to me, thank you for your contribution. 👍

PR #2462, opened 2024-04-16
feat(mono-v2): Enable mono v2 on RedM

thorium-cfx:
### Goal of this PR
Enable mono v2 scripting runtime on RedM.

Mind you, it's still in beta, breaking changes and crashes are expected.

### How is this PR achieving the goal
By adding it to the components

### This PR applies to the following area(s)
RedM, ScRT: C#

### Successfully tested on
RedM client up to start up

**Game builds:** None

**Platforms:** Windows

### Checklist

- [x] Code compiles and has been tested successfully.
- [x] Code explains itself well and/or is documented.
- [x] My commit message explains what the changes do and what they are for.
- [x] No extra compilation warnings are added by these changes.

PR #2464, opened 2024-04-16
fix(mono-v2): disallow KeyMap to be used on RedM

AvarianKnight:
- keymap natives are not currently supported by RedM, using these would result in RedM crashing

### Goal of this PR

Block the usage of KeyMap on RedM until input mappings are made

Throw a not implemented exception the native is called

ScRT: C#

### Successfully tested on
RedM

**Game builds:** N/A

**Platforms:** Windows

### Checklist

### DIFF
--- a/code/client/clrcore-v2/Attributes.cs
+++ b/code/client/clrcore-v2/Attributes.cs
@@ -58,13 +58,14 @@ namespace CitizenFX.Core
 		}
 	}
 
-#if !IS_FXSERVER
+// TODO: revert this commit (blame check) when RedM has KeyMapping support, also don't make changes to this comment.
+#if GTA_FIVE
 	/// <summary>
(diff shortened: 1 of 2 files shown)


code/client/clrcore-v2/Attributes.cs
```diff
@@ -58,13 +58,13 @@ public CommandAttribute(string command, bool restricted = false)
 		}
 	}
 
-#if !IS_FXSERVER
```
thorium-cfx, line 61: Can you add a note above this one describing:
`TODO: revert this commit when RedM has KeyMapping support, also don't make changes to this comment.`

Thanks

thorium-cfx (approved):

thorium-cfx:
Can be merged together with #2462

PR #2479, opened 2024-04-23
Add NodeJS v20 script runtime

martonp96:
### Goal of this PR
The goal of this PR is to add a new NodeJS v20 script runtime, so developers can experiment and migrate their server scripts to a newer version, enabling them to have the up-to-date NodeJS experience.

By creating a new server side component, so old and new node runtimes can work independently side by side.
Most of the code is reused from the old runtime in a well organized manner.

Server, ScRT: Node

### Successfully tested on
Windows, Linux

**Platforms:** Windows, Linux

### Checklist


code/components/citizen-scripting-node/src/NodeScriptRuntime.cpp
```diff
@@ -0,0 +1,533 @@
...
+
+	m_context.Reset();
+	return FX_S_OK;
+}
+
+result_t NodeScriptRuntime::Tick()
+{
+	SharedPushEnvironment pushed(this);
+	
+	uv_run(m_uvLoop, UV_RUN_NOWAIT);
```
blattersturm, line 251: Dropping the EnvironmentScope changeset here and using a nested libuv loop here seems risky - this would lead to any libuv callbacks being delayed by the server main tick rate, which could slow down _any_ nested IO operations by a _lot_. I'm hoping that you folks have deliberated this and are _certain_ you wish to go this way.

martonp96: @blattersturm
Indeed, thanks for pointing it out. I added a workaround for this case, however i'm not too happy about this solution. In general an optimal scenario is when workarounds are not needed, neither in the component nor within the nodejs src. Hopefully in the future we can work on this.

@thorium-cfx 
Node has its own libuv and v8 dependencies built with the binary, in order to have it easier from the embedder side, we have to rely on those instead of external ones like before, thats why its there.

blattersturm:
A "high frequency tick method" is not the way to go here either. People at times run _hundreds_ of resources, and this runs counter to e.g. the tickless work done prior, even though this was meant to focus on _client_ performance.

Tight coupling to reduce ScRT enumeration overhead _still adds per-runtime cost_, makes a mess of the code, and still adds timer overhead since event signaling _is not timer-bound at all_. There really isn't any way to cheap out of something like the EnvironmentScope stuff, _unless_ upstream nowadays has something like this via e.g. async hooks. If maintaining this patch set is problematic, consider suggesting this use case to upstream Node.js maintainers instead, given it's a common embedder use case to wish to embed in an existing libuv event loop.

code/components/citizen-scripting-node/src/NodeScriptRuntime.cpp
blattersturm: A "high frequency tick method" dedicated to pumping every Node resource's uv loop (even worse if it _also runs the main scheduler tick!_) is not the way to go here either. People at times run _hundreds_ of resources, and this runs counter to e.g. the tickless work done prior, even though this was meant to focus on _client_ performance, the same would apply to server-side overhead.

Tight coupling to reduce ScRT enumeration/tick overhead _still adds per-runtime cost_, makes a mess of the code, and still adds timer overhead/delay since event signaling _is not timer-bound at all_. There really isn't any way to cheap out of something like the EnvironmentScope stuff, _unless_ upstream nowadays has something like this via e.g. async hooks. If maintaining this patch set is problematic, consider suggesting this use case to upstream Node.js maintainers instead, given it's a common embedder use case to wish to embed in an existing libuv event loop (and the uv dependency isn't the painful one to wish to sync with upstream- it's the v8 one).

Perhaps a cheap trick could be to somehow hoist events from the Node.js event loop into the server event loop and wrap callbacks _that way_ rather than patching Node.js to enter/exit environments cleanly on callbacks.

(moved here from a top-level comment)

martonp96:
> A "high frequency tick method" dedicated to pumping every Node resource's uv loop is not the way to go here either. People at times run _hundreds_ of resources, and this runs counter to e.g. the tickless work done prior, even though this was meant to focus on _client_ performance, the same would apply to server-side overhead.
> 
> Tight coupling to reduce ScRT enumeration/tick overhead _still adds per-runtime cost_, makes a mess of the code, and still adds timer overhead/delay since event signaling _is not timer-bound at all_. There really isn't any way to cheap out of something like the EnvironmentScope stuff, _unless_ upstream nowadays has something like this via e.g. async hooks. If maintaining this patch set is problematic, consider suggesting this use case to upstream Node.js maintainers instead, given it's a common embedder use case to wish to embed in an existing libuv event loop.
> 
> Perhaps a cheap trick could be to somehow hoist events from the Node.js event loop into the server event loop and wrap callbacks _that way_ rather than patching Node.js to enter/exit environments cleanly on callbacks.

For this approach, i'm entirely relying on past experience, where it just works fine with a high amount of resources, even though most gamemode logic was on server side. Since using this ScRT is optional, i would like to see if this punishes performance with real world examples while in the meantime we can work on alternative solutions like the ones you recommended.

code/components/citizen-scripting-node/src/NodeScriptRuntime.cpp
martonp96: For this approach, i'm entirely relying on past experience, where it just works fine with a high amount of resources, even though most gamemode logic was on server side. Since using this ScRT is optional, i would like to see if this punishes performance with real world examples while in the meantime we can work on alternative solutions like the ones you recommended.

SymbianMoe:
finally <3

Yum1x:
@martonp96 do you have plans to implement Node 22 as it'll also be LTS

martonp96:
> @martonp96 do you have plans to implement Node 22? it'll also be [LTS](https://nodejs.org/en/blog/announcements/v22-release-announce)

@Yum1x yes, its planned when it becomes LTS officially

DigitalWolf98:
Hi all, I know that the question will sound a bit silly, but I'll ask it nevertheless - Does anyone know when there will be an update with Node.js (a specific release date, for example, a month or at least a year) to understand exactly whether it is worth waiting for it in the near future ?
Everywhere I asked the question, answered something like "in the near future will be...", but this answer is not satisfactory because there are teams of developers who seriously want to open their own project, and even formalize the official documents to open a sole proprietorship, which entails financial costs even for such a case as this, which makes the business idle.

DaniGP17:
> Does anyone know when there will be an update with Node.js 

In the experimental section of the forum they said: "Currently we are planning to investigate issue with performance when writing to statebags and then looks like nothing prevent us from pushing this to Canary."

DigitalWolf98:
Well again, even you said the phrase "In the near future, something will happen there", so this near future has already lasted long enough, if you take ES6 and specifically the implementation in Node.js, it appeared in version 12.0.0, which was released on April 23, 2019. But in FiveM this "functionality" is cut out for some reason, leaving only the error throwing - TypeError: Invalid host defined options.

prikolium-cfx:
Closing this PR as it was moved to private git. For updates please follow the topic on forum in experiments section.

Whisperisgod:
NTA :( ♥

CONGGOR:
sx

PR #2486, opened 2024-04-25
feat(core/profiler): add natives exposing start/stop recording

AvarianKnight:
- currently trying to do recordings can only be done via console commands, which can be tedious
- this adds `PROFILER_START_RECORDING` and `PROFILER_STOP_RECORDING` to allow automating profiler captures

### Goal of this PR
Allow resources to start/stop recording to allow for easier automation of testing performance of code

Exposes `PROFILER_START_RECORDING` and `PROFILER_STOP_RECORDING`

Natives, SCRT

### Successfully tested on
C#

**Platforms:** Windows

### Checklist

### DIFF
--- a/code/components/citizen-scripting-core/src/Profiler.cpp
+++ b/code/components/citizen-scripting-core/src/Profiler.cpp
@@ -949,6 +949,20 @@ static InitFunction initFunction([]()
 		ctx.SetResult<int>(profiler->IsRecording());
 	});
 
+	fx::ScriptEngine::RegisterNativeHandler("PROFILER_START_RECORDING", [](fx::ScriptContext& ctx)
+	{
+		static auto profiler = fx::ResourceManager::GetCurrent()->GetComponent<fx::ProfilerComponent>();
+		const int frames = ctx.GetArgument<int>(0);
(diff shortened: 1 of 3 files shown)


AvarianKnight:
Here's some test code.

Note on the client you'll have to set `profiler_allowSaveNatives true`

```lua
CreateThread(function()
    local profilerStarted = false
    local profilerId = 0
    while true do
        if not profilerStarted then
            profilerStarted = true
            print("started")
            ProfilerStartRecording(250)
        elseif not ProfilerIsRecording() then
            print("save")
            profilerId = profilerId + 1
            ProfilerSaveToJson(("test_profiler_%s.json"):format(profilerId))
            ProfilerSaveToMsgpack(("test_profiler_%s.msgpack"):format(profilerId))
            profilerStarted = false
        end
        
        Wait(250)
    end
end)
```

AvarianKnight:
The implementation has swapped from taking a string value as the file name, to defining one based on the current time of day to not allow the functions to be used maliciously.

This also adds a `profiler:/` vfs to save profiler recordings to so they wont set in the base directory.

PR #2511, opened 2024-05-10
Reorganize KVP natives and fix Lua example for the "find" functions

Mathu-lmn:
### Goal of this PR

This PR puts all the KVP related native decls in the kvp folder
It also provides a fix for the Lua example script that wasn't working due to a wrong boolean check and a missing Wait

By moving all the KVP files in the KVP folder and fixing the Lua example in both StartFindKvp and StartFindExternalKvp

Natives

**Game builds:** X

**Platforms:** X

### Checklist


AvarianKnight (commented):
These don't need a `Wait`, doing so would just slow down getting the KVPs (especially server side, since it ticks once every 50ms).

Mathu-lmn:
Have you tried it ? I can't right now as I'm away but I can next week if needed !

AvarianKnight:
Yes, the example works fine, you don't need to wait to get the kvp string though.

Mathu-lmn:
I wonder why I added it then. I think the client crashed with no wait but I can't check now

AvarianKnight:
Likely crashed with the `until key` because there would never be a key so it would be an infinite loop

Mathu-lmn:
That's why I changed it to `until not key` 😬

Mathu-lmn:
But wouldn't the client crash in a loop with no wait ?

AvarianKnight:
> But wouldn't the client crash in a loop with no wait ?

The old loop (which was `until key`) *would* crash with a endless loop because it would wait for `key to be valid, which it would never be.

The new loop can't crash because if there isn't a key it will break out of the loop.

Mathu-lmn:
Yes, what I meant is that in regular while loops, if you don't put a Wait(0), the Fivem client freezes and you ultimately crash. I'm wondering if it's the case with the repeat operator too

Mathu-lmn:
I've tested without the Wait and it works fine, thanks for the feedback @AvarianKnight

FabianTerhorst (approved):
Thanks for your contribution.  👍

PR #2548, opened 2024-05-15
feat(extra-natives/rdr3): Add extra natives for train tracks

### LINKED ISSUE #2424: [RedM] Cannot override base-game railway data

Sage-of-Mirrors:
### Goal of this PR

This PR provides three new natives for RedM for working with train tracks:

* **GET_TRACK_COUNT:** Returns the number of tracks that are currently loaded in the world.
* **GET_TRACK_FROM_INDEX:** Returns the name hash of the track at the given index, or 0 if that index is invalid.
* **LOAD_TRACKS_FROM_FILE:** Disposes of the currently-loaded tracks and loads new ones from a given `traintracks.xml` file. This allows for fully customized rail networks.

The extra natives can be find in `extra-natives-rdr3/src/TrackNatives.cpp`.

`GET_TRACK_COUNT` and `GET_TRACK_FROM_INDEX` are just directly accessing the track data.

`LOAD_TRACKS_FROM_FILE` is based on FiveM's [LOAD_WATER_FROM_PATH](https://github.com/citizenfx/fivem/blob/2d73aeef4863b3ea3d6001ae9cf34cd063fd257c/code/components/extra-natives-five/src/WaterExtraNatives.cpp#L737).

While this approach appears to work without issue, I am not sure how vehicle replication works. If a connecting client doesn't run a resource containing a `LOAD_TRACKS_FROM_FILE` call *before* the server tells it to spawn a replicated train object, the client will crash. I ran a few tests on my local machine with multiple clients, and I did not get any crashes upon logging onto a local server. However, I currently cannot rule out the possibility that a higher ping or larger set of resources to load may delay the execution of a script long enough to cause a crash. More thorough testing may be required unless someone with more intimate knowledge of CFX's vehicle replication system can confirm that this will never be an issue. 

RedM, Natives

### Successfully tested on

**Game builds:** 1491.50

**Platforms:** Windows

### Checklist

### Fixes issues

Fixes #2424


Sage-of-Mirrors:
This resource can be used to test the three extra natives: [tracktest.zip](https://github.com/citizenfx/fivem/files/15327413/tracktest.zip).

Sage-of-Mirrors:
At the request of @Disquse, I did some testing to determine if a bad actor could exploit `LOAD_TRACKS_FROM_FILE` to intentionally cause clients to crash. Unfortunately, I found that this is possible. If a client spawns a train on tracks that they have locally, but other clients do not have, those clients without the tracks will crash.

What can I do to prevent this? There's a [patch for train config crashes](https://github.com/citizenfx/fivem/blob/2d73aeef4863b3ea3d6001ae9cf34cd063fd257c/code/components/gta-core-rdr3/src/PatchTrainConfigCrash.cpp), should I develop a similar patch for trains? Ensuring that the track index `CREATE_MISSION_TRAIN` gets is valid?

Sage-of-Mirrors:
So, an update: it turns out that the crashes I was seeing were not the result of my new natives - they were because the resource I was using to test the natives handles track switching. When I spawned trains via the resource, it was kicking off code that tried to reset junction states - on tracks that no longer existed!

This behavior is part of the original game. Question is, will I be required fix the original junction natives to get this PR approved?

Sage-of-Mirrors:
Closing this MR because I'm bad at git. Will open a new, clean one in a moment.

PR #2550, opened 2024-05-18
tweak(natives/rpc): Use client-native names and apply aliases

### LINKED ISSUE #2447: Error on SetPedHairColor - 0xBB43F090
### What happened?

When trying to set the ped hair colour via the server side I get (global SetPedHairColor) doesn't exist
image

### Expected result

execute the native SetPedHairColor

### Reproduction steps

When trying to apply SetPedHairColor on character load it crashes.

### Importancy

Crash

### Area(s)

FiveM

### Specific version(s)

FiveM/7827

### Additional information

_No response_

### LINKED ISSUE #2549: Server side RPC native `SetPedHairColor` doesn't exist
### What happened?

Updated to the latest version of the server 8187 and the native no longer exists.

When looking at the documention for the client sided which still has SetPedHairColor (Ref: 0xBB43F090) and clicking the link to the server side version it takes me to SetPedHairTint (Ref: 0x4CFFC65454C93A49) however when using this too, it gives the same error about not existing.

image

image

### Expected result

To be able to set the player hair colour from the server side without crashing.

### Reproduction steps

1. Download the latest version of the server or version 8187
2. Try setting the ped hair colour server side using `SetPedHairColor(index, colour, highlight)` or `SetPedHairTint(index, color, highlight)`

### Importancy

Crash

### Area(s)

FXServer, OneSync

### Specific version(s)

Server 8187 on both linux and windows

### Additional information

 Seems to be an issue with all artifacts from the last two or three months.

tens0rfl0w:
### Goal of this PR

Commit https://github.com/citizenfx/natives/commit/e6f473573f6eb7e2921b2ef48668117cbbe84bbb changed the native name from '_SET_PED_HAIR_COLOR' to 'SET_PED_HAIR_TINT' which makes the RPC native definition invalid.

We now also check for alias matches when generating RPC decl. in codegen and apply client-native aliases to RPC natives. This allows us to keep RPC native naming in sync with client-native changes while maintaining backward compatibility with old native naming (like on the client side).

FiveM, Natives (RPC)

### Successfully tested on

**Game builds:** 2699 

**Platforms:** Windows

### Checklist

### Fixes issues
fixes #2549
fixes #2447
fixes https://forum.cfx.re/t/5226975

### DIFF
--- a/ext/natives/codegen.lua
+++ b/ext/natives/codegen.lua
@@ -168,22 +168,26 @@ end
 
 local function getNative(nativeName)
 	-- get the native
-	local n
+	for _, v in ipairs(natives) do
+		if #v.apiset == 0 or v.apiset[1] == 'client' then
+			if v.name == nativeName then
+				return v
+			end
 
-	for k, v in ipairs(natives) do
-		if v.name == nativeName and (#v.apiset == 0 or v.apiset[1] == 'client') then
-			n = v
-			break
+			for _, aliasName in ipairs(v.aliases) do
+				if nativeName == aliasName then
+					return v
+				end
+			end
 		end
 	end
-
-	return n
+	return nil
 end
 
 local function getServerNative(nativeName)
 	-- get the native
 	local n
-	
+
 	for k, v in ipairs(natives) do
 		if v.name == nativeName and #v.apiset > 0 and v.apiset[1] == 'server' then
 			n = v
@@ -251,14 +255,14 @@ function rpcEnvironment.context_rpc(nativeName)
 			if not ctx then
 				if v.pointer then
 					ctx = { idx = k - 1, type = 'ObjDel' }
-					
+
 					table.insert(args, {
 						translate = true,
 						type = 'ObjDel'
 					})
 				else
 					ctx = { idx = k - 1, type = 'ObjRef' }
-					
+
 					table.insert(args, {
 						translate = true,
 						type = 'ObjRef'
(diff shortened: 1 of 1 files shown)


DaniGP17:
Why some functions names start with `_`?

AvarianKnight:
Names that start with underscore mean that the name isn't known.

freedy69:
maybe `CALL_MINIMAP_SCALEFORM_FUNCTION` can also be renamed to `BEGIN_SCALEFORM_MOVIE_METHOD_ON_MINIMAP`?

freedy69:
> > maybe `CALL_MINIMAP_SCALEFORM_FUNCTION` can also be renamed to `BEGIN_SCALEFORM_MOVIE_METHOD_ON_MINIMAP`?
> 
> 
> 
> Why should this native be renamed? This is a custom native, not a base-game native.
> 
> 
> 
> Also, PRs are *generally* not the best place to suggest other stuff as this PR is about RPC natives.

Im just saying because it would fit naming formatting of other scaleform natives. Would make it easier to read, im not really suggesting, more so pointing out

gottfriedleibniz:
To avoid breaking compatibility here, it *should* be safe (i.e., requires testing) to update [getNative](https://github.com/citizenfx/fivem/blob/ddd4810161d4285918d52772b6d2cfce12816282/ext/natives/codegen.lua#L177) to also search through aliases, e.g.,

```lua
local function getNative(nativeName)
	-- get the native
	for k, v in ipairs(natives) do
		if #v.apiset == 0 or v.apiset[1] == 'client' then
			if v.name == nativeName then
				return v
			end

			for _,aliasName in ipairs(v.aliases) do
				if nativeName == aliasName then
					return v
				end
			end
		end
	end
	return nil
end
```

Updating this script to use a reverse name lookup would probably be better. That can be done later though.

ext/natives/codegen.lua
```diff
@@ -281,67 +285,70 @@ function rpcEnvironment.context_rpc(nativeName)
...
-		codeEnvironment.returns('void')
-		curNative.rpc = true
-		curNative.ogHash = n.hash
+	removeServerNative(n.name)
+	codeEnvironment.native(n.name)
+	codeEnvironment.arguments(n.arguments)
+	codeEnvironment.apiset('server')
+	codeEnvironment.returns('void')
+	for _, aliasName in ipairs(n.aliases) do
+		codeEnvironment.alias(aliasName)
```
gottfriedleibniz, line 294: Including aliases will also require `removeServerNative(aliasName)`, same applies to the other cases `codeEnvironment` below. Otherwise the server files will contain duplicate definitions and MonoV2 won't compile.

gottfriedleibniz: Can you double check after clearing all intermediate files, e.g., `rm -rf ext/natives/inp ext/natives/out`? MonoV2's compilation will give me, for example,

```cs
public static void SetPedHairColor(int ped, int colorID, int highlightColorID)
{
    NativesImpl.void_0xbb43f090(N64.Val(ped), N64.Val(colorID), N64.Val(highlightColorID));
}

// ...

public static void SetPedHairColor(int ped, int colorID, int highlightColorID) => SetPedHairTint(ped, colorID, highlightColorID);
```

gottfriedleibniz: Ah. Apparently `native-decls/native_md` not being cleaned properly - leaving old definitions around (on both computers I tested).

PR #2556, opened 2024-05-21
tweak(gamestate/server): allow blocking of ScriptEntityStateChangeEvent

### LINKED ISSUE #2553: Exploit using "SetVehicleExclusiveDriver_2"
### What happened?

**Their are cheaters currently executing the code below**

```lua
local vehicles = GetGamePool('CVehicle')

for i = 1, #vehicles do
    SetVehicleExclusiveDriver_2(vehicles[i], PlayerPedId(), 1)
end
```

**and this is making it where if they are in a car near them it kicks them out, and if they are trying to get in they are NOT able to get back into the driver seat.**

### Expected result

Work the same but should own the vehicle before setting it.

### Reproduction steps

1. Load up into any server
2. Go to an area populated with vehicles
3. Run the provided snippet above
4. You will be able to see no one able to get into driver seat

### Importancy

Slight inconvenience

### Area(s)

FiveM, OneSync, Natives, ScRT: Lua

### Specific version(s)

FiveM Artifacts 8151, Server Game Build 2802, windows

### Additional information

_No response_

tens0rfl0w:
### Goal of this PR
* This game event is sent when a client tries to change state values on a remotely owned entity.
* Allow to block this game event from being routed to target clients to prevent potential abusive uses.

Introducing a new ConVar called 'sv_enableNetworkedScriptEntityStates' that allows blocking of the 'SCRIPT_ENTITY_STATE_CHANGE_EVENT' game event (routing is enabled by default).

FiveM, RedM, Server

### Successfully tested on
**Game builds:** 2699

**Platforms:** Windows

### Checklist

### Fixes issues
fixes #2553

### DIFF
--- a/code/components/citizen-server-impl/src/state/ServerGameState.cpp
+++ b/code/components/citizen-server-impl/src/state/ServerGameState.cpp
@@ -95,6 +95,9 @@ static bool g_networkedSoundsEnabled;
 static std::shared_ptr<ConVar<bool>> g_networkedPhoneExplosionsEnabledVar;
 static bool g_networkedPhoneExplosionsEnabled;
 
+static std::shared_ptr<ConVar<bool>> g_networkedScriptEntityStatesEnabledVar;
+static bool g_networkedScriptEntityStatesEnabled;
+
 static std::shared_ptr<ConVar<int>> g_requestControlVar;
 static std::shared_ptr<ConVar<int>> g_requestControlSettleVar;
 
@@ -7125,6 +7128,14 @@ std::function<bool()> fx::ServerGameState::GetGameEventHandler(const fx::ClientS
 			return false;
 		};
 	}
+
+	if (eventType == SCRIPT_ENTITY_STATE_CHANGE_EVENT)
+	{
+		return []()
+		{
+			return g_networkedScriptEntityStatesEnabled;
+		};
+	}
 #endif
 
 #ifdef STATE_FIVE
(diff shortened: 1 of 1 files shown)


code/components/citizen-server-impl/src/state/ServerGameState.cpp
```diff
@@ -7074,6 +7074,28 @@ std::function<bool()> fx::ServerGameState::GetRequestControlEventHandler(const f
...
+	uint16_t stateType = 0;
+	rl::MessageBuffer msg;
+
+	if (ParseEvent(std::move(buffer), &msg))
+	{
+		entityId = msg.Read<uint16_t>(13);
+		stateType = msg.Read<uint16_t>(4);
+	}
+
+	return [this, client, entityId, stateType]()
```
FabianTerhorst, line 7089: Are lambda captures needed for this at all? Couldn't you just return a lambda always returning false when the `stateType` is 9 and otherwise return a lambda that is always returning true.

FabianTerhorst: Yes that's how I imagined it. Please only add captures if they are absolutely necessary and not for future usage.

FabianTerhorst (approved):

gottfriedleibniz:
Under no circumstances should we be blocking events outright when there is the potential for legitimate use cases.

Pedro-Lucas14:
@gottfriedleibniz Hello, is there a date for this fix to go into production, a lot of hacks are using this flaw?

gottfriedleibniz (approved):

PR #2577, opened 2024-06-06
fix(ros): update RGSC_GET_VERSION_INFO to pretend to be 2.3.2.5

ur3k-lw:
### Goal of this PR
I have encountered an issue where the client incorrectly thinks it is running version 2.1.6.5. Due to a new update that requires version 2.3.2.5, this discrepancy is causing compatibility problems and unexpected behavior in dependent systems.

To address this issue, updating the `RGSC_GET_VERSION_INFO` function to return version 2.3.2.5, regardless of the actual version. This change will stop the client from thinking it's on version 2.1.6.5 and ensure compatibility with the new update.

FiveM, ROS

**Platforms:** Windows

### Checklist

### Fixes issues
Didn't find any reports.

### DIFF
--- a/code/components/ros-patches-five/src/LegitimacyNui.cpp
+++ b/code/components/ros-patches-five/src/LegitimacyNui.cpp
@@ -553,7 +553,7 @@ function RGSC_GET_TITLE_ID()
 function RGSC_GET_VERSION_INFO()
 {
 	return JSON.stringify({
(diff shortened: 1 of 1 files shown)


nihonium-cfx:
Fixed in https://github.com/citizenfx/fivem/commit/38dd75dec7d49b74761fec7d1b6982a931c457b8.

PR #2603, opened 2024-06-26
feat(data): Increased PoolSizes `ROPE` 12 -> 32 & `ropeData` 8 -> 24

JajarGaming:
### Goal of this PR
When using ropes, we ran into some limitations and the poolSize might've been the cause. Also when using the `SetRopesCreateNetworkWorldState` native, we were receiving errors of the pool size being full. I couldn't find anything related to the `CNetworkRopeWorldStateData` pool and how it could be edited/increased, so thought it might be using the normal `ROPE` pool

A simple pool size increase in `gameconfig.xml`. RDR has a size of 48, GTA originally had a size of 12 and I bumped it up to 32.

FiveM

### Successfully tested on
N/A

**Game builds:** N/A

**Platforms:** Windows

### Checklist

Wasn't able to compile - didn't have the environment setup for it, but this change _shouldn't_ impact compilation anyways?

### DIFF
--- a/data/client/citizen/common/data/gameconfig.xml
+++ b/data/client/citizen/common/data/gameconfig.xml
@@ -3131,7 +3131,7 @@
 						</Item>
 						<Item>
 							<ResourceTypeName>ROPE</ResourceTypeName>
-							<ExpectedMaximum value="12"/>
+							<ExpectedMaximum value="32"/>
 						</Item>
 						<Item>
 							<ResourceTypeName>CAMERA</ResourceTypeName>
--- a/data/client/citizen/common/data/gameconfig.xml
+++ b/data/client/citizen/common/data/gameconfig.xml
@@ -686,6 +686,10 @@
 							<PoolName>scrGlobals</PoolName>
 							<PoolSize value="15080000"/>
 						</Item>
+						<Item>
+							<PoolName>ropeData</PoolName>
+							<PoolSize value="24"/>
+						</Item>
 					</Entries>
 				</PoolSizes>
 


Legacy-TacticalGamingInteractive:
<3

prikolium-cfx:
Maybe it also worth to extend Rope definitions pool to allow adding of new rope types with custom texture and width? You can do it in same PR

Legacy-TacticalGamingInteractive:
> Maybe it also worth to extend Rope definitions pool to allow adding of new rope types with custom texture and width? You can do it in same PR. Currently we are limited to 8 types, but I think nothing bad will happen if we extend it to 24

This would be awesome!

JajarGaming:
> Maybe it also worth to extend Rope definitions pool to allow adding of new rope types with custom texture and width? You can do it in same PR. Currently we are limited to 8 types, but I think nothing bad will happen if we extend it to 24

I'm not exactly sure where the definitions pool is located. It's my first time working with the fivem repo and C++.

prikolium-cfx:
> > Maybe it also worth to extend Rope definitions pool to allow adding of new rope types with custom texture and width? You can do it in same PR. Currently we are limited to 8 types, but I think nothing bad will happen if we extend it to 24
> 
> I'm not exactly sure where the definitions pool is located. It's my first time working with the fivem repo and C++.

This pool is not mentioned in gameconfig because it's using default value (8). You need to add 
```
<Item>
	<PoolName>ropeData</PoolName>
	<PoolSize value="24"/>
</Item>
```
To PoolSizes

JajarGaming:
Will do

prikolium-cfx (approved):

PR #2609, opened 2024-07-01
tweak(scripting/mono-v2): bump pilot date

FabianTerhorst:


DaniGP17:
Hello @prikolium-cfx, this pull request is going to be uploaded to the Beta or the Latest(Unstable) update channel?
I've one script in my server using the mono rt2 working right in the beta, in the latest is not working because the new msgpack (de)serializer apparently broke the export system so users in the Latest update channel get errors with that, but users with the beta have the script working well, today the pilot date has ended so I don't know if I've to remove the script until the exports system is fixed or if the beta update channel is going to work well like before.
Thank you

ook3D:
What a great idea, make it so it breaks on new years eve when no one is around to fix it.

FabianTerhorst:
> What a great idea, make it so it breaks on new years eve when no one is around to fix it.

The same time was chosen last year. It won't matter, when the time is increased before it happens or the time check is removed.

junredzik:
@prikolium-cfx  bump, also have a script that works very well but I am waiting for a pull request

prikolium-cfx:
> @prikolium-cfx bump, also have a script that works very well but I am waiting for a pull request

It's already merged in Canary like week ago

junredzik:
> > @prikolium-cfx bump, also have a script that works very well but I am waiting for a pull request
> 
> It's already merged in Canary like week ago

and there are already plans to move it to official?

PR #2618, opened 2024-07-07
fix(system-resources/chat): scroll to bottom and show chat on new message

minemalox:
### Goal of this PR

Currently, when using the included system-resource chat, it does not update the scroll position of the chat window, nor does it unhide the chat window when new chat messages are received.

With the change from Vue 2 to Vue 3, watchers were changed to be shallow, so they only react when a new value is assigned but won't trigger when changing a nested value. By setting the watcher to be a deep watcher again, the chat now scrolls automatically to the last message and shows up briefly when in 'when active' mode.

Server, system-resources: chat

### Successfully tested on

**Server build:** 8823 

**Game builds:** N/A

**Platforms:** Windows

### Checklist

### DIFF
--- a/ext/system-resources/resources/chat/html/App.ts
+++ b/ext/system-resources/resources/chat/html/App.ts
@@ -115,19 +115,22 @@ export default defineComponent({
     window.addEventListener("message", this.listener);
   },
   watch: {
-    messages() {
-      if (this.hideState !== ChatHideStates.AlwaysHide) {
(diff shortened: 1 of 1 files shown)


jakub-cfx:
Was able to repro the issue, as well as confirm that the proposed fix is working. Thanks for the contribution 😊

operationairstrike:
Exactly what I need

PR #2620, opened 2024-07-08
fix(game/five): fixes a crash caused by colliding objects without a drawhandler

slashkeyvalue:
### Goal of this PR

When colliding, some objects may hit a codepath that expects a crCreature to be created and that requires the entity to have a valid drawhandler, which invisible objects might not/won't have.

Checks for a valid crCreature before running some object collision logic.

FiveM

### Successfully tested on

**Game builds:** 2372, 3095.

**Platforms:** Windows.

### Checklist

### Fixes issues
Fixes uncle-oklahoma-king

### DIFF
--- a/code/components/gta-core-five/src/CrashFixes.DrawHandler.cpp
+++ b/code/components/gta-core-five/src/CrashFixes.DrawHandler.cpp
@@ -1,0 +1,45 @@
+#include <StdInc.h>
+
+#include <jitasm.h>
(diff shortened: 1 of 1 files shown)


Pedro-Lucas14:
The cheater is spawning a lot of props and knocking down all the players that some date to fix goes into production ? @prikolium-cfx

prikolium-cfx (approved):

PR #2624, opened 2024-07-08
feat(rdr3): Manifest statement for custom train tracks

### LINKED ISSUE #2424: [RedM] Cannot override base-game railway data

Sage-of-Mirrors:
### Goal of this PR

This PR adds the manifest statement `replace_traintrack_file`, which allows resource developers to override the default railway config file that is located at `common:/data/levels/rdr3/traintracks.xml`. This enables developers to create entirely custom railway networks without having to work around the base game's train tracks.

Since RDR2 was not initially designed to handle user-generated railroad data, this PR also includes several patches to improve stability.

The manifest statement is parsed from the `fxmanifest` file in the same location as the other statements, in `code/components/citizen-resources-gta/src/ResourcesTest.cpp`. Because this functionality is RedM specific, it has been placed in an `ifdef` guard checking for `IS_RDR3`. When a developer includes `replace_traintrack_file` in a resource, the path they specify is stored in the static variable `g_trainTrackOverridePath`.

A trampoline function has been inserted in front of calls to the track XML parser, which checks whether 1) the game is attempting to load `traintracks.xml`, and 2) whether `g_trainTrackOverridePath` has a non-empty value. If both of those checks succeed, the trampoline will call the XML parsing function with the overridden path; otherwise, the original path that was passed into the function is used instead. Note that the first check is necessary because the same function is also used to load the network of cables for the trolley in St. Denis.

The stability patches are:

* Track junctions pointing to invalid tracks will now be deleted after the tracks XML is loaded.
* The base-game native `CREATE_MISSION_TRAIN` will now fail to create a train if the client it is executed on has no tracks loaded.
* The base game only has dedicated space for 50 total track objects; the track loading loop will now exit once the pool reaches that maximum.

RedM, Lua

### Successfully tested on

**Game builds:** 1491

**Platforms:** Windows

### Checklist

### Fixes issues

Resolves #2424

### DIFF
--- a/code/components/citizen-resources-gta/src/ResourcesTest.cpp
+++ b/code/components/citizen-resources-gta/src/ResourcesTest.cpp
@@ -57,6 +57,11 @@ namespace streaming
 	void RemoveDataFileFromLoadList(const std::string& type, const std::string& path);
 
 	void SetNextLevelPath(const std::string& path);
+
+#if defined(IS_RDR3)
+	// RDR3 only: Set the path for a train track XML file to load instead of the default one.
+	void SetTrainTrackOverridePath(const std::string& path);
+#endif
 }
 #endif
 
@@ -187,6 +192,14 @@ static InitFunction initFunction([] ()
 			{
 				streaming::SetNextLevelPath(resourceRoot + meta.second);
 			}
+
+#if defined(IS_RDR3)
+			// RDR3 only: allow a creator to replace the default train tracks with custom ones.
+			for (auto& meta : metaData->GetEntries("replace_traintrack_file"))
+			{
+				streaming::SetTrainTrackOverridePath(resourceRoot + meta.second);
+			}
+#endif
 
 			if (!RangeLengthMatches(metaData->GetEntries("data_file"), metaData->GetEntries("data_file_extra")))
 			{
--- a/code/components/gta-core-rdr3/src/PatchTrainTrackCrashes.cpp
+++ b/code/components/gta-core-rdr3/src/PatchTrainTrackCrashes.cpp
@@ -1,0 +1,215 @@
+#include <StdInc.h>
+
+#include "Hooking.h"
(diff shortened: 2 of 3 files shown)


Sage-of-Mirrors:
This resource can be used to test the PR: [tracktest.zip](https://github.com/user-attachments/files/16132195/tracktest.zip).

### Testing Steps

1. Enable the resource on the server and log into it. **The message "Replacing common:/data/levels/rdr3/traintracks.xml with resources:/tracktest/data/traintracks.xml" should be written to the debug console during map load, and the game should not crash.**
2. Exit the game, and navigate to `traintracks.xml`, which is located at `<resources>/tracktest/data/`. Open `traintracks.xml` in a text editor, and remove the contents of the `<train_tracks>` XML element. Save the file, and log back into the game. **The game should not crash.**
3. Repeat Step 3, but instead of removing the contents of the `<train_tracks>` element, copy one of the `<train_track>` elements within it and paste it multiple times - enough to push the number of `<train_track>` elements within the `<train_tracks>` node over 50. Save the file and log back into the game. **The message "Track pool is full - no more tracks can be loaded." should be written to the debug console during map load, and the game should not crash.**

Sage-of-Mirrors:
@prikolium-cfx @martonp96 Pinging for visibility.

martonp96:
@Sage-of-Mirrors
Looks good, i'll test it then get back to you.

martonp96:
Tested, works fine. 
Thanks for the contribution.

Linertank:
Hi Sage, I have the message that replaces the traintrack but if I make sapwn a train the game crashes would you have an idea why?

PR #2625, opened 2024-07-08
feat(net): support dynamic server endpoints using handover

PichotM:
### Goal of this PR
Allow the server to pass a specific endpoint using deferrals to a specific player.

- Fetching handover json data and overriding the server's endpoints if an endpoint value exists in handover

FiveM & RedM

### Successfully tested on
Tested passing a custom endpoint

### Checklist

Documentation about handover https://docs-backend.fivem.net/docs/scripting-manual/nui-development/loading-screens/

```lua
-- Server script
AddEventHandler('playerConnecting', function(_, _, deferrals)
    local source = source

    deferrals.handover({
        endpoints = { "127.0.0.1" }
    })
end)
```

### DIFF
--- a/code/components/net/src/NetLibrary.cpp
+++ b/code/components/net/src/NetLibrary.cpp
@@ -1192,7 +1192,7 @@ concurrency::task<void> NetLibrary::ConnectToServer(const std::string& rootUrl)
 					m_connectionState = CS_IDLE;
 					return true;
 				}
-				
+
 				auto rawEndpoints = (node.find("endpoints") != node.end()) ? node["endpoints"] : json{};
 
 				auto continueAfterEndpoints = [=, capNode = node](const json& capEndpointsJson)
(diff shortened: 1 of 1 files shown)


prikolium-cfx:
That's good feature and I like it, but issues that I can see now is that player will need to restart his game twice. First time because of game build change, and second time because of pureLevel change. Am I right?

PichotM:
> That's good feature and I like it, but issues that I can see now is that player will need to restart his game twice. First time because of game build change, and second time because of pureLevel change. Am I right?

It will, except if you already are on the right build
I don't believe we can come up with a solution without changing the whole logic of the change-build stuff and doing it way later just like the second pure level check

PichotM:
PR updated to only support dynamic endpoints
(custom pure level has been dropped)

prikolium-cfx (approved):

PsychoShock:
[Doc](https://github.com/citizenfx/fivem-docs/pull/456) will also be updated with the newest PR goal.

racistnoob:
This might need looked into. Rarely some players just don't receive a pure mode switch pop-up and can load into a pure mode enabled server while their client isn't in pure mode.

Possibly reproducible by attempting to spam connect to a server and restarting fivem until you get connected without a pure mode switch pop-up.

PR #2634, opened 2024-07-11
fix(game/five): fixes a crash caused by invalid clips

slashkeyvalue:
### Goal of this PR

Invalid clips aren't properly checked for nullptrs when CTaskGeneralSweep updates, so a crash occurs.

Checks for a valid clip or bails from the task

FiveM

### Successfully tested on

**Game builds:** 2372, 3095.

**Platforms:** Windows.

### Checklist

### Fixes issues
Fixes `gee-single-sink`

### DIFF
--- a/code/components/gta-core-five/src/CrashFixes.TaskGeneralSweepClip.cpp
+++ b/code/components/gta-core-five/src/CrashFixes.TaskGeneralSweepClip.cpp
@@ -1,0 +1,34 @@
+#include <StdInc.h>
+
+#include <Hooking.h>
+
+static bool (*g_origCTaskGeneralSweep__Unk)(void* task, void* clip, float* unk1, float* unk2, float* unk3);
+static bool CTaskGeneralSweep__Unk(void* task, void* clip, float* unk1, float* unk2, float* unk3)
+{
+	if (!clip)
+		return false;
+
+	return g_origCTaskGeneralSweep__Unk(task, clip, unk1, unk2, unk3);
+}
+
+static HookFunction hookFunction([]
+{
+	auto patterns = hook::pattern("48 8B ? 48 8B CB 48 89 44 24 20 E8 ? ? ? ? 84").count_hint(5);
+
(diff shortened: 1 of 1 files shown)


Pedro-Lucas14:
The cheater is using this glitch 
 take down all the players that some date for fixing goes into production? @prikolium-cfx

code/components/gta-core-five/src/CrashFixes.TaskGeneralSweepClip.cpp
```diff
@@ -0,0 +1,34 @@
...
+{
+	if (!clip)
+		return false;
+
+	return g_origCTaskGeneralSweep__Unk(task, clip, unk1, unk2, unk3);
+}
+
+static HookFunction hookFunction([]
+{
+	auto patterns = hook::pattern("48 8B ? 48 8B CB 48 89 44 24 20 E8 ? ? ? ? 84").count_hint(5);
```
prikolium-cfx, line 16: We can't be sure that occurences will be always aligned as we need. Can we switch to just finding 3 needed targets with unique pattern then?
`48 8B D5 48 8B CB 48 89 44 24 ? E8 ? ? ? ? 84 C0 75`
`48 8B D5 48 8B CB 48 89 44 24 ? E8 ? ? ? ? 84 C0 74`
`48 8B D7 48 8B CB 48 89 44 24 ? E8 ? ? ? ? 84 C0`

prikolium-cfx: You don't need to do anything, I've fixed it in the merge process.

prikolium-cfx (approved):

code/components/gta-core-five/src/CrashFixes.TaskGeneralSweepClip.cpp
slashkeyvalue: Checked across multiple builds and wasn't sure if that could happen, will keep that in mind, thanks :)

PR #2645, opened 2024-07-16
tweak(data/rdr3): Increase netGameEvent, atDNetEventNode pool size and add them in poolEntriesTable

Identity-labs:
Increase netGameEvent, atDNetEventNode pool size and add them in poolEntriesTable

### Goal of this PR

Reduce amount of crash related to netGameEvent
Like it was done for FiveM before https://github.com/citizenfx/fivem/commit/273ad0546f7aa6e58e4320553ba6c6787a094f7d

Capture_decran_2024-07-11_015227

image

RedM

### Successfully tested on

**Game builds:** 1491

**Platforms:** Windows

### Checklist


prikolium-cfx:
I'm not sure that pool size is the root of problem. As you can see in both cases you have only one event taking 99% of space in pool. Can you better describe how often does it happen and if possible provide steps for reproduction.

Identity-labs:
This crash is only happening when we have 32 or more players in scope, when they are whistling, or looting composite.

How to reproduce ?
You can flood TaskWhistleAnim
When we are few in the scope, i can take 20/300 of the pool by just flooding whistle

This pool is also used by lot of networked natives like RequestNetworkControl, and lot of Tasks natives.

Increase this pool can enhance global usage on networked natives, and may be avoid or reduce crashes.
We have only 60 crashes of this type "Ran out of rage::netGameEvent pool space." since 4 years, with average of 200+ players.

prikolium-cfx (approved):
Well, if it's not frequent issue and just pool expanding can solve this, then ok.

PR #2649, opened 2024-07-18
fix(ui-build): Fix for `Cannot read properties of null (reading 'focus')`

MuDzYNN:
### Goal of this PR
When NUI is initializing in nui console you always see:
```
Uncaught TypeError: Cannot read properties of null (reading 'focus')
    at root.html:108:26
(anonymous) @ root.html:108
setTimeout (async)
frameEscapeFunctions.<computed> @ root.html:106
```
This error is regarding to line: `frame.contentWindow.focus();` and i'm just added question marks to avoid displaying this error.

As above I'm just added question marks, so when frame or frame.contentWindow is null nothing happen.

ui-build

### DIFF
--- a/ext/ui-build/data/root.html
+++ b/ext/ui-build/data/root.html
@@ -105,7 +105,7 @@ registerFrameFunction(function(msg, frameName, frameUrl)
 		frameEscapeFunctions[frameName] = () => {
 			setTimeout(() => {
 				if (document.activeElement == document.body) {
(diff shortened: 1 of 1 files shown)


FabianTerhorst:
It might be good to know why its null. Maybe its a symptom of another bug.

prikolium-cfx:
Rebased and merged.

PR #2504, opened 2024-05-07
fix(natives): Deleting docs for non-existent native

ModoSN:
NON-EXISTENT NATIVE 000000001e35dbba

### Goal of this PR
Reduce confusion

### How is this PR achieving the goal
Deleting non-existent native doc

### This PR applies to the following area(s)
Natives

### Successfully tested on
N/A

### Checklist

- [x] Code compiles and has been tested successfully.
- [x] Code explains itself well and/or is documented.
- [x] My commit message explains what the changes do and what they are for.
- [x] No extra compilation warnings are added by these changes.

### Fixes issues
fixes #2501

PR #2666, opened 2024-07-24
tweak(decrapped/native): remove files not supported

PsychoShock:
### Goal of this PR

Following natives that can't be removed from native decls for compatibility, it will now hide them from the list. This will prevent users to use natives not supported anymore.

Hide natives not supported anymore from the list.

Natives

### Checklist

### Fixes issues

This can close https://github.com/citizenfx/fivem/pull/2504


Mathu-lmn:
Hi, the feat(native/visibility): hide commit needs a description longer than 8 characters for the PR to be valid. Try to be more descriptive

PsychoShock:
Switching to draft since it might break codegen - will do some tests.

FabianTerhorst (approved):
Thanks for your contribution. Work fine for me.

AvarianKnight:
This would break code gen, the whole reason these haven't been removed is for ABI stability in C#

code/components/citizen-scripting-core/src/ScriptHost.cpp
```diff
@@ -340,7 +340,7 @@ result_t SAFE_BUFFERS TestScriptHost::InvokeNative(fxNativeContext & context)
 	}
 	else
 	{
-		trace("WARNING: NON-EXISTENT NATIVE %016llx\n", context.nativeIdentifier);
```
AvarianKnight, line 343: Another thing to note is this gets called if *any* native is called incorrectly, saying they're 'not supported' is, imo, less useful than saying the native you tried to call didn't exist

PR #2670, opened 2024-07-25
feat(server/impl): add onServerStop event

DaniGP17:
### Goal of this PR

Added an event to handle necessary actions when the server is stopped. Currently I need to use txadmin to know when the server is going to stop with the `txAdmin:events:serverShuttingDown` event.

Created an event that is executed when the quit is requested and when the quit is forced using CTRL+C.

Server

### Successfully tested on

**Platforms:** Windows

### Checklist


Yum1x:
<sup><sub>IMO</sup></sub>
this should be managed by another application (such as TxAdmin), as the way you proposed is not capable of handling when a crash occours.

DaniGP17:
> IMO this should be managed by another application (such as TxAdmin), as the way you proposed is not capable of handling when a crash occours.

Does txadmin handle when a crash occours?
I've crashed my server using txadmin and the event is not triggered, the same as in my way
```
[           resources] Stopping resource test
[    c-scripting-core] Creating script environments for test
[           resources] Started resource test
[         script:test] CRASHING SERVER

=================================================================
        Native Crash Reporting
=================================================================
Got a SIGSEGV while executing native code. This usually indicates
a fatal error in the mono runtime or one of the native libraries 
used by your application.
=================================================================

=================================================================
        Managed Stacktrace:
=================================================================
=================================================================
[ citizen-server-impl] network thread hitch warning: timer interval of 248 milliseconds
[ citizen-server-impl] sync thread hitch warning: timer interval of 252 milliseconds
[ citizen-server-impl]
```

Yum1x:
> > IMO this should be managed by another application (such as TxAdmin), as the way you proposed is not capable of handling when a crash occours.
> 
> Does txadmin handle when a crash occours? I've crashed my server using txadmin and the event `txAdmin:events:serverShuttingDown` is not triggered, the same as in my way
> 
> ```
> [           resources] Stopping resource test
> [    c-scripting-core] Creating script environments for test
> [           resources] Started resource test
> [         script:test] CRASHING SERVER
> 
> =================================================================
>         Native Crash Reporting
> =================================================================
> Got a SIGSEGV while executing native code. This usually indicates
> a fatal error in the mono runtime or one of the native libraries 
> used by your application.
> =================================================================
> 
> =================================================================
>         Managed Stacktrace:
> =================================================================
> =================================================================
> [ citizen-server-impl] network thread hitch warning: timer interval of 248 milliseconds
> [ citizen-server-impl] sync thread hitch warning: timer interval of 252 milliseconds
> [ citizen-server-impl]
> ```

That's I said .... u wont be able to hear `txAdmin:events:serverShuttingDown`<sup><sub>(or any other event like u proposed)</sup></sub> in FxServer when a crash occurs. U should write ur own solution to handle this case.

AvarianKnight:
> Added an event to handle necessary actions when the server is stopped. Currently I need to use txadmin to know when the server is going to stop with the txAdmin:events:serverShuttingDown event.

Did you actually test to see if this works? I'm not quite sure how FXServer handles shutting down but will the resources last long enough to even handle this event?

One might assume that someone would try to asynchronous tasks with this too (like saving players) which would also probably not work because the server could be shut down before the resource gets a chance to execute.

> That's I said .... u wont be able to hear txAdmin:events:serverShuttingDown(or any other event like u proposed) in FxServer when a crash occurs. U should write ur own solution to handle this case or ask this feature in TxAdmin repo.

This seems like a very weird take, the server *can't* trigger events if its crashing so *nothing* can save. Saving anything when the server is in an inconsistent state doesn't sound like something that should be supported by anything anyways.

DaniGP17:
> Did you actually test to see if this works? I'm not quite sure how FXServer handles shutting down but will the resources last long enough to even handle this event?

Yeah, in my case I'm saving some info of the player into DB, this is a bit random because I get different times, may be the delay that take the server to close the process.

PR #2678, opened 2024-07-27
fix(voip/mumble): Only clear voice target channels if mumble is conne…

niekschoemaker:
…cted

This is the same behavior as all other target natives

Really small change, seems to be an oversight, not sure if this causd

### Goal of this PR
This makes the behavior of all the voice target natives the same. This is the only native which calls UpdateVoiceTarget when not connected.

By not Updating the voice targets if mumble is not connected.

FiveM & RedM

### Successfully tested on

**Game builds:** 9073

**Platforms:** Windows, Linux

### Checklist

### Fixes issues
None


prikolium-cfx:
Thanks for your contribution. Can you please provide reproduction steps to verify that this change works?

PR #2679, opened 2024-07-28
fix(clcore): Adjust "==" and "!=" operators to fix null reference exceptions

### LINKED ISSUE #1312: "==" throwing null reference exception on Models which are null
As discovered by: https://canary.discord.com/channels/192358910387159041/553225456296525833/948031244967510067

https://github.com/citizenfx/fivem/blob/cbe56f78f86bebb68d7960a38c3cdc31c7d76790/code/client/clrcore/External/Model.cs#L405-L408 Will throw an exception if left is null since null.Equals() does not exist.

This would fix the issue:
```
public static bool operator ==(Model left, Model right)
{
  if (left == null)
    return right == null;

  return left.Equals(right);
}
```

Keyinator:
### Goal of this PR

Fix potential null reference exceptions in clcore.

Instead of using ``left.Equals(right)`` where left could be null, leading to a null reference exception, we check both left and right for being null before.

FiveM, RedM, Client, ScRT: C#

### Checklist

### Fixes issues

fixes #1312

### DIFF
--- a/code/client/clrcore/External/Game.cs
+++ b/code/client/clrcore/External/Game.cs
@@ -1025,7 +1025,7 @@ namespace CitizenFX.Core
 
 			public static bool operator ==(WeaponHudStats left, WeaponHudStats right)
 			{
-				return left.Equals(right);
+				return ReferenceEquals(left, null) ? ReferenceEquals(right, null) : left.Equals(right);
 			}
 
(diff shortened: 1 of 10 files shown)


Keyinator:
The errors from the action don't seem to be caused by my code if I am not mistaken

DaniGP17:
Yeah, for me is compiling well

PR #2682, opened 2024-07-29
feat(voip/mumble): Add MUMBLE_DOES_CHANNEL_EXIST native

### LINKED ISSUE #2677: Voice Targets not applied when channel doesn't exist
### What happened?

When a channel doesn't exist on a mumble server, the call to MUMBLE_ADD_VOICE_TARGET_CHANNEL is ignored (see [MumbleClient.cpp#L356](https://github.com/citizenfx/fivem/blob/e8e15540b6ffbea6b1f1c20c4b382d8f4da2c19e/code/components/voip-mumble/src/MumbleClient.cpp#L356)).

This causes people not to be able to hear eachother, since the target doesn't get added.

### Expected result

The target should be added or there should be a way to validate the result of the target update.

### Reproduction steps

1. Add a channel to a voice target.
2. Set channel of other person to the channel just added.
3. Observe that voice target doesn't work.

### Importancy

Slight inconvenience

### Area(s)

FXServer

### Specific version(s)

Server 9073 linux/windows

### Additional information

For some context, I internally cache which voice targets I have added, to avoid network overhead, this works fine, except for the case given above. I did think of some possible solutions, which I listed below. I'm willing to implement them myself, I'm just not sure which one would be the "most correct" one.

Possible solution:

1. Add a native to check if a channel exists, this way it can be assured that the channel exists.
2. Another "solution" would be to simply constantly Call MUMBLE_ADD_VOICE_TARGET_CHANNEL, this however would cause a lot of packets to be sent to the server for no reason, and would probably hit the "MAX_VOICE_TARGET_CHANNELS" limit on the mumble server which is 256, since there's no check if the target is already added anywhere as far a I could tell.
3. Don't clear m_pendingVoiceTargetUpdates, but remove each one only if they are actually handled. This however adds the possibility that a voice target is added after it is no longer needed, causing it to get stuck in the voice targets unless manually cleared.
4. Only add the voicetarget update to m_pendingVoiceTargetUpdates if the channel exists and return a boolean from the native, this would allow a plugin to validate if the target is actually going to be added.
~~5. Looking at all the code, the check if a channel actually exists doesn't actually seem necessary for correct behavior (see [client.cpp#L1011](https://github.com/citizenfx/fivem/blob/e8e15540b6ffbea6b1f1c20c4b382d8f4da2c19e/code/components/voip-server-mumble/src/client.cpp#L1011) so removing the check would also fix the issue.~~

I think option 1 or 4 would be the easiest to implement, but would require some testing to see if there's any implications to this.

niekschoemaker:
### Goal of this PR
To add the ability to check if channels exist, which is usefull to avoid issue #2677

Additionally changed the insert call for channels to emplace (same way as done in users) and cache the widened channel name before a for loop where this was not done yet.

It adds the native `MUMBLE_DOES_CHANNEL_EXIST` which allows scripts to check if a mumble channel exists before executing a native on said channel ensuring that the native actually has effect.

FiveM & RedM (voip/mumble)

### Successfully tested on

**Game builds:** 9073

**Platforms:** Windows

### Checklist

### Fixes issues
fixes #2677

### DIFF
--- a/ext/native-decls/MumbleDoesChannelExist.md
+++ b/ext/native-decls/MumbleDoesChannelExist.md
@@ -1,0 +1,14 @@
+---
+ns: CFX
+apiset: client
+---
+## MUMBLE_DOES_CHANNEL_EXIST
+
+```c
+void MUMBLE_DOES_CHANNEL_EXIST(int channel);
+```
+
+Check whether specified channel exists on the Mumble server.
+
+## Parameters
+* **channel**: A game voice channel ID.
--- a/ext/native-decls/MumbleDoesChannelExist.md
+++ b/ext/native-decls/MumbleDoesChannelExist.md
@@ -5,10 +5,13 @@ apiset: client
 ## MUMBLE_DOES_CHANNEL_EXIST
 
 ```c
-void MUMBLE_DOES_CHANNEL_EXIST(int channel);
+BOOL MUMBLE_DOES_CHANNEL_EXIST(int channel);
 ```
 
 Check whether specified channel exists on the Mumble server.
 
 ## Parameters
 * **channel**: A game voice channel ID.
+
+## Return value
+True if the specific channel exists. False otherwise.
--- a/code/components/gta-net-five/src/MumbleVoice.cpp
+++ b/code/components/gta-net-five/src/MumbleVoice.cpp
@@ -1151,6 +1151,20 @@ static HookFunction hookFunction([]()
 			}
 		});
 
+		fx::ScriptEngine::RegisterNativeHandler("MUMBLE_DOES_CHANNEL_EXIST", [](fx::ScriptContext& context) {
+			auto channel = context.GetArgument<int>(0);
+
+			if (g_mumble.connected)
+			{
+				auto channelName = fmt::sprintf("Game Channel %d", channel);
+				context.SetResult<bool>(g_mumbleClient->DoesChannelExist(channelName));
+			}
+			else
+			{
+				context.SetResult<bool>(false);
+			}
+		});
+
 		fx::ScriptEngine::RegisterNativeHandler("MUMBLE_ADD_VOICE_TARGET_PLAYER", [](fx::ScriptContext& context)
 		{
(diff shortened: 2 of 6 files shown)


niekschoemaker:
Force pushed to fix commit title from feature to feat

ext/native-decls/MumbleDoesChannelExist.md
```diff
@@ -0,0 +1,14 @@
+---
+ns: CFX
+apiset: client
+---
+## MUMBLE_DOES_CHANNEL_EXIST
+
+```c
+void MUMBLE_DOES_CHANNEL_EXIST(int channel);
```
FabianTerhorst, line 8: Please declare the return type correctly: `BOOL`.

ext/native-decls/MumbleDoesChannelExist.md
```diff
@@ -0,0 +1,14 @@
...
+```
+
+Check whether specified channel exists on the Mumble server.
+
+## Parameters
+* **channel**: A game voice channel ID.
```
FabianTerhorst, line 14: Add a return value description
```

## Return value
True if the specific channel exists. False otherwise.
```

FabianTerhorst (changes requested):

ext/native-decls/MumbleDoesChannelExist.md
PsychoShock: If nothing changed, natives never had description when name was obvious.

AvarianKnight: Even when they were unneeded `## Return value` is usually added for every native declaration that will return something.

ext/native-decls/MumbleDoesChannelExist.md
niekschoemaker: Oops, that was an oversight on my part, fixed it with latest commit

FabianTerhorst (approved):
Thanks for your contribution. 👍

PR #2686, opened 2024-08-01
feat(scripting/v8): check event existence

piitszk:
### Goal of this PR
Add the native DoesEventExist function to Lua and JavaScript in FiveM, allowing verification of whether an event created by RegisterNetEvent exists.

This PR modifies the scheduler file to add the DoesEventExist function, making it available in Lua and JavaScript. This allows developers to check for the existence of a specific event, improving event management in scripts.

- FiveM & RedM
- Lua & JS

### Checklist

### DIFF
--- a/data/shared/citizen/scripting/lua/scheduler.lua
+++ b/data/shared/citizen/scripting/lua/scheduler.lua
@@ -334,6 +334,10 @@ function RegisterNetEvent(eventName, cb)
 	end
 end
 
+function DoesEventExist(eventName)
+	return eventHandlers[eventName] or false
+end
+
 function TriggerEvent(eventName, ...)
 	local payload = msgpack_pack_args(...)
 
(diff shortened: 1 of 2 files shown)


AvarianKnight:
Can you give an example of how this improves event handling?

piitszk:
> Can you give an example of how this improves event handling?

In some very specific contexts related to server security, for example, I once wanted to manipulate the artifact (build) and create an event that only exists in that version to return something, like authentication, validation, or something similar. With DoesEventExist, I could basically check if the event exists; otherwise, it would mean that the build being used is not 'compatible'.

And of course, I could check this in other ways, such as using the TriggerEvent and waiting a few seconds to see if a callback occurs. However, let's be honest, that's neither efficient nor a clean approach to do this.

AvarianKnight:
Isn't this the whole point of versioning in your fxmanifest?

This would also not handle this correctly anyways as it would only work *for the current resource*.

PR #2695, opened 2024-08-05
feat(data): Add fwPortalSceneGraphNode pool to gameconfig

ook3D:
Currently by default, fwPortalSceneGraphNode is set to 400, servers with many interiors and in turn lots of portals, this can cause a game crash. its possible to simply add this pool to the config to increase it.

currently it says "Unknown Pool full - 400", which makes sense because fwPortalSceneGraphNode is missing from the config therefore its unknown.

### Goal of this PR

Adding the pool to the gameconfig and increasing its value to prevent game crashes with many interiors loaded

...

making the pool accessable to the gameconfig.
...

FiveM
...

### Successfully tested on

**Game builds:** .. 

**Platforms:** Windows, Linux

### Checklist

### DIFF
--- a/data/client/citizen/common/data/gameconfig.xml
+++ b/data/client/citizen/common/data/gameconfig.xml
@@ -293,6 +293,10 @@
 							<PoolSize value="900"/>
 						</Item>
 						<Item>
+							<PoolName>fwPortalSceneGraphNode</PoolName>
+							<PoolSize value="1500"/>
+						</Item>
+						<Item>
 							<PoolName>InteriorProxy</PoolName>
 							<PoolSize value="9060"/>
 						</Item>


FabianTerhorst (approved):
Thanks for your contribution 👍

prikolium-cfx:
As I can see it doesn't work, because IPL related items allocated from another precalculated Heap.

ook3D:
> As I can see it doesn't work, because IPL related items allocated from another precalculated Heap. If you still wish to fix it you better take a look on `fwSceneGraph::InitPools` method, then I hope you will understand what you need to do.

Hmm interesting, I'll take a look, thanks

DaniGP17:
Hi @ook3D  Have you started looking into that yet? Or can I try it myself?

PR #2717, opened 2024-08-16
tweak(core/sbag): delete state bags that get passed as nil

AvarianKnight:
### Goal of this PR
Delete state bags that have their value gets set to nil

Check for `0xc0` and if the first byte is it, remove it from the state bag map

Server

### Successfully tested on

**Game builds:** b3095

**Platforms:** Windows

### Checklist


code/components/citizen-resources-core/src/StateBagComponent.cpp
```diff
@@ -245,12 +245,16 @@ void StateBagImpl::SetKey(int source, std::string_view key, std::string_view dat
...
 
+// https://github.com/msgpack/msgpack/blob/master/spec.md#formats
+constexpr char MsgPackNil = static_cast<char>(0xc0);
 void StateBagImpl::SetKeyInternal(int source, std::string_view key, std::string_view data, bool replicated)
 {
 	{
 		std::unique_lock _(m_dataMutex);
-
-		if (auto it = m_data.find(key); it != m_data.end())
+		if (data[0] == MsgPackNil) {
```
FabianTerhorst, line 254: Bracket needs to be in the next line.
```
if (data[0] == MsgPackNil)
{
```

code/components/citizen-resources-core/src/StateBagComponent.cpp
```diff
@@ -245,12 +245,17 @@ void StateBagImpl::SetKey(int source, std::string_view key, std::string_view dat
...
 
+// https://github.com/msgpack/msgpack/blob/master/spec.md#formats
+constexpr char MsgPackNil = static_cast<char>(0xc0);
 void StateBagImpl::SetKeyInternal(int source, std::string_view key, std::string_view data, bool replicated)
 {
 	{
 		std::unique_lock _(m_dataMutex);
-
-		if (auto it = m_data.find(key); it != m_data.end())
+		if (data[0] == MsgPackNil)
```
FabianTerhorst, line 254: Just to be sure please add a check that data is not empty.

FabianTerhorst: Data won't be empty here.

FabianTerhorst (approved):
Thanks for the contribution. Tested and works well. 👍

PR #2728, opened 2024-08-19
tweak(conhost): use circular_buffer instead of ImVector

AvarianKnight:
### Goal of this PR
Fix crash when printing a large amount of things to console (2501+) due to `ItemTimes` using a `ImVector` instead of `circular_buffer`.

```lua
-- this expects you to have run `con_miniconChannels "script:*"` prior
CreateThread(function()
  for i = 1, 5000 do
    print("test", i)
  end
end)
```

Replace `ImVector` with a `circular_buffer` so that it properly caps out at `2500` items instead of allowing it to infinity expand as a Vector, causing a crash later when it tries to remove `Items` and `ItemKeys` because they are properly restricted to `2500` items and will try to remove past the bounds if `Items`.

https://github.com/citizenfx/fivem/blob/705cd7901ac03c5a0d1e1355d94e84046309b80d/code/components/conhost-v2/src/ConsoleHostGui.cpp#L729-L734

FiveM/RedM

### Successfully tested on

**Game builds:** N/A

**Platforms:** N/A

### Checklist

### DIFF
--- a/code/components/conhost-v2/src/ConsoleHostGui.cpp
+++ b/code/components/conhost-v2/src/ConsoleHostGui.cpp
@@ -684,7 +684,7 @@ auto msec()
 
 struct MiniConsole : CfxBigConsole
 {
-	ImVector<std::chrono::milliseconds> ItemTimes;
+	boost::circular_buffer<std::chrono::milliseconds> ItemTimes{ 2500 };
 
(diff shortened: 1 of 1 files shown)


FabianTerhorst (approved):
Thanks, works fine.

PR #2731, opened 2024-08-20
tweak(mono-rt2): swap back to using Prop instead of Object

AvarianKnight:
### Goal of this PR
Using `CitizenFX.Core.Object` alongside the regular `System.Object` leads to annoying cases where you have to disambiguate between if you want to use `CitizenFX.Core.Object` or `System.Object`

This goes back to the default that rt1 uses which is to have `Object` named `Prop` so you don't have to do any disambiguation with CLR

Removes object class for rt2 and just uses the original rt1 `Prop` class

FiveM, C#

### Successfully tested on
This hasn't been tested.

**Game builds:** .. 

**Platforms:** Windows

### Checklist


FabianTerhorst (approved):
Thanks for the contribution. Breaking changes in mono rt2 are fine 👍

PR #2753, opened 2024-08-26
fix(nui): hide mouse cursor when the css cursor property is set to none

kajkul:
### Goal of this PR
Hide the mouse cursor when it is over an element with css property `cursor: none`.

It checks in the `OnCursorChange` handler if the type is `CT_NONE` and changes `g_isCursorHidden` accordingly, which is added to the checks performed in `NUIRenderCallback.cpp` to determine whether the cursor should be displayed.

FiveM, RedM, NUI

### Successfully tested on

**Platforms:** Windows

### Checklist

### DIFF
--- a/code/components/nui-core/include/CefOverlay.h
+++ b/code/components/nui-core/include/CefOverlay.h
@@ -365,6 +365,8 @@ namespace nui
 	bool OVERLAY_DECL HasCursor();
 	bool OVERLAY_DECL HasFocus();
 	bool OVERLAY_DECL HasFocusKeepInput();
+	void OVERLAY_DECL SetCursorHidden(bool isHidden);
+	bool OVERLAY_DECL IsCursorHidden();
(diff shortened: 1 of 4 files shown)


kajkul:
Don't know why the check failed. It doesn't look like anything related to this PR.

Mathu-lmn:
> Don't know why the check failed. It doesn't look like anything related to this PR.

Yup, the cfx team is aware that the pipeline for linux server build is broken for now. A fix might come soon !

prikolium-cfx:
Is there any specific use case where it can be useful?

kajkul:
> Is there any specific use case where it can be useful?

I wanted to use it to hide the mouse while dragging an item in the player's inventory, because of my setup it would be tricky to correctly align the cursor and also track the actual drop position, so just hiding it would be the best option. It could be more generally useful when you want to have a custom mouse pointer in your ui. It is also the way that chromium handles this css property by default.

itzcodex24:
Is this something still being looked at? Our use case include using an image as a cursor overlay and hiding the default pointer.

PR #2756, opened 2024-08-28
fix(state/scripting): fixes player natives trying to reference non-existent player entity

AvarianKnight:
### Goal of this PR
Fix the server crashing if the player entity doesn't exist when the native is called

Using the proper `MakePlayerEntityFunction` wrapper that checks if the entity exist before calling the function

Server

### Successfully tested on
**Platforms:** Windows, Linux

### Checklist


Nobelium-cfx (approved):
LGTM

PR #2763, opened 2024-09-01
Building : Add known issue for C2039

Mathu-lmn:
### Goal of this PR

This PR adds a known issue and how to fix it when trying to build the FiveM client locally

Building docs

### Checklist

### DIFF
--- a/docs/building.md
+++ b/docs/building.md
@@ -81,3 +81,7 @@ If you don't know how to do that, here's how:
 - error CS0246: The type or namespace name 'Hash' could not be found (are you missing a using directive or an assembly reference?)
 
   `fxd gen` ran before `prebuild`. Close the solution, re-run `fxd gen` and then reopen & build the solution again.
+
+- error C2039: 'basic_string' is not a member of 'std'
+
+  Try to add an `#include <string>` at the beginning of the vendor\range-v3\include\range\v3\range\primitives.hpp - This could be explained by how your compiler's standard libraries


FabianTerhorst:
There should never be a workaround for building that involves editing vendor files.

DaniGP17:
I always get that error while trying to compile FiveM in Release instead of debug, so I don't know what's the good solution for this.

Mathu-lmn:
Maybe I should try opening a PR on the range repo then. But the activity looks stale for the last 6 months 😬

Mathu-lmn:
Tho I agree you should never modify a submodule

DaniGP17:
But exactly, what is the reason why this error has started to appear, I don't remember that in the past when compiling it in the release version I got this error

Mathu-lmn:
I'm not 100% sure, but it comes from your compiler that does not includes the string library by default... That's why it works for some people but not everyone.

Mathu-lmn:
> There should never be a workaround for building that involves editing vendor files.

I completely agree, but then what would your advice be ? Cloning the repo on the citizenfx org and making this change (while changing the submodule remote) ?

Nobelium-cfx:
> Cloning the repo on the citizenfx org and making this change (while changing the submodule remote) ?

I think the best way to go is to try making a PR in the vendor repository. See if they accept it. Worth trying at least.

If not - cloning seems to be an option. But I don't think it worth the effort and added complexity to the project.

IIUC we primarily target MSVC (correct me if Im wrong) and it should build fine there.  Can we just specify which compiler to use? And if the user decides to use something else - they are on their own.

FabianTerhorst:
Also its always worth a try to update the submodule, because we are 3,5 years behind.

prikolium-cfx:
Issue is related to Visual Studio toolset major update (in minor version). Our team will completely move to v144 toolset soon and it will be fixed.

Mathu-lmn:
Ok, surprisingly, my PR got merged in a few hours, so I think it's worth looking into updating the submodule to its latest version

PR #2340, opened 2024-01-14
feat(extra-natives/five) `GET_VEHICLE_HAS_FLAG`

kams3:
Adds an extra native to query vehicle.meta flags
The complete list of flags can be found at https://gtamods.com/wiki/Vehicles.meta#flags
Tested on b2802
EDIT: Tested on b1604 and b3095 on every vehicle class. Also tested on empty vehicles and vehicles driven by npcs.

FiveM_2024-01-14_02-22-34
FiveM_2024-01-14_02-21-12
FiveM_2024-01-14_02-38-49

<details>
  <summary>Sample Resource in Typescript</summary>

  ```ts
  RegisterCommand("print-flags", () => console.log(getAllVehicleFlags()), false);

  function getAllVehicleFlags(mVehicle = GetVehiclePedIsIn(PlayerPedId(), false)) {
    const flags = [];
    for (let i = 0; i < 204; i++) {
      if (GetVehicleHasFlag(mVehicle, i)) flags.push(VEHICLE_FLAGS[i]);
    }
    return flags;
  }

  export enum VEHICLE_FLAGS {
    FLAG_SMALL_WORKER,
    FLAG_BIG,
    FLAG_NO_BOOT,
    FLAG_ONLY_DURING_OFFICE_HOURS,
    FLAG_BOOT_IN_FRONT,
    FLAG_IS_VAN,
    FLAG_AVOID_TURNS,
    FLAG_HAS_LIVERY,
    FLAG_LIVERY_MATCH_EXTRA,
    FLAG_SPORTS,
    FLAG_DELIVERY,
    FLAG_NOAMBIENTOCCLUSION,
    FLAG_ONLY_ON_HIGHWAYS,
    FLAG_TALL_SHIP,
    FLAG_SPAWN_ON_TRAILER,
    FLAG_SPAWN_BOAT_ON_TRAILER,
    FLAG_EXTRAS_GANG,
    FLAG_EXTRAS_CONVERTIBLE,
    FLAG_EXTRAS_TAXI,
    FLAG_EXTRAS_RARE,
    FLAG_EXTRAS_REQUIR,
    FLAG_EXTRAS_STRONG,
    FLAG_EXTRAS_ONLY_BREAK_WHEN_DESTROYED,
    FLAG_EXTRAS_SCRIPT,
    FLAG_EXTRAS_ALL,
    FLAG_EXTRAS_MATCH_LIVERY,
    FLAG_DONT_ROTATE_TAIL_ROTOR,
[...]

PR #2780, opened 2024-09-08
feat(natives/vehicle): implement getter/setter for vehicle flags

spacevx:
### Goal of this PR
Implement natives to set and get vehicle flags. This is the PR of @kams3 with the addition of the setter.

By reading the specific offset in this signature: 48 85 C0 74 3C 8B 80 ? ? ? ? C1 E8 0F, which exposes vehicle flags on the vehicle. 

This rely on https://github.com/citizenfx/fivem-docs/pull/480

FiveM

### Successfully tested on

**Game builds:** 3095

**Platforms:** Windows

### Checklist

### DIFF
--- a/code/components/extra-natives-five/src/VehicleExtraNatives.cpp
+++ b/code/components/extra-natives-five/src/VehicleExtraNatives.cpp
@@ -281,6 +281,7 @@ static int VisualHeightSetOffset = 0x07C;
 static int LightMultiplierGetOffset;
 static int VehiclePitchBiasOffset;
 static int VehicleRollBiasOffset;
+static int VehicleFlagsOffset;
 
 static int VehicleDamageParentOffset;
 static int VehicleHandlingOffset;
@@ -753,6 +754,11 @@ static HookFunction initFunction([]()
 	{
 		VehiclePitchBiasOffset = *hook::get_pattern<uint32_t>("0F 2F F7 44 0F 28 C0 F3 44 0F 58 83", 12);
 		VehicleRollBiasOffset = VehiclePitchBiasOffset - 4;
+	}
+
+	{
+		auto location = hook::get_pattern<char>("48 85 C0 74 3C 8B 80 ? ? ? ? C1 E8 0F");
+		VehicleFlagsOffset = *(uint32_t*)(location + 7);
 	}
 
 	{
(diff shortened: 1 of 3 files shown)


martonp96:
Hey @ahcenezdh,
I'm not sure i understand the problem with flagIndex 224 well, since the documentation only mentions the valid range as "0-203".
Also wouldn't it be better to cast the memory to an "std::bitset<32>*" type so the code is more human readable when you work with it?

ahcenezdh:
> Hey @ahcenezdh, Thank you for your contribution. I'm not sure i understand the problem with flagIndex 224 well, since the documentation only mentions the valid range as "0-203". Also wouldn't it be better to cast the memory to an "std::bitset<32>*" type so the code is more human readable when you work with it?

Hey, the problem with flagIndex is that if R* adds new flags, then the flagIndex should be updated since the value is hardcoded, in the old PR Disquse [mentionned](https://github.com/citizenfx/fivem/pull/2340#discussion_r1471969437) a way to get the number of flags directly from rage parser

martonp96:
> > Hey @ahcenezdh, Thank you for your contribution. I'm not sure i understand the problem with flagIndex 224 well, since the documentation only mentions the valid range as "0-203". Also wouldn't it be better to cast the memory to an "std::bitset<32>*" type so the code is more human readable when you work with it?
> 
> Hey, the problem with flagIndex is that if R* adds new flags, then the flagIndex should be updated since the value is hardcoded, in the old PR Disquse [mentionned](https://github.com/citizenfx/fivem/pull/2340#discussion_r1471969437) a way to get the number of flags directly from rage parser

Then i think its fine to cast it to `std::bitset<256>*` and flagIndex should be less than 256. Accessing anything above 203 should be irrelevant as writing or reading them has no effect.
Additionally, there are helper functions in this file. Getting the model info ptr for example can be done as `readValue<uint64_t>(vehicle, ModelInfoPtrOffset)` like it was done for GET_VEHICLE_DRAWN_WHEEL_ANGLE_MULT. 
If you can do these changes, then the PR can be merged.

ahcenezdh:
> > > Hey @ahcenezdh, Thank you for your contribution. I'm not sure i understand the problem with flagIndex 224 well, since the documentation only mentions the valid range as "0-203". Also wouldn't it be better to cast the memory to an "std::bitset<32>*" type so the code is more human readable when you work with it?
> > 
> > 
> > Hey, the problem with flagIndex is that if R* adds new flags, then the flagIndex should be updated since the value is hardcoded, in the old PR Disquse [mentionned](https://github.com/citizenfx/fivem/pull/2340#discussion_r1471969437) a way to get the number of flags directly from rage parser
> 
> Then i think its fine to cast it to `std::bitset<256>*` and flagIndex should be less than 256. Accessing anything above 203 should be irrelevant as writing or reading them has no effect. Additionally, there are helper functions in this file. Getting the model info ptr for example can be done as `readValue<uint64_t>(vehicle, ModelInfoPtrOffset)` like it was done for GET_VEHICLE_DRAWN_WHEEL_ANGLE_MULT. If you can do these changes, then the PR can be merged.

Should be good, tell me if something is not, i just need to re test everything to see if there is no problems

martonp96:
> > > > Hey @ahcenezdh, Thank you for your contribution. I'm not sure i understand the problem with flagIndex 224 well, since the documentation only mentions the valid range as "0-203". Also wouldn't it be better to cast the memory to an "std::bitset<32>*" type so the code is more human readable when you work with it?
> > > 
> > > 
> > > Hey, the problem with flagIndex is that if R* adds new flags, then the flagIndex should be updated since the value is hardcoded, in the old PR Disquse [mentionned](https://github.com/citizenfx/fivem/pull/2340#discussion_r1471969437) a way to get the number of flags directly from rage parser
> > 
> > 
> > Then i think its fine to cast it to `std::bitset<256>*` and flagIndex should be less than 256. Accessing anything above 203 should be irrelevant as writing or reading them has no effect. Additionally, there are helper functions in this file. Getting the model info ptr for example can be done as `readValue<uint64_t>(vehicle, ModelInfoPtrOffset)` like it was done for GET_VEHICLE_DRAWN_WHEEL_ANGLE_MULT. If you can do these changes, then the PR can be merged.
> 
> Should be good, tell me if something is not, i just need to re test everything to see if there is no problems

Looks good to me

spacevx:
> Tested in b3407 and works nice

Thanks

Nobelium-cfx (approved):
LGTM. Sorry, it should have been merged after the initial approval, but got lost because flag was not set. Marking as ready for merge now.

PR #2792, opened 2024-09-15
feat(natives): implement `GET_MOUNTED_FILE_HASH`

PichotM:
### Goal of this PR
This PR introduces a new native `GET_MOUNTED_FILE_HASH`.

By implementing the native using the same logic as puremode's hashes.

FiveM & RedM client

### Successfully tested on
FiveM and will most likely work in RedM too.

**Game builds:** 3095

**Platforms:** Windows

### Checklist


code/components/extra-natives-five/src/FileNatives.cpp
```diff
@@ -0,0 +1,92 @@
...
+					hashFunction.update(fixedHeader);
+					std::vector<uint8_t> hash(hashFunction.output_length());
+					hashFunction.final(hash.data());
+
+					result = ToHexString(hash.data(), hash.size());
+				}
+			}
+		}
+
+		context.SetResult(result.c_str());
```
iridium-cfx, line 90: SetResult doesn't copy the string, which will be destroyed once the native returns. You need to make `result` static so its contents lasts long to be copied.

code/components/extra-natives-five/src/FileNatives.cpp
```diff
@@ -0,0 +1,92 @@
...
+			{
+				auto handle = device->Open(path, true);
+
+				if (handle != -1)
+				{
+					auto len = device->GetFileLength(handle);
+					std::vector<char> data(len);
+
+					device->Read(handle, data.data(), len);
+					device->Close(handle);
```
iridium-cfx, line 64: Reading and hashing the whole file is almost certainly going to cause noticeable stuttering, since this will be running on the main thread.

code/components/extra-natives-five/src/FileNatives.cpp
```diff
@@ -0,0 +1,92 @@
...
+						auto x = *(uint32_t*)(&fixedHeader[i + 4]);
+						if (x != 0x7fffff00 && (x & 0x80000000) == 0)
+						{
+							auto y = (uint32_t*)(&fixedHeader[i + 12]);
+							if (*y)
+							{
+								*y = 1;
+							}
+						}
+					}
```
iridium-cfx, line 78: This code is specifically for dealing with decrypted RPF headers, and is meaningless for other files.

iridium-cfx (commented):
I don't think this is a good idea to have as a native, given the performance implications.

PichotM:
> I don't think this is a good idea to have as a native, given the performance implications.

Do you have any ideas for a better implementation? 
I guess this native will only be used for a few small files on join. That won't necessarily be too expensive to run

iridium-cfx:
> > I don't think this is a good idea to have as a native, given the performance implications.
> 
> Do you have any ideas for a better implementation?
Not off the top of my head, no

> I guess this native will only be used for a few small files on join. That won't necessarily be too expensive to run
Perhaps. LOAD_RESOURCE_FILE exists so I guess loading files inside of natives isn't unheard of.

slashkeyvalue:
More weird anticheat "detection" attempts? :/

ahcenezdh:
> More weird anticheat "detection" attempts? :/

Some people are editing their client files to get advantages in PvP servers, I guess it's to block this

p1u3o:
What about a method to find out what files are loaded in the mods or citizen folder?

Heisenberg691:
it would be nice if there is a method to identify modified files from mods/citizen/game files or blocking certain files to be modified. For example i want my players to use NVE graphics mod, but not to use custom audio files like louder footsteps or headshot sounds for advantage in pvp. Maybe a new pure mode level 3 that will block all modified client files except known graphics mods can be a solution for that.

manzarek123:
> More weird anticheat "detection" attempts? :/

99% of cheaters these days are people using modified files.
By just dragging a rpf file in your mod folder you can have infinite stamina, perfect accuracy, bullet penetration, no bush, ect.
It doesn't even feel like cheating because it is so easy.

A simple native like this could prevent this and remove 99% of cheaters from FiveM and bring more positivity.
Sure this is not the perfect system but it starting to feel more and more like the perfect system is never going to happen. Just like the perfect system for dynamic pool resizing never happened.
So I think we should accept this kind of compromise and give more power to server owners.

As for performance issue, this will be checked during the loading process so this is not really an issue. Maybe adding a note in the native description could be a good idea then.

thisneq:
I have thought about something similar some time ago. In my opinion, there should be an additional pure mode which would allow server owners to specify the hashes of allowed files in the server configuration. Native function for that is not the most convenient solution.

PichotM:
> I have thought about something similar some time ago. In my opinion, there should be an additional pure mode which would allow server owners to specify the hashes of allowed files in the server configuration. Native function for that is not the most convenient solution.

your solution is the same thing but requires a game restart for every join, not really convenient

william-des:
> By just dragging a rpf file in your mod folder you can have infinite stamina, perfect accuracy, bullet penetration, no bush, ect.
> It doesn't even feel like cheating because it is so easy.

Furthermore, unlike a traditional cheat, players don't risk being banned from the platform

>  i want my players to use NVE graphics mod, but not to use custom audio files like louder footsteps or headshot sounds for advantage in pvp. Maybe a new pure mode level 3 that will block all modified client files except known graphics mods can be a solution for that.

It's the perfect example of why Pure Mode is not widely used and why we need something like this instead. Each game server has different requirements regarding the files that should be blocked

Keyinator:
> > By just dragging a rpf file in your mod folder you can have infinite stamina, perfect accuracy, bullet penetration, no bush, ect.
> > It doesn't even feel like cheating because it is so easy.
> 
> Furthermore, unlike a traditional cheat, players don't risk being banned from the platform
> 
> > i want my players to use NVE graphics mod, but not to use custom audio files like louder footsteps or headshot sounds for advantage in pvp. Maybe a new pure mode level 3 that will block all modified client files except known graphics mods can be a solution for that.
> 
> It's the perfect example of why Pure Mode is not widely used and why we need something like this instead. Each game server has different requirements regarding the files that should be blocked

Seconding exactly what @william-des and @manzarek123 said.
There's multiple categories of files that would be allowed/denied by certain servers like:
- Sounds
- Skys
- Bushes
- Hitboxes
- Tracers
- Kill Efects
- Shaders

While I think this native would be useful too, I think a better approach would be to detect modification on certain file-names (or better yet, "modification-classes") then allow server owners to either transparently allow/black these classes or do it via a self-coded script that will kick a user for disallowed files.

Even better yet would be if denied "classes" can be disabled by the server-owner and the vanilla data is used respectively.
I don't know if that's possible or not though.

prikolium-cfx:
Yeah, initially it was my idea. Including the fact that all the content files are read-only, can we make async hashing possible in another thread and waiting for result?

p1u3o:
I think sounds would make be tricky having just categories, for example I can see custom sirens being acceptable, but weapon sounds not so much (can make silenced weapons/footsteps easier to hear).

Perhaps servers don't even need a hash, simply having a list of files overriden by rpfs in the mod folder, and relying on puremode to make sure the game files aren't touched, would still be better than the status quo.

cloudyyy0273:
please add this already, no one can stand pure mode xD

ook3D:
> please add this already, no one can stand pure mode xD

? I can't see why you would dislike pure mode, unless you are trying to let players load whatever random files they want into your server.

PsychoShock:
Maybe because pure mode is full restricted and doesn't allow as a server owner to accept some files you may want players to have. Especially for PvP servers.

ook3D:
> Maybe because pure mode is full restricted and doesn't allow as a server owner to accept some files you may want players to have. Especially for PvP servers.

Fair enough I guess, didn't really think of such case

prikolium-cfx:
@PichotM What do you think about defining hashes list on server side in cfg file? So we can guarantee that check will be done on connect time and won't affect performance in play time.

manzarek123:
> @PichotM What do you think about defining hashes list on server side in cfg file? So we can guarantee that check will be done on connect time and won't affect performance in play time.

We then need a way to generate the hash without this native so that we can put it in the cfg.

code/components/extra-natives-five/src/FileNatives.cpp
```diff
@@ -0,0 +1,144 @@
...
+	const char* path = pathStr->c_str();
+	rage::fiDevice* device = rage::fiDevice::GetDevice(path, true);
+
+	if (device)
+	{
+		auto handle = device->Open(path, true);
+
+		if (handle != -1)
+		{
+			auto len = device->GetFileLength(handle);
```
prikolium-cfx, line 68: Use chunked reading and hash calculation here

PichotM:
will be back soon

DaniGP17:
Would it be better to add more options to be able to configure pure mode to specify what you want and don't want to restrict?

p1u3o:
> Would it be better to add more options to be able to configure pure mode to specify what you want and don't want to restrict?

I don't think that's a proper alternative to this feature, as this feature allows server owners to whitelist popular graphics mods like NVE.

william-des:
> Would it be better to add more options to be able to configure pure mode to specify what you want and don't want to restrict?

iirc the issue with pure mode is that you need to deny whole rpfs instead of individual file like this pr was meant to

DaniGP17:
> I don't think that's a proper alternative to this feature, as this feature allows server owners to whitelist popular graphics mods like NVE.

This, more than whitelisting, allows you to create your "own hash validation", so you would allow all files except those you validate with the native ones or those listed in cfg.

> iirc the issue with pure mode is that you need to deny whole rpfs instead of individual file like this pr was meant to

That makes more sense, yes, I guess the best option is a list of hashes in the cfg as prikolium said, what I have doubt is how the server owners will obtain those hashes, do we leave these native ones so they can use them to obtain the hashes?

william-des:
If the team really want to avoid this native being called at runtime to avoid performance issues etc you can still restrict this native to dev mode or similar

DaniGP17:
Yeah I'll do that then

PR #2803, opened 2024-09-21
feat(extra-natives): add `CNetObject` to `GET_GAME_POOL`

AvarianKnight:
### Goal of this PR
Give a fast path for calls to `GET_GAME_POOL` that only want to get networked objects (which can be a max of 80 objects), reducing the serialization/de-serialization overhead for this native call.

Getting entities from `GetGamePool("CObject")` can be costly due to the fact that this pool will include map objects too, which can lead to this table being thousands of objects long.

This doesn't include networked variants for `CPed` and `CVehicle` because these are more often times going to be networked, where as `CObject` is more likely to be client local objects

This adds a new fast path `CNetObject` which should only return objects that are networked

<details>
   <summary>Test Example</summary>
   
```lua
CreateThread(function()
    local networkedObjects = GetGamePool("CNetObject")
    for i = 1, #networkedObjects do
        local obj = networkedObjects[i]
        SetEntityAsMissionEntity(obj, true, false)
        DeleteEntity(obj)
    end
end)

local trackedObjectIds = {}
local logged = {}
CreateThread(function()
    local model = `prop_boombox_01`

    RequestModel(model)
    while not HasModelLoaded(model) do
        Wait(0)
    end

    while true do
        local networkedObjects = GetGamePool("CNetObject")
        print("netobj", #networkedObjects)

        for i = 1, #networkedObjects do
            local obj = networkedObjects[i]
            if trackedObjectIds[obj] then
                trackedObjectIds[obj] = false
                print("got tracked object", obj)
            end
        end

        local obj = GetGamePool("CObject")
        print("obj", #obj)

        local pedPos = GetEntityCoords(PlayerPedId())
        local obj = CreateObject(model, pedPos, true, true, false)
        if obj ~= 0 then
            trackedObjectIds[obj] = true
        end

        Wait(1000)
    end
end)
```
</details>

FiveM

### Successfully tested on

**Game builds:** N/A

**Platforms:** Windows

### Checklist

### DIFF
--- a/code/components/citizen-server-impl/src/state/ServerGameState_Scripting.cpp
+++ b/code/components/citizen-server-impl/src/state/ServerGameState_Scripting.cpp
@@ -962,7 +962,7 @@ static void Init()
 			desiredType = EntityType::Ped;	
 		else if (poolName == "CVehicle")
 			desiredType = EntityType::Vehicle;
-		else if (poolName == "CObject")
+		else if (poolName == "CObject" || poolName == "CNetObject")
 			desiredType = EntityType::Object;
 
 		for (auto& entity : gameState->m_entityList)
--- a/code/components/extra-natives-five/src/PoolTraversalNatives.cpp
+++ b/code/components/extra-natives-five/src/PoolTraversalNatives.cpp
@@ -141,7 +141,7 @@ struct PickupPoolTraits
 	}
 };
 
-template<typename TTraits>
+template<typename TTraits, bool NetworkOnly = false>
 static void SerializePool(fx::ScriptContext& context)
 {
 	std::vector<uint32_t> guids;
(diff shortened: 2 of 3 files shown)


Legacy-TacticalGamingInteractive:
is it possible for an increase beyond 80?

AvarianKnight:
I have no idea.

I don't think using networked objects is typically a good idea anyways, most of the use cases that networked objects will be used for (such as attaching props to a player) are better off with using state bags anyways, or manual syncing (in cases like weed scripts that usually have a static position and wont move)

FlawwsX:
This is a top notch change!

FabianTerhorst (approved):
Thanks for the contribution 👍

PR #2804, opened 2024-09-21
tweak(server/http): provide error response headers

D4isDAVID:
### Goal of this PR

In the source code, an HTTP request failure is both failure to connect, but also response codes such as 4xx or 5xx. In both cases, response headers are not provided to scripts.

The problem is that when HTTP servers respond with an error code such as 4xx or 5xx, they may also provide response headers which may contain valuable information - such as the Discord API providing rate limit information when getting rate limited.

The goal is to provide those response headers when requests fail.

Always creates the response data, not only when the request is successful, and passes it to the callback event also when the request has failed.

Seeing as this may be constituted as a breaking change, I passed it as a new callback parameter instead of the existing `headers` parameter.

Server, ScRT: Lua

### Successfully tested on

**Platforms:** Windows

### Checklist


Roms1383:
Any ETA for review of this PR ?

D4isDAVID:
Any updates on a review on this? Most of the PR is moving some of the code outside an if statement, and the rest is minimal additions.

PR #2824, opened 2024-09-29
feat(server): add sv_httpProxyOnly convar

kasuganosoras:
### Goal of this PR
This PR added a new convar `sv_httpProxyOnly`, it can restrict the IP addresses requesting the file server.
﻿
When you are using `fileserver_add` to add your custom proxy server, you can use this convar to prevent the client access the original file server directly.

Example:
```ini
set sv_httpProxyOnly "123.123.123.123"            # Your proxy server IP address
fileserver_add "https://my-cdn.example.com/files" # Your proxy server
```

Then the server will reject any requests that are not coming from your proxy server (`123.123.123.123`), when they are trying to access, it will result in error: `HTTP 403: Host <Client IP> is not allowed to access this endpoint.`
﻿
Modified the `code/components/citizen-server-impl/src/FilesHttpHandler.cpp` file, added a check for the client address.
﻿
Server
﻿
### Successfully tested on
* **Game builds:** 3095
* **Platforms:** Windows
﻿
### Checklist
﻿
### Fixes issues
None

### DIFF
--- a/code/components/citizen-server-impl/src/FilesHttpHandler.cpp
+++ b/code/components/citizen-server-impl/src/FilesHttpHandler.cpp
@@ -14,6 +14,7 @@
 #include <filesystem>
 
 constexpr const size_t kFileSendSize = 128 * 1024;
+std::shared_ptr<ConVar<std::string>> g_httpProxyOnly;
 
 namespace fx
 {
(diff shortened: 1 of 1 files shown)


Korioz:
It would be better to use the already implemented approach with `sv_proxyIPRanges` which will permits to match multiple IP address ranges.

Just adding a convar like `sv_httpFileServerOnlyProxy` set to `true` to keep backward compatibility, and you will be good.

kasuganosoras:
> It would be better to use the already implemented approach with `sv_proxyIPRanges` which will permits to match multiple IP address ranges.
> 
> Just adding a convar like `sv_httpFileServerOnlyProxy` set to `true` to keep backward compatibility, and you will be good.

That is a good idea, now I pushed a new commit to implement this.

FabianTerhorst (approved):
Works fine, thanks for your contribution.

PR #2785, opened 2024-09-11
feat(server): add getters for neons

ahcenezdh:
### Goal of this PR

* Added 2 getters GetVehicleNeonColour(), GetVehicleNeonEnabled()

### How is this PR achieving the goal
By reading/write bits in CVehicleAppearanceDataNode

For crew things needs more test/check some data could be wrong here

### This PR applies to the following area(s)

FiveM

### Successfully tested on

**Game builds:** 3095

**Platforms:** Windows, Linux

### Checklist

- [x] Code compiles and has been tested successfully.
- [x] Code explains itself well and/or is documented.
- [x] My commit message explains what the changes do and what they are for.
- [x] No extra compilation warnings are added by these changes.

PR #2834, opened 2024-10-03
fix: vehicle sync tree causing getters to return the wrong values

minemalox:
### Goal of this PR

- Fixing getting the wrong values returned when using server getters like `GET_VEHICLE_NUMBER_PLATE_TEXT`

- Removing the second serialization of `data.primaryColour` which seems to be a small copy & paste mistake. Added with the following pr: #2785

- FiveM, Server

### Successfully tested on

- Couldn't test it myself, but comparing the new and old synctree only showed this difference between them.

### Checklist

### Fixes issues

- Not working server getters like `GET_VEHICLE_NUMBER_PLATE_TEXT`

### DIFF
--- a/code/components/citizen-server-impl/include/state/SyncTrees_Five.h
+++ b/code/components/citizen-server-impl/include/state/SyncTrees_Five.h
@@ -671,7 +671,6 @@ struct CVehicleAppearanceDataNode : GenericSerializeDataNode<CVehicleAppearanceD
         s.Serialize(8, data.wheelColour);
         s.Serialize(8, data.interiorColour);
         s.Serialize(8, data.dashboardColour);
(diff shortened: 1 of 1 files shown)


ahcenezdh:
Good catch, indeed it was a mistake i didn't see this one

FabianTerhorst:
Please make sure to only commit the relevant line changes. You made many whitespace changes here that are not relevant for this fix.

minemalox:
> Please make sure to only commit the relevant line changes. You made many whitespace changes here that are not relevant for this fix.

Didn't realize yesterday that there were still some whitespace changes, should now be fixed. @FabianTerhorst

FabianTerhorst (approved):
Thanks for your contribution 👍

PR #2835, opened 2024-10-03
feat(extra-natives/five): add IS_FULLSCREEN

Nyakudev:
### Goal of this PR
This PR introduces a new native function that determines whether the client window is in fullscreen mode. This feature can benefit web developers and be useful for some functionalities.

By adding the native function ```IsFullscreen```, developers can easily check if the client window is in fullscreen mode.

- FiveM 
- Natives

### Successfully tested on
- 3258
- 3095
- 2944

**Game builds:**
all

**Platforms:**
Windows

### Checklist

### Fixes issues
none


Nyakudev:
Sorry for shitty commits, it's because i have modified the native-decls in github web.

AvarianKnight:
Can you explain what this is needed for? 

> This feature can benefit web developers and be useful for some functionalities.

Isn't very specific.

A feature request like this got shut down before out of fear that people would abuse this for cheat detection

Nyakudev:
> A feature request like this got shut down before out of fear that people would abuse this for cheat detection

What is the problem if it's used to detect cheaters? It's the server problem if it lead to a fake ban.

matehun00:
You can already do this with
```lua
print(GetConvar("profile_vidScreenType", 1))
```

Nyakudev:
> You can already do this with
> 
> ```lua
> print(GetConvar("profile_vidScreenType", 1))
> ```

oh ok good to know but very few people know it adding the native would make it “easier”.

AvarianKnight:
> > A feature request like this got shut down before out of fear that people would abuse this for cheat detection
> 
> What is the problem if it's used to detect cheaters? It's the server problem if it lead to a fake ban.

Because this will lead to servers trying to force players to play in "Fullscreen" or risk being banned because of the *possibility* of the player cheating. 

There is also *no need* for the server to have access to this information, unless you have another reason other than "to detect cheats"

prikolium-cfx:
Thank you for your contribution but our current common goal is to not increase codebase without proper reason for that. I can't find any real use case for that. If you have such use case feel free to provide it, but for now I will close this PR.

PR #2836, opened 2024-10-03
fix(ci/windows): Downgrade MSVC build tools

tens0rfl0w:
### Goal of this PR
Github Actions running on 'windows-latest' get shipped with VS 2022 C++ x64/x86 build tools v14.41-17.11 which includes various breaking changes that make building on the CI pipeline fail.

Github Actions doesn't allow to use older image versions, so we downgrade VS build tools to v14.40-17.10 manually as a temporary fix. This only downgrades necessary components to avoid increasing the build time any further.

This can be reverted once the build tool chain got updated.

CI, Building

### Successfully tested on

**Platforms:** Github Actions - Windows Latest

### Checklist

### Fixes issues
/

### DIFF
--- a/.github/workflows/ci_build.yml
+++ b/.github/workflows/ci_build.yml
@@ -44,6 +44,22 @@ jobs:
           git pull origin $GITHUB_REF:PR_BRANCH --depth=1
           git submodule update --jobs=$JOB_SLOTS --init --depth=1
         
+      - name: Downgrade MSVC build tools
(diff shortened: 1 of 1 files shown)


FabianTerhorst:
Does that mean we don't need this pull request anymore with #2837?

FabianTerhorst:
Ok then this probably can be closed, because merging both at the same time won't be needed.

PR #2841, opened 2024-10-05
feat(scripting/net): Unreliable network events

tens0rfl0w:
### Goal of this PR
There are certain scenarios where resource developers may want to send frequent data that is **not** state-critical (i.e., can be lost). Currently, all script-transmitted data is treated as reliable (ACK, resends, etc.), which can introduce a noticeable hit on general server performance when used for data that should not get sent as reliable.

- Introducing a new API to send unreliable client and server network events
- All new API functions are equal to the existing reliable send functions:
  - Lua: `TriggerUnreliableClientEvent` and `TriggerUnreliableServerEvent`
  - JS;  `emitUnreliableNet`, `TriggerUnreliableClientEvent` and `TriggerUnreliableServerEvent`
  - C# V1:  `Player.TriggerUnreliableEvent`, `TriggerUnreliableClientEvent` and `TriggerUnreliableServerEvent`
  - C# V2:   `Player.TriggerUnreliableEvent`, `Events.TriggerUnreliableAllClientsEvent`, `Events.TriggerUnreliableClientEvent`, and `Events.TriggerUnreliableServerEvent`
 
Implementation has been tested on all available ScRTs and should not introduce any breaking changes (`*__INTERNAL` natives are already declared as internal functions).

FiveM, Server, Natives, ScRT: Lua, ScRT: C#, ScRT: JS

### Successfully tested on

**Platforms:** Windows, Linux

### Checklist

### Fixes issues
/


manzarek123:
This is going to break a lot of server since people are using TRIGGER_CLIENT_EVENT_INTERNAL directly

Maybe last arg should be "unreliable" instead of "reliable"

https://github.com/overextended/ox_lib/blob/a2a711498658d36e51bd9853f681cacac502f9bb/imports/triggerClientEvent/server.lua#L14

PR #2847, opened 2024-10-07
tweak(gamestate/server): Apply routing bucket changes to child attachments

tens0rfl0w:
### Goal of this PR
Changing the routing bucket of an entity does not update the routing buckets of possible attached child entities. This causes various problems, one of which is broken collision data for entities that get re-attached to other entities with the same handle.

Sync data is only parent-focused, so we have no info about possible child attachments. This makes it necessary to iterate through the server's entity list and traverse possible chained attachments to make sure we catch every possible attachment path to the root entity.

I tried to keep recursive calls as low as possible with tracking already traversed paths, which should also prevent cyclic attachments from causing infinite recursions.

https://github.com/user-attachments/assets/df8aaa42-ca78-4c2d-a568-47fb93fd3fd4

FiveM, Server

### Successfully tested on
**Game builds:** 3258

**Platforms:** Windows

### Checklist

### Fixes issues
/

### DIFF
--- a/code/components/citizen-server-impl/src/state/ServerGameState_Scripting.cpp
+++ b/code/components/citizen-server-impl/src/state/ServerGameState_Scripting.cpp
@@ -17,6 +17,102 @@ namespace fx
 void DisownEntityScript(const fx::sync::SyncEntityPtr& entity);
 }
 
+namespace
+{
+	bool IsConnectedToRootEntity(const fx::sync::SyncEntityPtr& entity,
+								const fx::sync::SyncEntityPtr& rootEntity,
+								const fwRefContainer<fx::ServerGameState>& gameState,
+								std::unordered_set<uint32_t>& visitedEntities,
+								std::unordered_set<uint32_t>& rootConnectedEntities)
+	{
+		if (!entity)
+		{
+			return false;
+		}
+
+		if (rootConnectedEntities.find(entity->handle) != rootConnectedEntities.end())
+		{
+			return true;
+		}
+
+		if (visitedEntities.find(entity->handle) != visitedEntities.end())
+		{
+			return false;
+		}
+
+		visitedEntities.insert(entity->handle);
+
+		const auto& syncTree = entity->syncTree;
+		if (syncTree)
+		{
+			fx::sync::SyncEntityPtr parentEntity = {};
+			fx::sync::SyncEntityPtr vehicleEntity = {};
+			bool isAttached = false;
+			bool isOccupant = false;
+
+			const auto attachment = syncTree->GetAttachment();
+			if (attachment && attachment->attached && attachment->attachedTo)
+			{
+				parentEntity = gameState->GetEntity(0, attachment->attachedTo);
+				if (parentEntity && parentEntity->handle == rootEntity->handle)
+				{
+					isAttached = true;
+				}
+			}
+
+			const auto pedGameState = syncTree->GetPedGameState();
+			if (pedGameState && pedGameState->curVehicle)
+			{
+				vehicleEntity = gameState->GetEntity(0, pedGameState->curVehicle);
+				if (vehicleEntity && vehicleEntity->handle == rootEntity->handle)
+				{
+					isOccupant = true;
+				}
+			}
+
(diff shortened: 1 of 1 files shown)


CeebDev:
Could be great to also change routing buckets of passagers when we change the routing bucket of a vehicle instead of iterate on them

Identity-labs:
Great news :)
Any solution for non attached entities like pet ? or maybe vehicle or horse ?
to be switched automatically as well ?

JnKTechstuff:
> > Could be great to also change routing buckets of passagers when we change the routing bucket of a vehicle instead of iterate on them QoL
> 
> Yeah, this definitely could be added, though I'm not sure if this is a generally desired behavior. Would be great to get some additional feedback on this.

Well would it be desired behavior to set a vehicle into a bucket and all passengers fall out and the vehicle disappear? Logically thinking about it setting a vehicle's bucket should set all players and entities inside said vehicle into the same bucket. If people wish to remove people from a vehicle plenty of methods exist that don't involve removing it from that player's existence.

I think its safe to do the same with vehicle and pax if not already done.

Identity-labs:
> Could you elaborate on what you mean with 'entities like pet'? Not sure If I got you right here, are you also referring to automatically applying bucket changes to vehicle passengers?

Today if you have peds following you, with relationship/range & co you have to remove the ped before changing bucket.
Because ped will no longer have owner in original bucket and stay alone in this bucket.

May be add a native to tag entities to be linked to root entity, like that they are migrated in bucket when root entity is changing bucket same way as attached entity ?

Shinzuh:
> Could be great to also change routing buckets of passagers when we change the routing bucket of a vehicle instead of iterate on them QoL

There is scenarios where we want to change the routing bucket of the driver and not the passengers.
The best option would be to let a way to chose.

CeebDev:
> There is scenarios where we want to change the routing bucket of the driver and not the passengers. The best option would be to let a way to chose.
Give example of case where you only need to change bucket of vehicle but not passengers

Shinzuh:
Thinking of it I don't have any special exemples for my use case but the backward compatibilty scare me a bit.
Scripts wern't expecting that a player could change the bucket of other players in some special situations.
A bool to activate this new comportment seems to be the strict necessary

Gravxd:
I can't think of any scripts that would be designed to *expect* the players to magically teleport out of the player's vehicle by changing the bucket?

I can think of more use cases in which automatically switching the bucket for passengers / driver when changing the bucket of the vehicle they are in, is preferred /desired.

Gravxd:
> Thinking of it I don't have any special exemples for my use case but the backward compatibilty scare me a bit. Scripts weren't expecting that a change of a bucket for a player could change it for other players too. A bool to activate this new comportment seems to be the strict necessary

Also keep in mind the original proposal idea was to change to change the passenger's bucket (if any passengers) when the vehicle entity's bucket was changed, not the driver. So i really couldn't see any unintended behaviour as if you are modifying a vehicle's bucket, youd probably want to keep the driver.

draobrehtom:
Regarding entity bucket changes and related entities, I would follow an approach similar to SetEntityCoords. For example, when changing a vehicle's coordinates, all passengers should remain inside. However, if we change a passenger's coordinates, they should be teleported out of the vehicle. The same logic applies to attached entities.

CeebDev:
Any update on this ? If you want full backward compat, add a convar

jellyton255:
> > Could be great to also change routing buckets of passagers when we change the routing bucket of a vehicle instead of iterate on them QoL
> 
> Yeah, this definitely could be added, though I'm not sure if this is a generally desired behavior. Would be great to get some additional feedback on this.

An example use case of this would be if you had some sort of garage system, like GTA:O, where players can get in to a vehicle and drive out, it would be useful to have the passengers change buckets along with the vehicle otherwise they'll be kicked out when transferring buckets (leaving the garage) since the vehicle won't exist for them for x amount of time until the bucket swap is complete.

BennoArnautovic:
This is still important, why is this waiting this long for implementation...

PR #2851, opened 2024-10-08
tweak(codegen): set VAR_STRING as an unsupported type for OAL

AvarianKnight:
### Goal of this PR
Allow `VAR_STRING` to be used on RDR when using OAL

Marks it as 'unsupported' so it wont be generated as a supported native for OAL

RedM

### Successfully tested on
This was only tested that it doesn't add to the map in `NativesRDR.h`

**Game builds:** .. 

**Platforms:** Windows, Linux

### Checklist

### Fixes issues
Fixes the issue @outsider31000 reported

### DIFF
--- a/ext/natives/codegen_out_native_lua.lua
+++ b/ext/natives/codegen_out_native_lua.lua
@@ -9,7 +9,10 @@ local unsupList = {
 	'SET_OBJECT_AS_NO_LONGER_NEEDED',
 	'SET_PED_AS_NO_LONGER_NEEDED',
 	'SET_VEHICLE_AS_NO_LONGER_NEEDED',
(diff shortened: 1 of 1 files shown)


outsider31000:
didnt know that list existed.
does this mean VarString will still work right? just wont be considered by the OAL?

AvarianKnight:
> didnt know that list existed. does this mean VarString will still work right? just wont be considered by the OAL?

Yes, it'll fallback to the normal NativeInvoke.

outsider31000:
yes thanks, I read the description wrong.

outsider31000:
someone merge this please.

iridium-cfx (approved):
Looks good, thanks for the fix!

PR #2861, opened 2024-10-12
fix(scripting/mono-v2): validate app domain before removing ref

DaniGP17:
### Goal of this PR

Fix a crash that occurred when an export declared in lua code was executed from C# v2 from the same resource multiple times, the issue with more details is here: https://github.com/thorium-cfx/mono_v2_get_started/issues/32

It seems that there are times that `m_appDomain` is destroyed when the script is stopped, but RemoveRef is still called a little after the C# script is destroyed, therefore RemoveRef is called when `m_appDomain` is null and results in the crash.
I don't know if there is a better solution for this.

ScRT: C#

### Successfully tested on

**Game builds:** .

**Platforms:** Windows

### Checklist

### Fixes issues

fixes [#32](https://github.com/thorium-cfx/mono_v2_get_started/issues/32)


FabianTerhorst (approved):
👍

PR #2869, opened 2024-10-19
tweak(mono-rt2): Use null coalescing for Write & WriteLine

niekschoemaker:
### Goal of this PR

This fixes an error if a null value is accidently passed into Debug.WriteLine

Using simple Null coalescing on the ToString call

ScRT: C# v2

### Successfully tested on
Tested the code seperate from the game, it prints nothing instead of causing a NullReferenceException.

### Checklist

### DIFF
--- a/code/client/clrcore-v2/Debug.cs
+++ b/code/client/clrcore-v2/Debug.cs
@@ -51,12 +51,12 @@ namespace CitizenFX.Core
 
 		[SecuritySafeCritical]
 		public static void Write(string data) => ScriptInterface.Print(s_debugName, data);
-		public static void Write(object obj) => Write(obj.ToString());
+		public static void Write(object obj) => Write(obj?.ToString());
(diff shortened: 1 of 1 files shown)


niekschoemaker:
This is a somewhat opiniated change, so not sure which behavior is more desirable, printing nothing or Debug.WriteLine causing an Exception.

DaniGP17:
Wouldn't it be better to print "null" instead of an empty string like in unity?

niekschoemaker:
> Wouldn't it be better to print "null" instead of an empty string like in unity?

Was thinking the same but would add some extra logic, and this reflects the behavior of default C#, Console.WriteLine(null) in C# prints nothing.

DaniGP17:
Ohh okay, I've been checking how it was done in V1 of C# and it doesn't have this problem because the function expects a string instead of an object, so if you want to print it yourself you have to do the .ToString() and therefore The error is in your code and not in the Debug.WriteLine function itself

PR #2870, opened 2024-10-19
tweak(mono-rt2): Return Complete Coroutine when no data is returned

niekschoemaker:
This allows awaiting exports without a null check.

### Goal of this PR
In the current implementation of exports exports might return null instead of a coroutine, which is the case for exports which execute within a tick (so no wait calls) and don't return a value.
Simply returning a Completed Coroutine with a null value fixes this issue.

By returning a completed coroutine with a null value.

ScRt: C# v2

### Checklist

### DIFF
--- a/code/client/clrcore-v2/Interop/LocalFunction.cs
+++ b/code/client/clrcore-v2/Interop/LocalFunction.cs
@@ -71,7 +71,7 @@ namespace CitizenFX.Core
 				return Coroutine<object>.Completed(result);
 			}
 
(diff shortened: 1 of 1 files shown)


niekschoemaker:
Another one which is somewhat opiniated.

In my opinion being able to call await on the export without a null check and then handling null values after the await seems like a cleaner solution than:
```cs
var result = Exports["test", "test"]();
if (result != null) {
    await result;
}
```

PR #2872, opened 2024-10-21
feat(net): Configurable peer timeout

tens0rfl0w:
### Goal of this PR
Clients that don't send or accept data for a longer period of time can have a negative impact on the gameplay behavior of specific game modes (especially PVP-focused), due to the nature of existing RAGE net code and missing patches for any server-authoritative control over synced entity data that is game-state critical (position correction, damage control system, etc.)

Some examples:
- Player 'teleporting': clients that reconnect before timing out do not get reset to the last known server position
- Dropped damage data: clients that don't receive game frame updates don't have their entity states updated afterward.
- ...

While the goal should be to have some sort of correction for these types of desync issues, this is probably not accomplishable in the short run. In the meantime, we can allow servers to define shorter (or longer) peer timeout values to at least give them some sort of countermeasure for the above-described problems, as the default value of 30 seconds is quite long.

This change introduces a new replicated ConVar named 'sv_peerTimeout' that allows to configure a per-server peer timeout, which can be set via the (server-sided) console command 'set_peer_timeout [timeout]'.

The (client-sided) enet timeout will get reset to the default value of 30 seconds after disconnect.

FiveM, Server, Net

### Successfully tested on

**Platforms:** Windows, Linux

### Checklist

### Fixes issues
/

### DIFF
--- a/code/components/citizen-server-impl/include/GameServer.h
+++ b/code/components/citizen-server-impl/include/GameServer.h
@@ -274,6 +274,12 @@ namespace fx
 
 		ServerInstanceBase* m_instance;
 
+		uint32_t m_peerTimeout = 30000;
+
+		std::shared_ptr<ConVar<uint32_t>> m_peerTimeoutVar;
+
+		std::shared_ptr<ConsoleCommand> m_peerTimeoutCmd;
+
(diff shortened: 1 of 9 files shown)


n3xuuu:
any news on this pr?

PR #2891, opened 2024-10-30
fix(scripting/exploit): Redefine os.system to prevent command execution on the server.

ekaliclua:
### Goal of this PR

The goal of this PR is to redefine the `os.execute` function to prevent it from being used to execute commands on the server. This is done to enhance security by ensuring that no arbitrary system commands can be run through this function.

This PR achieves the goal by redefining the `os.execute` function to raise an error with a message indicating that the function has been disabled for security reasons. This prevents any attempt to use `os.system` to execute commands.

- Server
- ScRT: Lua

### Successfully tested on

**Game builds:** Latest stable build

**Platforms:** Windows, Linux

### Checklist

### Fixes issues

/

### DIFF
--- a/data/shared/citizen/scripting/lua/scheduler.lua
+++ b/data/shared/citizen/scripting/lua/scheduler.lua
@@ -1115,3 +1115,8 @@ end
 if not isDuplicityVersion then
 	LocalPlayer = Player(-1)
 end
+
+-- Redefine os.execute to prevent command execution on the server
+function os.execute()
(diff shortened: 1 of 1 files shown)


FabianTerhorst:
Can you make it depend on a ConVar? We can't do this by default without breaking preexisting scripts.

p1u3o:
Can this not be tied to some kind of ace permission similar to ExecuteCommand?

Saying that, the current implementation of this can by bypassed by the various io.* functions

PR #2893, opened 2024-10-30
fix(server/state): properly reassign entities that are set to `EntityOrphanMode::KeepEntity`

AvarianKnight:
- this swaps `MoveEntityToCanidate` to use `ShouldServerKeepEntity` instead of having specific logic for orphan mode
- makes 'entitiesToDestroy' use `ShouldServerKeepEntity` instead of `IsOwnedByServerScript`so we reassign if orphan mode is set to keep the entity

### Goal of this PR
Properly reassign the entity to the server in certain situations where it would previously not happen

Add a new function that uses the original script check & entity orphan checks

Server

### Successfully tested on
**Platforms:** Windows

### Checklist

### DIFF
--- a/code/components/citizen-server-impl/include/state/ServerGameState.h
+++ b/code/components/citizen-server-impl/include/state/ServerGameState.h
@@ -963,6 +963,14 @@ public:
 		return scriptHash;
 	}
 
+	/// <summary>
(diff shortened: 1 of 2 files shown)


FabianTerhorst (approved):
👍

PR #2902, opened 2024-11-01
Minor fixes for trains, use `IsEntityValid` for other calls checking if the entity exists

AvarianKnight:
### Goal of this PR
Fix some oversights in some code regarding entity validation, trains, and how deletions for trains are handled, and how setting orphan mode is handled for trains

Make code that iterates over trains more generic so we can re-use it throughout more code.

Make calls that were previously just checking for `!entity` also check for if the entity is being deleted, fixes some cases

When called on trains `SET_ENTITY_ORPHAN_MODE` will also set the orphan mode for all trains attached, fixing the possibility of partial deletes from a train not being relevant.

Change `DELETE_ENTITY` to not allow it to be called on train carriages, as we don't have a way to write to the sync node of the current train that one of its links no longer exists, this could lead to broken state and possibly cause weird issues.

Add `DELETE_TRAIN` which when called on *any* part of the train will delete the entire train.

Server

### Successfully tested on
**Platforms:** Windows

### Checklist

### DIFF
--- a/code/components/citizen-server-impl/include/state/ServerGameState.h
+++ b/code/components/citizen-server-impl/include/state/ServerGameState.h
@@ -857,7 +857,6 @@ enum EntityOrphanMode : uint8_t
 	DeleteOnOwnerDisconnect = 1,
 	KeepEntity = 2,
 };
-
 struct SyncEntityState
 {
 	using TData = std::variant<int, float, bool, std::string>;
@@ -1357,6 +1356,10 @@ public:
 	
 	void GetFreeObjectIds(const fx::ClientSharedPtr& client, uint8_t numIds, std::vector<uint16_t>& freeIds);
 
+#ifdef STATE_FIVE
+	void IterateTrainLink(const sync::SyncEntityPtr& train, std::function<bool(sync::SyncEntityPtr&)> fn, bool callOnInitialEntity = true);
+#endif
+
 	void ReassignEntity(uint32_t entityHandle, const fx::ClientSharedPtr& targetClient, std::unique_lock<std::shared_mutex>&& lock = {});
 
 	bool SetEntityStateBag(uint8_t playerId, uint16_t objectId, std::function<std::shared_ptr<StateBag>()> createStateBag) override;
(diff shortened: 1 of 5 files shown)


AvarianKnight:
[train-test.zip](https://github.com/user-attachments/files/17601519/train-test.zip)

This was the script used to test this, making sure this doesn't effect migration is a bit harder.

code/components/citizen-server-impl/src/state/ServerGameState.cpp
```diff
@@ -2654,42 +2695,22 @@ void ServerGameState::ReassignEntity(uint32_t entityHandle, const fx::ClientShar
...
+		// game code works as follows:
+		// -> if train isEngine, enumerate the entire list backwards and migrate that one along
+		// -> if not isEngine, migrate the engine
+
+
+		// This call expects link-handle != entityHandle
+		// This will prevent
+		// 1. double-locking clientMutex
+		// 2. reassigning the same entity twice
+		this->IterateTrainLink(train, [=](const fx::sync::SyncEntityPtr& link) {
```
FabianTerhorst, line 2707: Redundant 'this->'. Same as in the other places where you have added it.

ext/native-decls/DeleteEntity.md
```diff
@@ -1,14 +1,16 @@
...
-
-```c
-void DELETE_ENTITY(Entity entity);
-```
-
-Deletes the specified entity.
-
-## Parameters
-* **entity**: The entity to delete.
+---
```
FabianTerhorst, line 1: Are the line changes caused by the encoding change?

code/components/citizen-server-impl/src/state/ServerGameState.cpp
```diff
@@ -2644,6 +2626,65 @@ void ServerGameState::ReassignEntityInner(uint32_t entityHandle, const fx::Clien
...
+			recurseTrain(initialTrain);
+		}
+		else if (trainState->engineCarriage && trainState->engineCarriage != initialTrain->handle)
+		{
+			if (auto engine = GetTrain(this, trainState->engineCarriage))
+			{
+				if (!fn(engine))
+				{
+					return;
+				};
```
FabianTerhorst, line 2681: ':' not needed here. Also add a new line after the '}'.

code/components/citizen-server-impl/include/state/ServerGameState.h
```diff
@@ -1371,7 +1375,49 @@ class ServerGameState : public ServerGameStatePublic, public fx::IAttached<fx::S
 	void ReassignEntityInner(uint32_t entityHandle, const fx::ClientSharedPtr& targetClient, std::unique_lock<std::shared_mutex>&& lock = {});
 
 public:
-	void DeleteEntity(const fx::sync::SyncEntityPtr& entity);
+
+	template<bool IgnoreTrainChecks = false>
+	void DeleteEntity(const fx::sync::SyncEntityPtr& entity)
+	{
+		if (entity->type != sync::NetObjEntityType::Player && entity->syncTree)
```
FabianTerhorst, line 1382: Do early return here to reduce the nesting.

ext/native-decls/DeleteEntity.md
AvarianKnight: Yea something weird with my editor I'll recommit with white space changes removed

AvarianKnight: Should be fixed

FabianTerhorst (approved):
👍  Thanks a lot for the contribution.

PR #2907, opened 2024-11-03
feat(core/five): sanitization for attachments on remote peds

tens0rfl0w:
### Goal of this PR
This PR introduces the ability to toggle checks that prevent entity-to-ped attachments on remotely owned peds. These checks are disabled by default in FiveM.

A new native function, `OnesyncEnableRemoteAttachmentSanitization`, has been added to allow toggling of these attachment checks at runtime.

The value resets to the default state ('off') upon player disconnect.

FiveM

### Successfully tested on
**Game builds:** 2545, 3258

### Checklist

### Fixes issues
Requested by @slashkeyvalue in: https://discord.com/channels/779705925577080842/779739506286002196/1301027817483538453 (Cfx.re Engineering Group)

### DIFF
--- a/code/components/gta-core-five/src/PatchAttachCD2545.cpp
+++ b/code/components/gta-core-five/src/PatchAttachCD2545.cpp
@@ -1,11 +1,12 @@
 #include <StdInc.h>
 #include <Hooking.h>
 #include "CrossBuildRuntime.h"
+#include <ScriptEngine.h>
+#include "ICoreGameInit.h"
 
 //
-// This patch forces a tunable (seen starting at b2545) preventing applying
-// of ped-to-ped attachment data to networked peds to be 'off', returning the
-// old behavior for compatibility.
+// This patches multiple checks that prevent the application of entity-to-ped
+// attachment data to networked peds so it can be dynamically toggled.
 //
 
 static bool ReturnFalse()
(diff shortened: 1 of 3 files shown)


slashkeyvalue:
Missed...?

code/components/gta-core-five/src/PatchAttachCD2545.cpp
```diff
@@ -1,32 +1,54 @@
...
 
-		auto attachEntityToEntityTunable = hook::get_pattern<char>("BA 37 89 3D 8A B9 BD C5 AF E3");
+		const auto attachEntityToEntityTunable = hook::get_pattern<char>("BA 37 89 3D 8A B9 BD C5 AF E3");
 		hook::call(attachEntityToEntityTunable + 23, ReturnFalse);
 	}
 });
+
+static InitFunction initFunction([]()
+{
+	fx::ScriptEngine::RegisterNativeHandler("ONESYNC_ENABLE_REMOTE_ATTACHMENT_SANITIZATION", [](fx::ScriptContext& context)
```
prikolium-cfx, line 44: What do you think if instead of adding new native function for each such flag we just add
SET_CFX_CONFIG_FLAG, GET_CFX_CONFIG_FLAG and RESET_CFX_CONFIG_FLAG (will reset to default) and there will be a service that will control such flags. We already have many of this, like FLY_THROUGH_WINDSHIELD, ragdoll on R and so on.
If you can't/don't want to do it - I will do it anyway, but can't say when

tens0rfl0w: Sounds good! Just to make sure: Are we still handling values internally with a ConVar then?

Like your example:
We keep `game_enableFlyThroughWindscreen` as a ConVar and allow interaction with the new exposed Setters/Getters? My only question then would be how we'd handle existing behavior for flags like this one, as they're currently treated as replicated ConVars which would allow no client-side set operation.

Will take a look!

prikolium-cfx: At least I want to see it consistent. Some of setters we have on server side, some on client. Let's add everything on server side and replicate if it already works like this

Pedro-Lucas14:
There are many cheaters using this feature that has been disabled. in the new versions fivem . Is there a date when this will go into production to fix this problem? @prikolium-cfx @tens0rfl0w

tens0rfl0w:
I understand that this one is a little bit more urgent. Will try to finish my proposal for the config flag system today👍🏼

Pedro-Lucas14:
Good thing, do you have a date for when it will go into production? @tens0rfl0w @prikolium-cfx

Pedro-Lucas14:
@iridium-cfx @dalekenium-cfx

Pedro-Lucas14:
Do you have any estimate for when this function will be implemented? We are trying to disable props on the server to prevent cheaters from teleporting players or locking them to the ground ?

prikolium-cfx (approved):

PR #2911, opened 2024-11-05
fix(net/five): correct train track node on ownership change

Ehbw:
### Goal of this PR

Fixes scenarios where a train becomes owned by the server (orphan natives) and upon a player re-entering the trains scope and  becoming owner of the train, The train resets to the node 0 of the track it is currently on

By forcing the train to find its current track node based on the trains current position, only upon ownership change to a player from a server and if the new client has no knowledge of the trains current track node.

FiveM

### Successfully tested on

**Game builds:** 1604, 2372, 2545, 2802, 3258

**Platforms:** Windows

### Checklist

### DIFF
--- a/code/components/gta-net-five/src/CloneObjectManager.cpp
+++ b/code/components/gta-net-five/src/CloneObjectManager.cpp
@@ -12,6 +12,11 @@
 #include <CloneManager.h>
 #include <CrossBuildRuntime.h>
 
+#include <netBlender.h>
+#include <netSyncTree.h>
+#include <EntitySystem.h>
+#include <netPlayerManager.h>
+
 static ICoreGameInit* icgi;
 
 extern void CD_AllocateSyncData(uint16_t objectId);
@@ -81,6 +86,20 @@ static void netObjectMgrBase__DestroyNetworkObject(rage::netObjectMgr* manager, 
 	}
 }
 
+#ifdef GTA_FIVE
+static int g_trainTrackNodeIndexOffset;
+
+static hook::cdecl_stub<void(CVehicle*, int, int)> CTrain__SetTrainCoord([]()
+{
+	return hook::pattern("44 8B C2 48 83 C4 ? 5B").count(1).get(0).get<void>(8);
+});
+
+static hook::cdecl_stub<bool(CVehicle*)> CTrain__IsCarriageEngine([]()
+{
+	return hook::get_call(hook::get_pattern("E8 ? ? ? ? 80 A3 ? ? ? ? ? 24 ? 02 C0 08 83 ? ? ? ? F6 83 ? ? ? ? ? 74 ? 8A 05"));
+});
+#endif
+
 static void(*g_orig_netObjectMgrBase__ChangeOwner)(rage::netObjectMgr*, rage::netObject*, CNetGamePlayer*, int);
 
 static void netObjectMgrBase__ChangeOwner(rage::netObjectMgr* manager, rage::netObject* object, CNetGamePlayer* targetPlayer, int migrationType)
@@ -96,6 +115,24 @@ static void netObjectMgrBase__ChangeOwner(rage::netObjectMgr* manager, rage::net
 	object->PostMigrate(migrationType);
 
 	CloneObjectMgr->ChangeOwner(object, oldOwnerId, targetPlayer, migrationType);
+	// Handle scenarios where a train was previously owned by the server
+#ifdef GTA_FIVE
+	// Make sure that this is a train and that we are now the new owner of it
+	if (object->objectType == (uint16_t)NetObjEntityType::Train && targetPlayer->physicalPlayerIndex() == rage::GetLocalPlayer()->physicalPlayerIndex())
+	{
+		if (CVehicle* train = (CVehicle*)object->GetGameObject())
+		{
+			// Ensure this is the engine and the client has no knowledge of the trains current track node.
+			if (CTrain__IsCarriageEngine(train) && *(int*)((uintptr_t)train + TrainTrackNodeIndexOffset) == 0)
+			{
+				//Find the trains track node based on its current location
+				CTrain__SetTrainCoord(train, -1, -1);
+				// Force blend to apply location
+				object->GetBlender()->Update();
+			}
+		}
+	}
+#endif
 }
 
 static rage::netObject* (*g_orig_netObjectMgrBase__GetNetworkObject)(rage::netObjectMgr* manager, uint16_t id, bool evenIfDeleting);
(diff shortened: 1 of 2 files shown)


Legacy-TacticalGamingInteractive:
Excellent

code/components/gta-net-five/src/CloneObjectManager.cpp
```diff
@@ -96,6 +110,25 @@ static void netObjectMgrBase__ChangeOwner(rage::netObjectMgr* manager, rage::net
...
+	// Handle scenarios where a train was previously owned by the server
+#ifdef GTA_FIVE
+	// Make sure that this is a train and that we are now the new owner of it
+	if (object->objectType == (uint16_t)NetObjEntityType::Train && targetPlayer->physicalPlayerIndex() == rage::GetLocalPlayer()->physicalPlayerIndex())
+	{
+		// Ensure that the vehicle isn't a nullptr
+		if (CVehicle* train = (CVehicle*)object->GetGameObject())
+		{
+			// Ensure this is the engine and the client has no knowledge of the trains current track node.
+			if (CTrain__IsCarriageEngine(train) && *(int*)((char*)train + TrainTrackNodeIndexOffset) == 0)
```
Nobelium-cfx, line 122: Nit: `(uintptr_t)train`

code/components/gta-net-five/src/CloneObjectManager.cpp
```diff
@@ -96,6 +110,25 @@ static void netObjectMgrBase__ChangeOwner(rage::netObjectMgr* manager, rage::net
...
+			{
+				//Find the trains track node based on its current location
+				CTrain__SetTrainCoord(train, -1, -1);
+				// Force blend to apply location
+				object->GetBlender()->m_30();
```
Nobelium-cfx, line 127: Nit: we can also rename `m_30()` to `Update()` in `netBlender`. So there is more context on what the function we are calling is about.

Nobelium-cfx (changes requested):
Thank you for your contribution! It looks like a really valuable change. Left some minor comments.

But also, do you have a good way to reproduce/test it? I tried to play with it locally:
- Put one player in a carriage of a train. Make sure that this player is the owner.
- Get second player on the train as well.
- See `netObjectMgrBase__ChangeOwner` being called, but the code never reaches the `CTrain__SetTrainCoord(train, -1, -1);` part. Seems like `(CTrain__IsCarriageEngine(train) && *(int*)((char*)train + TrainTrackNodeIndexOffset) == 0)` check is always false in my case.
- Disconnect the first player.
- After some time train disappears even tho second player in in a carriage.

Do I misunderstand the use case or this? Am I doing something wrong?

Ehbw:
> Thank you for your contribution! It looks like a really valuable change. Left some minor comments.
> 
> But also, do you have a good way to reproduce/test it? I tried to play with it locally:
> 
> * Put one player in a carriage of a train. Make sure that this player is the owner.
> * Get second player on the train as well.
> * See `netObjectMgrBase__ChangeOwner` being called, but the code never reaches the `CTrain__SetTrainCoord(train, -1, -1);` part. Seems like `(CTrain__IsCarriageEngine(train) && *(int*)((char*)train + TrainTrackNodeIndexOffset) == 0)` check is always false in my case.
> * Disconnect the first player.
> * After some time train disappears even tho second player in in a carriage.
> 
> Do I misunderstand the use case or this? Am I doing something wrong?

This issue isn't related to ownership changing between players as in most (if not all cases) the client becoming owner should already have knowledge of the track node the train is located on. However, in cases where the server was the previous owner (in cases where no player is in the trains scope to have ownership over the train) the client becoming owner will have no knowledge of the trains current track node. Causing the train to reset to node 0 regardless of where it actually is.
The issue resolved with this PR can be replicated with 1 player. 

* Create a train doesn't matter the track location or train config. Applying the server-native SET_ENTITY_ORPHAN_MODE on the engine and all of it's carriages (this is simplified in #2902)
* Once the train begins moving, move the player far enough away from the train to where the player leaves the trains scope (trains have a much larger scope then normal entities, the entity culling natives can be used here to make the scope smaller and make it easier to repro)
* When re-entering the trains scope. Without this patch the train will begin calculating its next position from node 0 of the track it is on (usually 0 for metro lines, 3 for main train tracks) causing the train to teleport to node 0. With this patch, the train will continuing moving along based on the current track node it is located at.

I've also attached the repro resource i used to test this. (find a train track, do /train on the client to create a train. Leave the scope and return)
[trainnode-repro.zip](https://github.com/user-attachments/files/17679379/trainnode-repro.zip)

code/components/gta-net-five/src/CloneObjectManager.cpp
```diff
@@ -81,6 +81,20 @@ static void netObjectMgrBase__DestroyNetworkObject(rage::netObjectMgr* manager,
 	}
 }
 
+#ifdef GTA_FIVE
+static int TrainTrackNodeIndexOffset;
+
+static hook::cdecl_stub<void(CVehicle*, int, int)> CTrain__SetTrainCoord([]()
```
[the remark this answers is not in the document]
Ehbw: Fixed now, some how forgot to commit that

Nobelium-cfx (approved):
I've tested it locally and can confirm that it works and makes the difference. Thank you for the great job!

PS: For some reason couldn't trigger train owner change with just one player - everything worked even without fix. But if I use one player to spawn the train, disconnect and use second to fly close to the train - then without the fix train disappears, while with the fix it stays where it should.

Ehbw:
>For some reason couldn't trigger train owner change with just one player - everything worked even without fix. But if I use one player to spawn the train, disconnect and use second to fly close to the train - then without the fix train disappears, while with the fix it stays where it should.

There was some instances when testing this patch where I had to decrease the trains Culling distance to 424 (onesync's default scope range).  Since even going 4000+ units away from the train's position would still cause it to be relevant for the owning client. I haven't had time to investigate why this would happen though and it does seem to be inconsistent.

Though from testing and feedback from others using trains in a production environment. Trains behave very weirdly in OneSync / GTA:O especially with ownership reassignment (issues such as randomly detaching train carriages, negative speeds breaking train blending). Hoping to address some more of these issues in the future.

Legacy-TacticalGamingInteractive:
@Ehbw for spawning trains will we have to do anything differently now to be able to utilize this commit or not really? Currently on client we run this snippet for NPC trains. We use another different system for drivable trains though.

```
		SwitchTrainTrack(0, true) -- Setting the Main train track(s) around LS and towards Sandy Shores active
		SwitchTrainTrack(3, true) -- Setting the Metro tracks active
		SetTrainTrackSpawnFrequency(0, 720000) -- The Train spawn frequency
		SetTrainTrackSpawnFrequency(3, 720000) -- The Metro spawn frequency
		SetRandomTrains(true) -- Telling the game we want to use randomly spawned trains
		SetTrainsForceDoorsOpen(false)
end)
```

Ehbw:
> @Ehbw for spawning trains will we have to do anything differently now to be able to utilize this commit or not really? Currently on client we run this snippet for NPC trains. We use another different system for drivable trains though.
> 
> ```
> Citizen.CreateThread(function()
> 		SwitchTrainTrack(0, true) -- Setting the Main train track(s) around LS and towards Sandy Shores active
> 		SwitchTrainTrack(3, true) -- Setting the Metro tracks active
> 		SetTrainTrackSpawnFrequency(0, 720000) -- The Train spawn frequency
> 		SetTrainTrackSpawnFrequency(3, 720000) -- The Metro spawn frequency
> 		SetRandomTrains(true) -- Telling the game we want to use randomly spawned trains
> 		SetTrainsForceDoorsOpen(false)
> end)
> ```

If you are already using the server orphan natives. No.

PR #2915, opened 2024-11-07
tweak(legacy-net/resources): check for gameinit on server-command

tens0rfl0w:
### Goal of this PR
This ensures that the "same thread" constraint in SendNetPacket stays valid. (We don't need to send server-commands before full game init. anyway).

Blocking server-commands before we are fully initialized (Render->MainThrd switch).

FiveM

### Successfully tested on

### Checklist

### Fixes issues
/

### DIFF
--- a/code/components/citizen-legacy-net-resources/src/ResourceNetBindings.cpp
+++ b/code/components/citizen-legacy-net-resources/src/ResourceNetBindings.cpp
@@ -775,10 +775,19 @@ void NetLibraryResourcesComponent::AttachToObject(NetLibrary* netLibrary)
 		});
 	});
 
+	static bool gameLoaded = false;
+
+	Instance<ICoreGameInit>::Get()->OnGameFinalizeLoad.Connect([]()
+	{
+		gameLoaded = true;
+	});
+
 	static bool inSessionReset = false;
 
 	Instance<ICoreGameInit>::Get()->OnShutdownSession.Connect([]()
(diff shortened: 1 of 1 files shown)


Nobelium-cfx:
What will be the visible effect of this change? Do we currently see any issues related to this?

But overall looks reasonable to me, but I'm not 100% confident. @FabianTerhorst can you take a quick look as well?

tens0rfl0w:
> What will be the visible effect of this change? Do we currently see any issues related to this?

Recent changes to `SendNetPacket` now include a thread ID check to ensure it isn't called from multiple threads. However, `CS_ACTIVE` may be set before the main game thread is fully initialized. As a result, if a command is entered during game initialization (e.g., loadingscreen), the `SendNetPacket` call can be triggered from the Render thread rather than the Main thread, causing this constraint to fail.

Since sending server commands isn't necessary until the game is fully loaded, adding a guard to delay these calls until the main thread is ready should resolve the issue and maintain the thread safety constraint.

FabianTerhorst:
> What will be the visible effect of this change? Do we currently see any issues related to this?
> 
> But overall looks reasonable to me, but I'm not 100% confident. @FabianTerhorst can you take a quick look as well?

This PR just prevents the warning print.  Because the check inside the Send method prevents the unsafe send.

Nobelium-cfx (approved):

PR #2916, opened 2024-11-08
feat(extra-natives/five): train station natives

Ehbw:
### Goal of this PR

Allows for mission trains (trains created with CREATE_MISSION_TRAIN) to stop at stations. Enabling for the same behaviour as non-mission trains. This removes the need for scripts to implement their own logic which may not play well with multiple players/owners 

By creating SET_TRAIN_STOP_AT_STATIONS which allows ScRT to directly toggle a flag deciding if a train should stop at stations

FiveM, Natives

### Successfully tested on

**Game builds:**  1604, 2372, 2699, 3258

**Platforms:** Windows

### Checklist

### DIFF
--- a/code/components/extra-natives-five/src/VehicleExtraNatives.cpp
+++ b/code/components/extra-natives-five/src/VehicleExtraNatives.cpp
@@ -316,6 +316,7 @@ static int VehicleDamageStructOffset;
 static bool* g_trainsForceDoorsOpen;
 static int TrainDoorCountOffset;
 static int TrainDoorArrayPointerOffset;
+static int TrainFlagOffset;
 
 static int VehicleRepairMethodVtableOffset;
 
@@ -588,6 +589,7 @@ static HookFunction initFunction([]()
 		WheelHealthOffset = *hook::get_pattern<uint32_t>("75 24 F3 0F 10 ? ? ? 00 00 F3 0F", 6);
 		LightMultiplierGetOffset = *hook::get_pattern<uint32_t>("00 00 48 8B CE F3 0F 59 ? ? ? 00 00 F3 41", 9);
 		VehicleRepairMethodVtableOffset = *hook::get_pattern<uint32_t>("C1 E8 19 A8 01 74 ? 48 8B 81", -14);
+		TrainFlagOffset = *hook::get_pattern<uint32_t>("80 8B ? ? ? ? ? 8B 05 ? ? ? ? FF C8", 2);
 	}
 
 	{
@@ -1465,6 +1467,10 @@ static HookFunction initFunction([]()
 
 		GetTrainDoor(train, doorIndex)->ratio = ratio;
 	}));
+
+	fx::ScriptEngine::RegisterNativeHandler("SET_TRAIN_STOP_AT_STATIONS", std::bind(writeVehicleMemoryBit<&TrainFlagOffset, 4>, _1, "SET_TRAIN_STOP_AT_STATIONS"));
+
+	fx::ScriptEngine::RegisterNativeHandler("DOES_TRAIN_STOP_AT_STATIONS", std::bind(readVehicleMemoryBit<&TrainFlagOffset, 4>, _1, "DOES_TRAIN_STOP_AT_STATIONS"));
 
 	fx::ScriptEngine::RegisterNativeHandler("GET_VEHICLE_WHEEL_X_OFFSET", makeWheelFunction([](fx::ScriptContext& context, fwEntity* vehicle, uintptr_t wheelAddr)
 	{
--- a/ext/native-decls/DoesTrainStopAtStations.md
+++ b/ext/native-decls/DoesTrainStopAtStations.md
@@ -1,0 +1,16 @@
+---
+ns: CFX
+apiset: client
+game: gta5
+---
+## DOES_TRAIN_STOP_AT_STATIONS
+
+```c
+BOOL DOES_TRAIN_STOP_AT_STATIONS(Vehicle train);
+```
+
+## Parameters
+* **train**: The train handle
+
+## Return value
+True if the train stops at stations. False otherwise
--- a/ext/native-decls/SetTrainStopAtStations.md
+++ b/ext/native-decls/SetTrainStopAtStations.md
@@ -1,0 +1,16 @@
+---
+ns: CFX
+apiset: client
+game: gta5
+---
+## SET_TRAIN_STOP_AT_STATIONS
+
+```c
+void SET_TRAIN_STOP_AT_STATIONS(Vehicle train, BOOL state);
+```
+
+Toggles a train's ability to stop at stations
+
+## Parameters
+* **train**: The train handle
+* **state**: True to make the train stop at stations. False to make the train not stop at stations


code/components/extra-natives-five/src/VehicleExtraNatives.cpp
```diff
@@ -316,6 +316,7 @@ static int VehicleDamageStructOffset;
 static bool* g_trainsForceDoorsOpen;
 static int TrainDoorCountOffset;
 static int TrainDoorArrayPointerOffset;
+static int g_trainFlagOffset;
```
Nobelium-cfx, line 319: Nit: while the global style guide is `g_trainFlagOffset` - keeping consistent style inside a single file is a priority. Lets use `TrainFlagOffset` in this case :).

ext/native-decls/SetTrainStopAtStations.md
```diff
@@ -0,0 +1,18 @@
...
+```c
+void SET_TRAIN_STOP_AT_STATIONS(Vehicle train, BOOL state)
+
+```
+
+Toggles a train's ability to stop at stations
+
+## Parameters
+* **vehicle**: The train handle
+* **state**: True to make the train stop at stations. False to make train resume default behavior
```
Nobelium-cfx, line 18: Nit: I assume theoretically this native can also be used with regular trains, not only the `CREATE_MISSION_TRAIN` ones. E.g. to make those trains not stop at the station.

Lets change the wording to remove the `resume default behavior` part. E.g. to something like `True to make the train stop at stations. False to make the train go without stops`

ext/native-decls/DoesTrainStopAtStations.md
```diff
@@ -0,0 +1,18 @@
+---
+ns: CFX
+apiset: client
+game: gta5
+---
+## DOES_TRAIN_STOP_AT_STATIONS
+
+
+```c
+void DOES_TRAIN_STOP_AT_STATIONS(Vehicle train)
```
Nobelium-cfx, line 10: Return value should be `BOOL` not `void`

Nobelium-cfx, line 10: Missing `;` natives won't compile. Same for `void SET_TRAIN_STOP_AT_STATIONS(Vehicle train, BOOL state)`

PS: such errors are usually visible when you run `.\prebuild.cmd`. E.g
```
25/809 - CREATE_RUNTIME_TEXTURE_FROM_DUI_HANDLE                                                     Error: C code snippet contained wrong amount of functions.
```

ext/native-decls/DoesTrainStopAtStations.md
```diff
@@ -0,0 +1,18 @@
...
+```
+
+
+## Parameters
+* **vehicle**: The train handle
```
Nobelium-cfx, line 15: Should be `train`. Same in `SET_TRAIN_STOP_AT_STATIONS`

code/components/extra-natives-five/src/VehicleExtraNatives.cpp
```diff
@@ -1466,6 +1468,14 @@ static HookFunction initFunction([]()
 		GetTrainDoor(train, doorIndex)->ratio = ratio;
 	}));
 
+	fx::ScriptEngine::RegisterNativeHandler("SET_TRAIN_STOP_AT_STATIONS", makeTrainFunction([](fx::ScriptContext& context, fwEntity* train)
+	{
+		bool state = context.GetArgument<bool>(1);
+		*(int*)((uintptr_t)train + g_trainFlagOffset) |= 16 * (state & 1);
```
Nobelium-cfx, line 1474: This won't work to set the state to false.

You can use something similar to how we handle `vehicleTankEmptyFlags`:
```
		uint8_t* trainFlags = (uint8_t*)((uintptr_t)train + g_trainFlagOffset);
		*trainFlags ^= (*trainFlags ^ (uint8_t)state) & 1;
```

Nobelium-cfx (changes requested):
For some reason the fix doesn't work locally for me. I use the `train` script from your previous PR and add
```
    print(DoesTrainStopAtStations(train))
    SetTrainStopAtStations(true)
    print(DoesTrainStopAtStations(train))
```
To it. Both `DoesTrainStopAtStations` calls return False and the train doesn't stop at the station (I test it around Puerto Del Sol).

Not sure yet what could be the issue here, from the look of the code it should work.

Does it work for you locally? Maybe some changes are missing?

ext/native-decls/DoesTrainStopAtStations.md
Ehbw: I haven't been able to get prebuild.cmd to work on my machine. But corrected the issues in both native declaration and made sure prebuild.cmd was able to compile the natives on another machine

code/components/extra-natives-five/src/VehicleExtraNatives.cpp
Ehbw: This file is a bit of a mess in terms of style. But I've changed it to TrainFlagOffset

Ehbw:
> For some reason the fix doesn't work locally for me. I use the `train` script from your previous PR and add
> 
> ```
>     print(DoesTrainStopAtStations(train))
>     SetTrainStopAtStations(true)
>     print(DoesTrainStopAtStations(train))
> ```
> 
> To it. Both `DoesTrainStopAtStations` calls return False and the train doesn't stop at the station (I test it around Puerto Del Sol).
> 
> Not sure yet what could be the issue here, from the look of the code it should work.
> 
> Does it work for you locally? Maybe some changes are missing?

It works locally for me. Even when doing it on train creation (using the same resource from my previous PR)

Using it in a command also works for me (
```lua
--- /setStopsAtStation 1
RegisterCommand("setStopsAtStation", function(_, args)
    local value = tonumber(args[1]) or 0
    print(DoesTrainStopAtStations(train))
    SetTrainStopAtStations(train, value == 1)
    print(DoesTrainStopAtStations(train))
end, false)
```

code/components/extra-natives-five/src/VehicleExtraNatives.cpp
Nobelium-cfx: I actually found that we have a function to set a vehicle bit. So we can replace this native implementation with:
```
fx::ScriptEngine::RegisterNativeHandler("SET_TRAIN_STOP_AT_STATIONS", std::bind(writeVehicleMemoryBit<&g_trainFlagOffset, 4>, _1, "SET_TRAIN_STOP_AT_STATIONS"));
```

It looks better that way.

Nobelium-cfx:
> It works locally for me. Even when doing it on train creation (using the same resource from my previous PR)
> 
> Using it in a command also works for me (
> 
> ```lua
> --- /setStopsAtStation 1
> RegisterCommand("setStopsAtStation", function(_, args)
>     local value = tonumber(args[1]) or 0
>     print(DoesTrainStopAtStations(train))
>     SetTrainStopAtStations(train, value == 1)
>     print(DoesTrainStopAtStations(train))
> end, false)
> ```

The new `SET_TRAIN_STOP_AT_STATIONS` implementation works fine for me. I'm not sure whats the difference was, but I guess it doesn't matter so much.

Left a comment tho - I think it better to reuse existing `writeVehicleMemoryBit` function. I've tested it with `writeVehicleMemoryBit` as well - as it also works just fine. Please apply that and we can set it as ready to merge :)

Ehbw:
> > It works locally for me. Even when doing it on train creation (using the same resource from my previous PR)
> > Using it in a command also works for me (
> > ```lua
> > --- /setStopsAtStation 1
> > RegisterCommand("setStopsAtStation", function(_, args)
> >     local value = tonumber(args[1]) or 0
> >     print(DoesTrainStopAtStations(train))
> >     SetTrainStopAtStations(train, value == 1)
> >     print(DoesTrainStopAtStations(train))
> > end, false)
> > ```
> 
> The new `SET_TRAIN_STOP_AT_STATIONS` implementation works fine for me. I'm not sure whats the difference was, but I guess it doesn't matter so much.

I believe you were missing the train argument (at least from the snippet you posted in the first comment).

> 
> Left a comment tho - I think it better to reuse existing `writeVehicleMemoryBit` function. I've tested it with `writeVehicleMemoryBit` as well - as it also works just fine. Please apply that and we can set it as ready to merge :)

Done

Nobelium-cfx:
> I believe you were missing the train argument (at least from the snippet you posted in the first comment).

Oh, I think you are right ^^. My bad, sorry.

ext/native-decls/DoesTrainStopAtStations.md
```diff
@@ -0,0 +1,16 @@
...
+## DOES_TRAIN_STOP_AT_STATIONS
+
+```c
+BOOL DOES_TRAIN_STOP_AT_STATIONS(Vehicle train);
+```
+
+## Parameters
+* **train**: The train handle
+
+# Return value
```
Nobelium-cfx, line 15: Nit: Should be `##`

Nobelium-cfx (approved):
LGTM. Thank you for the contribution! ^^

PR #2918, opened 2024-11-09
feat(mono-rt2): allow default parameters for dynfunc

DaniGP17:
### Goal of this PR
This allows parameters to have default values ​​for DynFuncs

### Example
```cs
[EventHandler("testing_event", Binding.Local)]
void TestEvent(int a, string b = "default value")
{
    Debug.WriteLine("Test func a="+a+", b="+b);
}
```

Declaring at the beginning a local variable to store the size of the argument array. Within the for, if the index of the current parameter is greater than the size of the array arguments, it will use the optional parameter. To do this I have seen the implementation that was made at https://github.com/citizenfx/msgpack-cs/blob/master/MsgPack/MsgPackFunc.cs#L138

ScRT: C#

### Successfully tested on

**Game builds:** ..

**Platforms:** Windows

### Checklist

### DIFF
--- a/code/client/clrcore-v2/DynFunc.cs
+++ b/code/client/clrcore-v2/DynFunc.cs
@@ -114,8 +114,6 @@ namespace CitizenFX.Core
 				return (DynFunc)existingMethod.CreateDelegate(typeof(DynFunc), target);
 			}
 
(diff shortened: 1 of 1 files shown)


DaniGP17:
Closing this as https://github.com/citizenfx/msgpack-cs/pull/2 get merged and it already contains the support for default params.

PR #2920, opened 2024-11-09
Soft dependency support for resources

r3ps4J:
### Goal of this PR
This PR adds support for soft dependencies, indicated by the manifest entries `soft_dependency` and `soft_dependencies`. Soft dependencies only get started if they exist, making it possible to add optional dependencies to a resource. This is useful when a resource can work fine without a dependency, but can benefit from having access to that dependency.

See the proposal on the forum for more details: https://forum.cfx.re/t/soft-dependencies-for-resources/5281288/1

This PR implements soft dependencies by loading them the same way as normal dependencies, but if one is missing or fails to start it skips it and continues to the next one. It also still prints the same information as with regular dependencies, and only tries to load the soft dependencies after the hard dependencies have been checked and loaded to prevent starting them when the resource can't start anyway.

This PR specifically adds the `const bool soft = false` parameter to the `loadDeps` function. The paramter has a default value of false, to remain backwards compatible. When `soft` is set to `true` the following things are different as opposed to loading hard dependencies:
- Constraints are not checked, as having optional constraints makes little sense.
- When a soft dependency is not found, the loop continues instead of returning `false` early to stop the resource from loading.
- The soft dependency does not get stored in `resourceDependants`, as stopping the soft dependency should not prevent the resource from running.
- When a soft dependency fails to start, the loop continues instead of returning `false` early to stop the resource from loading and storing the dependency in `resourceDependencies` to start it later.

This applies to the server and clients. The changes were made in `citizen-resources-core`.

### Successfully tested on
The changes were tested on the server and FiveM client.

**Game builds:** b3095

**Platforms:** Windows

### Checklist

### Fixes issues
None.

### DIFF
--- a/code/components/citizen-resources-core/src/ResourceDependencyLoader.cpp
+++ b/code/components/citizen-resources-core/src/ResourceDependencyLoader.cpp
@@ -33,10 +33,12 @@ static InitFunction initFunction([]()
 			bool success = true;
 			std::vector<std::string> errorList;
 
-			auto loadDeps = [&] (const std::string& type)
+			auto loadDeps = [&] (const std::string& type, const bool soft = false)
 			{
 				for (const auto& dependency : metaData->GetEntries(type))
 				{
+					// Skip constraints when handling soft dependencies, as these can't be soft.
+					if (!soft)
 					{
 						std::string constraintError;
 						auto constraintResult = constraintsComponent->MatchConstraint(dependency.second, &constraintError);
@@ -57,12 +59,22 @@ static InitFunction initFunction([]()
 
 					if (!other.GetRef())
 					{
-						trace("Could not find dependency %s for resource %s.\n", dependency.second, resource->GetName());
+						trace("Could not find %s %s for resource %s.\n", soft ? "soft dependency" : "dependency", dependency.second, resource->GetName());
+
+						// Continue to the next dependency if we're handling soft dependencies, as these dependencies are optional.
+						if (soft)
+						{
+							continue;
+						}
+
 						return false;
 					}
 
-					// store in a list for use in OnStop
-					resourceDependants.insert({ other->GetName(), resource->GetName() });
+					if (!soft)
+					{
+						// store in a list for use in OnStop
+						resourceDependants.insert({ other->GetName(), resource->GetName() });
+					}
 
 					if (other->GetState() == fx::ResourceState::Starting || other->GetState() == fx::ResourceState::Started)
 					{
@@ -73,7 +85,13 @@ static InitFunction initFunction([]()
 
 					if (!success)
 					{
-						trace("Could not start dependency %s for resource %s.\n", dependency.second, resource->GetName());
+						trace("Could not start %s %s for resource %s.\n", soft ? "soft dependency" : "dependency", dependency.second, resource->GetName());
+
+						// Continue to the next dependency if we're handling soft dependencies, just like we do when one is missing.
+						if (soft)
+						{
+							continue;
+						}
 
 						// store for deferred use (e.g. build systems)
 						resourceDependencies.insert({ other->GetName(), resource->GetName() });
@@ -103,6 +121,10 @@ static InitFunction initFunction([]()
 
 				return false;
 			}
+
+			// Load soft dependencies defined with `soft_dependency` or `soft_dependencies`.
+			// We ignore the result, as we don't want to cancel loading the resource if a soft dependency is missing.
+			loadDeps("soft_dependency", true) && loadDeps("soft_dependencie", true); // soft_dependencies without s
 
 			return true;
 		}, -9999);


r3ps4J:
Okay I wanted to rename the branch on my repo, but github doesn't seem to like it.

Nobelium-cfx:
Thank you for the contribution!

I understand the motivation behind this PR and the code looks reasonable. But I don't have enough context to make the final call on if it's actually needed. @prikolium-cfx WDYT?

prikolium-cfx:
> Thank you for the contribution!
> 
> I understand the motivation behind this PR and the code looks reasonable. But I don't have enough context to make the final call on if it's actually needed. @prikolium-cfx WDYT?

Should be discussed together with team. Can't say right now pros and cons of such decision

r3ps4J:
Sounds good! I added some quotes around the resource names to clarify what is and what is not part of the resource names. This also makes it match the other traces in the dependency loader.

r3ps4J:
@prikolium-cfx any updates on this?

code/components/citizen-resources-core/src/ResourceDependencyLoader.cpp
```diff
@@ -57,12 +59,22 @@ static InitFunction initFunction([]()
 
 					if (!other.GetRef())
 					{
-						trace("Could not find dependency %s for resource %s.\n", dependency.second, resource->GetName());
+						trace("Could not find %s '%s' for resource '%s'.\n", soft ? "soft dependency" : "dependency", dependency.second, resource->GetName());
```
prikolium-cfx, line 62: Why do we notify user about missed soft dependency?

code/components/citizen-resources-core/src/ResourceDependencyLoader.cpp
```diff
@@ -73,7 +85,13 @@ static InitFunction initFunction([]()
 
 					if (!success)
 					{
-						trace("Could not start dependency %s for resource %s.\n", dependency.second, resource->GetName());
+						trace("Could not start %s '%s' for resource '%s'.\n", soft ? "soft dependency" : "dependency", dependency.second, resource->GetName());
```
prikolium-cfx, line 88: Same here

code/components/citizen-resources-core/src/ResourceDependencyLoader.cpp
```diff
@@ -104,6 +122,10 @@ static InitFunction initFunction([]()
 				return false;
 			}
 
+			// Load soft dependencies defined with `soft_dependency` or `soft_dependencies`.
+			// We ignore the result, as we don't want to cancel loading the resource if a soft dependency is missing.
+			loadDeps("soft_dependency", true) && loadDeps("soft_dependencie", true); // soft_dependencies without s
```
prikolium-cfx, line 127: Please leave only 1 way of setting soft dependencies

r3ps4J: This is the same way but the plural form. Just like is done for regular dependencies. You've got `dependency` and `dependencies`. Normally you use `server_script` and it would append an s, resulting in the correct `server_scripts`. But since the plural of dependency is dependencies, we need to use `dependencie` as well.

This same principle was applied here for `soft_depenency` and `soft_dependencie` resulting in `soft_dependencies`.

prikolium-cfx: > This is the same way but the plural form. Just like is done for regular dependencies. You've got `dependency` and `dependencies`. Normally you use `server_script` and it would append an s, resulting in the correct `server_scripts`. But since the plural of dependency is dependencies, we need to use `dependencie` as well.
> 
> This same principle was applied here for `soft_depenency` and `soft_dependencie` resulting in `soft_dependencies`.

To be honest today we discussed with team this naming and soft_dependency doesn't make much sense and possible name for that field that we found would be `load_after` so it's not confusing creators. Probably you can suggest better name

code/components/citizen-resources-core/src/ResourceDependencyLoader.cpp
r3ps4J: So they know why something is not working even though they think it will. Say your resource uses ox_target and registers eye options if it's present. If it has been renamed or something users might expect it to still work but then they can see this message and investigate.

Normal depencies also log when they don't get found and then it proceeds to cancel starting the resource. This is pretty much the same except that it doesn't cancel starting the resource, but still logs that it's missing something.

code/components/citizen-resources-core/src/ResourceDependencyLoader.cpp
r3ps4J: See other comment

r3ps4J: See other comment. Here it's even more important to log. If a soft dependency is present people will expect it to work, but if it can't be started for some reason they should be informed about that.

code/components/citizen-resources-core/src/ResourceDependencyLoader.cpp
r3ps4J: `load_after` makes very little sense to me as it just implies that the resource should load after another one. The whole point of this PR is to introduce a soft dependency as is used in many other types of software. It also starts the soft dependency if it is found while load after implies that it starts this one if and only if the other one is already started. That's not what this feature does so it makes very little sense to me to name it that.

I think load_after would create more confusion for developers. Most developers should already be familiar with how soft dependencies work as well for example from npm packages, php or minecraft plugins. But also many many other software systems.

Could you explain why it doesn't make much sense to you? As a software engineer I've seen this type of dependency used a lot.

Here is a link to how it works in magento, maybe that makes it more clear for you: https://developer.adobe.com/commerce/php/architecture/modules/dependencies/

Please also take the time to read the full proposal in the forum post, as it explains way more than what I've said here: https://forum.cfx.re/t/soft-dependencies-for-resources/5281288/1

r3ps4J:
Hey @prikolium-cfx @Nobelium-cfx and the rest of the team! Just checking in since it's been a while. Is this ever going to get merged?

prikolium-cfx:
> Hey @prikolium-cfx @Nobelium-cfx and the rest of the team! Just checking in since it's been a while. Is this ever going to get merged?

Current manifest system is already overcomplicated and such change will make it more confusing. Thanks for your efforts, but we won't accept such change.

r3ps4J:
> > Hey @prikolium-cfx @Nobelium-cfx and the rest of the team! Just checking in since it's been a while. Is this ever going to get merged?
> 
> Current manifest system is already overcomplicated and such change will make it more confusing. Thanks for your efforts, but we won't accept such change.

That's a shame, I feel like it would make it a lot easier for developers. Why do you think adding this would make it more confusing?

MrTurtles:
@r3ps4J @prikolium-cfx
This pull request directly addresses a real pain point many FiveM creators are experiencing. From a developer perspective, it improves workflow and removes unnecessary friction without breaking existing behavior. This is exactly what we need.

I’d strongly encourage a technical review of the implementation itself rather than dismissing the use case. This would be a meaningful quality-of-life improvement for a large portion of the community.

I had to rewrite my comment using AI because my anger wrote a different kind of message. Again, please review this pull request.

lLukDreams:
@MrTurtles soft depencies doesnt make sense, either you need or not. With a good documentation you solve most of the issues, people are not stupid and this is just encouraging stupid behaviour

MrTurtles:
@lLukDreams, I think you're missing the technical necessity of load-order management.

This isn't a documentation issue, it’s a resource graph issue. Documentation can tell a human what to do, but it cannot tell the server engine how to prioritize initialization. Currently, if resource B has an optional integration for resource A, there is no way to programmatically ensure resource A starts first without making it a hard requirement.

Without soft dependencies, we are forced to rely on users perfectly ordering their server.cfg. If a user makes a manual mistake and resource B starts first, its GetResourceState or export checks will fail during initialization. Even if resource A is present, the "integration" is already broken because the timing was wrong.

Soft dependencies aren't "encouraging stupid behavior", they are a standard industry tool (used in NPM, Cargo, and Linux package managers) to handle modularity. They allow a single script to support multiple frameworks or optional add-ons (like ox_target vs qb-target) without forcing a "one-size-fits-all" installation or breaking the startup sequence.

Dismissing a request for better automation as "solving for stupid behavior" ignores the fundamental goal of good software: to be robust by design. We should be building a platform that handles complexity for the creator, not one that relies on manual workarounds.

lLukDreams:
@MrTurtles since you used ai to improve your answer, let the war start: 
I think you’re overloading load-order convenience into a feature that changes the dependency model in unsafe ways.

A few core issues:

1. This blurs the meaning of “dependency”
If a resource’s behavior meaningfully changes based on another resource being present at startup, that is a dependency. Making it “soft” doesn’t change the coupling, it just hides it from the engine’s dependency graph. At that point the system can no longer reason correctly about resource relationships.

Package managers like npm or Cargo still resolve optional dependencies explicitly and deterministically. They don’t silently skip failed deps at runtime — that’s a key difference.

2. Skipping failed starts creates undefined runtime states
With soft dependencies:

A resource may partially initialize assuming an integration exists

The dependency might start later

No re-initialization occurs unless authors manually handle it

That pushes complexity into every resource author, forcing them to implement state reconciliation, event listeners, retries, etc. The engine no longer guarantees a consistent startup contract.

Hard deps fail fast. Soft deps fail silently. Silent failure is worse.

3. Load-order problems are being solved at the wrong layer
The real issue you’re describing is conditional integration, not dependency resolution.

A better pattern is:

Declare hard dependencies for anything required at startup

Detect optional integrations at runtime via events or explicit “ready” signals

React dynamically when those integrations become available

That keeps the dependency graph honest while still allowing modular behavior.

4. “User misordering server.cfg” is not a justification
The engine already has a dependency solver. If people bypass it by not declaring deps, that’s a configuration error — not a platform deficiency. Designing around misconfiguration by weakening guarantees leads to fragile systems.

Robust software doesn’t mean “never error”; it means “error early, clearly, and predictably.”

5. This encourages ambiguous resource design
Soft deps make it unclear:

What functionality is guaranteed

What is optional

What state the resource is in after startup

That ambiguity makes debugging, maintenance, and tooling harder — especially for larger servers.

In short:
This PR trades deterministic startup guarantees for convenience, and shifts complexity from the engine (where it belongs) into userland scripts (where it will be inconsistently implemented).

Optional integrations are valid.
Optional dependencies at load time are not.

If the goal is better modularity, the solution should strengthen lifecycle signaling — not weaken dependency semantics.

r3ps4J:
Why is there a war here lmao?

And man do I hate AI, this is just difficult to read through already. But okay let me just clarify what is what. It seems to me like @MrTurtles ran into a similar problem I had when I thought of this and found this PR. Although his wording was changed with AI which shows, he does get the point of my PR.

> @MrTurtles soft depencies doesnt make sense, either you need or not. With a good documentation you solve most of the issues, people are not stupid and this is just encouraging stupid behaviour

@lLukDreams there are actually use cases as I explain in this PR and the proposal on the forum. With documentation you can tell server owners the load order sure, but you cant enforce that resources run before yours if and only if they are present. I don't think this encourages stupid behaviour, its used alot in software development and makes it easier for developers. And you also don't need to use it if you don't want to? So in my opinion it would just be nice to have available.

Not gonna get too deep into the rest because AI.
 
> This blurs the meaning of “dependency”
If a resource’s behavior meaningfully changes based on another resource being present at startup, that is a dependency. Making it “soft” doesn’t change the coupling, it just hides it from the engine’s dependency graph. At that point the system can no longer reason correctly about resource relationships.

What? No? If a dependency is not present the resource won't work. That's a dependency.

> Package managers like npm or Cargo still resolve optional dependencies explicitly and deterministically. They don’t silently skip failed deps at runtime — that’s a key difference.

Did you read the PR? Or the code? Because this doesn't silently skip failed soft depencies. It does exactly what you say it doesn't...

The rest also seems incorrect.

> This PR trades deterministic startup guarantees for convenience, and shifts complexity from the engine (where it belongs) into userland scripts (where it will be inconsistently implemented).

What so my PR shifts the complexity away from the engine, aka the place it belongs according to you, even though I add it there? And you suggested users should implement it and now you say will be inconsistently implemented. So you just gave a valid argument yourself for why it would be useful to have soft dependencies. Thanks I guess.

PR #2923, opened 2024-11-11
fix(gamestate/server): add array bounds checks and fix native decl.

tens0rfl0w:
### Goal of this PR
This PR addresses two issues: missing bounds checks that allowed reading out-of-bounds memory, and a missing parameter in the native declaration of `GET_VEHICLE_DOOR_STATUS`.

- Introduced checks to ensure that array indices are within valid bounds before accessing data, preventing out-of-bounds reads.
- Updated the native declaration of `GET_VEHICLE_DOOR_STATUS` to include the missing `doorIndex` parameter.
- Corrected array size for `doorPositions` in `CVehicleGameStateNodeData`.

FiveM, Server

### Successfully tested on
**Game builds:** 3258

**Platforms:** Windows (Server)

### Checklist

### Fixes issues
Reported here: https://discord.com/channels/779705925577080842/779739477642838036/1304152704137826304 (Cfx.re Engineering Group)

### DIFF
--- a/code/components/citizen-server-impl/src/state/ServerGameState_Scripting.cpp
+++ b/code/components/citizen-server-impl/src/state/ServerGameState_Scripting.cpp
@@ -652,7 +652,13 @@ static void Init()
 
 		if (context.GetArgumentCount() > 1 && doorsOpen)
 		{
-			doorStatus = vn->doorPositions[context.GetArgument<int>(1)];
+			const int index = context.GetArgument<int>(1);
+			if (index < 0 || index > 6)
+			{
+				return doorStatus;
+			}
+
+			doorStatus = vn->doorPositions[index];
 		}
 
 		return doorStatus;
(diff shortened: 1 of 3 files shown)


ext/native-decls/GetVehicleDoorStatus.md
```diff
@@ -5,12 +5,14 @@ apiset: server
 ## GET_VEHICLE_DOOR_STATUS
 
 ```c
-int GET_VEHICLE_DOOR_STATUS(Vehicle vehicle);
+int GET_VEHICLE_DOOR_STATUS(Vehicle vehicle, int doorIndex);
```
AvarianKnight, line 8:
```suggestion
int GET_VEHICLE_DOOR_STATUS(Vehicle vehicle, cs_split int doorIndex);
```

We need to split the declaration so we don't break ABI compatibility with C# (even though this wouldn't have worked anyways.

tens0rfl0w: Good catch, thx!

Nobelium-cfx:
Should be merged after https://github.com/citizenfx/fivem/pull/2925

code/components/citizen-server-impl/src/state/ServerGameState_Scripting.cpp
```diff
@@ -718,7 +724,12 @@ static void Init()
 
 		if (!wheelsFine && context.GetArgumentCount() > 1)
```
AvarianKnight, line 725: Are the argument count checks here required? Usually this is only done for natives that return pointers.

tens0rfl0w: Yeah, no clue why this was added here initially. Seems like those count checks are included in a lot of native definitions while not being required. Will remove here as we're already changing the implementation👍🏼

FabianTerhorst (approved):
👍

PR #2940, opened 2024-11-18
tweak(gamestate/server): Parse CPickupCreationDataNode

### LINKED ISSUE #2924: entityCreating Event: Ambient Pickup GetEntityModel returns 0
### What happened?

I created an ambient pickup on the client. I listened to the entityCreating event on the server.
The creation triggered the event, but when I run GetEntityModel(entity) it returns 0.

### Expected result

GetEntityModel(entity) should return the pickup hash

### Reproduction steps

1. Create this client command and run it:
```
RegisterCommand('test_pickup', function()
    local pickupCoords = GetEntityCoords(PlayerPedId())
    local weaponModel = GetHashKey("weapon_pistol")
    local pickupHash = GetHashKey("PICKUP_WEAPON_PISTOL")

        local pickup = CreateAmbientPickup(pickupHash, pickupCoords.x, pickupCoords.y, pickupCoords.z + 2.0, 0, 100, weaponModel, true, true)
    end)
end)
```
2. Listen to the entityCreating event on the server and fetch the entity model:
```
on('entityCreating', (entity) => {
    console.log(GetEntityModel(entity))
})
```
3. It should return the same pickup hash when the command is ran, but it returns 0.

### Importancy

Unknown

### Area(s)

FiveM, FXServer, Natives

### Specific version(s)

Client 10939 / FXServer 10367 Windows

### Additional information

_No response_

tens0rfl0w:
### Goal of this PR
Allowing to get the entity model of a pickup object on the server-side.

Parse `CPickupCreationDataNode` partially and return the custom model hash (if set) on `GET_ENTITY_MODEL`.

FiveM, Server

### Successfully tested on
**Game builds:** 2060, 2699, 3258

### Checklist

### Fixes issues
resolves #2924


spacevx:
With this data node you could create a server setter no?

tens0rfl0w:
Without verifying: Yeah that should work, though this node has a lot more serialization stuff going on, so completing this isn't super easy.

william-des:
This bug made it possible to easily block the creation of pickups by canceling any entity with a model hash of 0 in the entityCreating event.
I don't think there is any other way to to distinguish between pickups and regular objects server side, and maybe a native can be created for that ?

tens0rfl0w:
The modelhash will still be 0 if no custom model is supplied, as the game will then just use the metadata of the pickup type to determine which model to render. This is a limitation that will be hard to overcome as we have no metadata on the server (yet). I can have a look if we can implement something like `GET_PICKUP_HASH` on the server-side to check what type of pickup this object is.

PR #2945, opened 2024-11-18
feat(gamestate/server): add GET_ENTITY_TYPE_SPECIFIC native

tens0rfl0w:
### Goal of this PR
While working on #2940 it was mentioned that there is currently no way to get the specific entity type on the server-side.
This implements a new custom (server) native to get the specific `NetObjEntity` type of a networked entity.

Reading `entity->type` and returning the specific type as an int on calling `GET_ENTITY_TYPE_SPECIFIC`.

FiveM, RedM, Natives, Server

### Successfully tested on
**Game builds:** 3258 

**Platforms:** Windows

### Checklist

### Fixes issues
/


AvarianKnight:
GET_ENTITY_TYPE_RAW might be a better name here, it matches [other natives](https://docs.fivem.net/natives/?_0xDE73BC10)

ext/native-decls/GetEntityTypeSpecific.md
```diff
@@ -0,0 +1,69 @@
...
+| 21    | Herd               |
+| 22    | Horse              |
+| 23    | WorldState         |
+| 24    | WorldProjectile    |
+| 25    | Incident           |
+| 26    | Guardzone          |
+| 27    | PedGroup           |
+| 28    | CombatDirector     |
+| 29    | PedSharedTargeting |
+| 30    | Persistent         |
```
AvarianKnight, line 63: Same here

tens0rfl0w:
I agree that `_RAW` fits better here, I also changed the return value for an invalid entity to -1 to match the behavior of `GET_VEHICLE_TYPE_RAW`.

Thanks for the feedback!

iridium-cfx (commented):
I'm not a huge fan of returning the raw integer value. Maybe use GetNetObjEntityName to return the string value instead?

AvarianKnight:
> I'm not a huge fan of returning the raw integer value. Maybe use GetNetObjEntityName to return the string value instead?

Would it not be better to just make two natives new natives for this and have one return the string and one return the raw integer value? 

`GET_NET_OBJ_TYPE` -> returns as string (though maybe not specifically from `GetNetObjEntityName` as the `CNetObj` substring is a bit redundant
`GET_NET_OBJ_TYPE_RAW` -> returns as int

I don't think many people (though i could be a bit biased here) really care about the string representation unless they're debugging.

slashkeyvalue:
I also prefer strings due to the fact that we don't have enums or equivalents

prikolium-cfx:
I honestly prefer Int because of these reasons:
1. Single operation comparison
2. Less chance for mistake, as string can be misinterpreted, compared with capitalized, uppercase and so.

prikolium-cfx (approved):

PR #2952, opened 2024-11-20
fix(game/five): updateBusySpinner thread constraint check

### LINKED ISSUE #2850: updateBusySpinner leads to crash
### What happened?

I've had this issue with multiple other players as well where the client crashes after the loading screen.
Apparently it is cause by the `updateBusySpinner` being called from the wrong thread.
The repro for this bug is pretty random as it requires `OnLookAliveFrame` to be called from `LoadDataFiles` (while on Render).
image
This issue was seen happening on April 2024. I'll try to get a dump for this issue asap

### Expected result

Not crashing

### Reproduction steps

1. Connect to the server
2. Wait until loading screen is done
3. Crash or don't... as said above it can't be predicted

### Importancy

Crash

### Area(s)

FiveM

### Specific version(s)

Gamebuild 3095 - Canary

### Additional information

_No response_

tens0rfl0w:
### Goal of this PR
Prevent crashes caused by reading from a nullptr when `updateBusySpinner` is called from the Render thread.

Although the root cause of this issue requires further investigation, this PR provides a temporary fix to prevent crashes when `updateBusySpinner` is unexpectedly called from the Render thread instead of the Main thread. This call causes a subsequent function to return a nullptr due to failing thread constraint checks in native game code.

The only potential side effect of this fix is missing updates to the busy spinner during the loading of addon data files, but this is still better than outright crashing.

FiveM

### Successfully tested on
**Game builds:** 3258

### Checklist

### Fixes issues
fixes #2850

### DIFF
--- a/code/components/loading-screens-five/src/HookLoadingScreens.cpp
+++ b/code/components/loading-screens-five/src/HookLoadingScreens.cpp
@@ -483,6 +483,12 @@ static InitFunction initFunction([] ()
 		{
 			if (*g_scaleformMgr)
 			{
+				if (GetCurrentThreadId() != g_mainThreadId)
+				{
+					trace("Error: BusySpinner update called from non-main thread!\n");
+					return;
+				}
+
 				setupBusySpinner(1);
 				*spinnerDep = true;
 				updateBusySpinner();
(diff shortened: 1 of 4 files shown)


Mathu-lmn:
Any plans to get this reviewed ? We're having new players reporting this issue

Mathu-lmn:
Hey, I didn't see any reports of this crash for a few months. Is this still planned to ever get a merge or would it be preferable to find the "root cause" and fix it ?

PR #2953, opened 2024-11-21
tweak(conhost-v2): switch CPU utilization tracking to dynamic scaling

tens0rfl0w:
### Goal of this PR
Starting with Windows 8, Microsoft updated CPU performance counters to use "Processor Utility" instead of "Processor Time" to reflect dynamic frequency scaling (e.g., Intel Turbo Boost, AMD Precision Boost). This causes the in-game metrics to display wrong CPU utilization values.

This change ensures in-game metrics display those more accurate values.

FiveM, RedM

### Successfully tested on

### Checklist

### Fixes issues
Reported here: https://discord.com/channels/779705925577080842/1260757053635297381 (Cfx.re Engineering Group)

### DIFF
--- a/code/components/conhost-v2/src/DrawPerf.cpp
+++ b/code/components/conhost-v2/src/DrawPerf.cpp
@@ -206,7 +206,7 @@ static InitFunction initFunction([]()
 		if (!cpuQuery)
 		{
 			PdhOpenQuery(NULL, NULL, &cpuQuery);
-			PdhAddEnglishCounter(cpuQuery, L"\\Processor Information(_Total)\\% Processor Time", NULL, &cpuTotal);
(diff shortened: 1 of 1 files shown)


iridium-cfx:
I suppose this depends on what you class as the right/wrong values.

Does this now match what you see in task manager?

tens0rfl0w:
Yes, when this change was introduced in Win8, MSFT also changed the metrics source for CPU utilization in the task manager to this new metric. (https://learn.microsoft.com/en-us/troubleshoot/windows-client/performance/cpu-usage-exceeds-100)

iridium-cfx (approved):
Looks good, I'd agree it makes sense for the average user to try and match what task manager says.

IllInuz:
Yeah this makes more sense to the average user i would say, this has also been done by triple a's to avoid the confusion.

PR #2965, opened 2024-11-27
fix(game/five): crash when no medium lod exists on a vehicle

packfile:
### Goal of this PR
Fixes a crash that occurs when a vehicle only has the highest lod set up. Usually people work around this by duplicating the model file & appending `_hi` (as further explained in code)

If the original function does not return a valid lod, we find the first suitable one and return it to prevent a crash later on.

FiveM

### Successfully tested on
**Game builds:** 1604, 2189, 2699, 3095, 3258

**Platforms:** Windows

### Checklist

### DIFF
--- a/code/components/gta-core-five/src/CrashFixes.DecalCallbacks.cpp
+++ b/code/components/gta-core-five/src/CrashFixes.DecalCallbacks.cpp
@@ -1,0 +1,46 @@
+#include <StdInc.h>
+#include <Hooking.h>
+#include <Hooking.FlexStruct.h>
+#include <Hooking.Stubs.h>
+
+static void* (*g_orig_CDecalCallbacks__GetLod)(void* self, hook::FlexStruct* entity, hook::FlexStruct* drawable);
+
+static void* CDecalCallbacks__GetLod(void* self, hook::FlexStruct* entity, hook::FlexStruct* drawable)
+{
+	void* foundLod = g_orig_CDecalCallbacks__GetLod(self, entity, drawable);
+	if (foundLod)
+	{
+		return foundLod;
+	}
+
+	// Check entity is ENTITY_TYPE_VEHICLE as only these require the patch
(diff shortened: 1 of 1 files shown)


st860923:
This is a great improvement, and I appreciate the effort put into preventing crashes due to missing LOD setups.

As a suggestion, perhaps adding debug information, such as a warning message when the module lacks LOD_MED, could provide developers with more helpful information. It would also highlight best practices for setting up about modelling.

code/components/gta-core-five/src/CrashFixes.DecalCallbacks.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+static HookFunction hookFunction([]
+{
+	// Some calls to CDecalCallbacks::GetLod assume that a valid pointer is returned.
+	// 
+	// For vehicles, this function returns LOD_HIGH if a _hi (HD) file exists;
+	// However if this file does not exist, it will return the LOD_MED instead.
+	// 
+	// Due to a large percentage of content creators only defining a LOD_HIGH, this will crash if no _hi (HD) file is present;
+	// As a workaround many people duplicate the #ft and append _hi to force the LOD_HIGH to be used.
```
Firecul, line 42:
```suggestion
	// As a workaround many people duplicate the yft and append _hi to force the LOD_HIGH to be used.
```

coblyox:
Just tested this change. Seems to be working for vehicle spawn, but without `_hi` if you crash a vehicle where the glass needs to brake or you shoot vehicle glass, you crash instantly.

Always the same crash, we've tested it with 12180 artifacts. Here's dump, hopefully it helps.
[CfxCrashDump_2025_01_02_22_40_07.zip](https://github.com/user-attachments/files/18294998/CfxCrashDump_2025_01_02_22_40_07.zip)

PR #2973, opened 2024-12-01
feat(extra-natives/five): track natives

Ehbw:
### Goal of this PR

This PR introduces a set of track natives to allow for scripts to get and set information relating to the track. This PR also resolves 2 track related issues tied to trains.

- Disabled tracks being considered when finding the nearest track to spawn mission trains on.
-  The tracks maximum speed being constantly set to all trains that can stop at stations while the train is in a moving state. Preventing scripts from being able to modify it, this change does not impact trains when slowing down at stations or speeding up to leave a station.

By creating new natives that retrieve track information from the game.  Hooking into the function responsible for finding the closest track index and it's track node and making sure it accounts for disabled tracks.

FiveM, Natives

### Successfully tested on

**Game builds:**  1604, 3258 

**Platforms:** Windows

### Checklist

### DIFF
--- a/code/components/extra-natives-five/src/TrackNatives.cpp
+++ b/code/components/extra-natives-five/src/TrackNatives.cpp
@@ -1,0 +1,186 @@
+/*
+* This file is part of the CitizenFX project - http://citizen.re/
+*
+* See LICENSE and MENTIONS in the root of the source tree for information
+* regarding licensing.
+*/
+
+#include "StdInc.h"
+#include <ScriptEngine.h>
+#include <Hooking.h>
+#include <limits>
+#include <MinHook.h>
+#include <rageVectors.h>
+
+struct TrackNode
+{
+public:
+	float m_x, m_y, m_z;
+	float m_unk;
+	uint8_t m_unk1;
+	int m_station;
+};
+
+struct TrackData
+{
+public:
+	uint32_t m_hash;
+	bool m_enabled;
+	bool m_isLooped;
+	bool m_stopsAtStation;
+	bool m_MPStopsAtStation;
+	uint32_t m_speed;
+	uint32_t m_brakeDistance;
+
+	int m_nodeCount;
+	TrackNode* m_nodes;
+
+	uint8_t m_pad[8];
+
+	bool m_disableAmbientTrains;
+	// and a bunch of other fields...
+};
+
+static hook::cdecl_stub<TrackData* (uint32_t)> getTrainTrack([]
+{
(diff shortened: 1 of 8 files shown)


code/components/extra-natives-five/src/TrackNatives.cpp
```diff
@@ -0,0 +1,186 @@
...
+	if (trackIndex < 0 || trackIndex > kMaxTrainTracks)
+	{
+		trace("Invalid track index %i passed to %s\n", trackIndex, nn);
+		context.SetResult(0);
+		return NULL;
+	}
+
+	TrackData* track = getTrainTrack(trackIndex);
+
+	if (!track || *(int*)track == NULL)
```
dalekenium-cfx, line 115: Nit: Is `*(int*)track == NULL` meant to be doing `track->m_hash == 0`? If so, checking `m_hash` directly would be best since it's clearer as to what is going on.

code/components/extra-natives-five/src/TrackNatives.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+	fx::ScriptEngine::RegisterNativeHandler("GET_TRACK_BREAKING_DISTANCE", [](fx::ScriptContext& context)
+	{
+		if (TrackData* track = getAndCheckTrack(context, "GET_TRACK_BREAKING_DISTANCE"))
+		{
+			context.SetResult<int>(track->m_breakDistance);
+		}
+	});
+
+	fx::ScriptEngine::RegisterNativeHandler("SET_TRACK_BREAKING_DISTANCE", [](fx::ScriptContext& context)
```
dalekenium-cfx, line 153: Nit: `break` is not technically the correct word here - it should be `brake` instead. Would appreciate if you could make that switch wherever `break` has been used.

code/components/extra-natives-five/src/TrackNatives.cpp
dalekenium-cfx, line 115: Nit: Is `*(int*)track == NULL` meant to be doing `track->m_hash == 0`? If so, checking `m_hash` directly would be best since it's clearer as to what is going on.

dalekenium-cfx (commented):
Hey Ehbw, thank you for your contribution. I just have a couple of small things for you to take a look at, and then I can approve this MR.

code/components/extra-natives-five/src/TrackNatives.cpp
Ehbw: I've changed this to be track->m_hash

dalekenium-cfx: Great, thank you!

PR #2980, opened 2024-12-04
tweak(system_resources/chat):Hide invalid command in chat

outsider31000:
### Goal of this PR

users often type invalid commands and this fallback allows to still be posted as message and everyone can see it.
it will fill the whole chat just with these and can become quite annoying
many users in my community ask how to remove this.

removes the fallback message when a command is not valid 

RedM/FiveM

### Successfully tested on

**Game builds:** .. 

**Platforms:** Windows, Linux

### Checklist

### DIFF
--- a/ext/system-resources/resources/chat/sv_chat.lua
+++ b/ext/system-resources/resources/chat/sv_chat.lua
@@ -212,12 +212,10 @@ AddEventHandler('_chat:messageEntered', function(author, color, message, mode)
 end)
 
 AddEventHandler('__cfx_internal:commandFallback', function(command)
-    local name = GetPlayerName(source)
-
-    -- route the message as if it were a /command
-    routeMessage(source, name, '/' .. command, nil, true)
-
     CancelEvent()
+    -- local name = GetPlayerName(source)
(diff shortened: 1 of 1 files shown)


titanium-cfx:
Hey, you can use the [registerMessageHook export](https://docs.fivem.net/docs/resources/chat/exports/registerMessageHook/) to achieve the same result by checking if the message starts with a `/` and if so, cancel it.

outsider31000:
This commit is for messages typed directly in the chat script.

If I understood correctly the export is to be used on other scripts? 

The idea is to avoid telling everyone  what they need to do to avoid it,  if it can be done in the default.

Not sure what this fallback is for,  But in live servers I don't see the purpose of spamming a chat with invalid commands because users often don't know how to follow chat suggestions xD

outsider31000:
I guess I could add the export in our framework to do this. But then not everyone would benefit from it, if that makes sense.

outsider31000:
can we get a review on this please. tired of users asking me how to remove /messages on their chat script.
making modifications in the chat it self since now it comes with artifacts its a hassle.

PR #2982, opened 2024-12-05
tweak(extra-natives/five): sanitize mission train creation

Ehbw:
### Goal of this PR

Prevent client crashes when using the CREATE_MISSION_TRAIN native.

By adding checks that are missing with the original native implementation. Such as invalid train variations (that do not exist in trains.xml), required carriage models not being loaded, and checking to ensure that there is at least a single track enabled to create a train on (this wouldn't crash the client, but would rather return a native error in the console which could be confusing to users).

This PR also moves some Train structs from TrackNatives to a Train header file to reduce code duplication 

FiveM

### Successfully tested on

**Game builds:** 1604, 2060, 2372, 2545, 2699, 2802, 3095, 3258, 3407

**Platforms:** Windows

I've created a repro resource to reproduce the crashes and their new behaviour with this PR. 
[train-crash-test.zip](https://github.com/user-attachments/files/18026708/train-crash-test.zip)

### Checklist

### DIFF
--- a/code/components/extra-natives-five/include/Train.h
+++ b/code/components/extra-natives-five/include/Train.h
@@ -1,0 +1,59 @@
+#pragma once
+
+#include <atArray.h>
+
+namespace rage
+{
+struct CTrackNode
+{
+public:
+	float m_x, m_y, m_z;
+	float m_unk;
+	uint8_t m_unk1;
+	int m_station;
+};
+
+struct CTrainTrack
+{
+public:
+	// This has never changed.
+	static const int kMaxTracks = 27;
+
+	uint32_t m_hash;
+	bool m_enabled;
+	bool m_isLooped;
+	bool m_stopsAtStation;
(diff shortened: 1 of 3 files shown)


Ehbw:
> There was a bug that broke building the server project, but it has been fixed in master. Can you merge that fix into your branch? I will do a review once that's done. Thank you!

Done

dalekenium-cfx (approved):
The only thing that jumps out at me is whether `train.h` should be in `extra-natives-five`, but I don't think that's a big enough reason to not approve this MR at this point. Looks good.

Legacy-TacticalGamingInteractive:
hey @Ehbw in case you did not notice yet but @prikolium-cfx merged some of your other commits recently :)
keep up the awesome work guys! <3

prikolium-cfx:
Could you please rebase it from master?

PR #2988, opened 2024-12-07
fix(game/five): fix vehicle cloth not rendering

### LINKED ISSUE #2977: Vehicle Flags stopped working
### What happened?

If you spawn car with visual flags it not displaying for some reason.

### Expected result

Spawn car with flags

### Reproduction steps

1. Join FiveM test server (2699) or any server on 3258 (Only two versions i've tested)
2. Spawn car `monster`
3. The car has been spawned without flags

### Importancy

Slight inconvenience

### Area(s)

FiveM

### Specific version(s)

FiveM

### Additional information

image (21)
obraz_2024-12-03_201709813

alexguirre:
### Goal of this PR

The patch introduced in https://github.com/citizenfx/fivem/commit/4f294e49262fb3b095163553d463d089f30c83cd was checking `type != 3` instead of the intended `type == 3`, which disabled the vehicle cloth rendering.

Correct the patch to check for `type == 3`.

FiveM

### Successfully tested on

**Game builds:** 3258

**Platforms:** Windows

### Checklist

### Fixes issues

Fixes #2977.

### DIFF
--- a/code/components/gta-core-five/src/CrashFixes.FakeParachuteProp.cpp
+++ b/code/components/gta-core-five/src/CrashFixes.FakeParachuteProp.cpp
@@ -134,7 +134,7 @@ static HookFunction hookFunction([]
 				mov(al, byte_ptr[rax + 0xA]);		//
 				and(al, 0xF);						//
 				cmp(al, 3);							//        if ( ((shader + 0xA) & 0xF) == 3 /* vehicle */ )
-				jz("fail");							//        {
+				jnz("fail");						//        {
 													//
 				mov(rax, retSuccess);				//
 				jmp(rax);							//            [run original code]


slashkeyvalue:
That patch is being revisited as i've chatted with dalekenium a bit about it, it was supposed to fix a crash and it did fix it in some away by being wrong

dalekenium-cfx:
As @slashkeyvalue said, the incorrect instruction `jz` does fix the crash, however this is because it prevents _all_ instances of cloth sims on vehicles from rendering. As stated in this PR, the intention of the patch was to only render the cloth sims if the shader type is 3 (`VEHICLE`).

With the `jz` switched to `jnz`, vehicle cloth sims now render correctly but the crash with parachute model overrides still occurs. Even when the model's parent entity is not a vehicle, the _shader_ still has type `VEHICLE`; thus, the renderer still attempts to access the parent entity as a vehicle, which causes undefined behavior that inevitably leads to a crash.

We are currently working on a proper fix for the crash that also preserves vehicle cloth sims, so I am going to close this PR.

PR #2990, opened 2024-12-07
Mumble fixes

AvarianKnight:
### Goal of this PR
1. Fix a bunch of issues where players would not be able to talk to other users, or hear other users
2. Fix music mode intent having cut outs for quieter music

Fixes a bunch of different implementation errors, each commit should explain what it does relatively well in the description.

FiveM, RedM

### Successfully tested on

**Game builds:** 3258

**Platforms:** Windows

### Checklist

### DIFF
--- a/code/components/voip-mumble/src/MumbleAudioInput.cpp
+++ b/code/components/voip-mumble/src/MumbleAudioInput.cpp
@@ -47,6 +47,9 @@ static InputIntentMode g_lastIntentMode = InputIntentMode::SPEECH;
 void MumbleAudioInput::Initialize()
 {
 	m_bitrateVar = std::make_shared<ConVar<int>>("voice_inBitrate", ConVar_None, m_curBitrate, &m_curBitrate);
+	// NOTE: If the maximum here is ever increased we need to properly split audio packets into smaller buffers so we don't send packets
+	// that are to large to the server, for FiveM mumble sending too large of a packet will result in the client being dropped, in the mumble-server
+	// implementation this will just cause the packet to be ignored.
 	m_bitrateVar->GetHelper()->SetConstraints(16000, 128000);
 
 	m_denoiseVar = std::make_shared<ConVar<bool>>("voice_enableNoiseSuppression", ConVar_Archive, m_denoise, &m_denoise);
@@ -404,6 +407,8 @@ void MumbleAudioInput::HandleData(const uint8_t* buffer, size_t numBytes)
 	SendQueuedOpusPackets();
 }
 
+constexpr uint16_t kMaxUdpPacket = 1024;
+
 void MumbleAudioInput::EnqueueOpusPacket(std::string&& packet, int numFrames)
 {
 	m_opusPackets.push({ std::move(packet), numFrames });
@@ -423,14 +428,20 @@ void MumbleAudioInput::SendQueuedOpusPackets()
 
 		const auto& [packet, frames] = packetChunk;
 
-		char outBuf[16384];
+		// Maximum size of Opus audio frames is 8191, but we *can't* actually send that much, FiveMs mumble will drop the player, while other implementations will just ignore the packet.
+		// This should be split into multiple packets for higher voice bitrates.
+		// https://mumble-protocol.readthedocs.io/en/latest/voice_data.html#packet-format 
+		// https://mumble-protocol.readthedocs.io/en/latest/voice_data.html#opus-audio-frames
+		char outBuf[kMaxUdpPacket];
 		PacketDataStream buffer(outBuf, sizeof(outBuf));
 
 		buffer.append((4 << 5) | (m_client->GetVoiceTarget() & 31));
 
 		buffer << m_sequence;
 
-		bool bTerminate = false;
+		// This fixed stuttering a while back 
+		// https://github.com/citizenfx/fivem/commit/6b341dc0d71a63a8992c18afe8e6048418978adc
+		constexpr bool bTerminate = false;
 
 		buffer << (packet.size() | (bTerminate ? (1 << 13) : 0));
 		buffer.append(packet.c_str(), packet.size());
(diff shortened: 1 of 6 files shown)


dalekenium-cfx (approved):
Nothing jumps out at me, so I will approve this MR; however, I'd like another reviewer to take a look since this is a big set of changes.

code/components/voip-mumble/src/MumbleAudioInput.cpp
```diff
@@ -417,14 +428,20 @@ void MumbleAudioInput::SendQueuedOpusPackets()
 
 		const auto& [packet, frames] = packetChunk;
 
-		char outBuf[16384];
+		// Maximum size of Opus audio frames is 8191, but we *can't* actually send that much, FiveMs mumble will drop the player, while other implementations will just ignore the packet.
+		// This should be split into multiple packets for higher voice bitrates.
+		// https://mumble-protocol.readthedocs.io/en/latest/voice_data.html#packet-format 
+		// https://mumble-protocol.readthedocs.io/en/latest/voice_data.html#opus-audio-frames
+		static char outBuf[kMaxUdpPacket] = {};
```
iridium-cfx, line 435: Probably shouldn't make this static (and remove the zero-initializer)

code/components/voip-mumble/src/MumbleClient.cpp
```diff
@@ -784,19 +761,24 @@ void MumbleClient::SendUDP(const char* buf, size_t size)
...
+	if (size > kMaxUdpPacket)
 	{
+		trace("We recieved a packet that was too large, max packet size is %d bytes, got sent %d bytes\n", kMaxUdpPacket, size);
 		return;
 	}
 
-	uint8_t outBuf[8192];
-	if (!m_crypto->Decrypt(buf, outBuf, size))
+
+	static uint8_t outBuf[kMaxUdpPacket] = {};
```
iridium-cfx, line 778: And this one

code/components/voip-mumble/src/MumbleClient.cpp
```diff
@@ -732,15 +731,14 @@ void MumbleClient::SetListenerMatrix(float position[3], float front[3], float up
 
 void MumbleClient::SendVoice(const char* buf, size_t size)
 {
-	if ((msec() - m_lastUdp) > kUDPTimeout)
+	// If we don't have UDP then we should send the packets over a TCP tunnel
+	if (!m_hasUdp)
```
joelwurtz, line 729: since it's false at the start, will this send tcp packet on the start until it resolve the first udp ping ?

code/components/voip-mumble/src/MumbleClient.cpp
```diff
@@ -980,6 +962,27 @@ MumbleConnectionInfo* MumbleClient::GetConnectionInfo()
...
+		m_crypto.m_remoteLate = ping.late();
+		m_crypto.m_remoteLost = ping.lost();
+		m_crypto.m_remoteResync = ping.resync();
+
+		if (m_hasUdp && (m_crypto.m_remoteGood == 0 || m_crypto.m_localGood == 0) && (msec() - m_lastUdp) > 2s)
+		{
+			console::PrintWarning("mumble", "UDP packets can *not* be received. Switching to TCP tunnel mode.\n");
+			m_hasUdp = false;
+		}
+		else if (!m_hasUdp && m_crypto.m_remoteGood > 3 && m_crypto.m_localGood > 3)
```
joelwurtz, line 979: maybe there should be an additional check for initialization, from what i understand this means that on start we will have to wait for 3 correct ping before it can switch to udp packet ? or maybe i'm missing something ?

joelwurtz:
Nice to see some improvements, feel correct for most part, and i think having a check on the udp packet like mumble does will indeed improve some connection issue that were not detected before.

If you want to try other implementations, there is this one https://github.com/SOZ-Faut-etre-Sub/ZUMBLE that mimic fivem server mumble implementation, (we use it at lot for some years now without any major issue, don't have the ability to compile this but since we have user on expiremental build we will be able to give you feedback once it's merged)

code/components/voip-mumble/src/MumbleAudioInput.cpp
AvarianKnight: Is there any reason not to here, anything used by these are copied out

1. The usage here is initialize buffer
2. Use buffer for `PacketDataStream`
3. Append packet to backing buffer
4. Send to `SendVoice` which will call `SendUDP` (or `Send` for TCP tunnel if UDP isn't available), 

5a. if its send via UDP it will be copied out into a new `outBuf` when its encrypted
5b. if its send via TCP it will be encrypted via TLS (which will block until encryption is done)

This all happens sequentially so there isn't any risk of this being used by multiple threads.

code/components/voip-mumble/src/MumbleClient.cpp
AvarianKnight: That is what this means, this shouldn't really be much of an issue considering on the initial setup you'll still be loading in, and a ping is sent every second by the server, which means it would be 3 seconds maximum where the client can't use UDP (this copies the behavior of mumble-server)

AvarianKnight:
> Nice to see some improvements, feel correct for most part, and i think having a check on the udp packet like mumble does will indeed improve some connection issue that were not detected before.
> 
> If you want to try other implementations, there is this one https://github.com/SOZ-Faut-etre-Sub/ZUMBLE that mimic fivem server mumble implementation, (we use it at lot for some years now without any major issue, don't have the ability to compile the fivem client but since we have user on expiremental build we will be able to give you feedback once it's merged)

Most of my testing was done [on my fork](https://github.com/avarianKnight/ZUMBLE) of your project, although I did test to make sure the internal uMurmur didn't break because of any of these changes.

IllInuz:
Looks all good from what i can tell, most of it has been addressed above.
Thanks @AvarianKnight

PR #2991, opened 2024-12-08
fix(extra-natives-five): Fix unable to set last pixel of runtime texture

Lucas7yoshi:
### Goal of this PR

Fix not being able to set the last/bottom right pixel of runtime textures.

Fixing a logic error (>= when it should be >)

FiveM, Natives

### Successfully tested on

**Game builds:** 3258

**Platforms:** Windows

Note: tested and was unable to set outside bounds of a 8x4 texture through either x/y.
Would be very surprised if this broke based on build.

### Checklist

### Fixes issues
N/A


dalekenium-cfx (approved):
Simple enough. Looks good to me.

PR #3001, opened 2024-12-14
tweak(server): ignore resources that have a name starting with '.' in resource scan

jellyton255:
### Goal of this PR

Like the .git folder, we ignore the .github folder from the resource scan so git repos with a .github folder do not result in a warning during a resource scan.

Adding an additional check to the resource scan loop so that resources with the name ".github" are skipped.

Server

### Successfully tested on

**Platforms:** Windows

### Checklist

### DIFF
--- a/code/components/citizen-server-impl/src/ServerResourceList.cpp
+++ b/code/components/citizen-server-impl/src/ServerResourceList.cpp
@@ -110,6 +110,13 @@ void ServerResourceList::ScanResources(const std::string& resourceRoot, ScanResu
 					else if (scannedNow.find(findData.name) == scannedNow.end())
 					{
 						const auto& resourceName = findData.name;
+
+						// ignore hidden folders
+						if (resourceName[0] == '.')
+						{
+							continue;
+						}
+
 						scannedNow.emplace(resourceName, resPath);
 
 						auto oldRes = m_manager->GetResource(resourceName, false);


Mathu-lmn:
Why not ignoring all folders starting with a dot ?

jellyton255:
> Why not ignoring all folders starting with a dot ?

Since resource scan is a critical component, I wouldn't want to change something there unless there was a specific desire to do so, like there is with `.github`. It may be the case that it's a good idea to ignore all folders with a `.` in the front, but I personally wouldn't be comfortable with making that change w/o an express desire from Cfx.

DRAIKDev:
Also will be great to add .vscode : )

keeganwut:
Wouldn't it be better (if possible) to leave this up to a server convar? Being able to define different types of files or folders to ignore? I'm sure this could cause issues but if used correctly potentially not.

DRAIKDev:
I think the first server resource checks is done before server convars is settled, also seems a bit unnecessary for the vast majority of servers.

eoshorizon:
Wouldnt it be much better to add a file like .gitignore where we can define content that should be skipped? if it doesnt exist then use the current defaults?

AvarianKnight:
It would probably just be better to ignore any file starting with a `.`, windows already treats these as a hidden file there's no reason we shouldn't do the same imo.

Maximus7474:
Definetly the best approach to just ignore all the folders starting with `.`
There are so many that can be used for various reasons, can also be useful without it being linked ignoring non resource folders

AvarianKnight:
This should probably be specifically done for directories, not every file.

jellyton255:
True, lemme fix that real quick.

jellyton255:
@tabarra If/when this gets merged, you'll eventually want to update https://github.com/tabarra/txAdmin/blob/8cbc767aa24e614fe2dc801810cb3ea58df282fd/core/extras/serverDataScanner.ts to mirror this change.

thelindat:
You've messed up your PR and are reverting changes to the FiveM repo since your initial commit was made.

FabianTerhorst:
> This should probably be specifically done for directories, not every file.

More specifically it should ignore all childs inside a directory starting with "." and any files starting with ".".

tabarra:
@jellyton255 Thanks for the heads up, but my scanner implementation will be deprecated soon, as I'll switch to the new  ScanResourceRoot native:  
https://docs.fivem.net/natives/?_0x636F097F

PR #3006, opened 2024-12-15
tweak(scripting/lua): speed up lua event serialization.

icy4471:
If there are no event handlers for the event, there is no need to unpack or set the source as they will never be used.

### Goal of this PR

Significantly speeds up event serialization when there are no handlers available.

By returning early when there are no event handlers available.

FiveM, ScRT: Lua

### Successfully tested on

**Game builds:** b3095

**Platforms:** Windows, Linux

### Checklist


thelindat:
Worth noting that this function is only triggered in functions which have registered the event previously, so this only helps with removed events. Possibly worth being able to remove resources from the event registry altogether?

AvarianKnight:
Another thing to note here is `msgpack_unpack` should be called after the net checks.

PR #3008, opened 2024-12-16
feat(resources-metadata): Add GetConvar function

ImHoppy:
### Goal of this PR

This PR serves more as a proof of concept than ready-to-merge code. It introduces functionality to dynamically load files based on a defined ConVar. 

For context, our base is compatible with both RDR3 and GTA5, as well as multiple servers. To separate files, we currently use file headers with return conditions ([example screenshot](https://img.flashfa.fr/file/hoppy/f5deb993-2ff6-43aa-84c1-7320a42395ce.png)). 

With this functionality, we can, for instance, load only the files in the `gta5` folder when the `gamename` variable is set to `gta5`.

Here is an example of a resource that takes advantage of this new functionality:
fxmanifest:
```lua
fx_version 'cerulean'
games { 'rdr3', 'gta5' }
version '1.0.0'

if GetConvar('gamename', 'gta5') == 'rdr3' then
    server_script 'redm.lua'
else
    server_script 'fivem.lua'
end
```

This PR achieves its goal by adding a GetConvar function in the fxmanifest. The function takes two parameters:
- varName: The name of the ConVar to retrieve.
- defaultValue: The value to use if the ConVar is not defined.
Using this function allows conditional logic to determine which files are loaded. This approach simplifies the management of multiple game environments and servers.

Server on the resources load

### Successfully tested on
Custom server build

**Game builds:** tested on 3095 but should work on any

**Platforms:** Windows

### Checklist

### Fixes issues
not related to any issues

NOTE: I am fully open to making changes as needed. I am also ready to update the documentation.

### DIFF
--- a/code/components/citizen-resources-metadata-lua/src/LuaMetaDataLoader.cpp
+++ b/code/components/citizen-resources-metadata-lua/src/LuaMetaDataLoader.cpp
@@ -8,6 +8,7 @@
 #include "StdInc.h"
 #include "Resource.h"
 #include "ResourceMetaDataComponent.h"
+#include <CoreConsole.h>
 
 #include "VFSManager.h"
 
@@ -227,6 +228,30 @@ std::optional<std::string> LuaMetaDataLoader::LoadMetaData(fx::ResourceMetaDataC
 		lua_pushnil(m_luaState);
 		lua_setglobal(m_luaState, removeThat);
 	}
+
+	lua_pushcfunction(m_luaState, [](lua_State* L)
+	{
+		const char* varName = lua_tostring(L, 1);
+		const char* defaultValue = lua_tostring(L, 2);
+
+		if (!varName)
+		{
+			lua_pushstring(L, defaultValue);
+			return 1;
+		}
+
+		const auto var = console::GetDefaultContext()->GetVariableManager()->FindEntryRaw(varName);
+		if (!var)
+		{
+			lua_pushstring(L, defaultValue);
+			return 1;
+		}
+
+		lua_pushstring(L, var->GetValue().c_str());
+
+		return 1;
+	});
+	lua_setglobal(m_luaState, "GetConvar");
 
 	auto fileNameAttempts = { "fxmanifest.lua"s, "__resource.lua"s };
 	bool attemptResults[] = { true, true };


FabianTerhorst:
Maybe just do a method that returns the game name instead of allowing ConVar reads inside the meta data loader?

ImHoppy:
That's good too, but ConVar reads let us filter files dynamically per server, like `servername == "Unnamed RP"`.

ImHoppy:
But it's true that in reality we only need two variables. And I don't see the use of other variables.

FabianTerhorst:
Yes, then add the server name and game name as two methods and then it should be fine. Don't really see the need allowing to read any existing ConVar here.

pedr0fontoura:
If `.cfg` is still the recommended way of defining server config, reading ConVars could be relevant.

To illustrate the use case, imagine a resource with an export API and some UI. If the UI is optional, you could use a ConVar to avoid loading all the UI-related files.

prikolium-cfx:
I would not recommend to build any logic in fxmanifest.lua. At some point we plan to move to declarative fxmanifest and it means that you won't be able to execute code there.

ImHoppy:
And it's not possible to have this pr until you release the v3 of the fxmanifest?

prikolium-cfx:
> And it's not possible to have this pr until you release the v3 of the fxmanifest?

For sure it's possible. But probably it's possible to satisfy your need in more declarative way. For example you can add optional argument to server_script function and pass game name to it or something like this. In that case this feature can be transferred to the next fxmanifest version

FabianTerhorst:
Wouldn't it be better to split it to two resources at that point? One for rdr and one for gta.

ImHoppy:
Unfortunately, this isn't a solution for us, splitting the code would force us to maintain the same code twice. To illustrate, here's a breakdown of the line count for our single resource. And excluding GTA5, we still have 160,643 lines of code:

Language | Files | Blank | Comment | Code
-- | -- | -- | -- | --
Lua | 1277 | 31,242 | 4,737 | 258,310
SQL | 14 | 248 | 256 | 1,566
Total | 1291 | 31,490 | 4,993 | 259,876

But the idea from Prikolium also seems like a good one, especially if it seems more coherent for the next version

tabarra:
I'm big time against needing to execute lua to get the resource metadata, which might be fine for fxserver, but any other codebase (such as txAdmin or Cfx backends) that wants to read it *properly* will then need to evaluate lua code. But the need for game-specific definitions is valid.  
  
In monitor/txAdmin, I also want to be able to have different manifest directives depending on the game name.

Issue #3010, opened 2024-12-16
Sandboxed 'debug' library missing some functions [v1.0.0.12078]


iNont:
### What happened?

On [v1.0.0.12078](https://github.com/citizenfx/fivem/releases/tag/v1.0.0.12078), make my resources cannot working since I was using some the debug functions.

These functions are missing from 'debug'
- `debug.setupvalue`
- `debug.getlocal`
- `debug.setlocal`
- `debug.gethook`
- `debug.sethook`

### Expected result

All missing 'debug' functions work normally

### Reproduction steps

1. Try to use the missing functions for ex. `debug.getlocal`

### Importancy

Crash

### Area(s)

FiveM, Natives, ScRT: Lua

### Specific version(s)

v1.0.0.12078

iNont:
Just reported because it is already launched on Latest version, and some players got error when joined in server

iridium-cfx:
What are your use-cases for these library functions?

iNont:
> What are your use-cases for these library functions?

1. **Anticheat** : cheaters modified some value on `scheduler.lua`, so my server using these library functions to detect the changes (it seems like many developers in my region got the same issue here)
2. **Avoid dumper** : I'm using these functions to make my resources to be more complex (like using `debug.setlocal`, `debug.setupvalue`)

eryxiz:
> > What are your use-cases for these library functions?
> 
> 1. **Anticheat** : cheaters modified some value on `scheduler.lua`, so my server using these library functions to detect the changes (it seems like many developers in my region got the same issue here)
> 2. **Avoid dumper** : I'm using these functions to make my resources to be more complex (like using `debug.setlocal`, `debug.setupvalue`)

Me too, I'm using the debug library for cheat detection and have encrypted my code to make it harder for cheaters to understand after dumping resources.

Mycroft-Studios:
> Anticheat : cheaters modified some value on scheduler.lua, so my server using these library functions to detect the changes (it seems like many developers in my region got the same issue here)

the entire point of sandboxing, _is_ to stop you from doing that.
"Ah yes, peoples modify a file in skechy ways, so let my resource also read and write to those files in totally none skechy ways!"
the entire intent is for resources to be confined to themselves only, so you cant do sketchy stuff exactly what you just described.
also `debug.setlocal` to make your code "complex", be real brother xD

nacknattavat:
[...]

PR #3016, opened 2024-12-18
feat(scripting/lua): add getnparams to client debug sandbox

Peter-Greek:
### Goal of this PR

Add additional functionality to the new client sandboxed lua debug lib as much of its functionality was removed. 

Added the `debug.getnparams()` function to be used on the client. This is a bypass for the unsafe debug.getlocal(), so devs can still get the parameter names of a function passed as the argument to the new function.

New lua docs for the debug function
```
---
--- Returns a table with the function parameter names from 1 to `nparams`, or **nil** if `nparms` 
--- is not available. Variable arguments are ignored but will still return parameters before it.
--- `self` is included as the first parameter if the function is a method.
---@param f function
---@return table
function debug.getnparams(f) end
```

ScRT: Lua

### Successfully tested on

Unable to build fivem on my laptop but did test it on Windows 10, using a simple C++ script and lua script and the code compiled and ran as expected. Used the functions in the file `LuaDebug.cpp` as reference to ensure consistency.

**Game builds:** 

**Platforms:** 

### Checklist

### Fixes issues

Addresses #3010

### DIFF
--- a/code/components/citizen-scripting-lua/src/LuaDebug.cpp
+++ b/code/components/citizen-scripting-lua/src/LuaDebug.cpp
@@ -268,6 +268,42 @@ int db_traceback(lua_State* L)
 	return 1;
 }
 
+int db_getparametersnames(lua_State* L)
+{
+	int arg;
+	lua_Debug ar;
+	lua_State* L1 = getthread(L, &arg);
+	luaL_checktype(L, arg + 1, LUA_TFUNCTION);
+	lua_pushvalue(L, arg + 1);
+	if (!lua_getinfo(L1, ">u", &ar)) /* Retrieve debug info */
+	{
+		lua_pushnil(L); /* Return nil if lua_getinfo fails */
+		return 1;
+	}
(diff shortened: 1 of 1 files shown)


iridium-cfx:
So, I am tempted to allow the name-only path of getlocal, but I would still like to understand what people's use-cases for it are. Though if we do add it back, it will be easier to just remove the unwanted parts of the original getlocal, rather than adding a new function.

thelindat:
https://forum.cfx.re/t/why-disable-the-debug-lib-without-warning/5291361

Peter-Greek:
> So, I am tempted to allow the name-only path of getlocal, but I would still like to understand what people's use-cases for it are. Though if we do add it back, it will be easier to just remove the unwanted parts of the original getlocal, rather than adding a new function.

I mentioned a few times some use cases but quick synopsis is being able to get the names of function parameters, this can be used in things like function overloading (calling functions with the same name or functionality based on the parameters it accepts). Atm without the use of getlocal it is impossible to do this.

PR #3018, opened 2024-12-19
feat(v8/scripting): swap from msgpack-lite to msgpackr

AvarianKnight:
### Goal of this PR
Swap JS to msgpackr to allow serializing classes & other js builtin objects (i.e. `Map`, `Set`, `Error`, etc) and also increase performance.

### How is this PR achieving the goal
Swap out the custom version of msgpack-lite to use [msgpackr](https://github.com/kriszyp/msgpackr)

### This PR applies to the following area(s)
ScRT: JS

### Successfully tested on

**Game builds:** N/A

**Platforms:** Windows, Linux

### Checklist

- [x] Code compiles and has been tested successfully.
- [x] Code explains itself well and/or is documented.
- [x] My commit message explains what the changes do and what they are for.
- [x] No extra compilation warnings are added by these changes.

### Fixes issues
Supersedes PR #2931

PR #3019, opened 2024-12-19
feat(scripting/v8): support unpacking Lua vectors

thelindat:
### Goal of this PR

Allows the JS runtime to unpack vectors received from Lua as arrays, rather than requiring manual handling of buffers.

**However**, I would like to add default Vector classes instead of using arrays if acceptable.

Adds a msgpackr extension for the vector types (20, 21, 22), dependant on #3018.

ScRT: JS

### Successfully tested on

**Game builds:** N/A

**Platforms:** Windows

### Checklist

### Fixes issues

Supercedes #2946

### DIFF
--- a/data/shared/citizen/scripting/v8/main.js
+++ b/data/shared/citizen/scripting/v8/main.js
@@ -3,6 +3,9 @@
 
 const EXT_FUNCREF = 10;
 const EXT_LOCALFUNCREF = 11;
+const EXT_VECTOR2 = 20;
+const EXT_VECTOR3 = 21;
+const EXT_VECTOR4 = 22;
 
 (function (global) {
 	let boundaryIdx = 1;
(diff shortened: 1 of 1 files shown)


Korioz:
It will break my scripts and i don't think of any way to fix the breaking, my use case is : JS script receive data in any form or types from Lua (db documents) and use the current extension type 20, 21, 22 in the variable to check if the variable was a vector in Lua.

thelindat:
Refusing to support vectors because of backwards compatibility is stupid.

> i don't think of any way to fix the breaking

Not having support for arrays or objects with xyz properties, strictly expecting vectors only, seems like a bad idea.
Check the object type, if it's an array or doesn't have the buffer and type properties then fallback to old behaviour.

p1u3o:
I would find this very useful, could it be put behind a newer manifest version so existing scripts don't break?

Korioz:
> Refusing to support vectors because of backwards compatibility is stupid.

I approve that point and i am open to changes but these changes will break my use cases, simply because there will be no way to identify a Lua vector from a 2, 3, 4 numbers array.

Lucas7yoshi:
I'm being nosey but i'd tend to agree that sounds like the kind of thing that would be best put behind a new manifest version beyond cerulean

thelindat:
As if we'll ever get a manifest version again.

Identity-labs:
Why not adding native to allowing us to set function to pack/unpack every type like middleware client & server side
On my side i have custom value object for every type, but i do manually it every time
Like that default case will be like now, if you want your own Vector3 you can.

thelindat:
> Why not adding native to allowing us to set function to pack/unpack every type

That's part of what #3018 is allowing - swapping to msgpackr and exposing addExtension globally; so you absolutely can replace the packer and unpacker with your own functions.

PR #3026, opened 2024-12-21
feat(scripting/lua): add missing debug func (getlocal, setlocal, setupvalue) in lua debug sandbox

eryxiz:
### Goal of this PR

Add missing functions in debug library to new sandboxed lua debug.
- `debug.setupvalue`
- `debug.getlocal`
- `debug.setlocal`

Added the functions (`debug.setupvalue`, `debug.getlocal`, `debug.setlocal`) based on lua debug API.
The functions in the debug library can't access variables, code, or files outside the scope of a resource, so they shouldn’t have any impact on exploiting servers.

ScRT: Lua

### Successfully tested on

**Platforms:** Windows

### Checklist

### Fixes issues

Related issue #3010 
(`debug.gethook`, `debug.sethook` not including in this PR)

### DIFF
--- a/code/components/citizen-scripting-lua/src/LuaDebug.cpp
+++ b/code/components/citizen-scripting-lua/src/LuaDebug.cpp
@@ -1,4 +1,4 @@
-﻿/*
+/*
 ** $Id: ldblib.c $
 ** Interface from Lua to its debug API
 ** See Copyright Notice in lua.h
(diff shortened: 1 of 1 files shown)


Mathu-lmn:
Hey, these function were removed for a reason, I'll let a member of the cfx team confirm that but it's not planned to add them !

eryxiz:
> Hey, these function were removed for a reason, I'll let a member of the cfx team confirm that but it's not planned to add them !

I understand that the reason for removing it is to prevent server exploits. However, as far as I know, the debug library currently operates within the scope of its own resource. So, I’m curious—how could someone exploit a server using these debug library functions when they’re limited to their own resource? This is different from the io library, which previously allowed full permission to read/write files anywhere.

AvarianKnight:
Having these debug hooks can give users access to the scheduler, which isn't desired behavior.

iridium-cfx:
If you think a specific function should be partially/fully re-added, please create an issue stating why (i.e what is your use-case, and why is it not possible without this function?).

PR #3033, opened 2024-12-23
tweak(font-renderer): force watermark bottom right

alberttheprince:
### Goal of this PR
Watermark currently randomly moves between the top right and bottom right of the player's screen, which is unnecessary. A set position (bottom right was chosen, as it's just a standard/out-of-the-way spot common across the industry) would allow more reliable expectations for the UI.

Position is hardcoded to bottom right, rather than top and bottom right randomly.

FiveM client

### Successfully tested on

**Game builds:** 3258

**Platforms:** Windows

### Checklist

### Fixes issues
Quality of life update, frees up the top right of the screen and sets a position for the watermark going forward.


alberttheprince:
Reopening with correct format

PR #3034, opened 2024-12-24
tweak(font-renderer): force watermark bottom right

alberttheprince:
### Goal of this PR
Watermark currently randomly moves between the top right and bottom right of the player's screen, which is unnecessary. A set position (bottom right was chosen, as it's just a standard/out-of-the-way spot common across the industry) would allow more reliable expectations for the UI.

Position is hardcoded to bottom right, rather than top and bottom right randomly.

FiveM client

### Successfully tested on

**Game builds:** 3258

**Platforms:** Windows, Linux

### Checklist

### Fixes issues
Quality of life update, frees up the top right of the screen and sets a position for the watermark going forward.

### DIFF
--- a/code/components/font-renderer/src/GtaGameInterface.cpp
+++ b/code/components/font-renderer/src/GtaGameInterface.cpp
@@ -476,23 +476,7 @@ static InitFunction initFunction([] ()
 			BottomLeft = 3,
 		};
 
-		static auto anchorPosBase = ([]()
-		{
-			if (launch::IsSDKGuest())
-			{
-				return AnchorPos::BottomRight;
-			}
-			else
-			{
-				std::random_device rd;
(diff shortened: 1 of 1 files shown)


keeganwut:
Would be nice to have the mods loaded thing be added to this. It's the biggest pest in the FiveM UI.

Korioz:
I'm almost sure it was done like this to prevent people from hiding it easily on stream, videos and such.

prikolium-cfx:
That was done intentionally and we don't plan to change this decision right now.

alberttheprince:
> I'm almost sure it was done like this to prevent people from hiding it easily on stream, videos and such.

Can be hidden now, it used to be 1 corner, 4 corners and then 2 corners, really no reason to have this moving ad its super imposed.

PR #3038, opened 2024-12-24
tweak(docs): update and rename code/LICENSE to LICENSE

Michael21107:
Allows to view the license directly on the repo's main page. (Also updated the readme to match this change.)


tabarra:
That is not correct. As PulseHarbor said, the licensing is scoped to specific directories.  
I do think there should be a `./LICENSE` file, but pointing to the other licenses.

PR #3043, opened 2024-12-26
fix(state): fix not being able to use any entity native calls in 'entityRemoved'

### LINKED ISSUE #3042: entityRemoved event broken since newer artifacts
### What happened?

**Server Artifact :** b12160 (Linux) - Maybe also on some olders
**Categorie :** Server side issue
**Issue :** entityRemoved entity args cannot be use as before, using GetEntityType or Owner send an error.
"SCRIPT ERROR: native 000000000b1bd08d: Tried to access invalid entity: 157447"
Native being dependant on which native you are using ofc.

### Expected result

No error entitytype returned or 0

### Reproduction steps

1) Update to recent artifacts
2) Add this snippet server side :
```
> AddEventHandler('entityRemoved', function(_entity)
>     local entity = _entity
>     --
>     local type = GetEntityType(entity)
> end)
> 
```
3) Create a vehicle and then delete it (should also work on objects/peds but only tried on vehicles)
It will send the error.

### Importancy

Unknown

### Area(s)

FiveM

### Specific version(s)

FiveM b12160 or older

### Additional information

Don't know since which artifact though.
Here if you need any infos.
Peace

AvarianKnight:
### Goal of this PR
Fix oversight in 8957209b72626bacd970f07412f3dff6ccac2bf0 which added the `IsEntityValid` to `makeEntityFunction`, this is only needed for checks that an entity is valid (i.e. `DOES_ENTITY_EXIST`, and to a lesser extent, `NETWORK_GET_ENTITY_FROM_NETWORK_ID`) and pool getters.

Swap back to the old check for `makeEntityFunction`

Server

### Successfully tested on
**Platforms:** Windows, Linux

### Checklist

### Fixes issues
Fixes #3042

### DIFF
--- a/code/components/citizen-server-impl/src/state/ServerGameState_Scripting.cpp
+++ b/code/components/citizen-server-impl/src/state/ServerGameState_Scripting.cpp
@@ -20,8 +20,11 @@ void DisownEntityScript(const fx::sync::SyncEntityPtr& entity);
 static void Init()
 {
 
+
+	// If the entity is in its deleting/finalizing state we should not allow access to them
+	// This should only be used for natives that are not expected to work when an entity is actively being deleted
+	// i.e. setters, getting entity by network id / does exist, or any pool getter natives
 	static auto IsEntityValid = [](const fx::sync::SyncEntityPtr& entity) {
-		// if we're deleting or finalizing our deletion then we don't want to be included in the list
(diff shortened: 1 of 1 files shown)


FabianTerhorst (approved):
Thanks a lot 👍

PR #3056, opened 2025-01-06
fix(commerce): allow the server to refresh sku's after initial load

AvarianKnight:
Was talking with @the-neco about the usage of some of the commerce natives but found that their usage was extremely limited because they could not be updated after the first request.

### Goal of this PR
Allow LoadPlayerCommerceDataExt to reload player SKUs while in game, this currently only fetched once and cannot be updated until the player rejoins, this adds unnecessary friction.

This allows `LoadPlayerCommerceDataExt` to fetch the players packages again once every 10 seconds.

Calls to `LoadPlayerCommerceDataExt` after 10 seconds will refresh the clients SKUs

Server

### Successfully tested on
**Platforms:** Windows

### Checklist

### DIFF
--- a/code/components/citizen-server-impl/src/ServerExtCommerce.cpp
+++ b/code/components/citizen-server-impl/src/ServerExtCommerce.cpp
@@ -471,6 +471,8 @@ private:
 	fx::Client* m_client;
 
 	bool m_commerceDataLoaded;
+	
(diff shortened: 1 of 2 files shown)


FabianTerhorst (approved):
Thanks 👍

PR #3063, opened 2025-01-12
fix(server/gamestate): memory leak

tens0rfl0w:
### Goal of this PR
This PR addresses an issue where creating an object, ped, or vehicle via server-setters does not validate the availability of free object IDs. This oversight can lead to memory leaks and undefined behavior, as overwriting an existing entity handle keeps the previous entity pointer alive in the sync pool, resulting in "ghost entities" and inaccessible allocated memory.

Checking if we got a valid / free object id, and if not, returning a non-valid GUID ('0') indicating failed creation (Client logic parity).

Server

### Successfully tested on

**Platforms:** Windows (Server)

### Checklist

### Fixes issues
/


FabianTerhorst (approved):
Thanks 👍

csskroubledev:
Thanks, cannot wait for this to be merged :)

enzoo27:
Hi, I have the same problem. What am I supposed to do?

PR #3077, opened 2025-01-18
fix(citizen-script-lua): Fix crash with nil value as name in CreateTh…

niekschoemaker:
…readNow

### Goal of this PR
Fix a crash that happens when a nil value is passed to Citizen.CreateThreadNow as the name.

...

It checks to see if lua_tostring returns a null pointer and if it does leaves the name empty.
...

ScRT: LUA

...

### Successfully tested on

**Game builds:** 12206

**Platforms:** Windows

### Checklist


niekschoemaker:
Tried to merge, but merge commits aren't allowed by title, so I undid those changes, that's the mess with forced pushes

PR #3088, opened 2025-01-21
fix(game/five): SET_HELI_*_ROTOR_HEALTH sync fixes

### LINKED ISSUE #2997: Synchronization issue with helicopter blade breaking using the native SetHeliMainRotorHealth
### What happened?

Synchronization issue with helicopter blade breaking using the native `SetHeliMainRotorHealth`

Entity owner's POV (driver): https://streamable.com/oonw7j
Another player's POV: https://streamable.com/2ab34k

### Expected result

That the rotors are correctly synchronized for the helicopter driver.

### Reproduction steps

1. Create a networked vehicle client-side (e.g., a Frogger helicopter) (this issue might also occur with server-side vehicle creation ?).
2. Use the native function SetHeliMainRotorHealth to break the helicopter's rotor blades (the same issue may also occur with SetHeliTailRotorHealth ??).

### Importancy

Slight inconvenience

### Area(s)

FiveM, FXServer

### Specific version(s)

FiveM 3095 / FXServer-master SERVER v1.0.0.11840 win32

### Additional information

I remain available for any additional information or to answer your questions if needed.

tens0rfl0w:
### Goal of this PR
This PR addresses an issue where helicopter rotors do not break off correctly when their health is set to `0.0` via scripts using the natives `_SET_HELI_MAIN_ROTOR_HEALTH` or `_SET_HELI_TAIL_ROTOR_HEALTH`.

Under normal game scenarios, such as damage from collisions or bullets, a rotor with `0.0` health breaks off the helicopter as intended. However, when the health is altered via these natives, the necessary checks to trigger rotor detachment are missing. This results in an inconsistent synchronization state: the rotor remains attached on the entity owner's client, while it appears broken off for remote players (as health sync node updates for `NetObjHeli` contain the required checks already).

This PR ensures that rotor health changes made via script are handled consistently, preserving sync integrity across all clients.

We hook the native handlers for `_SET_HELI_MAIN_ROTOR_HEALTH` and `_SET_HELI_TAIL_ROTOR_HEALTH`. After the native invocation, the new rotor health is checked. If the health is set to `0.0`, we manually call the rotor break-off functions to ensure that the rotor detaches correctly and the synchronization state remains consistent for all clients within the sync range.

FiveM, Natives

### Successfully tested on
**Game builds:** 2060, 3407

### Checklist

### Fixes issues
fixes #2997

### DIFF
--- a/code/components/extra-natives-five/src/NativeFixes.cpp
+++ b/code/components/extra-natives-five/src/NativeFixes.cpp
@@ -540,12 +540,88 @@ static void FixMissionTrain()
 	});
 }
 
+// PatchVehicleHoodCamera.cpp
+enum eVehicleType : uint32_t
+{
+	VEHICLE_TYPE_CAR = 0,
+	VEHICLE_TYPE_PLANE,
+	VEHICLE_TYPE_TRAILER,
+	VEHICLE_TYPE_QUADBIKE,
+	VEHICLE_TYPE_DRAFT,
+	VEHICLE_TYPE_SUBMARINECAR,
+	VEHICLE_TYPE_AMPHIBIOUS_AUTOMOBILE,
+	VEHICLE_TYPE_AMPHIBIOUS_QUADBIKE,
+	VEHICLE_TYPE_HELI,
+	VEHICLE_TYPE_BLIMP,
+	VEHICLE_TYPE_AUTOGYRO,
+	VEHICLE_TYPE_BIKE,
+	VEHICLE_TYPE_BICYCLE,
+	VEHICLE_TYPE_BOAT,
+	VEHICLE_TYPE_TRAIN,
+	VEHICLE_TYPE_SUBMARINE,
+};
+
+static int g_vehicleTypeOffset;
+
+using CHeli_BreakOffRotor_t = void(__fastcall*)(fwEntity* heli);
+static CHeli_BreakOffRotor_t CHeli_BreakOffMainRotor;
+static CHeli_BreakOffRotor_t CHeli_BreakOffTailRotor;
+
+static void SetHeliRotorHealthHandler(const uint64_t nativeHash, void(*breakOffRotorFunc)(rage::fwEntity*))
+{
+	auto handler = fx::ScriptEngine::GetNativeHandler(nativeHash);
+
+	if (!handler)
+	{
+		return;
+	}
+
+	fx::ScriptEngine::RegisterNativeHandler(nativeHash, [handler, breakOffRotorFunc](fx::ScriptContext& ctx)
+	{
+		handler(ctx);
+
+		const auto rotorHealth = ctx.GetArgument<float>(1);
+		if (rotorHealth > 0.0f && !std::isnan(rotorHealth))
+		{
+			return;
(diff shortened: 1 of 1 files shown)


ImBaphomettt:
This seems to fix my previous report. Great work !

https://github.com/citizenfx/fivem/issues/2997

code/components/extra-natives-five/src/NativeFixes.cpp
```diff
@@ -20,6 +20,29 @@
...
+	VEHICLE_TYPE_BLIMP,
+	VEHICLE_TYPE_AUTOGYRO,
+	VEHICLE_TYPE_BIKE,
+	VEHICLE_TYPE_BICYCLE,
+	VEHICLE_TYPE_BOAT,
+	VEHICLE_TYPE_TRAIN,
+	VEHICLE_TYPE_SUBMARINE,
+};
+
+static int g_vehicleTypeOffset;
```
Nobelium-cfx, line 44: Nit: lets move it close to `SetHeliRotorHealthHandler`. Seems like thats the style this file follows.

code/components/extra-natives-five/src/NativeFixes.cpp
```diff
@@ -540,13 +563,66 @@ static void FixMissionTrain()
...
+static void SetHeliRotorHealthHandler(const uint64_t nativeHash, void(*breakOffRotorFunc)(rage::fwEntity*))
+{
+	auto handler = fx::ScriptEngine::GetNativeHandler(nativeHash);
+
+	if (!handler)
+	{
+		return;
+	}
+
+	fx::ScriptEngine::RegisterNativeHandler(nativeHash, [handler, breakOffRotorFunc](fx::ScriptContext& ctx)
```
Nobelium-cfx, line 579: Nit: I think this part could be a bit more readable if we got rid of some of the nesting below and replace `reinterpret_cast<char*>` with just `(char*)` (seems like this file uses both styles).

WDYT about something like:
```
		handler(ctx);

		const auto rotorHealth = ctx.GetArgument<float>(1);
		if (rotorHealth > 0.0f && !std::isnan(rotorHealth))
		{
			return;
		}

		const auto vehicleHandle = ctx.GetArgument<uint32_t>(0);
		const auto vehicle = rage::fwScriptGuid::GetBaseFromGuid(vehicleHandle);
		if (vehicle && vehicle->IsOfType<CVehicle>())
		{
			const auto vehicleType = *reinterpret_cast<eVehicleType*>((char*)vehicle + g_vehicleTypeOffset);
			if (vehicleType == VEHICLE_TYPE_HELI || vehicleType == VEHICLE_TYPE_BLIMP)
			{
				breakOffRotorFunc(vehicle);
			}
		}
```

Nobelium-cfx (commented):
Thank you for the contribution. Thats a really nice change!

Just commented on some minor style nits.

tens0rfl0w:
> Just commented on some minor style nits.

Thanks for the review! Implemented the requested changes, let me know if that works.

Nobelium-cfx (approved):

PR #3089, opened 2025-01-23
fix(extra-natives/five): xmas snow grip factor natives

tens0rfl0w:
### Goal of this PR
The previous implementation hooked the subtraction instruction (`subss`) instead of the initial multiplication instruction (`mulss`) bypassing the total snow amount multiplier. It also caused game crashes for NPC controlled vehicles when calculating wheel collision vectors and had no impact on player controlled vehicles.

We now hook the initial multiplication instruction which keeps reference to the total snow amount (which can be overwritten). This calculation occurs in two separate sections, both of which are now patched.

Values are clamped between `0.0` and `2.0` to make sure physics don't get broken.

GameValueStub's used for less code duplication.

FiveM, Natives

### Successfully tested on
**Game builds:** 3407

### Checklist

### Fixes issues
/

### DIFF
--- a/code/components/extra-natives-five/src/VehicleExtraNatives.cpp
+++ b/code/components/extra-natives-five/src/VehicleExtraNatives.cpp
@@ -38,6 +38,7 @@
 
 #include "DeferredInitializer.h"
 #include "EntitySystem.h"
+#include "GameValueStub.h"
 
 using namespace winrt::Windows::Gaming::Input;
 
@@ -242,7 +243,7 @@ static void writeVehicleMemory(fx::ScriptContext& context, std::string_view nn)
 }
 
 static float* PassengerMassPtr;
-static float* SnowGripFactor;
+static GameValueStub<float> SnowGripFactor;
 
 static int StreamRenderGfxPtrOffset;
 static int HandlingDataPtrOffset;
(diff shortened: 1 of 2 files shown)


code/components/extra-natives-five/src/VehicleExtraNatives.cpp
```diff
@@ -1683,7 +1682,7 @@ static HookFunction initFunction([]()
 		g_globalFuelConsumptionMultiplier = 1.f;
 
 		*PassengerMassPtr = 0.05f;
```
Yum1x, line 1684: not related to this commit but that wouldnt override the value from [#_0x1c47f6ac](https://docs.fivem.net/natives/?_0x1C47F6AC) ?

AvarianKnight: Yes thats why its done `OnNetworkKill` i.e. when the user uses `disconnect` it will reset the value back to default (which is 0.05f)

Yum1x: lol I've got tricked by git line collapsing

ext/native-decls/SetVehicleXmasSnowFactor.md
```diff
@@ -8,4 +8,4 @@ game: gta5
 void SET_VEHICLE_XMAS_SNOW_FACTOR(float gripFactor);
 ```
 ## Parameters
-* **gripFactor**: amount of grip strength vehicle wheels have when xmas weather is active, 1.0 being normal weather grip. 0.2 is the default.
\ No newline at end of file
+* **gripFactor**: amount of grip strength vehicle wheels have when xmas weather is active, 0.0 being normal weather grip. 0.2 is the default.
```
Nobelium-cfx, line 11: Nit: is 0.2 still the default? I think default should also be 0, but I'm now not use why it used to be 0.2. If update - dont forget about GET_VEHICLE_XMAS_SNOW_FACTOR as well.

Nobelium-cfx (commented):
Thank you for the change! Looks good to me.

ext/native-decls/SetVehicleXmasSnowFactor.md
tens0rfl0w:
```
auto location = hook::get_pattern<uint32_t>("F3 0F 59 05 ? ? ? ? 48 23 C8", 4);
hook::get_address<float*>(location);
```
still returns `0.2` as the factor here if unchanged. I think the previous implementation just hooked the wrong instruction by accident. As we now properly hook the grip reduce factor this *should* still be `0.2`, maybe I'm missing something here?😄

Nobelium-cfx: Actually, I just confused as of which part we are hooking exactly. Disregard this comment and the one above :)

Nobelium-cfx (approved):

CoreBitX:
how to fix xmas cresh ? i have cresh i have my server on bulid 3407... no fix this...

PR #3093, opened 2025-01-23
feat(citizen-scripting-lua): table.size function

someone005:
### Goal of this PR
Add table.size function to lua runtime so developers can use it to get size of tables. Mostly useful for hex tables because # will always return 0.

By declaring **Lua_TableSize** function and setting it in global **table**.

ScRT: Lua

### Successfully tested on

**Game builds:** .. 

**Platforms:** Linux

### Checklist

- [ x ] Code compiles and has been tested successfully.
- [ x ] Code explains itself well and/or is documented.
- [ x ] My commit message explains what the changes do and what they are for.
- [ x ] No extra compilation warnings are added by these changes.

### DIFF
--- a/code/components/citizen-scripting-lua/src/LuaScriptRuntime.cpp
+++ b/code/components/citizen-scripting-lua/src/LuaScriptRuntime.cpp
@@ -806,6 +806,25 @@ static int Lua_Require(lua_State* L)
 	return luaL_error(L, "module '%s' not found", name);
 }
 
+static int Lua_TableSize(lua_state* L)
+{
+    if (lua_gettop(L) != 1 || !lua_istable(L, 1))
+    {
+        return luaL_error(L, "table expected got %s", lua_typename(L, lua_type(L, 1)));
+    }
+
+    lua_Integer count = 0;
+    lua_pushnil(L);
+    while (lua_next(L, 1))
+    {
+        count++;
+        lua_pop(L, 1);
+    }
+
+    lua_pushinteger(L, count);
+    return 1;
+}
+
 //
 // Scheduling
 //
@@ -1410,6 +1429,14 @@ result_t LuaScriptRuntime::Create(IScriptHost* scriptHost)
 
 	lua_pushcfunction(m_state, Lua_Require);
 	lua_setglobal(m_state, "require");
+
+	lua_getglobal(m_state, "table");
+	if (lua_istable(m_state, -1))
+	{
+		lua_pushcfunction(m_state, Lua_TableSize);
+		lua_setfield(m_state, -2, "size");
+	}
+	lua_pop(m_state, 1);
 
 #if LUA_VERSION_NUM >= 504
 	lua_setwarnf(m_state, Lua_Warn, nullptr);


Mathu-lmn:
Any benchmark to compare the performance of using this instead of doing the size in Lua ?

AvarianKnight:
Is there really a reason to add this vs the user having their own extension functions (similar to how gmod does it)

someone005:
> Any benchmark to compare the performance of using this instead of doing the size in Lua ?

I did a benchmark and c++ version is faster than Lua.

```
Lua table.size: 0.216406 seconds
C++ table.size: 0.0784628 seconds
```

> Is there really a reason to add this vs the user having their own extension functions (similar to how gmod does it)

I would say it can help developers so they don't have to write these functions in every script they make. I know that frameworks such as ESX provide functions like this but people who makes standalone scripts or "multi-framework" have to make their own. It will simply make their work a little easier/faster.

Mathu-lmn:
May I ask what the table size was used for those results ?

FabianTerhorst:
Isn't the recommended way to store the size inside a extra value? Then you do not traverse the table to calculate it.

someone005:
> May I ask what the table size was used for those results ?

```lua
function table.size(tbl)
    local count = 0
    for _ in pairs(tbl) do
        count = count + 1
    end
    return count
end
```

I think there won't be any huge difference with `count+=1` method. If i'm wrong u can correct me.

> Isn't the recommended way to store the size inside an extra value in lua? Then you do not traverse the table to calculate it.

I have tried one thing but i had a problem with implementing it. I only found a way to do this but it's inside lua source code, so I would have to make pr there. If u have some idea how to make it, show/describe it.

AvarianKnight:
> > Is there really a reason to add this vs the user having their own extension functions (similar to how gmod does it)
> 
> I would say it can help developers so they don't have to write these functions in every script they make. I know that frameworks such as ESX provide functions like this but people who makes standalone scripts or "multi-framework" have to make their own. It will simply make their work a little easier/faster.

My main concern here is that people will constantly want more and more functions added for convenience which adds more possible breakage points for Cfx if there's ever another Lua version upgrade.

thelindat:
It just seems so unnecessary. It's not hard to implement in your own resource or load from an external resource.

> I think there won't be any huge difference with count+=1 method.

I would assume he meant how large was the table being tested, rather than what was the Lua function.

someone005:
> It just seems so unnecessary. It's not hard to implement in your own resource or load from an external resource.

Of course it's not hard but implementing it for the x time is frustrating.

> I would assume he meant how large was the table being tested, rather than what was the Lua function.
10.000.000

someone005:
> My main concern here is that people will constantly want more and more functions added for convenience which adds more possible breakage points for Cfx if there's ever another Lua version upgrade.

So their functions will simply be assessed in terms of usefulness as this one.

> It just seems so unnecessary. It's not hard to implement in your own resource or load from an external resource.

Of course it's not hard but implementing it for the x time is frustrating.

> I would assume he meant how large was the table being tested, rather than what was the Lua function.

10.000.000

PR #3098, opened 2025-01-25
feat(citizen-scripting-lua): ClearTimeout function

poco8537:
### Goal of this PR

Adds a ClearTimeout function to the Lua runtime so that scripts can cancel a timer that was previously set with SetTimeout.

Implements ClearTimeout which removes a scheduled timer from the queue.

ScRT: Lua

### Successfully tested on

**Game builds:** .. 

**Platforms:** Windows

### Checklist

### DIFF
--- a/code/components/citizen-scripting-lua/include/LuaScriptRuntime.h
+++ b/code/components/citizen-scripting-lua/include/LuaScriptRuntime.h
@@ -330,6 +330,16 @@ public:
 
 	lua_State* GetRunningThread();
 
+	lua_State* GetState()
+	{
+		return m_state.Get();
+	}
+
+	auto& GetPendingBookmarks()
+	{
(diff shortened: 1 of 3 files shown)


martonp96 (approved):
Looks good, thanks

outsider31000:
There is an issue that if you add a `Wait` the `SetTimeout callback` won't be canceled same happens when using `ClearTimeout` inside a `CreateThread`

```lua
RegisterCommand("timeout", function()
    local id = SetTimeout(2000, function(id)
        print("Hello", id) -- the id here is not the same as waht it was returned
    end)
    print("clearing timeout", id)
    Wait(0) -- or use a CreateThread to clear the timeout, same result...
    ClearTimeout(id)
end, false)
```

PR #3099, opened 2025-01-25
fix(gta-core-five): crash when get invalid player bone matrix

### LINKED ISSUE #2963: fivem.exe+13CBFF9 - pluto-jupiter-low Crash
### What happened?

Players are getting random crashes while playing on server attached is numerous dumps from multiple timeframes and players.
[CfxCrashDump_2024_11_25_02_11_14.zip](https://github.com/user-attachments/files/17913316/CfxCrashDump_2024_11_25_02_11_14.zip)
[CfxCrashDump_2024_11_26_00_51_59.zip](https://github.com/user-attachments/files/17913317/CfxCrashDump_2024_11_26_00_51_59.zip)
[CfxCrashDump_2024_11_26_05_31_59.zip](https://github.com/user-attachments/files/17913318/CfxCrashDump_2024_11_26_05_31_59.zip)
[CfxCrashDump_2024_11_26_05_21_22.zip](https://github.com/user-attachments/files/17913319/CfxCrashDump_2024_11_26_05_21_22.zip)
[CfxCrashDump_2024_11_26_03_58_44.zip](https://github.com/user-attachments/files/17913320/CfxCrashDump_2024_11_26_03_58_44.zip)
[CfxCrashDump_2024_11_24_05_53_27.zip](https://github.com/user-attachments/files/17913321/CfxCrashDump_2024_11_24_05_53_27.zip)
[CfxCrashDump_2024_11_24_05_53_29.zip](https://github.com/user-attachments/files/17913322/CfxCrashDump_2024_11_24_05_53_29.zip)
[CfxCrashDump_2024_11_24_04_24_37.zip](https://github.com/user-attachments/files/17913323/CfxCrashDump_2024_11_24_04_24_37.zip)

### Expected result

Should not crash

### Reproduction steps

Random, not sure any way to reproduce

### Importancy

Crash

### Area(s)

FiveM

### Specific version(s)

fx: b9956 gamebuild: 3095

### Additional information

_No response_

### LINKED ISSUE #3095: New CRASH need URGENT FIX
### What happened?

Hello there is actually new type of crash that can crash all players on server here is cfx dump so please fix it fast

[CfxCrashDump_2025_01_25_01_05_03.zip](https://github.com/user-attachments/files/18544358/CfxCrashDump_2025_01_25_01_05_03.zip)

### Expected result

they should not crash

### Reproduction steps

Player comes to you and crash you

### Importancy

Crash

### Area(s)

FiveM

### Specific version(s)

tx: v7.3.2 | fx: b12135

### Additional information

_No response_

DaniGP17:
### Goal of this PR
It should fix a crash caused when trying to get the global matrix of a bone in a skeleton, I guess it can occur when trying to get the matrix of a bone id that does not exist.

I haven't been able to test on my local server if this works correctly because I don't know how cheaters do it but it should work. This is my first time doing this type of fix so I don't know if what I've done is a good solution.

It hooks the GetGlobalMtx function and checks if m_Objects and all columns are null, if they are valid the original function is called.

FiveM

### Successfully tested on
**Game builds:** 1604, 2060, 2189, 3095, 3258

**Platforms:** Windows

### Checklist

### Fixes issues
fixes #3095
fixes #2963

### DIFF
--- a/code/components/gta-streaming-five/src/PatchSkeletonCrash.cpp
+++ b/code/components/gta-streaming-five/src/PatchSkeletonCrash.cpp
@@ -162,6 +162,44 @@ static HookFunction hookFunction([]()
 		}
 	}
 
+	// crSkeleton::_GetGlobalMatrix (patch bone index out-of-bounds)
+	{
+		auto location = hook::get_pattern<char>("48 81 EC ? ? ? ? 4D 8B D0 4C 8B 41");
+
+		static struct : jitasm::Frontend
+		{
+			uintptr_t retSuccess;
+			uintptr_t retFail;
+
+			void Init(uintptr_t location)
+			{
+				retSuccess = location + 7;		 // mov    r10, r8
+			}
+
+			void InternalMain() override
+			{
+				sub(rsp, 0xD0);			 // rsp, 0xD0 (original code)
+
+				mov(r10, qword_ptr[rcx + 0x20]); // mov  r10, QWORD PTR [rcx+0x20] (m_boneCount)
+				movsxd(rdx, edx);				 // movsxd rdx, edx (boneIdx)
+				cmp(rdx, r10);			 // cmp  rdx, r10(boneIdx >= m_boneCount)
+				jge("crash");
+				
+				mov(r10, retSuccess);			 // mov    r10, r8
+				jmp(r10);						 // jmp    r10
+
+				L("crash");
+				add(rsp, 0xD0);			 // rsp, 0xD0 (Clean up stack)
+				pop(rbp);						 // pop   rbp (Restore base pointer)
+				ret();
+			}
+		} patchStub;
+
+		patchStub.Init(reinterpret_cast<intptr_t>(location));
+
+		hook::jump(location, patchStub.GetCode());
+	}
+
 #if 0
 	// fwEntity::_GetGlobalMatrix: If required hook the dynamic-dispatch bits.
 	{


florijanftw:
Hi, can you share any code that we can implement right now in fivem?

d22tny:
> Hi, can you share any code that we can implement right now in fivem?

.. the fix stands in gta-core, there's nothing you can implement to fix it on your side.

Gogsi:
I briefly took a look at the crash a few days ago and came to the same conclusion as you about why it happens. However I'm not sure this is the cleanest fix, because this function probably gets called hundreds of times every frame. Hooks are kinda slow and on top of that you have multiple branches or a call back to the original function.

Despite all of this I'm not sure what the best fix is. Maybe the GetGlobalMtx method can somehow be patched in place to add a conditional jump? Can you try to do a benchmark with and without your fix to see if performance is actually affected, maybe I'm overestimating how expensive it is? Both from custom builds would be preferable to make sure adhesive or compiler settings aren't interfering.

Gogsi:
> > Hi, can you share any code that we can implement right now in fivem?
> 
> .. the fix stands in gta-core, there's nothing you can implement to fix it on your side.

Until the proper fix is implemented, as a temporary hack maybe you can solve it by just kicking players that don't have a human player model. That is of course if your server doesn't have some functionality to switch your player model, otherwise you'll get false positives

DaniGP17:
> I briefly took a look at the crash a few days ago and came to the same conclusion as you about why it happens. However I'm not sure this is the cleanest fix, because this function probably gets called hundreds of times every frame. Hooks are kinda slow and on top of that you have multiple branches or a call back to the original function.
> 
> Despite all of this I'm not sure what the best fix is. Maybe the GetGlobalMtx method can be patched in-place to add a conditional jump? Can you try to do a benchmark with and without your fix to see if performance is actually affected, maybe I'm overestimating how expensive it is? Both from custom Release builds would be preferable to make sure adhesive or compiler settings aren't interfering.


Yes, this function is being executed many times, I mostly assume it is because of FootIK, it is a good idea to directly patch the function, I will try to do it :sweat_smile:

florijanftw:
> > > Hi, can you share any code that we can implement right now in fivem?
> > 
> > 
> > .. the fix stands in gta-core, there's nothing you can implement to fix it on your side.
> 
> Until the proper fix is implemented, as a temporary hack maybe you can solve it by just kicking players that don't have a human player model. That is of course if your server doesn't have some functionality to switch your player model, otherwise you'll get false positives

Instructing thing is that player actually as you said did not have a human model, he was a fish if I remember correctly. So that is the way to tell who he is, right? Does he also have to be that fish to actually start crashing? Maybe we can fix it by just prohibiting that “fish” model.

And also I want to mention that yes we ban players like this, but for 1 person to have ability to disrupt like big server is for me crazy that can happen and still is not fixed 1 month later. So we have to be alert 0-24 for a kid that bought Lumia and has this crasher, I mean Its hard for me to wrap my head around it.

Gogsi:
> > > > Hi, can you share any code that we can implement right now in fivem?
> > > 
> > > 
> > > .. the fix stands in gta-core, there's nothing you can implement to fix it on your side.
> > 
> > 
> > Until the proper fix is implemented, as a temporary hack maybe you can solve it by just kicking players that don't have a human player model. That is of course if your server doesn't have some functionality to switch your player model, otherwise you'll get false positives
> 
> Instructing thing is that player actually as you said did not have a human model, he was a fish if I remember correctly. So that is the way to tell who he is, right? Does he also have to be that fish to actually start crashing? Maybe we can fix it by just prohibiting that “fish” model.
> 
> And also I want to mention that yes we ban players like this, but for 1 person to have ability to disrupt like big server is for me crazy that can happen and still is not fixed 1 month later. So we have to be alert 0-24 for a kid that bought Lumia and has this crasher, I mean Its hard for me to wrap my head around it.

The crash I saw was done with the fish model but I suspect other models will also work. Kicking/banning non-human models is probably better than targeting just the fish. Also I recommend at first kicking and logging the player, just to make sure you don't accidentally ban innocent people.

topordenis:
The problem arises from the task driveby -> vehicle gun, which relies on the vehicle's task to fire a gun, combined with the removal of mounted weapons from the inventory. When this happens, the system defaults to the player's firing logic. However, the firing position is determined based on the upper arm (derived from the matrix), which vehicles, of course, don’t have.

florijanftw:
> The problem arises from the task driveby -> vehicle gun, which relies on the vehicle's task to fire a gun, combined with the removal of mounted weapons from the inventory. When this happens, the system defaults to the player's firing logic. However, the firing position is determined based on the upper arm (derived from the matrix), which vehicles, of course, don’t have. //skeleton data is not streamed over network

So I will make a script that will just switch back to original PED if cheater tries to use fish or any other non-human entity, I guess that will fix the problem right.

tens0rfl0w:
I'm not sure if this patch will have the desired effect. After the early return, the callee will still reference `void* *outMtx`, which will lead to undefined behavior as the matrix datatype has no default initialization. To properly address this issue, the patch would need to modify the previous caller on the stack to ensure `outMtx` is handled correctly before the early return. (Maybe some sane default value instead of returning "nothing" will also work)

DaniGP17:
> I'm not sure if this patch will have the desired effect. After the early return, the callee will still reference `void* *outMtx`, which will lead to undefined behavior as the matrix datatype has no default initialization. To properly address this issue, the patch would need to modify the previous caller on the stack to ensure `outMtx` is handled correctly before the early return. (Maybe some sane default value instead of returning "nothing" will also work)

Not sure about this, but I think the `outMtx` is having a default value, I've tested without modifying it and the FootIK looks like is taking 0, 0, 0 or some specific position.
image

tens0rfl0w:
I was able to reproduce the crash in this specific section by spoofing the model hash in the `CPlayerAppearanceDataNode`:
lol
(lol)

Unfortunately, the proposed fix did not prevent the crashes in this area as intended.

After further investigation and experimenting with the hook (especially considering that we already have other hooks in place for similar calls, though specific), I believe I have found a solution that prevents crashes in this section:
https://github.com/citizenfx/fivem/commit/dced67dab9665064863dfde065f6ce1d79fa2ec8
(May require some additional cleanup).

DaniGP17:
> After further investigation and experimenting with the hook (especially considering that we already have other hooks in place for similar calls, though specific), I believe I have found a solution that prevents crashes in this section: [3e05874](https://github.com/citizenfx/fivem/commit/3e05874d50419446067bb2ee435709a98565f214) (May require some additional cleanup and maybe some different approach than using a trampoline regarding performance concerns).

Nice, the different approach refers to patching the function as Gogsi said?

PR #3113, opened 2025-01-30
feat(extra-natives-five): add DrawCorona native

DaniGP17:
### Goal of this PR
Added native `DRAW_CORONA` to render more advanced light effects, like volumetric lights.

image

Creating the `CCoronas` structure that contains the list of coronas, and a method to add new ones.
The patterns have been taken from https://github.com/thorium-cfx/fivem/commit/26f7e2d61251f798b87d5aba0750372993e32d22 and the `CF_NO_RENDER_UNDERGROUND` flag from https://github.com/alexguirre/Spotlight/blob/master/Source/Core/Memory/GameStructs.cs#L296

FiveM, Natives

### Successfully tested on
**Game builds:** 3095, 3407

**Platforms:** Windows, Linux

### Checklist

### DIFF
--- a/code/components/extra-natives-five/src/VfxNatives.cpp
+++ b/code/components/extra-natives-five/src/VfxNatives.cpp
@@ -24,6 +24,13 @@ namespace rage
 	using Vec4V = DirectX::XMVECTOR;
 }
 
+static hook::cdecl_stub<void(void*, Vector3*, float, uint32_t, float, float, Vector3*, float, float, float, uint16_t)> addLightCorona([]()
+{
+	return hook::get_call(hook::get_pattern("F3 0F 11 44 24 28 F3 0F 11 7C 24 20 E8 ? ? ? ? E8", 12));
+});
+
+static void* g_coronas;
+
 static hook::cdecl_stub<void(CVehicle*, int32_t, rage::Mat34V*, int32_t*)> _getExhaustMatrix([]()
 {
 	return hook::get_pattern("4D 8B E1 49 8B F0 48 8B D9 44 0F 29 48", -0x33);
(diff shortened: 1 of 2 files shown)


ext/native-decls/DrawCorona.md
```diff
@@ -0,0 +1,51 @@
+---
+ns: CFX
+apiset: client
+game: gta5
+---
+## DRAW_CORONA
+
+```c
+void DRAW_CORONA(float posX, float posY, float posZ, float size, int red, int green, int blue, int alpha, float intensity, float zBias, float dirX, float dirY, float dirZ, float viewThreshold, float innerAngle, float outerAngle, int flags);
+```
```
spacevx, line 10: Maybe you could add a description?

DaniGP17: This looks good? `Allows drawing advanced light effects, known as coronas, which support flares, volumetric lighting, and customizable glow properties.`

spacevx: 👍

code/components/extra-natives-five/src/VfxNatives.cpp
```diff
@@ -24,6 +24,34 @@ namespace rage
 	using Vec4V = DirectX::XMVECTOR;
 }
 
+static hook::cdecl_stub<void(void*, Vector3*, float, uint32_t, float, float, Vector3*, float, float, float, uint16_t)> addLightCorona([]()
+{
+	return hook::get_call(hook::get_pattern("F3 0F 11 44 24 28 F3 0F 11 7C 24 20 E8 ? ? ? ? E8", 12));
+});
+
+struct Corona_t
```
Nobelium-cfx, line 32: Nit: I don't have a strong opinion here - so feel free to discard if you disagree. But I feel like this struct definition and the `CCoronas` below are unnecessary. We only really need the `addLightCorona` function and `g_coronas` which could be just `static void* g_coronas;`. And then we call `addLightCorona(g_coronas, ....)` inside `DRAW_CORONA`. I personally prefer not to add code that is not used at the moment - if needed we can define these types later.

Nobelium-cfx (approved):
Looks great, thank you for the contribution!

Just a few minor typos and nits. PTAL.

code/components/extra-natives-five/src/VfxNatives.cpp
DaniGP17: Done

DaniGP17:
Should be good now

PR #3121, opened 2025-02-02
feat(gta-streaming-five): Increase fwPortalSceneGraphNode pool

DaniGP17:
### Goal of this PR
Increase the size of the `fwPortalSceneGraphNode` pool as suggested in the [forum](https://forum.cfx.re/t/expansion-of-fwportalscenegraphnode-pool/5302881).
Recently it was attempted to add them to the gameconfig(#2695) but as prikolium said they are allocated from another precomputed heap and should be looked for in `fwSceneGraph::InitPools`.
This was already done by blattersturm [here](https://github.com/citizenfx/fivem/commit/ba52c98749a83d901af4dd4f5f415fc71d4611d0) but 2 days later they [reverted](https://github.com/citizenfx/fivem/commit/85be134d37afa561e669a9049c52d162bcf550e8) his change due to `Potential cause for 'missing doors after a while'.` What they did was hook the PortalSceneGraphPool constructor to multiply the size and allocate the memory elsewhere.

What I'm doing is first hooking the constructor call to check in the F8 if the pool sizes are as expected. I get the sizeof the fwPortalSceneGraphNode class, then I read the `poolStorageFlagsSize` which is the size of the memory that `poolStorageSize` and `poolFlagsSize` add together. Then I read the `poolStorageSize` individually without the flags. Once I have this information I have created a variable that is the number of new entries that are going to be added to the pool, in this case 50, and the `poolStorageFlagsSize` is calculated, which is adding the current size of the `fwPortalSceneGraphNode` class multiplied by the amount of addition, then we do the same but with `poolStorageSize`.
Once we have the new `poolStorageFlagsSize`, `poolStorageSize` we put those values ​​in the `AssignStorageMemory` function. Finally we look for the `portalPoolCount` which is the default pool count (400), we add `poolAddition` to it, and finally we put that value from where we have obtained it.

I think it may be a better option to establish each size of these pools from fivem instead of replacing exactly the one we want to modify, this way we will have more control over all of them and not just one, I don't know what you think.

### Waiting For
This is just a test right now and from what I've been able to test it works fine, but I'd like to ask someone where I should put the new pool value, should I get it from gameconfig.xml? Or should I multiply it by some static number?

FiveM

### Successfully tested on
**Game builds:** 1604, 2060, 3095, 3258, 3407

**Platforms:** Windows

### Checklist

### DIFF
--- a/code/components/gta-streaming-five/src/Streaming.cpp
+++ b/code/components/gta-streaming-five/src/Streaming.cpp
@@ -1,5 +1,6 @@
 #include <StdInc.h>
 #include <Hooking.h>
+#include "Hooking.Stubs.h"
 #include <Streaming.h>
 
 static void* g_storeMgr;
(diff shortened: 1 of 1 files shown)


DaniGP17:
I've closed this to create a separate PR that will allow editing most of the SceneGraphNode pools.

PR #3124, opened 2025-02-03
feat(clrcore): add setter for ped's IsInStealthMode

DatGuy1:


DaniGP17:
Could you complete the PR template?

code/client/clrcore/External/Ped.cs
```diff
@@ -2,15 +2,15 @@
 using CitizenFX.Core.Native;
 using System.Security;
 
-#if MONO_V2
-using CitizenFX.Core;
+#if MONO_V2
+using CitizenFX.Core;
```
DaniGP17, line 6: Can you avoid the modification of these lines without any change?

PR #3101, opened 2025-01-26
fix(handling-loader): fix broken physics

MoskalykA:
### Goal of this PR

Correct broken physics by modifying handling with nils values

### How is this PR achieving the goal

Switch from GetArgument to CheckArgument to be able to create an error if the value is not correct

### This PR applies to the following area(s)

FiveM

### Successfully tested on

**Game builds:** 3095

**Platforms:** Windows

### Checklist

- [x] Code compiles and has been tested successfully.
- [x] Code explains itself well and/or is documented.
- [x] My commit message explains what the changes do and what they are for.
- [x] No extra compilation warnings are added by these changes.

### Fixes issues

fixes #3097

PR #3134, opened 2025-02-06
Revert "fix(handling-loader): fix broken physics"

### LINKED ISSUE #3130: Error when using SetVehicleHandling natives with value of 0
### What happened?

The recent merge of PR https://github.com/citizenfx/fivem/pull/3101 stemming from the issue https://github.com/citizenfx/fivem/issues/3097 now prevents the use of `0`/`0.0` as a valid handling value.

From what I can tell from a look into the codebase and some basic knowledge of C++, the game sees `0` the same as `nil` and it functionally acts the same from my testing too. I can use `nil` (in Lua) or `0` interchangeably and get the same results. I understand the original issue's point, I suppose for some handling values `nil` or `0` can break the physics of the vehicle, but ultimately a lot of different values can completely screw up vehicles to the point where they cannot be fixed.

A float value of `0.0` is valid for various different handling values, and is the default value of many, such as `fSuspensionRaise`. An int value of `0` is commonly used for signifying no flags, such as for `strAdvancedFlags`. These also work fine and don't break the vehicle when run in-game.

_TL;DR_: Using `SetVehicleHandlingFloat` with a value (arg index 3) of `0.0`, the game now sees as invalid even though it should be valid. Same happens when running `SetVehicleHandlingInt` with a value of `0`, it throws the same error.

### Expected result

Using SetVehicleHandlingFloat with a value of 0.0 should run without an error.

### Reproduction steps

The following code, attempting to 'reset' `fSuspensionRaise` back to 0.0:
```lua
local vehicle = GetVehiclePedIsIn(PlayerPedId(), false)
SetVehicleHandlingFloat(vehicle, "CHandlingData", "fSuspensionRaise", 0.0)
```

Returns the following error, saying arg 3 is nil:
Image

Here's an example repro for `strAdvancedFlags`, setting a value of `0` to signify no advanced flags should be placed on the vehicle. The error returned is the essentially the same as the above.

```lua
local vehicle = GetVehiclePedIsIn(PlayerPedId(), false)
SetVehicleHandlingInt(vehicle, "CCarHandlingData", "strAdvancedFlags", 0)
```

### Importancy

Crash

### Area(s)

Natives, FiveM

### Specific version(s)

FiveM Client 12746 (Currently canary only)

### Additional information

_No response_

jrgrimshaw:
### Goal of this PR
As outlined in detail in https://github.com/citizenfx/fivem/issues/3130, the [following PR](https://github.com/citizenfx/fivem/pull/3101) attempts to 'validate' against NULL values when using `SetVehicleHandlingFloat` or `SetVehicleHandlingInt` to prevent vehicle physics from breaking. The issue is that NULL values are parsed as 0, meaning that you can also no longer use the value 0 in these natives (which is common).

Furthermore, the issue the PR was intending to fix wasn't actually caused by NULL values; it's isolated to the handling loader setting a high value for `fTractionCurveMax` when 0.0f/NULL is detected ([source](https://github.com/citizenfx/fivem/blob/7bcb23de344a68b53ff0b95e63c49f0a9cb5298a/code/components/handling-loader-five/src/HandlingLoader.cpp#L215-L218)).

Not allowing values of 0 breaks many scripts, including my own. Even if NULL _is_ used, it is parsed as 0 anyway, and therefore the commit doesn't protect against anything & doesn't fix the root issue. You can see this by taking a look at `CheckArgument`; in the case of setting a float, it checks whether the value is equal to an empty constructor of the type, which in the case of a float would be like running `float()`, which returns 0.0f.
https://github.com/citizenfx/fivem/blob/84444fe5a722acb14b3d33bd85f3352871caf0ab/code/components/scripting-gta/include/ScriptEngine.h#L70

Furthermore, the issue the PR was intending to fix wasn't actually caused by NULL values; it's isolated to the handling loader setting a high value for `fTractionCurveMax` when 0.0f/NULL is detected ([source](https://github.com/citizenfx/fivem/blob/7bcb23de344a68b53ff0b95e63c49f0a9cb5298a/code/components/handling-loader-five/src/HandlingLoader.cpp#L215-L218))

Reverts https://github.com/citizenfx/fivem/pull/3101 to allow 0 (same as NULL) values again.

FiveM

### Successfully tested on

**Game builds:** 3258

**Platforms:** Windows, Linux

### Checklist

### Fixes issues
fixes #3130

### DIFF
--- a/code/components/handling-loader-five/src/HandlingLoader.cpp
+++ b/code/components/handling-loader-five/src/HandlingLoader.cpp
@@ -334,17 +334,17 @@ static void SetHandlingDataInternal(fx::ScriptContext& context, CHandlingData* h
 				switch (member->m_definition->type)
 				{
 					case rage::parMemberType::Float:
-						setFloatField(handlingChar, offset, context.CheckArgument<float>(3), handlingField);
+						setFloatField(handlingChar, offset, context.GetArgument<float>(3), handlingField);
 						break;
 
 					case rage::parMemberType::UInt8:
-						*(uint8_t*)(handlingChar + offset) = uint8_t(context.CheckArgument<int>(3));
+						*(uint8_t*)(handlingChar + offset) = uint8_t(context.GetArgument<int>(3));
 						break;
 
 					case rage::parMemberType::UInt32:
 						// every string field (so far) is parsed to an int in memory
 					case rage::parMemberType::String:
-						setIntField(handlingChar, offset, context.CheckArgument<int>(3), handlingField);
+						setIntField(handlingChar, offset, context.GetArgument<int>(3), handlingField);
 						break;
 
 					case rage::parMemberType::Vector3_Padded:


CeebDev:
> It is not the core codebase's responsibility to prevent bad or untested scripting

I dont 100% agree with that, if you go that way, so in every native we just dont do any check on args passed.
But if NULL is parsed as 0.0f in these fields, then it's a GTA problem and I agree.

Imo we should have atleast a message when we pass NULL

DaniGP17:
> Imo we should have atleast a message when we pass NULL

I'm not sure if that's possible, but I do think it's a good idea to have a warning when 0.0 is applied in some handlings, which is the same as null in this case, because it will set a very large value that breaks the physics.
https://github.com/citizenfx/fivem/blob/7bcb23de344a68b53ff0b95e63c49f0a9cb5298a/code/components/handling-loader-five/src/HandlingLoader.cpp#L215-L218

jrgrimshaw:
> > It is not the core codebase's responsibility to prevent bad or untested scripting
> 
> 
> 
> I dont 100% agree with that, if you go that way, so in every native we just dont do any check on args passed.
> 
> But if NULL is parsed as 0.0f in these fields, then it's a GTA problem and I agree.
> 
> 
> 
> Imo we should have atleast a message when we pass NULL

I tested a decent range of different handling values and null/0.0 produce the same effect. 

I took a look at `CheckArgument` to confirm & this is correct. In the case of setting a float, it checks whether the value is equal to an empty constructor of the type, which in the case of a float would be like running `float()`, which returns 0.0f.

https://github.com/citizenfx/fivem/blob/84444fe5a722acb14b3d33bd85f3352871caf0ab/code/components/scripting-gta/include/ScriptEngine.h#L70

So right now, we are validating against using 0.0f. So for example if you wanted to reset a vehicle's suspension height with   `SetVehicleHandlingFloat(vehicle, "CHandlingData", "fSuspensionRaise", 0.0)`, this currently results in an error.

CeebDev:
> > Imo we should have atleast a message when we pass NULL
> 
> I'm not sure if that's possible, but I do think it's a good idea to have a warning when 0.0 is applied in some handlings, which is the same as null in this case, because it will set a very large value that breaks the physics.
> 
> https://github.com/citizenfx/fivem/blob/7bcb23de344a68b53ff0b95e63c49f0a9cb5298a/code/components/handling-loader-five/src/HandlingLoader.cpp#L215-L218

So this is the problem, why set this at 100000000.f and other handling values not ?

DaniGP17:
> > I'm not sure if that's possible, but I do think it's a good idea to have a warning when 0.0 is applied in some handlings, which is the same as null in this case, because it will set a very large value that breaks the physics.
> > https://github.com/citizenfx/fivem/blob/7bcb23de344a68b53ff0b95e63c49f0a9cb5298a/code/components/handling-loader-five/src/HandlingLoader.cpp#L215-L218
> 
> So this is the problem, why set this at 100000000.f and other handling values not ?

Well that's the same behaviour of the game.
image

Cocodrulo:
Smth I dont understand, why a handling parameter cannot be set with 0.0 or 0? What's the reason?

DaniGP17:
> Smth I dont understand, why a handling parameter cannot be set with 0.0 or 0? What's the reason?

Well that depend in the handling but if you look in the else of the screenshot I send you can see a division, if you divide by 0.0 ... :exploding_head:

Cocodrulo:
But using the native I get an error

DaniGP17:
> But using the native I get an error

That's the reason to revert the previous PR, because the code is considering 0.0 as null, if this PR is merged we'll get the previous behavior that is allowing the 0.0 values.

Cocodrulo:
Got it

Cocodrulo:
Or at least, don't allow null, but allow 0.0 and 0

DaniGP17:
> Or at least, don't allow null, but allow 0.0 and 0

https://github.com/citizenfx/fivem/pull/3134#issuecomment-2640125581

PR #3136, opened 2025-02-06
feat(gta-streaming-five): allow to edit scene graph pools

DaniGP17:
### Goal of this PR
Allow editing of most SceneGraphNodes pools, the increase of `fwPortalSceneGraphNode` was suggested in the [forum](https://forum.cfx.re/t/expansion-of-fwportalscenegraphnode-pool/5302881).
Recently it was attempted to add the pool `fwPortalSceneGraphNode` to the gameconfig(#2695) but as prikolium said they are allocated from another precomputed heap and should be looked for in `fwSceneGraph::InitPools`.
This was already done by blattersturm [here](https://github.com/citizenfx/fivem/commit/ba52c98749a83d901af4dd4f5f415fc71d4611d0) but 2 days later they [reverted](https://github.com/citizenfx/fivem/commit/85be134d37afa561e669a9049c52d162bcf550e8) his change due to `Potential cause for 'missing doors after a while'.` What they did was hook the PortalSceneGraphPool constructor to multiply the size and allocate the memory elsewhere.
I don't know if this is the approach that you want to take in order to increase the sizes of these pools.

First of all, I have defined a list of pools so that their sizes can be easily edited.
```cpp
v0 = Allocate(214473LL, 16LL);
qword_7FF6B50437F8 = v0;
AssignStorage(v0, 210864LL);
v1 = (unsigned int)dword_7FF6B5043808 + v0;
v2 = v1 + 800;
v3 = v1;
v4 = v1 + 802;
v5 = v1 + 1802;
v6 = v1 + 1803;
v7 = v1 + 2803;
v8 = v7 + 406;
```
As you can see in the code, by default the game allocates 214473 in memory which is the sum of the pool storage sizes and the pool flag sizes and in AssignStorageMemory the pool flag sizes are sent. Both values ​​are calculated with the data from the `pools` table and set in both arguments.
The memory displacement offsets are modified for each pool (v2, v4, v5, v6, etc.)
Later the value of all poolSizes is changed in each pool constructor.
And finally, other memory offsets of the AssignStorage function are changed.

FiveM

### Successfully tested on
**Game builds:** 1604, 2060, 2189, 3095, 3258, 3407(My tests have been to join the server and have a test hook to see the arguments that are sent to the pool constructor and they seemed correct.)

**Platforms:** Windows

### Checklist

### DIFF
--- a/code/components/gta-streaming-five/src/SceneGraphPools.cpp
+++ b/code/components/gta-streaming-five/src/SceneGraphPools.cpp
@@ -111,13 +111,16 @@ static HookFunction hookFunction([]()
 	hook::put<uint32_t>((char*)clearVisDataLocation + 20, sceneNodeCount * 4);
 
 	// Change pool size of FIRST_ROOM_SCENE_NODE_INDEX
-	
 	std::initializer_list<PatternPair> roomSceneNodeIndexLocations = {
-		{ "41 BC ? ? ? ? 0F BA E0 ? 73 ? 0F BA E0", 2 }, // AddVisibility
-		{ "48 89 5C 24 ? 48 89 6C 24 ? 48 89 74 24 ? 57 41 54 41 55 41 56 41 57 48 83 EC ? 41 BB", 61 } // ProcessEntityContainer
+		{ "48 89 5C 24 ? 48 89 6C 24 ? 48 89 74 24 ? 57 41 54 41 56 48 83 EC ? 4C 8B 91", 77 }, // AddVisibility
+		{ "48 89 5C 24 ? 48 89 6C 24 ? 48 89 74 24 ? 57 41 54 41 55 41 56 41 57 48 83 EC ? 41 BB", 61 }, // ProcessEntityContainer
+		{ "0F B7 42 ? 41 B8 ? ? ? ? 33 D2", 6 } // GetGbufScreenQuadPair
 	};
 	for(auto& entry : roomSceneNodeIndexLocations)
 	{
-		hook::put<uint32_t>(hook::get_pattern(entry.pattern, entry.offset), 1151);
+		hook::put<uint32_t>(hook::get_pattern(entry.pattern, entry.offset),
+			pools[eSceneGraphPool::FW_EXTERIOR_SCENE_GRAPH_NODE].poolSize +
+			pools[eSceneGraphPool::FW_STREAMED_SCENE_GRAPH_NODE].poolSize +
+			pools[eSceneGraphPool::FW_INTERIOR_SCENE_GRAPH_NODE].poolSize - 1);
 	}
 });
--- a/code/components/gta-streaming-five/src/SceneGraphPools.cpp
+++ b/code/components/gta-streaming-five/src/SceneGraphPools.cpp
@@ -83,16 +83,13 @@ static int FIRST_PORTAL_SCENE_NODE_INDEX =
 
 static void AddSkyPortal(rage::fwScanResultsSkyPortals* self, rage::fwPortalSceneGraphNode* skyPortalNode)
 {
-	__int64 v3;
-	__int8 v4;
-	if(self->m_skyPortalCount <= 63)
+	if(self->m_skyPortalCount >= 0 && self->m_skyPortalCount <= 63)
 	{
-		uint16_t portalIndex = skyPortalNode->m_index - FIRST_PORTAL_SCENE_NODE_INDEX;
-		v3 = (unsigned __int64)portalIndex >> 3;
-		v4 = skyPortalFlagArray[v3];
-		if(((unsigned __int8)(1 << (portalIndex & 7)) & v4) == 0)
+		const uint16_t nodeIndex = skyPortalNode->m_index - FIRST_PORTAL_SCENE_NODE_INDEX;
+		const uint8_t bitMask = 1 << (nodeIndex % 8);
+		if((skyPortalFlagArray[nodeIndex/8] & bitMask) == 0)
 		{
-			skyPortalFlagArray[v3] = v4 | (1 << (portalIndex & 7));
+			skyPortalFlagArray[nodeIndex/8] |= bitMask;
 			self->m_skyPortals[self->m_skyPortalCount++] = skyPortalNode;
 		}
 	}
--- a/code/components/gta-streaming-five/src/SceneGraphPools.cpp
+++ b/code/components/gta-streaming-five/src/SceneGraphPools.cpp
@@ -1,5 +1,8 @@
 #include "StdInc.h"
 #include <Hooking.h>
+#include <mutex>
+
+struct PatternPair;
 
 enum class eSceneGraphPool
 {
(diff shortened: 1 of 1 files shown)


code/components/gta-streaming-five/src/SceneGraphPools.cpp
```diff
@@ -0,0 +1,85 @@
...
+	for (const auto& pool : pools) {
+		total += storage ? (pool.poolSize * pool.classSize) : pool.poolSize;
+	}
+	return total;
+}
+
+static HookFunction hookFunction([]()
+{
+	assert(pools[1].poolSize <= 64 && "The pool size of FW_FIXED_ENTITY_CONTAINER can't be greater than 64");
+	auto initPoolsLocation = hook::get_pattern("48 8D B3 ? ? ? ? 48 8B FB");
```
Nobelium-cfx, line 45: Nit: its probably better to use patter that points to the beginning of the function even tho its a bit longer: `48 89 5C 24 ? 48 89 6C 24 ? 48 89 74 24 ? 57 41 54 41 55 41 56 41 57 48 83 EC ? BA ? ? ? ? B9`. This way we won't need to use negative offsets (like line 51) and I think it's a bit more maintainable (e.g. if I want to check if a function was modified - I will often try to search pattern that points to the beginning of it).

Nobelium-cfx (changes requested):
Thank you for the effort. This looks like a great start! I have some comments, let me know WDYT.

1. It seems that there are more places where those sizes are used and therefore probably should be updated as well. E.g. in this function: `4C 8B C1 0F B6 49 ? 85 C9`. But there is seems to be plenty. I didn't check it too deep, so idk if its feasible to track and update them all.
2. This is less of a priority that the 1st point and can be done later. I think it would make sense to move this code into `PoolManagement.cpp` and conditionally increase as we do with other pools based on `fx::PoolSizeManager::GetIncreaseRequest()`.
3. Left some style nits related to this code.

DaniGP17:
Thank you very much for the review, about the way I'm changing all the values making many hook::put is it okay? It seems to me that I am using it a lot and I don't know if there would be another better way.

Nobelium-cfx:
> Thank you very much for the review, about the way I'm changing all the values making many hook::put is it okay? It seems to me that I am using it a lot and I don't know if there would be another better way.

I think `hook::put` ok. Idk any better way to do it and to me it look good as it is.

code/components/gta-streaming-five/src/SceneGraphPools.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+	// Change pool size of FIRST_ROOM_SCENE_NODE_INDEX
+	
+	std::initializer_list<PatternPair> roomSceneNodeIndexLocations = {
+		{ "41 BC ? ? ? ? 0F BA E0 ? 73 ? 0F BA E0", 2 }, // AddVisibility
+		{ "48 89 5C 24 ? 48 89 6C 24 ? 48 89 74 24 ? 57 41 54 41 55 41 56 41 57 48 83 EC ? 41 BB", 61 } // ProcessEntityContainer
+	};
+	for(auto& entry : roomSceneNodeIndexLocations)
+	{
+		hook::put<uint32_t>(hook::get_pattern(entry.pattern, entry.offset), 1151);
```
iridium-cfx, line 121: What does 1151 correspond to? Is it supposed to be hardcoded?

iridium-cfx (commented):
We have also investigated this internally before.
There is at least one other place that needs patching - the function at `83 B9 ? ? ? ? ? 4C 8B CA 4C 8B C1`, which also then indexes into a fixed-size bitset based on the node index.

code/components/gta-streaming-five/src/SceneGraphPools.cpp
DaniGP17: That looks wrong, that should be the sum of some pools. I'll check it later.

code/components/gta-streaming-five/src/SceneGraphPools.cpp
```diff
@@ -0,0 +1,222 @@
...
+		void Init(intptr_t location)
+		{
+			this->location = location + 23;
+		}
+		
+		void InternalMain() override
+		{
+			lea(r8d, dword_ptr[rdx+0x32]);
+			add(rcx, 0x37C);
+			mov(rax, (uintptr_t)memset);
```
iridium-cfx, line 210: Looks like this matches the original code. I guess you meant to update it to work with the modified m_skyPortalFlags location?

code/components/gta-streaming-five/src/SceneGraphPools.cpp
```diff
@@ -0,0 +1,222 @@
...
+	{
+		hook::put<uint32_t>(hook::get_pattern(entry.pattern, entry.offset),
+			pools[eSceneGraphPool::FW_EXTERIOR_SCENE_GRAPH_NODE].poolSize +
+			pools[eSceneGraphPool::FW_STREAMED_SCENE_GRAPH_NODE].poolSize +
+			pools[eSceneGraphPool::FW_INTERIOR_SCENE_GRAPH_NODE].poolSize - 1);
+	}
+
+	// Patch AddSkyPortal
+	auto addSkyPortalLocation = hook::get_pattern("83 B9 ? ? ? ? ? 4C 8B CA 4C 8B C1");
+	static struct : jitasm::Frontend
```
iridium-cfx, line 141: It would be a lot cleaner to just rewrite the whole function in C++

code/components/gta-streaming-five/src/SceneGraphPools.cpp
```diff
@@ -0,0 +1,222 @@
...
+};
+
+static std::unordered_map<void*, std::unique_ptr<uint8_t[]>> skyPortalFlagMap;
+
+static uint8_t* getSkyPortalFlagAddress(void* self)
+{
+	auto& arrayPtr = skyPortalFlagMap[self];
+    
+	if (!arrayPtr) {
+		arrayPtr = std::make_unique<uint8_t[]>((pools[eSceneGraphPool::FW_PORTAL_SCENE_GRAPH_NODE].poolSize + 7) / 8);
```
iridium-cfx, line 58: This would need double-checking, but I think only 1 fwScan is ever run at a time, so you could probably avoid the map lookup entirely?

code/components/gta-streaming-five/src/SceneGraphPools.cpp
DaniGP17: Sorry, I forgot to finish it, it should be fine now

code/components/gta-streaming-five/src/SceneGraphPools.cpp
DaniGP17: Done

code/components/gta-streaming-five/src/SceneGraphPools.cpp
DaniGP17: In the short time I was testing I only saw one instance of that class, but I'm not 100% sure so I made a list containing all the possible arrays. I've already changed it to use the same array.

code/components/gta-streaming-five/src/SceneGraphPools.cpp
```diff
@@ -138,74 +176,23 @@ static HookFunction hookFunction([]()
...
 		}
 		
 		void InternalMain() override
 		{
-			lea(r8d, dword_ptr[rdx+0x32]);
+			mov(rcx, reinterpret_cast<uintptr_t>(skyPortalFlagArray.get()));
+			xor(edx, edx);
+			mov(r8, static_cast<size_t>(FIRST_PORTAL_SCENE_NODE_INDEX));
 			add(rcx, 0x37C);
 			mov(rax, (uintptr_t)memset);
```
iridium-cfx, line 197: I don't think this makes sense still. You want to be calling `memset(skyPortalFlagArray.get(), 0, sizeOfSkyPortalFlags)` (where `sizeOfSkyPortalFlags = (pools[eSceneGraphPool::FW_PORTAL_SCENE_GRAPH_NODE].poolSize + 7) / 8`) so
```
rcx = skyPortalFlagArray.get()
edx = 0
r8 = sizeOfSkyPortalFlags
```

iridium-cfx (commented):
I've added a couple others to review this as well.
It's quite a large change, so I'm not sure if it would be better to put it on a few branch for a while.

prikolium-cfx:
> > It's quite a large change, so I'm not sure if it would be better to put it on a few branch for a while.
> 
> Honestly, at first I had in mind only increasing the size of a pool (#3121), I saw that I could increase the others but then new locations to patch began to appear and now it is a big change, yes, I don't know if you have any suggestions on this subject

Are you ok with idea to push this to experimental branch?

DaniGP17:
> Are you ok with idea to push this to experimental branch?

Yes, it's fine for me

prikolium-cfx:
> > Are you ok with idea to push this to experimental branch?
> 
> Yes, it's fine for me

It will be published to `specific/feature/mlo-limits-increase` as soon as build finish. I will also test it with big MLO packs that was crashing earlier.

DaniGP17:
> It will be published to `specific/feature/mlo-limits-increase` as soon as build finish.

Nice, and if I want to make some changes I push a commit to my branch and that will automatically be published or I've to ask a cfx member to update that?


> I will also test it with big MLO packs that was crashing earlier.

Related to that could you share the MLO packs later(when you have time) so I can test with that?

prikolium-cfx:
> > It will be published to `specific/feature/mlo-limits-increase` as soon as build finish.
> 
> Nice, and if I want to make some changes I push a commit to my branch and that will automatically be published or I've to ask a cfx member to update that?
> 
> > I will also test it with big MLO packs that was crashing earlier.
> 
> Related to that could you share the MLO packs later(when you have time) so I can test with that?

I just have Gabz subscription and use all of them together, so it crash me after 5-10 minutes teleporting between them

DaniGP17:
> I just have Gabz subscription and use all of them together, so it crash me after 5-10 minutes teleporting between them

Oh ok, I have it too, I'll try it

prikolium-cfx:
> > I just have Gabz subscription and use all of them together, so it crash me after 5-10 minutes teleporting between them
> 
> Oh ok, I have it too, I'll try it

And when you want to update your branch - poke me, I will update it on our end.

Whisperisgod:
I have a very high number of maps, if you want, I can help you with the map and give you vps for you to test.

DaniGP17:
> I have a very high number of maps, if you want, I can help you with the map and give you vps for you to test.

Thanks, I appreciate the proposal but the VPS is not necessary. Are you getting a crash when having too many maps?

DaniGP17:
> Related to this, I need to check it out because it seems that some recent change has caused it to crash most of the time in a tunnel inside a mountain.

Already fixed this, now for me is not crashing, tested in 1604 & 3095.
@prikolium-cfx when you have time could you update the experimental branch to fix that crash?

prikolium-cfx:
> > Related to this, I need to check it out because it seems that some recent change has caused it to crash most of the time in a tunnel inside a mountain.
> 
> Already fixed this, now for me is not crashing, tested in 1604 & 3095. @prikolium-cfx when you have time could you update the experimental branch to fix that crash?

Merged your latest changes into feature branch

DaniGP17:
Hi, the latest commit solve another crash, since that commit I haven't received any crashes so I think it could be tested with many mappings.

prikolium-cfx:
Squashed and merged

ESK0:
> hat they did was hook the PortalSceneGraphPool constructor to multiply the size and allocate the memory elsewhere.

Thanks for merging. Doors and doors framers are missing since yesterday :)))

ESK0:
> Squashed and merged

Thanks for merging. Doors and doors framers are missing since yesterday :)))

DaniGP17:
> Thanks for merging. Doors and doors framers are missing since yesterday :)))

Sorry for this, I'm working for finding the issue

PR #3145, opened 2025-02-09
fix(core/rdr3): AMV hardcoded limit causing blinking

### LINKED ISSUE #2821: After a certain amount of custom/edited AMV for MLO/Interiors, vehicles (wagons & trains) start to blink
### What happened?

After around 22 custom/edited AMV for MLO/Interior vehicles (wagons & trains) start to blink 

here some videos of the issue: (sorry for low quality)

https://github.com/user-attachments/assets/42eae846-9ac0-4d15-9343-b79a4549942b

https://github.com/user-attachments/assets/e2516241-28a6-46e2-99f8-254f105f27c4

https://github.com/user-attachments/assets/d17c3f0d-e8e4-4087-890a-2b2754ec824d

https://github.com/user-attachments/assets/d7a0d477-372b-4a13-b16c-2cbb8c17c93a

here a video with/without more than 22 amv

https://github.com/user-attachments/assets/a39e57e4-9fe8-42f7-b2df-975b78c68fca

AMV limits and usage

amv_usage_around_30_custom_amv

5150 it's base amv pool and on both value is lower

It can be related to amv ytd / ymt entries or ymt files stremed but i already merged amv's to be under 22 and veh/trains seems to work OK

Would be great help to fix or make more tests as AMV in Rdr2 are very important for MLO/Interiors and many creators use them, so very soon servers will start to have issues related to wagons/trains blinking

### Expected result

No blinking on veh/wagons/trains

### Reproduction steps

1. Testing with more than 22 amv = blinking wagons and trains
2. Testing with 22 amv or less = NO blinking wagons and trains

(same video as before)

https://github.com/user-attachments/assets/e5585487-47e1-4b2c-b779-548e768bd7bc

Before focusing on AMV I tested Timecycles, different amount/number of custom MLO/Interiors, ydr's, ytd's but then the issue was resumed to AMV's.

More creators such as Ktos and SevenLife tested with same results, and more people/servers start to have issues with this.

### Importancy

Slight inconvenience

### Area(s)

RedM

### Specific version(s)

RedM ver.10011/prod & ver.10048/canary or latest versions

### Additional information

As many creators use now more and more AMV's in their projects in order to make proper interiors/MLO - servers will start to have this issue more frequently if the issue is not fixed, wagons and trains will blink. The only way I managed to workaround on my server was to remove extra AMV's and merge some of them to not exceed and like this wagons and trains are not bilinking anymore, but it is also a bit random, also it can happen that merging amv will make trains blink again or just some wagons/coaches, then i remove and remake amv and does not blink anymore, so that is why issue can be related to amv ytd / ymt entries or ymt files stremed. 
Hope the information I bring will help to understand the issue atm. Thank you for your help in fixing this, for further tests or anything im available to help.

UPDATE fev2025

The merging situation is out off control after a certain amount and even merging AMV's is not working anymore making all trains and wagons to blink...

alexguirre:
### Goal of this PR

There is a global bitset used to store the enabled/disabled state of every AMV. This bitset has a hardcoded size of 6016 entries, once there are more AMVs than the bitset size, it starts reading/writing out-of-bounds. This causes the AMVs to start blinking and can potentially crash.

This patch re-allocates that bitset with enough space to fit the capacity of all the AMV pools. The functions that access it are patched to point to the new buffer and use the new size constants.

RedM

### Successfully tested on

**Game builds:** 1491, 1436, 1355, 1311

**Platforms:** Windows

### Checklist

### Fixes issues

Fixes #2821.

### DIFF
--- a/code/components/gta-core-rdr3/src/PatchAMV.cpp
+++ b/code/components/gta-core-rdr3/src/PatchAMV.cpp
@@ -1,0 +1,66 @@
+#include <StdInc.h>
+#include <Hooking.h>
+#include <Hooking.Patterns.h>
+#include <Pool.h>
+
+//
+// There is a global bitset used to store the enabled/disabled state of every AMV. This bitset has a hardcoded size
+// of 6016 entries, once there are more AMVs than the bitset size, it starts reading/writing out-of-bounds. This causes
+// the AMVs to start blinking and can potentially crash.
+// 
+// Here we increase the size of that bitset to fit the capacity of all the AMV pools.
+//
+
+static HookFunction hookFunction([]()
+{
+	// keep in sync with gameconfig
+	constexpr size_t CAmbientMaskVolume_PoolSize = 6686;
+	constexpr size_t CAmbientMaskVolumeDoor_PoolSize = 1850;
+	constexpr size_t CAmbientMaskVolumeEntity_PoolSize = 150;
+
+	constexpr size_t TotalAMVs = CAmbientMaskVolume_PoolSize + CAmbientMaskVolumeDoor_PoolSize + CAmbientMaskVolumeEntity_PoolSize;
+
+	// Due to how AMVPresentBuffer is compiled, the number of blocks needs to match the form (N*32+28). AMVPresentBuffer copies the bitset
+	// from update thread to the render thread bitset. The copy loop is vectorized to copy 32 blocks per iteration, then has an epilogue to
+	// copy the remaining 28 blocks, which works out for the original 188 blocks. When resizing the bitset we need to pad the number of blocks
+	// so it doesn't read/write out-of-bounds.
+	constexpr size_t BitsetNumBlocksNoPadding = (TotalAMVs + 31) / 32; // minimum number of 32-bit blocks needed by the AMVs
+	constexpr size_t BitsetPresentCopyNumIterations = (BitsetNumBlocksNoPadding - 28 + 31) / 32; // smallest N such that (N*32+28) >= BitsetNumBlocksNoPadding
+	constexpr size_t BitsetNumBlocks = BitsetPresentCopyNumIterations * 32 + 28;
+	constexpr size_t BitsetNumBytes = BitsetNumBlocks * sizeof(uint32_t);
+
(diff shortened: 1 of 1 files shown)


alexguirre:
I simplified the patch by re-allocating the bitset buffer instead, should be less fragile than the stubs. Though now it hardcodes the AMV pool sizes, I didn't find any existing function to read the sizes from gameconfig and `GetPoolBase` is not initialized at that point. Let me know if there is a better way.

Ktos93:
Patch tested, verified, and working as expected. All tests passed, and the build is successful. Ready to merge.
image

code/components/gta-core-rdr3/src/PatchAMV.cpp
```diff
@@ -0,0 +1,66 @@
...
+// of 6016 entries, once there are more AMVs than the bitset size, it starts reading/writing out-of-bounds. This causes
+// the AMVs to start blinking and can potentially crash.
+// 
+// Here we increase the size of that bitset to fit the capacity of all the AMV pools.
+//
+
+static HookFunction hookFunction([]()
+{
+	// keep in sync with gameconfig
+	constexpr size_t CAmbientMaskVolume_PoolSize = 6686;
```
prikolium-cfx, line 17: Currently we don't allow to increase these 3 pools and such way is safe, but what should we do in future when anybody need to extend one of these pools?

alexguirre: > Currently we don't allow to increase these 3 pools and such way is safe, but what should we do in future when anybody need to extend one of these pools?

In case of enabling `increase_pool_size` with these pools, looking up the increase values in `fx::PoolSizeManager::GetIncreaseRequest()` and adding them to these pool size constants would be enough. The increase request map is already initialized when this hook function runs, so should be straightforward to add.

> And what will happen if we define bigger value here than real pool size? Probably we can set here maximum possible value, for example double pool size, then it's fine.

Nothing, just a bit of unused memory.

outsider31000:
@alexguirre  when you can you should squash the commits

PR #3149, opened 2025-02-11
tweak(gta-five-net): network synchronised scene id hacks

### LINKED ISSUE #2910: Network synchronized scenes will fail for remote clients with >3 clients present.
### What happened?

When running network synchronized scenes, remote clients will occasionally observe the player ped in the scene "fall out" of the scene, either by going into idle pose, or falling about 3 meters down (through collision)

This does not occur for just two clients, and appears to worsen the more clients that are there.

### Expected result

Network synchronized scenes should not fail for remote clients.

### Reproduction steps

1. Have four or more clients in one area
2. Have one, or all run synchronized scene(s) with or without prop elements
3. Observe the scenes aborting/failing/exiting.

Footage of four clients running scenes from the perspective of two. The scene has no props and only characters.
https://github.com/user-attachments/assets/6569ba32-cd37-4fba-a17a-2ac65346b96a
Code for above provided: https://gist.github.com/Lucas7yoshi/c9ff818f55ecd3b5540e0b3a3d30366e

Footage of four clients running scenes from the perspective of two. With props, same issue.

https://github.com/user-attachments/assets/4db9da59-f51d-48fa-8fd8-b272bdc45237

### Importancy

Unknown

### Area(s)

FiveM, FXServer, OneSync

### Specific version(s)

FiveM, All game builds, all game servers

### Additional information

When a character falls out of the scene for remote players, *all* remote players see that happen in sync. On the executing clients end, nothing is wrong.

This happens even with networksynchronizedscenes which contain only one ped, and no objects.

This seems to worsen with a lot of scenes firing off (as is often required for certain mechanics, especially interactive sequences seen in GTA Online like casino lockboxes and grabbing cash anims which will trigger like a scene a second.)

"Array handlers" were mentioned as being the culprit of this when i had mentioned it prior.

tens0rfl0w:
### Goal of this PR
The original game code assigns global IDs for synchronized network scenes by using the physical player index as a unique identifier. The ID space is padded to a maximum of 64 concurrent synchronized scenes `(physicalIndex * 64)`. However, in our network implementation, the local player's physical index is always set to `128`, leading to scene ID collisions when multiple machines trigger networked scenes simultaneously and also integer range overflows (`13-bit = 8191 / 64 = 127 < 128`).

1. Expansion of Scene ID Serialization:
    - The local scene ID variable was already of type `int32_t`, but the serialization process previously limited scene IDs to 13 bits.
    - This patch expands the serialization logic to utilize the full 32-bit space for scene IDs.

2. Replacement of Physical Player Index with Server ID:
    - Instead of using the physical player index (128 for all local players), we now use the player's server-id as an uniquifier.

FiveM, Server

### Successfully tested on
**Game builds:** 1, 1604, 2060, 2189, 2372, 2545, 2612, 2699, 2802, 2944, 3095, 3258, 3407

**Platforms:** Windows

### Checklist

### Fixes issues
fixes #2910

### DIFF
--- a/code/components/citizen-server-impl/src/state/ServerGameState.cpp
+++ b/code/components/citizen-server-impl/src/state/ServerGameState.cpp
@@ -5747,12 +5747,11 @@ struct CNetworkPtFXEvent
 
 struct CRequestNetworkSyncedSceneEvent
 {
-	uint16_t sceneId;
+	uint32_t sceneId; // Increased from uint16_t, see "NetworkSynchronisedSceneHacks.cpp"
 
 	void Parse(rl::MessageBufferView& buffer)
 	{
-		// FIXME: Scene ID length-hack workaround, see `CStartNetworkSyncedSceneEvent`.
-		sceneId = buffer.Read<uint16_t>(8) | (buffer.Read<uint16_t>(5) << 8);
+		sceneId = buffer.Read<uint32_t>(32);
 	}
 
 	inline std::string GetName()
@@ -5839,7 +5838,7 @@ private:
 	};
 
 public:
-	uint16_t sceneId;
+	uint32_t sceneId; // Increased from uint16_t, see "NetworkSynchronisedSceneHacks.cpp"
 	uint32_t startTime;
 
 	bool isActive;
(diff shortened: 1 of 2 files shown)


n3xuuu:
any news on this pr?

prikolium-cfx:
Thanks for your contribution but this change is going to break network protocol and requires net protocol version increase most likely. Let's see what @FabianTerhorst say

FabianTerhorst:
@prikolium-cfx We can merge this if its ok to break the synced scene events for mismatching client and server versions.

prikolium-cfx:
> @prikolium-cfx We can merge this if its ok to break the synced scene events for mismatching client and server versions.

Merge will force all servers to use latest FxServer but we currently have CPU spikes and NodeJS issues so probably it's better to delay that for now.

prikolium-cfx (approved):
PR looks good, but can't be merged right now because or internal reasons. Delayed for now by 2 weeks.

PR #3150, opened 2025-02-12
feat(extra-native-rdr3): `GET_MINIMAP_TYPE`

KadDarem:
### Goal of this PR
Get the minimap type in RDR3

...

Return the value 0,1,2 or 3 depending on whether the map is off, standard, expended, or simple. 
...

RedM natives

...

### Successfully tested on
Get the minimap type after edit it with SetMinimapType(): ✅
Get the minimap type after edit it in the pause menu: ✅

**Game builds:** 1491

**Platforms:** Windows

### Checklist

### Fixes issues
Fixed the missing native to get the minimap type set by SetMinimapType()

### DIFF
--- a/code/components/extra-natives-rdr3/src/HudNatives.cpp
+++ b/code/components/extra-natives-rdr3/src/HudNatives.cpp
@@ -11,6 +11,11 @@ static hook::cdecl_stub<void(void*, char)> g_uiMinimap_SetType([]()
 	return hook::get_pattern("48 81 EC ? ? ? ? 8B F2 48 8B E9 E8", -16);
 });
 
+static hook::cdecl_stub<uint32_t(void*)> g_uiMinimap_GetType([]()
+{
+	return hook::get_call(hook::get_pattern("E8 ? ? ? ? 83 F8 ? 74 ? 48 8D 15"));
+});
+
 static HookFunction hookFunction([]()
 {
 	{
@@ -27,6 +32,12 @@ static HookFunction hookFunction([]()
 
 		if (g_uiMinimap)
 			g_uiMinimap_SetType(g_uiMinimap, minimapType);
+	});
+
+	fx::ScriptEngine::RegisterNativeHandler("GET_MINIMAP_TYPE", [](fx::ScriptContext& context)
+	{
+		uint32_t minimapType = g_uiMinimap_GetType(g_uiMinimap);
+		context.SetResult<int>(minimapType);
 	});
 
 	/*
--- a/ext/native-decls/GetMinimapType.md
+++ b/ext/native-decls/GetMinimapType.md
@@ -1,0 +1,18 @@
+---
+ns: CFX
+apiset: client
+game: rdr3
+---
+## GET_MINIMAP_TYPE
+
+```c
+int GET_MINIMAP_TYPE();
+```
+
+Get the minimap type:
+```
+0 = Off,
+1 = Regular,
+2 = Expanded,
+3 = Simple,
+```


outsider31000:
just a note for the **Cfx team** *if it helps*, I was able to test this and works as intended
although I'm not sure if scanning patterns inside native handlers is a good idea...

@KadDarem  can you please resolve the comments I made? it's now outdated so it doesnt confuse the team, just to make it easier for them to merge this.

outsider31000:
The crash might be occurring because the pattern you're using to locate the address is probably dynamic
and that is also why it works inside the native handler, but this is just my guess...

KadDarem:
Do you have tips to find a better pattern ? 
I have the static pointer inside my Cheat engine. I think my method to convert it into a pattern is wrong.

outsider31000:
I have tried to find it for a couple hours and @Ktos93 also helped , but we weren't able to find one

KadDarem:
I can give you the CE pointer I found if it can help you.
This pointer is valid for the game built from the source code AND for the game from the launcher AND for another guy. Now, we have to found how to read it and it's done.

DaniGP17:
> I can give you the CE pointer I found if it can help you.

Can you share here so other people can try to help?

KadDarem:
All of these pointers trigger the correct value of the minimap type.

DaniGP17:
> All of these pointers trigger the correct value of the minimap type.

Nice, I'll try to help with that.

ekaliclua:
I found a pattern like that, now we have to find the correct way to use it

`48 8B C1 C3 CC 8B 81 A0 ?? ?? ?? C3 90 48 83 EC 28`

KadDarem:
I found a lot of interesting patterns from my CE pointers, but all crash my launcher if I put them outside the function.

DaniGP17:
Hi, I find a function that return the minimap type, the function only receives the instance of the class that we already have declared above in a variable called `g_uiMinimap`.
```cpp
static hook::cdecl_stub<uint32_t(void*)> g_uiMinimap_GetType([]()
{
	return hook::get_call(hook::get_pattern("E8 ? ? ? ? 83 F8 ? 74 ? 48 8D 15"));
});

fx::ScriptEngine::RegisterNativeHandler("GET_MINIMAP_TYPE", [](fx::ScriptContext& context)
{
        uint32_t minimapType = g_uiMinimap_GetType(g_uiMinimap);
        
        context.SetResult<int>(minimapType);
});
```

KadDarem:
Nice ! I will test that to check if it works in my side too.

KadDarem:
It works perfectly, thanks @DaniGP17 .

Can you confirm this refactor resolves your request @outsider31000  ?

DaniGP17:
> It works perfectly, thanks @DaniGP17 .

Glad to hear that :smile:

outsider31000:
seems to work yes it's a static one 
image

about the implementation thats up to `Cfx` wait for their review.

you should squash the commits thought @KadDarem

outsider31000:
I have made some tests and i have to clarify, that this native will only detect the type it has been set either by using the native `SetMinimapType` or through `settings`, it will not detect the current type of the minimap, just so people is aware.

for example:
1- if you press ESC the minimap hides as in OFF type 0 but still returns last type set
2- if you press and hold TAB it expands as in type 2 expanded but still returns last type set

which makes sense. thats how it should work. 

for a detection of current type we would need another native perhaps like `GET_CURRENT_MINIMAP_TYPE` so it detects the current state.

KadDarem:
1. I agree with you; this native is perfect for detecting which setting is used for the minimap. I'm looking for this native because IsRadarHidden() returns `true` if the minimap type is set to 0. It was impossible to detect if the minimap was hidden by the `DISPLAY_RADAR` native or by the `SET_MINIMAP_TYPE` native. 

It's now perfectly possible with this new native. 

2. The native name seems to be right for me. The result matches perfectly with SetMinimapType native or with the settings or with the "Radar" prompts. It's not affected by the game context (tab pressing or pause menu). It's perfect to save the minimap type and be able to apply it again later. 

3. If someone want try to find the `GET_CURRENT_MINIMAP_TYPE` native, feel free to do it in a new PR since this research for `GET_MINIMAP_TYPE` is done. 

PS : I also tested with the "Radar" prompts and the `GET_MINIMAP_TYPE` returns the correct result too.

outsider31000:
yeah this works as intended, just dont forget to squash the commits @KadDarem so it doesn't delay any revivews

KadDarem:
Commits squashed.

Thanks for your help everyone ;)

prikolium-cfx:

PR #3151, opened 2025-02-13
tweak(docs) update building instructions

ook3D:
### Goal of this PR
make instructions easier to understand
...

better wording
...

...

### Successfully tested on

**Game builds:** .. 

**Platforms:** Windows, Linux

### Checklist


docs/building.md
```diff
@@ -18,7 +18,7 @@ To build FiveM, RedM or FXServer on Windows you need the following dependencies:
 * [PowerShell 7](https://aka.ms/powershell-release?tag=stable) or higher.
 * [Python 3.8 or higher](https://python.org/) with the `py` launcher installed.
 * [MSYS2](https://www.msys2.org/) at `C:\msys64\` which is where the installer places it.
-* [Node.js](https://nodejs.org/en/download/) and [Yarn](https://classic.yarnpkg.com/en/docs/install/) in your PATH as `node` and `yarn`.
+* [Node.js](https://nodejs.org/en/download/) and [Yarn](https://classic.yarnpkg.com/en/docs/install/) in your PATH environment variables.
```
jag3dagster, line 21: This feels like 2 steps forward, 1 step back. 

Wouldn't something along the lines of the following be more clear between the two?

> In your environment variables as `node` and `yarn`, respectively.

ook3D: no, because it is incorrect. you should only put them in the path, otherwise you will run into issues with the prebuild

jag3dagster: Ah I see what you're saying. Disregard my original comment.

colistro123: What if people type nodejs as the environment variable instead of just `node`?

PR #3157, opened 2025-02-15
fix(shared): include chrono to fix breakage caused by msvc-170

AvarianKnight:
Microsoft stopped including chrono automatically internally, so we must now include it in our own standard include

https://learn.microsoft.com/en-us/cpp/overview/what-s-new-for-visual-cpp-in-visual-studio?view=msvc-170

FiveM, RedM, Server

### Successfully tested on
**Platforms:** Windows

### Checklist


iridium-cfx (approved):
Looks good, thanks!

PR #3162, opened 2025-02-17
feat(server): add `GET_PLAYER_PEER_STATISTICS` to get useful peer stats

AvarianKnight:
This adds the ability to get packet loss, packet loss variance, and round trip time variance, rtt is already revealed via GET_PLAYER_PING but this still re-adds it for consistency

Having some Prometheus endpoints here might make sense in the future, not sure if adding more endpoints is wanted outside of perf specific stuff though.

Server

### Successfully tested on
**Platforms:** Windows

### Checklist

### DIFF
--- a/code/components/citizen-server-impl/include/GameServerNet.h
+++ b/code/components/citizen-server-impl/include/GameServerNet.h
@@ -17,6 +17,33 @@ enum NetPacketType
 namespace fx
 {
 	class GameServer;
+
+	enum ENetPeerStatistics : uint8_t
+	{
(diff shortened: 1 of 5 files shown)


tabarra:
Cool PR, but why not return everything instead of single items per native call?

AvarianKnight:
> Cool PR, but why not return everything instead of single items per native call?

It could be done this way, I'm not sure what the overhead would be for this vs just a regular native call, it might make sense to have a version that does both.

FabianTerhorst:
Both is fine imo, a single native or a native for each.

FabianTerhorst (approved):

PR #3169, opened 2025-02-19
feat(cfx-ui): allow html in connecting status

csskroubledev:
### Goal of this PR

Add support for using HTML code in connecting status e.g. `deferral.update()`, like rejection status can do (deferral.done()), allowing server developers to create more customized update statuses than Adaptive Cards can provide. (excluding inputs of course).
...

PR is replacing current implementation of showing the status message, to one similar to rejection status.
...

FiveM, RedM
...

### Successfully tested on

**Game builds:** 3258, 3095

**Platforms:** Windows

### Checklist

### Preview
I'm attaching a PoC of my change.

image

### DIFF
--- a/ext/cfx-ui/src/cfx/apps/mpMenu/parts/LegacyConnectingModal/ConnectStatus.tsx
+++ b/ext/cfx-ui/src/cfx/apps/mpMenu/parts/LegacyConnectingModal/ConnectStatus.tsx
@@ -5,11 +5,13 @@ import {
   Modal,
   Text,
   noop,
+  linkify
 } from '@cfx-dev/ui-components';
 import { observer } from 'mobx-react-lite';
 
-import { $L, useL10n } from 'cfx/common/services/intl/l10n';
-import { nl2brx } from 'cfx/utils/nl2br';
+import { $L } from 'cfx/common/services/intl/l10n';
+import { nl2br } from 'cfx/utils/nl2br';
+import { html2react } from 'cfx/utils/html2react';
 
 import { ConnectState } from '../../services/servers/connect/state';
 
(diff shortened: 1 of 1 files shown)


AvarianKnight:
XSS?

csskroubledev:
> XSS?

I was informed that FiveM already prevents script tags from being used, an additionally html is allowed in connection rejection for a long time (used by txAdmin mostly).

I tried using those, and in fact didn’t succeed.

csskroubledev:
> > XSS?
> 
> I was informed that FiveM already prevents script tags from being used, an additionally html is allowed in connection rejection for a long time (used by txAdmin mostly).
> 
> I tried using those, and in fact didn’t succeed.

Just to make sure, I have done some scenarios where an potential bad actor could abuse the HTML ability.

Code that was tested:
image

Video of checking it out:
https://streamable.com/ceo6n8

csskroubledev:
> > > XSS?
> > 
> > 
> > I was informed that FiveM already prevents script tags from being used, an additionally html is allowed in connection rejection for a long time (used by txAdmin mostly).
> > I tried using those, and in fact didn’t succeed.
> 
> Did you try the way how txAdmin format connection status message? https://github.com/tabarra/txAdmin/blob/master/core/routes/player/checkJoin.ts#L26-L45

I might have a problem of understanding what do you mean by your message.
Do you want me to check if the same code that txAdmin uses the reject works for updating statuses too?

If it's about that they sanitize the ban reason, it doesn't matter because every other resource can reject the connection trying to use some XSS stuff, even though they seem to be already sanitized somewhere in between sending data from resource -> cfx-ui, which was proven by my example in https://github.com/citizenfx/fivem/pull/3169#issuecomment-2668457095

salam4n:
any progress here? would be useful tbh

PR #3177, opened 2025-02-20
fix(native-decls): fix native hyperlinks

ook3D:
it appears the hyperlink does not work on the native docs, i believe this is because it was previously in lowercase.

fix SET_GLOBAL_PASSENGER_MASS_MULTIPLIER and SET_VEHICLE_XMAS_SNOW_FACTOR hyperlinks.

### Goal of this PR

Fix hyperlinks between natives.
...

changing the case of the native hashes.
...

FiveM, Natives
...

### Successfully tested on

**Game builds:** .. 

**Platforms:** Windows, Linux

### Checklist

### DIFF
--- a/ext/native-decls/GetGlobalPassengerMassMultiplier.md
+++ b/ext/native-decls/GetGlobalPassengerMassMultiplier.md
@@ -9,7 +9,7 @@ game: gta5
 float GET_GLOBAL_PASSENGER_MASS_MULTIPLIER();
 ```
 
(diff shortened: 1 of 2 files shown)


AvarianKnight (commented):
LGTM, I'm guessing your reasoning here is probably right, these are probably looked up in a map whenever the docs are compiled (which would be case sensitive)

iridium-cfx (approved):
Good spot, thanks.

PR #3182, opened 2025-02-21
fix(gta-core-five): validate bounds collisions

### LINKED ISSUE #3173: Build 12933 and 12947, crashing GTA5_b3258.exe!sub_1414C81F4
### What happened?

We moved from b11427 to b12947, to get ourselves more up to date and get the new Node22, but something is causing players within the same area to crash. The problem did move area a few times, but mostly around pillbox probably due to  high population area, we do have custom MLO's for Pillbox and Casino.  All that we know is something happens and crashes all players in scope or distance from whatever causes the crash, if it's a player causing it we assign the other players have to be within the scope of that player.

We then rolled back to b12933 but the issue still persisted, so we've now just reverted back to b11427 since we can't keep rolling back the build until it doesn't happen because that isn't ideal for the player base.

Multiple of us tried to replicate the issue on our local development instances of fxserver, using the same build and same resources etc, but no joy in being able to recreate the problem.

Crash errors are;
GTA5_b3258.exe!sub_1414C81F4 (0x14) - Seen 952 times
GTA5_b3258.exe!sub_1414C81F4 (0x39) - Seen 370 times

Crashing happened to 400+ users as logged by our logging system

Crash DMP files, hopefully these can aid to what the problem is.

[GTA5_b3258.exe!sub_1414C81F4 (0x14) - 8c52c3e5-a966-4397-b283-1e03145292d1.dmp](https://github.com/user-attachments/files/18874437/8c52c3e5-a966-4397-b283-1e03145292d1.dmp)

[GTA5_b3258.exe!sub_1414C81F4 (0x39) - 793342b0-f620-442c-9361-458a4c9293f9.dmp](https://github.com/user-attachments/files/18874445/793342b0-f620-442c-9361-458a4c9293f9.dmp)

### Expected result

No crashing

### Reproduction steps

Unable to reproduce sadly.

### Importancy

Crash

### Area(s)

FiveM

### Specific version(s)

FiveM

### Additional information

_No response_

DaniGP17:
### Goal of this PR
Fix a game crash when two collisions collide with each other without having a valid m_Bounds. This is happening for specific objects, mostly testing objects that are not used in the map but are spawned by malicious players to make people nearby crash. By default when spawning the prop there is no problem, but if the player or something external hits a sub-object that makes it move and collide with another sub-object of the same object it will reproduce this crash.

Hooking the `ProcessSelfCollision` function and validating that the `boundComposite` and `boundComposite->m_Bounds` are valid

FiveM

### Successfully tested on
**Game builds:** 1, 1604, 2060, 2189, 2372, 2545, 3095, 3258, 3407

**Platforms:** Windows

### Checklist

### Fixes issues
fixes #3173

### DIFF
--- a/code/components/gta-core-five/src/CrashFixes.SelfCollision.cpp
+++ b/code/components/gta-core-five/src/CrashFixes.SelfCollision.cpp
@@ -1,0 +1,59 @@
+#include "StdInc.h"
+
+#include "Hooking.Patterns.h"
+#include "Hooking.Stubs.h"
+#include <DirectXMath.h>
+
+namespace rage {
+	struct alignas(16) Vec3V
+	{
+		float x;
+		float y;
+		float z;
+		float pad;
+
+		Vec3V()
+			: x(0), y(0), z(0), pad(NAN)
+		{
+		}
+
+		Vec3V(float x, float y, float z)
+			: x(x), y(y), z(z), pad(NAN)
+		{
+		}
+	};
+	struct phBound
+	{
+		uint8_t m_Type;
+		uint8_t m_Flags;
+		uint8_t m_PartIndex;
+		float m_RadiusAroundCentroid;
+	};
+	using Mat34V = DirectX::XMMATRIX;
+
+	struct phBoundComposite : phBound {
+		phBound** m_Bounds;
+		rage::Mat34V* m_CurrentMatrices;
+		rage::Mat34V* m_LastMatrices;
+		rage::Vec3V* m_LocalBoxMinMaxs;
+		uint32_t* m_TypeAndIncludeFlags;
+		uint32_t* m_OwnedTypeAndIncludeFlags;
+		uint16_t m_MaxNumBounds;
+		uint16_t m_NumBounds;
+	};
+}
+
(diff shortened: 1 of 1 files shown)


DaniGP17:
> for where you defined the phBound structs, would it be possible to use the already defined ones inside rage-formats-x component to prevent duplicating code? https://github.com/citizenfx/fivem/tree/master/code/components/rage-formats-x/include

I didn't realize that, I'll do it thanks

Gogsi:
> Can the `phBound` structure from the `rage-formats-x` project be easily referenced from `gta-core-five`?

You can add "rage:formats:x" as a dependency in [component.json](https://github.com/citizenfx/fivem/blob/0dec7bf96e953989dfe6e5a5bb15e1288ac6ee47/code/components/gta-core-five/component.json) and rerun `fxd gen -game five`.
That should be enough to reference the project, hopefully it doesn't cause other issues.

DaniGP17:
I've tried using `phBoundComposite` from the `rage-formats-x` project but the `m_childBounds` member is private, should I change it to public or leave the structure as I have it?

ook3D:
> I've tried using `phBoundComposite` from the `rage-formats-x` project but the `m_childBounds` member is private, should I change it to public or leave the structure as I have it?

Probably best to leave it how you have it originally then

xliel:
Is there a way to block this using natives for now until you get approval?

prikolium-cfx (approved):

PR #3183, opened 2025-02-21
fix(citizen-scripting-core): strict mode only for the client

MoskalykA:
### Goal of this PR

The strict state bags mode prevents writing by the server. My PR corrects this.

With a `#ifndef` to check that the execution is done on the client side

Server

### Successfully tested on

**Platforms:** Windows

### Checklist


AvarianKnight:
This isn't the proper way to fix this, you should check if the `m_sbac->GetRole() != StateBagRole::Server`

FabianTerhorst (approved):

PR #3186, opened 2025-02-22
fix(gta-streaming-five): sanitize ragdoll component based in attach bone

### LINKED ISSUE #3185: GTA5_b3258.exe!sub_141066CFC(0x5c5) Crash
### What happened?

when i was using the ‘CS_TaosTranslator’ model, i pulled out my weapon and got hit by a car, then i crashed, and I've seen some cheaters crash logs with this as well, i think this is their crash

### Expected result

fix that

### Reproduction steps

when i was using the ‘CS_TaosTranslator’ model, i pulled out my weapon and got hit by a car, then i crashed, and I've seen some cheaters crash logs with this as well, i think this is their crash

### Importancy

Crash

### Area(s)

FiveM

### Specific version(s)

Fivem b3258

### Additional information

[CfxCrashDump_2025_02_22_11_01_31.zip](https://github.com/user-attachments/files/18922033/CfxCrashDump_2025_02_22_11_01_31.zip)

DaniGP17:
### Goal of this PR
Validate that the component obtained from the attachment index is not -1, which represents an invalid position in the component array.

There are some peds that don't have the same attachment bones as the common ones and when the ragdoll is activated with a weapon it causes it to crash near people.

Patching a part of the `CPedWeaponManager::SwitchToRagdoll` function to validate that the `r13` register representing the component value is not -1.

FiveM

### Successfully tested on
**Game builds:** 1, 1604, 3407

**Platforms:** Windows

### Checklist

### Fixes issues
fixes #3185

### DIFF
--- a/code/components/gta-streaming-five/src/PatchSkeletonCrash.cpp
+++ b/code/components/gta-streaming-five/src/PatchSkeletonCrash.cpp
@@ -179,6 +179,42 @@ static HookFunction hookFunction([]()
 		}
 	}
 
+	// CPedWeaponManager::SwitchToRagdoll
+	{
+		static struct : jitasm::Frontend
+		{
+			intptr_t retSuccess = 0;
+			intptr_t retFail = 0;
+
+			void Init(intptr_t success, intptr_t fail)
+			{
+				this->retSuccess = success;
+				this->retFail = fail;
+			}
+
+			virtual void InternalMain() override
+			{
+				mov(rax, qword_ptr[rdx+0x118]); // Original code
+				movsxd(r13, dword_ptr[rax+rcx*4]); // Original code
+				cmp(r13, -1); // Check if the index of the bone map is -1 that represents a invalid position
+				je("fail");
+				mov(rax, this->retSuccess);
+				jmp(rax);
+				L("fail");
+				mov(rax, this->retFail);
+				jmp(rax);
+			}
+		} patchStub;
+		
+		auto location = hook::get_pattern("48 8B 82 ? ? ? ? 4C 63 2C 88");
+		hook::nop(location, 11);
+		const auto successPtr = reinterpret_cast<intptr_t>(location) + 11;
+		const auto failPtr = reinterpret_cast<intptr_t>(hook::get_pattern("4C 8D 85 ? ? ? ? E8 ? ? ? ? 4C 8D 9C 24 ? ? ? ? 49 8B 5B ? 41 0F 28 73", 12));
+		assert((failPtr - reinterpret_cast<intptr_t>(location)) < 3000);
+		patchStub.Init(successPtr, failPtr);
+		hook::jump(location, patchStub.GetCode());
+	}
+
 #if 0
 	// fwEntity::_GetGlobalMatrix: If required hook the dynamic-dispatch bits.
 	{


iridium-cfx:
The index is used a few times before this in the same function.
Would it be better to just return from the function if it's -1, right after getting the index?

iridium-cfx:
> > The index is used a few times before this in the same function. Would it be better to just return from the function if it's -1, right after getting the index?
> 
> I assume you mean jumping directly here, right? image Is it better to calculate that position with an offset from the current pattern or to get a new pattern for that location?

Yeah, and uh, it's quite a large function, so maybe sig scan with an assert that the distance is vaguely correct.

DaniGP17:
> Yeah, and uh, it's quite a large function, so maybe sig scan with an assert that the distance is vaguely correct.

This is the code looking for the pattern of the fail location:
```cpp
auto location = hook::get_pattern("48 8B 91 ? ? ? ? 48 8D 8D");
hook::nop(location, 7);
const auto successPtr = reinterpret_cast<intptr_t>(location) + 7;
const auto failPtr = reinterpret_cast<intptr_t>(hook::get_pattern("4C 8D 85 ? ? ? ? E8 ? ? ? ? 4C 8D 9C 24 ? ? ? ? 49 8B 5B ? 41 0F 28 73", 12));
trace("CPedWeaponManager::SwitchToRagdoll: %lld\n", static_cast<long long>(failPtr - reinterpret_cast<intptr_t>(location))); // 1375
patchStub.Init(successPtr, failPtr);
hook::jump(location, patchStub.GetCode());
```

But I didn't understand the assert, do you mean something like this?
```cpp
assert((failPtr - reinterpret_cast<intptr_t>(location)) < 2000);
```

iridium-cfx:
> > Yeah, and uh, it's quite a large function, so maybe sig scan with an assert that the distance is vaguely correct.
> 
> This is the code looking for the pattern of the fail location:
> 
> ```c++
> auto location = hook::get_pattern("48 8B 91 ? ? ? ? 48 8D 8D");
> hook::nop(location, 7);
> const auto successPtr = reinterpret_cast<intptr_t>(location) + 7;
> const auto failPtr = reinterpret_cast<intptr_t>(hook::get_pattern("4C 8D 85 ? ? ? ? E8 ? ? ? ? 4C 8D 9C 24 ? ? ? ? 49 8B 5B ? 41 0F 28 73", 12));
> trace("CPedWeaponManager::SwitchToRagdoll: %lld\n", static_cast<long long>(failPtr - reinterpret_cast<intptr_t>(location))); // 1375
> patchStub.Init(successPtr, failPtr);
> hook::jump(location, patchStub.GetCode());
> ```
> 
> But I didn't understand the assert, do you mean something like this?
> 
> ```c++
> assert((failPtr - reinterpret_cast<intptr_t>(location)) < 2000);
> ```

Yeah, that's fine.

DaniGP17:
I think it's better to leave it as it was before because now it seems to skip some function in charge of cleaning the weapon, and it stays floating when I get run over.
image

iridium-cfx:
Oh right, I mean return right after reading the index, `4C 63 2C 88 48 8B 82`, hopefully it won't break with that?

DaniGP17:
It should be good now, tested and seems to be working as expected

what11-ctrl:
When will it be fixed

PR #3044, opened 2024-12-26
feat(extra-natives/five): track switching

Ehbw:
### Goal of this PR

Introduce a way to register track switching. Track switching allows for a train to change the track index it is currently navigating along. Suggested by @glitchdetector 

### How is this PR achieving the goal

introduces ``REGISTER_TRACK_SWITCH`` and ``SET_TRACK_SWITCH_ACTIVE`` to allow for registration of a train switch and ability to enable/disable the track switch

### This PR applies to the following area(s)

FiveM, Natives

### Successfully tested on

A test resource showing the functionality of all new natives and test cases are provided  https://github.com/citizenfx/fivem/pull/3044#issuecomment-2671891066

**Game builds:** 1604, 2060, 3258, 3407

**Platforms:** Windows

### Checklist

- [X] Code compiles and has been tested successfully.
- [X] Code explains itself well and/or is documented.
- [X] My commit message explains what the changes do and what they are for.
- [X] No extra compilation warnings are added by these changes.

PR #3191, opened 2025-02-24
feat(five/streaming): expand train track array

Ehbw:
### Goal of this PR

Expands the hardcoded track limit of 27 tracks to 127 (though can be easily expanded in the future by modifying kMaxTracks in extra-natives-five/TrackNatives.cpp and gta-streaming-five/PatchTrackLimits.cpp.

With the introduction of track switching with https://github.com/citizenfx/fivem/pull/3044 There is a bigger need for having more available tracks (Base GTA takes up 12, Leaving 15 for custom tracks that may be taken up by different map resources and track switching)

By replacing all instances of the original array holding track data and replacing it with a larger allocated array. All checks for trackIndex being less then 27 have also been replaced with kMaxTracks or kMaxTracks - 1. 

Two other track related functions were also rewritten and placed into TrackNatives.cpp because of missing checks. This change only applies to newer game builds (>2545) due to differences with the older game builds that aren't likely to be supported for much longer.

FiveM

### Successfully tested on

Tests were done with the original tracks + custom tracks that exceeded the original 27 limit. Tested on older builds that don't have this patch to ensure the functionality is the same.

For testing. I've provided a resource that contains enough tracks to exceed the original 27 limit. With a client debugging script and commands for testing 
[Extended-Track-Repro.zip](https://github.com/user-attachments/files/18949352/Extended-Track-Repro.zip)

**Game builds:** 2189, 2372, 2545, 2699, 2802, 2944, 3095, 3258, 3407

**Platforms:** Windows

### Checklist

### DIFF
--- a/code/components/extra-natives-five/include/Train.h
+++ b/code/components/extra-natives-five/include/Train.h
@@ -16,9 +16,6 @@ public:
 struct CTrainTrack
 {
 public:
-	// This has never changed.
-	static const int kMaxTracks = 27;
-
 	uint32_t m_hash;
 	bool m_enabled;
 	bool m_isLooped;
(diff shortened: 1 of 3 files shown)


dalekenium-cfx (approved):
No complaints from me. Approved - thanks for the contribution. :)

PR #3193, opened 2025-02-24
tweak(extra-natives/five): track switching improvements

Ehbw:
### Goal of this PR

Fixes the direction check in track switching, an slightly incorrect warning message when attempting to remove train junctions, and correcting velocity on track switch.

Fixes flag check in CTrain::GetDirection, now will properly return the direction, this was missed in the original test cases I had for track switching. 

The trains current speed/velocity is now stored before beginning the track switch and is set back when finishing the switch. There was previous cases where the function used to smooth the transition would set the velocity to a negative speed, launching the train backwards and causing the train to jitter on remote clients.

FiveM

### Successfully tested on

**Game builds:** 3258, 3407

**Platforms:** Windows, Linux

### Checklist


dalekenium-cfx (approved):
Pretty straightforward, lgtm.

PR #3221, opened 2025-03-08
feat(citizen-devtools): added GET_RESOURCE_MONITOR_DATA

Nova-px:
### Goal of this PR

Providing the ability to get data from the resource monitor to visualize it in grafana or something else

Added the `GET_RESOURCE_MONITOR_DATA` native

Server, Natives

### Successfully tested on

**Platforms:** Windows

### Checklist

### DIFF
--- a/ext/native-decls/server/GetResourceMonitorData.md
+++ b/ext/native-decls/server/GetResourceMonitorData.md
@@ -1,0 +1,12 @@
+---
+ns: CFX
+apiset: server
+---
+## GET_RESOURCE_MONITOR_DATA
+
+```c
+object GET_RESOURCE_MONITOR_DATA();
+```
+
+## Return value
+An array containing the resource monitor data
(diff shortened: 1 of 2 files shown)


ext/native-decls/server/GetResourceMonitorData.md
prikolium-cfx, line 1: Could you please add proper structure to the documentation?

prikolium-cfx:
Thank you for this contribution. Could you also add client side version as well?

Nova-px:
> I'm generally curious on why expose like this vs adding more Prometheus endpoints

Also thought about that, but I don't have much knowledge of the codebase and how it should be implemented

prikolium-cfx:
> I'm generally curious on why expose like this vs adding more Prometheus endpoints

But in this case how should we provide client side data?

Nova-px:
> > I'm generally curious on why expose like this vs adding more Prometheus endpoints
> 
> But in this case how should we provide client side data?

Maybe add the native only on client-side and add another Prometheus endpoint on the server-side, I think that would also reduce the performance cost of the implementation to visualize / redirect the data from the native

AvarianKnight:
I don't think that the client side should be included in this PR at all, there should be some proper way for the client to send the server performance metrics if that is wanted, otherwise we'll end up with a weird "everyone makes there own, and there's no nice/standardized way to do it" and it will be a big mess. 

The server should add more endpoints (i.e. `/perf/resources`) and expand any performance/statistics logging from there.

PR #3237, opened 2025-03-13
fix(citizen-server-impl): add the ability to restart categories for command: restart

Kid-Sander:
### Goal of this PR

Add the ability to restart categories for command: restart
...

Added ability to use: “restart [resourceCategory]” like start/stop. Also added a new variation of “restart [resourceCategory] [isPrintStoppedResource]” to avoid the situation where a category may contain a resource that is not running. All other behavioral logic remains the same.

image
...

FiveM, RedM, Server
...

### Successfully tested on

**Platforms:** Windows

### Checklist

### DIFF
--- a/code/components/citizen-server-impl/src/ServerResources.cpp
+++ b/code/components/citizen-server-impl/src/ServerResources.cpp
@@ -742,7 +742,7 @@ static InitFunction initFunction([]()
 			}
 		});
 
-		auto restartCommandRef = [=](const std::string& resourceName, const bool isPrintStoppedResource) 
+		static auto restartCommandRef = instance->AddCommand("restart", [=](const std::string& resourceName) 
 		{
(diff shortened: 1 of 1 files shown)


STACHY225:
You've got typo in `isPrintStopedResource`, in 2nd commit (`Stopped` is missing double `p`).

prikolium-cfx:
Thank you for the contribution. We would like to not add additional complexity to resource manager system so I want to ask you to remove this additional argument and just do print something like `Skipping resource resource_name restart as it is not started`.

Kid-Sander:
Thanks for the review. Returned the standard behavior, when restarting one or more resources, only by changing the old message to a new one, which contains specifics about what resource was trying to restart now. 

An example of this message can be seen on the screenshot in the attachment.
image

PR #3238, opened 2025-03-13
tweak(input/five): usabillity improvements

### LINKED ISSUE #3230: InputHook.cpp: control bypasses not respected
### What happened?

When focus is stolen by NUI all the mouse buttons are reset, even if they shouldn't be. If your push to talk is bound to a mouse key and you are using it when focus is stolen PTT is turned off.

### Expected result

PTT should remain enabled until the key is no longer pressed

### Reproduction steps

1. Set PTT to a mouse key (eg M5)
2. Press it and keep it pressed
3. Trigger NUI to take focus (`SetNuiFocus(true, true)`)
4. Observe that PTT was disabled even though the key is still pressed

### Importancy

There's a workaround

### Area(s)

FiveM

### Specific version(s)

FiveM 13301+

### Additional information

Control bypasses are used to allow PTT to be pressed while NUI is open.

I believe this bug was introduced in #3076 @tens0rfl0w

tens0rfl0w:
### Goal of this PR
This fixes regressions introduced by some previous changes and improves inconsistencies across different input methods.

[tweak(nui/core): disable cursor clipping in main menu](https://github.com/citizenfx/fivem/commit/1cdd2262bb9266129d865a8fdb0a18ad030a05a3) 

Small regression introduced by a previous change done by me (https://github.com/citizenfx/fivem/pull/3083) that enabled cursor clipping in the main menu, which is an unwanted behavior in general.

[tweak(input/five): only flush mouse buttons on full focus loss](https://github.com/citizenfx/fivem/commit/0f9ddca23f842bc1a4a63c42375417b183dfdc9f) 

The current implementation is a bit confusing, when we set 'SetGameMouseFocus' in 'InputHook' to false, there are still cases where inputs get passed through, for example when a NUI has focus but no cursor set. In those cases we don't want to flush the mouse buttons as the game continues to receive input, which would require a button repress to keep the input active.

[tweak(input/five): exclude PTT button from flush on focus loss](https://github.com/citizenfx/fivem/commit/0f9ddca23f842bc1a4a63c42375417b183dfdc9f)

Another small regression introduced by a previous change done by me (https://github.com/citizenfx/fivem/pull/3076) that flushes mouse buttons on focus loss. This successfully fixes stuck mouse button, but also resets the PTT button state on focus loss, which of course isn't intended behavior.

[tweak(nui/core): block X-Buttons on WndProc](https://github.com/citizenfx/fivem/commit/fd6f1d94d0db1a99fcee9ae7972d45d88e26200b) 

This missed case handling for WM_XBUTTONDOWN and similar, causing the game camera to stutter and the PTT state to reset when a NUI had full focus and a user was pressing a X-Button (commonly used for PTT) while the game received other inputs (mouse move, ...). This only affects users that have their input setting set to "Windows".

FiveM

### Successfully tested on

### Checklist

### Fixes issues
fixes #3230

### DIFF
--- a/code/components/citizen-game-main/src/FrontendNui.cpp
+++ b/code/components/citizen-game-main/src/FrontendNui.cpp
@@ -305,7 +305,7 @@ public:
 	{
 	}
 
-	virtual void SetGameMouseFocus(bool val) override
+	virtual void SetGameMouseFocus(bool val, bool flushMouse = true) override
 	{
 	}
 
--- a/code/components/glue/src/GtaNui.cpp
+++ b/code/components/glue/src/GtaNui.cpp
@@ -39,6 +39,8 @@ private:
 
 	bool m_lastTargetMouseFocus = true;
 
+	bool m_flushMouse = true;
+
 public:
 	virtual void GetGameResolution(int* width, int* height) override;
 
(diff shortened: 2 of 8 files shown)


mninc:
thanks 👍

PR #3262, opened 2025-03-19
feat(scripting/v8): implement vector classes

thelindat:
### Goal of this PR
Improve compatibility of vector types across scripting runtimes.

Note this is a technically breaking change to some scripts, as a few developers have opted to depend on undefined behaviour (i.e. implementation details).

Implements simple vector classes which can be correctly encoded and decoded with msgpack.

ScRT: JS

### Successfully tested on
**Game builds:** 3407

**Platforms:** Windows

### Checklist

### Fixes issues
Supersedes #2946.

### DIFF
--- a/data/shared/citizen/scripting/v8/main.js
+++ b/data/shared/citizen/scripting/v8/main.js
@@ -752,6 +752,98 @@ const EXT_LOCALFUNCREF = 11;
 		global.LocalPlayer = Player(-1);
 	}
 
+	class Vector {
+		#length;
+	
+		constructor(...args) {
+			this.#length = args.length;
+			this.x = args[0];
+			this.y = args[1];
+		}
+	
+		*[Symbol.iterator]() {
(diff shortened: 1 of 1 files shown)


data/shared/citizen/scripting/v8/main.js
```diff
@@ -752,6 +752,104 @@ const EXT_LOCALFUNCREF = 11;
 		global.LocalPlayer = Player(-1);
 	}
 
+
+
+	if (!isDuplicityVersion) {
+		global.LocalPlayer = Player(-1);
+	}
```
AvarianKnight, line 759: You're redefining LocalPlayer

Korioz:
You should implement quaternion too as it was supported before, ext type is 23 and it's the same structure as a vec4

thelindat:
Can do, but would also like to know how cfx prefer these implementations (just basic, or with methods?)

PR #3264, opened 2025-03-20
tweak(voip-server-mumble): disallow external connections to uMurmur by default

AvarianKnight:
### Goal of this PR
Currently any client is able to join the internal uMurmur, this changeset looks to make it where only clients that are already connected to FXServer will be able to join the mumble server.

This is a stop-gap to prevent people joining from mumble clients.

Only allow clients that are connected to FXServer to join uMurmur, and add ConVar `mumble_allowExternalConnections` to allow opting out of this change.

Server

### Successfully tested on
**Platforms:** Windows

### Checklist


code/components/voip-server-mumble/src/server.cpp
```diff
@@ -551,26 +575,6 @@ static InitFunction initFunction([]()
 				return;
 			}
 
-			// is this a Mumble ping?
-			if (len == 12 && *(uint32_t*)data == 0)
```
FabianTerhorst, line 555: Just to be sure. The ping was only required for external connections?

AvarianKnight: Yes, this is only used for external clients to show the ping/user count on the "Mumble Server Connect" list

image

Us not providing this won't matter for proper clients, all of this information will be sent via [`ServerSync`](https://github.com/citizenfx/fivem/blob/52f3489dd3764ccbc58cb8ccf2ca21f711257872/code/components/voip-server-mumble/src/Mumble.proto#L100-L111) packet

FabianTerhorst (approved):

PR #3274, opened 2025-03-22
feat(core/sbag): allow state bag modifications in strict mode

n3xuuu:
### Goal of this PR

Allow modification of specific state bags in strict mode by introducing a way to 'whitelist' them.

- Added a new native, `ALLOW_CLIENT_STATE_BAG`, to allow client state bag modifications in `sv_stateBagStrictMode`.
- Updated `SET_STATE_BAG_VALUE` to check `allowedClientStateBags` before allowing modifications.

FiveM, Natives, Server

### Successfully tested on

**Game builds:** -

**Platforms:** -

### Checklist

### DIFF
--- a/code/components/citizen-scripting-core/src/ResourceScriptFunctions.cpp
+++ b/code/components/citizen-scripting-core/src/ResourceScriptFunctions.cpp
@@ -357,34 +357,45 @@ static InitFunction initFunction([] ()
 		context.SetResult(fx::SerializeObject(msgpack::type::nil_t{}));
 	});
 
+	static std::unordered_set<std::string> allowedClientStateBags;
+
 	fx::ScriptEngine::RegisterNativeHandler("SET_STATE_BAG_VALUE", [](fx::ScriptContext& context)
 	{
-		auto bagName = context.CheckArgument<const char*>(0);
-		auto keyName = context.CheckArgument<const char*>(1);
-		auto keyValue = context.CheckArgument<const char*>(2);
-		auto keySize = context.GetArgument<uint32_t>(3);
-		auto replicated = context.GetArgument<bool>(4);
+    	auto bagName = context.CheckArgument<const char*>(0);
+    	auto keyName = context.CheckArgument<const char*>(1);
+    	auto keyValue = context.CheckArgument<const char*>(2);
+    	auto keySize = context.GetArgument<uint32_t>(3);
+    	auto replicated = context.GetArgument<bool>(4);
 
-		#ifndef IS_FXSERVER
-			if (replicated && stateBagStrictModeVar.GetValue())
-			{
-				fx::scripting::Warningf("natives", "StateBags can't be modified from the client, because the StateBag strict mode is enabled. Disable it using setr sv_stateBagStrictMode false\n");
-				return;
-			}
-		#endif
+    	#ifndef IS_FXSERVER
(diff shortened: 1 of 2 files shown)


code/components/citizen-scripting-core/src/ResourceScriptFunctions.cpp
```diff
@@ -357,34 +357,45 @@ static InitFunction initFunction([] ()
...
+    	context.SetResult(false);
+	});
 
-			return;
-		}
+	fx::ScriptEngine::RegisterNativeHandler("ALLOW_CLIENT_STATE_BAG", [](fx::ScriptContext& context)
+	{
+    	#ifndef IS_FXSERVER
+        	return;
+    	#endif
```
766974616c79, line 395: wouldn't it be better to use a `#ifdef IS_FXSERVER` for the native definition?

AvarianKnight:
*I don't want you to think this is targeted towards you, just general thoughts, this just adds to them*

This entire system is really starting to feel like its pieced together, instead of adding proper allow/blocking for state bags that was suggested a while back by Thorium we're now slowly getting a system that is slowly just pieced together and barely usable, and is also almost unusable on larger servers due to how badly it scales because of how high contention this section is.

We should *really* think about doing more changes to this section, and perhaps actually add proper support for state bag read/write policies.

n3xuuu:
I totally understand what you're pointing out. Yes, this PR is a temporary solution, but until someone with the right resources can work on a proper policy to prevent malicious clients from modifying their state bags, this pull request implements a fix by allowing clients to modify only specific state bags.

code/components/citizen-scripting-core/src/ResourceScriptFunctions.cpp
n3xuuu: done

rol3x1337:
te pup vărule, e bună sugestia sper să o acceptați.

Gravxd:
It's a good PR but Avarian is referring to what should've been released by now - we were approx 11 months ago that a NUI/statebag update was coming where commands were going to be available to configure desired behaviour including whitelisting - if there was progress made on said update, it would be more beneficial to continue on that rather than introducing new natives / methods that would then potentially be dropped and then because of not wanting to break compatibility, they would stay in probably resulting in more issues/bugs.

thelindat:
> this PR is a temporary solution

Unfortunately every update made to statebags is a "temporary solution" and has been since they were first introduced.

n3xuuu:
> Unfortunately every update made to statebags is a "temporary solution" and has been since they were first introduced.

Yeah, but unfortunately, until the promised NUI/statebag update drops, I'd really like to have a way, even if it's just a temporary fix, to update 'the less sensitive' statebags.

tens0rfl0w:
I agree with @AvarianKnight—adding complexity to the current system, especially with new natives, makes maintenance harder when the new system is eventually released.

Regarding your changeset:
`allowedClientStateBags` exists on both the client and server, but you're only setting its value on the server, while the actual check occurs client-side. This wouldn't work as intended. If such a system were to be implemented, a server-side check would be necessary to maintain integrity and prevent clients from arbitrarily modifying state bags.

n3xuuu:
I agree with what you pointed out about the changeset, and since it's my first time working with FiveM's codebase, I will close this pull request until I have enough knowledge or the statebag update drops.

PR #3292, opened 2025-03-26
fix(net/sbac): don't duplicate the registration of the state bag component

AvarianKnight:
### Goal of this PR
`RegisterStateBag` gets called further down in `rage::netObjectMgr::RegisterNetworkObject(obj)` which is hooked to call `CloneManagerLocal::RegisterNetworkObject` which will *then* register the state bag in a queued mode where it will wait for the client to ack a packet from the server for the object before it would apply it.

Calling it here could lead to the client still getting state bag changes for this even if the entity didn't *actually* exist for the client.

Remove the call and rely on the call later down the call stack.

FiveM, RedM

### Successfully tested on

**Game builds:** .. 

**Platforms:** Windows, Linux

### Checklist

### DIFF
--- a/code/components/gta-net-five/src/CloneManager.cpp
+++ b/code/components/gta-net-five/src/CloneManager.cpp
@@ -995,11 +995,6 @@ bool CloneManagerLocal::HandleCloneCreate(const msgClone& msg)
 
 	auto& objectData = m_trackedObjects[msg.GetObjectId()];
 
-	if (!objectData.stateBag)
-	{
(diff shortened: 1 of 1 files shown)


FabianTerhorst:
@prikolium-cfx Do you think we can push this to a feature branch? I don' think we can easily test this.

prikolium-cfx:
> @prikolium-cfx Do you think we can push this to a feature branch? I don' think we can easily test this.

Yeah, sure.
`specific/feature/state-bag-registration-delay`. Will be deployed in ~1 hour.

Issue #3289, opened 2025-03-25
CreateVehicleServerSetter State Bags can become orphaned.


Lucas7yoshi:
### What happened?

Vehicles created with CreateVehicleServerSetter will rarely, but often enough to be extremely annoying, cause statebags to become orphaned. Once a netid's statebag because orphaned it will never be properly cleaned up by the server when the vehicle is deleted. This doesnt happen for most vehicles, but it compounds as data from one statebag applies to another vehicle, causing it to be wonky, making someone put it away (delete it), which then will have that netid reassigned and repeat... Gathering all sorts of statebag values.

This can be manually remedied by manually purging statebags, but this is a point of concern for race conditions and also should not be neccasary. Most of our vehicles do not have this problem, but when one becomes "infected" it spreads and start corrupting all sorts of vehicle data.

Another manual remedy is explicitly setting nil to problem statebag values, but again, this is tedious, and basically doubles our code to fix a annoying bug. This did not occur with other methods of creating vehicles, and regretfully even after weeks of attempting to debug the root cause we have not been able to determine what induces a netid's statebag to become broken. 

My hope is that some part of this will ring some bell in someones head and they will think of what might be the problem.

### Expected result

State bags should be properly wiped when the vehicle is deleted.

### Reproduction steps

This is exceptionally difficult to track. We've been attempting to track this for weeks now and have not been able to reliably find what induces a statebag to become orphaned. Once it becomes orphaned, it will NEVER be wiped (even if another vehicle takes on that netid, and is deleted properly). 

### Importancy

There's a workaround

### Area(s)

FXServer, FiveM, OneSync

### Specific version(s)

FiveM, Release as of 2025-03-25, FXServer 13183 and likely beyond.

### Additional information

The server this is occuring on has a consistent peak of about 250~ players. If needed we are willing to add logging or try ideas that may be had. At the moment we are unable to update server artifact due to msgpack related issues caused by our usage of apparently undefined behaviour, but I believe a PR is up to address that.

Heisenberg691:
I have the same issue. State bags are being transferred to random vehicles. For example, I use state bags to check whether a vehicle is blacklisted. However, sometimes a "blacklist" state bag is incorrectly assigned to random vehicles, preventing users from using vehicles that aren’t actually blacklisted.

AvarianKnight:
[...]

PR #3293, opened 2025-03-26
fix(core/sbag): make sure our value isn't being erased before we try …

AvarianKnight:
### Goal of this PR
It's possible that this is a race condition in the delete logic, if the entity gets created immediately during the `OnCloneRemove` its possible that the entity will keep its old state bag since `RegisterStateBag` doesn't check that a state bag was being erased before it just gives out the same state bag reference, and the `m_erasureList` is only called once per server tick.

Move the check for erasing the state bag id further up so we don't try and return the same reference if an entity is actively being deleted from the list.

Server, RedM, FiveM

### Successfully tested on

**Game builds:** .. 

**Platforms:** Windows, Linux

### Checklist

### Fixes issues
Possible fix for https://github.com/citizenfx/fivem/issues/3289

### DIFF
--- a/code/components/citizen-resources-core/src/StateBagComponent.cpp
+++ b/code/components/citizen-resources-core/src/StateBagComponent.cpp
@@ -490,24 +490,35 @@ std::shared_ptr<StateBag> StateBagComponentImpl::RegisterStateBag(std::string_vi
 	std::shared_ptr<StateBagImpl> bag;
 	std::string strId{ id };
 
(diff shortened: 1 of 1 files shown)


Nobelium-cfx (approved):
Idk if it will (or how to test that it will) fix the issue in question. But the change itself look like a good thing to me.

@FabianTerhorst can you please take a look as well, in case I'm missing some implications of this change.

Issue #2612, opened 2024-07-03
8-bit to 32-bit integer - potential fixes for sirensetting id (carcol) (vehicle lights)


Legacy-TacticalGamingInteractive:
### What happened?

Someone at Cfx mentioned that I should make an issue about this here when the fix got added for the 255 Global Props limit by changing the 8bit to 32bit integer issue. Perhaps this same methodology could be applied to the global limit of carcols sirensetting id's which are also capped at 255 for the same reason. You can see that information here: https://github.com/citizenfx/fivem/commit/3d66c889321ffa2b0095f09194be618deafee0c6

Additionally, here is a repo of some Single Player/FiveM client sided fixes that perhaps we can replicate to FiveM overall in some fashion:
[ASI Mod for FiveM and Singleplayer](https://www.lcpdfr.com/downloads/gta5mods/scripts/28560-sirensetting-limit-adjuster/)
[Source Code](https://github.com/cpast/SirenSetting_Limit_Remover)

Should anyone be willing to take a look see and make it so. This would be an incredible game changer to have these fixed server side synced with FiveM client build.

Cheers!

### Expected result

Ability to increase siren id's from 255 to 65535

### Reproduction steps

1. Load any server with unique carcols `id value` and carvariations `sirensettings value=` up to 255 and beyond.
2. After reaching 256+ the lights will be broken in game. The lighting will not function correctly as it overlaps with other ID's at that point.
3. While a client sided fix can be applied manually, it will not be synced across a server for all players or the FiveM client build. But the potential is here to test this and fix.

### Importancy

There's a workaround

### Area(s)

FiveM, FXServer

### Specific version(s)

FiveM 3095 / Windows Server / Build 8733

### Additional information

This would greatly solve a lot of issues for server owners, vehicle developers and the longevity of FiveM with roleplay servers especially with many Tebex assets these days. Meaning that cars can no longer be modified and bundled to use one set ID for a pack anymore by devs or server owners without original source files. Which used to be a way to at least mitigate the amount of ID's needed for a set of vehicles. But today, often the developers release them one at a time each with a unique ID value.. Thus hitting this limit is a lot faster and easier to get to these days than in previous years.

JajarGaming:
8 to 32 is a bg jump. Isn't 16-bit possible?

Legacy-TacticalGamingInteractive:
Well yeah 16 would be fine too.

Legacy-TacticalGamingInteractive:
(anything at this point)

[...]

PR #3313, opened 2025-04-04
tweak(five): edit carcols aka sirensettings value to allow for 16bit

tugamars:
Applies the changes proposed here https://github.com/citizenfx/fivem/pull/3273

With the formatting fixed. 

Originally purposed by Legacy, just fixed the formatting with his permission.

PR #3273, opened 2025-03-22
tweak(five): edit carcols aka sirensettings value to allow for 16bit

Legacy-TacticalGamingInteractive:
### Goal of this PR
Adjust from 8-bit to 16-bit carcols/sirensettings values to go beyond 255. Fixing servers that have reached more than 255 emergency vehicles in regards to conflicting carcols ID values.

This value is located in carcols.meta under <siren> <id value="123"> as well as carvariations.meta <sirensettings value="123"> and is currently an 8-bit value thus limiting people to only 255 unique values.

...

### How is this PR achieving the goal
- Attempt to allow reading of not only single bit values but 16-bit values as well.
- Read values properly.
...

### This PR applies to the following area(s)
FiveM, Server, C#
...

### Successfully tested on
Not yet tested.

**Game builds:** .. 

**Platforms:** Windows, Linux

### Checklist

- [ ] Code compiles and has been tested successfully.
- [X] Code explains itself well and/or is documented.
- [X] My commit message explains what the changes do and what they are for.
- [ ] No extra compilation warnings are added by these changes.

### Fixes issues

Allow for more than 255 emergency vehicles with unique siren - emergency light patterns.

Basic attempt [with little knowledge, apologies] to fix: https://github.com/citizenfx/fivem/issues/2612

PR #3314, opened 2025-04-04
tweak(five): patch carcols sirensettings allowing for 16bit values

d22tny:
### Goal of this PR
Adjust from 8-bit to 16-bit carcols/sirensettings values to go beyond 255. Fixing servers that have reached more than 255 emergency vehicles in regards to conflicting carcols ID values.

This value is located in carcols.meta under as well as carvariations.meta and is currently an 8-bit value thus limiting people to only 255 unique values.

This PR just has a slightly better formatting than https://github.com/citizenfx/fivem/pull/3273 and https://github.com/citizenfx/fivem/pull/3313

I did not compile and test the code.

For a quicker check on what happens to sync-trees, this is the change:

image

...

I think trying to patch the size of the sirensettings?!
...

...

### Successfully tested on
Not tested
**Game builds:** .. 
-
**Platforms:** Windows, Linux
-

### Checklist

### Fixes issues

https://github.com/citizenfx/fivem/issues/2612

### DIFF
--- a/code/components/gta-streaming-five/src/ModKitIdRelocation.cpp
+++ b/code/components/gta-streaming-five/src/ModKitIdRelocation.cpp
@@ -59,22 +59,22 @@ static HookFunction hookFunction([]()
 {
 	_vehicleModKitArray = (uint16_t*)hook::AllocateStubMemory(sizeof(uint16_t) * NUM_MODKIT_INDICES);
 	RelocateRelative(
-	{ { "66 3B F0 73 ? 48 8D", 8 },
-	{ "66 41 3B C0 73 ? 48 8D", 9 },
-	{ "45 33 C0 4C 8D 0D ? ? ? ? B9", 6 },
-	{ "B8 FF FF 00 00 48 8D 3D", 8 },
-	{ "7D ? 41 BC FF FF 00 00 4C 8D 3D", 11 } });
+	{ { "66 3B F0 66 73 ?? 48 8D", 8 },  // Adjusted to 16-bit search: 66 3B F0 66 73 ?? 48 8D
+	{ "66 41 3B C0 66 73 ?? 48 8D", 9 },  // Adjusted to 16-bit search: 66 41 3B C0 66 73 ?? 48 8D
+	{ "45 33 C0 4C 8D 0D ?? ?? ?? ?? B9", 6 },  // Adjusted for 16-bit match (??)
+	{ "66 B8 FF FF 00 00 48 8D 3D", 8 },  // Adjusted to 16-bit search: 66 B8 FF FF 00 00 48 8D 3D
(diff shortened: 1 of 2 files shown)


Nobelium-cfx:
The reason why diff in [SyncTrees_Five.h](https://github.com/citizenfx/fivem/pull/3314/files#diff-5691634c3e6016c75852762f798a5e5dcc0847f8ecb20ef74da7de4db49dd59a) is still huge seems to be because end line symbol has changed. This file has a mix of \n and \r\n. When you open it in most editors - the editor will unify all the enlines. That's where the diff comes from.

There are two ways of dealing with this problem:
- Add only the lines that were actually changed through `git add -p`
- Make a commit that unifies endlines in this file. Make the actual changes as a separate followup commit.

code/components/gta-streaming-five/src/ModKitIdRelocation.cpp
```diff
@@ -59,22 +59,22 @@ static HookFunction hookFunction([]()
...
+	{ "66 39 4B 2A 66 73 ?? 0F", 14 } });  // Adjusted to 16-bit search: 66 39 4B 2A 66 73 ?? 0F
+	hook::nop(hook::get_pattern("66 3B F0 66 73 ?? 48 8D", 0), 5);
+	hook::nop(hook::get_pattern("66 41 3B C0 66 73 ?? 48 8D", 0), 6);
+	hook::nop(hook::get_pattern("66 3B D1 66 73 ?? 8B C2", 0), 5);
+	hook::nop(hook::get_pattern("66 39 4B 2A 66 73 ?? 0F", 0), 6);
 	hook::nop(hook::get_pattern("41 81 F8 00 04 00 00 7C", 0), 9);
 
 	{
-		auto location = hook::get_pattern("B8 FF FF 00 00 48 8D 3D", 13);
+		auto location = hook::get_pattern("66 B8 FF FF 00 00 48 8D 3D", 13);
```
tens0rfl0w, line 77: Most changed patterns seem to be invalid, not too if that's the right approach for patching this.

d22tny: > Most changed patterns seem to be invalid, not too sure if that's the right approach for patching this.

that's what i thought of it as well, at this time there are 3 PR's with the same idea, i opened this to try and help see the better formatting, i didn't know if i should let the mixed line ends, that's why i added a unify with the mostly used one.

as i've said i didn't compile anything, i didn't test it, i just helped with formatting in a way that all line ends are the same and it doesn't change the entire file.

d22tny:
Hey @Nobelium-cfx , i just tried to help with the formatting, as i've explained into tens0rfl0w's review. 
In my opinion, all 3 PR's regarding this should be closed and maybe someone with a little bit more experience into reverse engineering takes a look at it ?

PR #3316, opened 2025-04-05
feat(kvp-client): add client asynchronous kvp natives

outsider31000:
### Goal of this PR

just like the server we add client asynchronous natives.
using async natives reduces the load on the client as it's not always  needed to wait for the operations to finish

the goal is to have it stream lined with the server natives that already have async natives

FiveM/RedM

### Successfully tested on

**Game builds:** .. 

**Platforms:** Windows, Linux
windows

### Checklist

### DIFF
--- a/code/components/citizen-resources-client/src/KVScriptFunctions.cpp
+++ b/code/components/citizen-resources-client/src/KVScriptFunctions.cpp
@@ -100,30 +100,32 @@ static std::string FormatKey(const char* key, const std::string& resource = {})
 	return "res:" + resName + ":" + key;
 }
 
+template<bool sync = true>
 static void PutResourceKvp(fx::ScriptContext& context, const char* data, size_t size)
 {
 	auto db = EnsureDatabase();
 	auto key = FormatKey(context.CheckArgument<const char*>(0));
 
 	leveldb::WriteOptions options;
-	options.sync = true;
+	options.sync = sync;
 
 	db->Put(options, key, leveldb::Slice{ data, size });
 }
 
-template<typename T>
+template<typename T, bool sync = true>
 static void SetResourceKvp(fx::ScriptContext& context)
 {
 	msgpack::sbuffer buffer;
 	msgpack::packer<msgpack::sbuffer> packer(buffer);
 	packer.pack(std::is_pointer_v<T> ? context.CheckArgument<T>(1) : context.GetArgument<T>(1));
 
-	PutResourceKvp(context, buffer.data(), buffer.size());
+	PutResourceKvp<sync>(context, buffer.data(), buffer.size());
 }
 
+template<bool sync = true>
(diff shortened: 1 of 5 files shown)


code/components/citizen-resources-client/src/KVScriptFunctions.cpp
```diff
@@ -319,12 +321,15 @@ static void EndFindKvp(fx::ScriptContext& context)
...
+template<bool sync = true>
 static void DeleteResourceKvp(fx::ScriptContext& context)
 {
 	auto db = EnsureDatabase();
 	auto key = FormatKey(context.CheckArgument<const char*>(0));
+	leveldb::WriteOptions options;
+	options.sync = sync;
 
-	db->Delete(leveldb::WriteOptions{}, key);
+	db->Delete(options, key);
```
AvarianKnight, line 332: `sync` defaults to false here, you're changing the default from non blocking to blocking for `DELETE_RESOURCE_KVP`, we should just leave it as is and add the native for the sake of compatibility

AvarianKnight: What I'm saying is you *shouldn't* this is already async/non-blocking you *shouldn't* change this to be `sync` *at all*, leave the options as its default value `leveldb::WriteOptions{}`

outsider31000: yeah I understood what you meant now.

outsider31000:
@prikolium-cfx  what you think how we should proceed ?
my changes would not be breaking scripts, I think everyone assumes the delete native on client is sync due to the name , when in the server is async and the docs is declared as shared with no mention of the client native being async.

So my question is:
Do we fix it with my changes (to correct it and avoid miss interpretation and keept it stream lined with server) or leave it as it was ?
if you could give some advice on this would be appreciated.
thanks

RealStonerGamer:
yeah I always thought the DeleteResourceKvp was sync lol

AvarianKnight:
It should also be noted changing this from being async -> sync will also have performance implications since the sync versions are notably slower, and people who used this not thinking it had any performance implications, will suddenly have to deal with the overhead of it being synchronous.

outsider31000:
People who use this don't even know it's async in the first place, don't think a few ms per call would  be more important to "keep" than correcting the mess and allow it to use it properly aligned with server natives and with no miss interpretations.

But that's just my opinion.

code/components/citizen-resources-client/src/KVScriptFunctions.cpp
prikolium-cfx: What was the initial point to make all methods sync and only Delete async? For me it looks like a mistake.

prikolium-cfx (approved):

PR #3318, opened 2025-04-06
fix(core/rdr3): block custom startup.meta file

Ehbw:
### Goal of this PR

Fix an issue where users having a custom startup.meta file inside of their Red Dead Redemption or Red Dead Online installation would lead to the game failing to load correctly.

By preventing the game from acknowledging the files existance. Tested with problematic [startup.meta](https://gist.github.com/Ehbw/32275b1ee13e5d6f1b880cf448e5873c) file placed in x64/data/ of my Red Dead Online installation (which is used by redm). In normal conditions this file would never exist and the function would normally return false as is. So no

RedM

### Successfully tested on

**Game builds:**  1491

**Platforms:** Windows, Linux

### Checklist

### DIFF
--- a/code/components/gta-core-rdr3/src/GameInitRage.cpp
+++ b/code/components/gta-core-rdr3/src/GameInitRage.cpp
@@ -9,6 +9,7 @@
 #include <GameInit.h>
 
 #include <Hooking.h>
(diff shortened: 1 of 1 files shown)


martonp96:

PR #3319, opened 2025-04-07
tweak(scripting/scheduler.lua): allow for trace function hooking

bleenobtw:
### Goal of this PR
Allows script developers and or server owners to hook the Citizen.Trace function, so that they can inherently catch stack traces for when resources or functions throw errors that aren't caught from within the resource. 

It creates a global alias for Citizen.Trace called `Trace` and uses it within scheduler when throwing errors

FiveM/RedM

### Successfully tested on
**Game builds:** .. 
**Platforms:** Windows

### Checklist

### Fixes issues


AvarianKnight:
How would this change anything? You could already overwrite the `Citizen.Trace` function

bleenobtw:
> How would this change anything? You could already overwrite the `Citizen.Trace` function

Yes, you can overwrite `Citizen.Trace`; if you overwrite it in a resource and then call it from that resource. However in scheduler.lua, `Citizen` is locally scoped so if you overwrite the function in a resource, it will not be called when it's used in scheduler.

bleenobtw:
See [this commit](https://github.com/citizenfx/fivem/commit/9ab9bb6c26b3cff434604153b28828ed56b58587) as to why it's not possible.

bleenobtw:
This commit will be very useful for developers and server owners because instead of having to ask each of their players to take a screenshot of their f8 console whenever they run into a bug/error. You can hook the `Trace` function, enumerate the parameters and check if they include common error key words such as "SCRIPT ERROR" or "bad argument", check which resource it's being called from and then deal with said error in what way you see fit.

PR #3327, opened 2025-04-09
tweak(client/launcher): Gen9-specific warnings

tens0rfl0w:
### Goal of this PR
Adds some specific errors/warnings during initial setup to inform users about no Enhanced support for FiveM, as this currently causes very unspecific load errors/crashes which confuses some users.

- Add warnings if a user selects an installation of GTA V Enhanced as their game directory (some users seem to rename the V executable to bypass filter checks in the game select window).
- Try to auto-detect GTA V Enhanced installs if auto-detection for Legacy fails to inform users about the need to have V Legacy installed.
- Clear wrong path entry from the private profile if it points to V Enhanced.

Early warning if auto-detect fails but V Enhanced seems installed:
image

Error on init/folder selection if the path points to V Enhanced:
image

FiveM, Launcher

### Successfully tested on

### Checklist

### Fixes issues
/

### DIFF
--- a/code/client/launcher/GameSelect.cpp
+++ b/code/client/launcher/GameSelect.cpp
@@ -152,6 +152,22 @@ std::optional<int> EnsureGamePath()
 
 		if (path[0] != L'\0')
 		{
+#if defined(GTA_FIVE)
+			// TEMP: Gen9 specific error, some users rename their .exe to bypass name checks
+			// Throw early to avoid unspecific error output
+			if (const std::wstring lastComponent = std::filesystem::path(path).filename().wstring();
+				lastComponent.find(L"V Enhanced") != std::wstring::npos || lastComponent.find(L"VEnhanced") != std::wstring::npos)
+			{
+				// Clear "wrong" path entry
+				WritePrivateProfileString(L"Game", pathKey, nullptr, fpath.c_str());
+
+				static constexpr auto GEN9_ERROR = L"Your selected game installation folder points to the Enhanced edition of GTA V, which is currently not supported by FiveM.\n\n"
+					L"Please select the installation folder for the Legacy version of GTA V.";
+				MessageBox(nullptr, GEN9_ERROR, PRODUCT_NAME, MB_OK | MB_ICONWARNING);
+			}
+			else
+			{
+#endif
 			// check stuff regarding the game executable
 			std::wstring gameExecutable = fmt::sprintf(L"%s\\%s", path, GAME_EXECUTABLE);
 
@@ -159,6 +175,9 @@ std::optional<int> EnsureGamePath()
 			{
 				return {};
 			}
+#if defined(GTA_FIVE)
+			}
+#endif
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


code/client/launcher/GameSelect.cpp
```diff
@@ -316,6 +386,17 @@ std::optional<int> EnsureGamePath()
 	std::wstring gamePath = resultPath;
 	auto exeNameLength = std::size(GAME_EXECUTABLE); // counts null terminator, but here we use that for a backslash
 
+#if defined(GTA_FIVE)
+	// TEMP: Gen9 specific error, some users rename their .exe to bypass name checks
+	if (gamePath.find(L"V Enhanced") != std::wstring::npos || gamePath.find(L"VEnhanced") != std::wstring::npos)
```
Nobelium-cfx, line 391: This check will return an error if there is and `V Enhanced` or `VEnhanced` anywhere in the path. Which could still point to a valid Legacy installation for some people (something like "X:/VEnhanced/games/Grand Theft Auto V/GTA5.exe").

Lets check only the last folder for this substrings. Same for the `if (wcsstr(path, L"V Enhanced") != nullptr || wcsstr(path, L"VEnhanced") != nullptr)` check above.

code/client/launcher/GameSelect.cpp
```diff
@@ -283,6 +301,58 @@ std::optional<int> EnsureGamePath()
 	}
 #endif
 
+#if defined(GTA_FIVE)
```
Nobelium-cfx, line 304: Nit: this block seems to be largely copied from the legacy installation check above. Can we refactor and move common parts into separate functions?

E.g. `checkForBasicInstall` and `proposeDirectory` can be a single function that accepts a list of files to check in the folder.

Nobelium-cfx (commented):
Thats a really useful change, thank you!

Left some comments but overall lgtm.

code/client/launcher/GameSelect.cpp
tens0rfl0w: Checking only the last folder in the path now👍🏼

code/client/launcher/GameSelect.cpp
tens0rfl0w: Refactored👍🏼

code/client/launcher/GameSelect.cpp
```diff
@@ -195,8 +215,56 @@ std::optional<int> EnsureGamePath()
 	fileDialog->SetFileTypes(1, &filter);
 
 #if defined(GTA_FIVE) || defined(IS_RDR3) || defined(GTA_NY)
+	auto proposeDirectory = [&](const std::wstring& gameRoot, const std::unordered_set<std::string>& filesToCheck)
```
Nobelium-cfx, line 218: Nit: we never search in `filesToCheck`, we only iterate through it. Lets make it a `std::vector` instead.

code/client/launcher/GameSelect.cpp
```diff
@@ -152,13 +153,32 @@ std::optional<int> EnsureGamePath()
...
+			// TEMP: Gen9 specific error, some users rename their .exe to bypass name checks
+			// Throw early to avoid unspecific error output
+			if (const std::wstring lastComponent = std::filesystem::path(path).filename().wstring();
+				lastComponent.find(L"V Enhanced") != std::wstring::npos || lastComponent.find(L"VEnhanced") != std::wstring::npos)
+			{
+				// Clear "wrong" path entry
+				WritePrivateProfileString(L"Game", pathKey, nullptr, fpath.c_str());
+
+				static constexpr auto GEN9_ERROR = L"Your selected game installation folder points to the Enhanced edition of GTA V, which is currently not supported by FiveM.\n\n"
+				L"Please select the installation folder for the Legacy version of GTA V.";
```
Nobelium-cfx, line 166: Nit: add tab so it's clear that it's a continuation of `GEN9_ERROR` line. Same for two other error messages below.

Nobelium-cfx (approved):
Thank you! LGTM. Just a few leftover style nits.

code/client/launcher/GameSelect.cpp
tens0rfl0w: Done👍🏼

code/client/launcher/GameSelect.cpp
tens0rfl0w: Done👍🏼

code/client/launcher/GameSelect.cpp
```diff
@@ -7,6 +7,7 @@
 
 #include "StdInc.h"
 #include <optional>
+#include <unordered_set>
```
Nobelium-cfx, line 10: Nit: this include is no longer needed.

tens0rfl0w:
@Nobelium-cfx FYI: Latest push was just to add Gen9 Xbox Game Pass installs to the registry search.

PR #3330, opened 2025-04-10
fix(client/mumble): request resync if we fail to decrypt a packet after 1 seconds

AvarianKnight:
### Goal of this PR
Send resync for crypt state in situations where we've failed to decrypt after a certain span of times

After 1s of not being able to decrypt a packet from the server we sent a resync request, copying behavior [from mumble](https://github.com/mumble-voip/mumble/blob/master/src/mumble/ServerHandler.cpp#L244-L254), notably the time we wait is different here, we send ping packets every 1 second, mumble only does it every 5 seconds.

This should hopefully fix some of the last few edge cases where people have reported "i can talk, but I can't hear anyone"

FiveM, RedM

### Successfully tested on
**Game builds:** Agnostic

**Platforms:** Windows

### Checklist

### DIFF
--- a/code/components/voip-mumble/src/MumbleCrypto.cpp
+++ b/code/components/voip-mumble/src/MumbleCrypto.cpp
@@ -8,6 +8,10 @@
 
 #include <array>
 
+inline std::chrono::milliseconds msec()
+{
+	return std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::high_resolution_clock::now().time_since_epoch());
+}
 
 void MumbleCrypto::Encrypt(const uint8_t* plain, uint8_t* cipher, size_t length)
 {
@@ -144,6 +148,8 @@ bool MumbleCrypto::Decrypt(const uint8_t* cipher, uint8_t* plain, size_t length)
 	m_localGood++;
 	m_localLate += late;
 	m_localLost += lost;
+
+	m_lastGoodUdp = msec();
 
 	return true;
 }
--- a/code/components/voip-mumble/include/MumbleClientImpl.h
+++ b/code/components/voip-mumble/include/MumbleClientImpl.h
@@ -86,6 +86,8 @@ public:
 	uint32_t m_localLost;
 	uint32_t m_localResync;
 
+	std::chrono::milliseconds m_lastGoodUdp;
+
 	void Encrypt(const uint8_t* plain, uint8_t* cipher, size_t length);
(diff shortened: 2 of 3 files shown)


code/components/voip-mumble/src/MumbleCrypto.cpp
```diff
@@ -145,6 +149,8 @@ bool MumbleCrypto::Decrypt(const uint8_t* cipher, uint8_t* plain, size_t length)
 	m_localLate += late;
 	m_localLost += lost;
 
+	m_lastGoodUdp = msec();
```
Nobelium-cfx, line 152: Why do we update `m_lastGoodUdp` in encryption. IIUC we should do it in decrypt. Since we are tracking time since last good message received from the server.

Our `voip-server-mumble` seems to be setting `tLastGood` in `CryptState_decrypt` as well.

Nobelium-cfx (commented):
Just so I understand the context of this change better: if mumble does the same check - why do we need to duplicate it on our side? Is it because of the difference in ping time? Or do we not use part of the mumble library that does this check?

AvarianKnight:
> Just so I understand the context of this change better: if mumble does the same check - why do we need to duplicate it on our side? Is it because of the difference in ping time? Or do we not use part of the mumble library that does this check?

At the base Mumble is just a specification on how to handle network traffic between clients, we have a number of [ProtoBufs](https://github.com/mumble-voip/mumble/blob/26a266c3a990f2159ef4f59a5767f71df07dbdaf/src/Mumble.proto#L6) we can use to send/receive TCP traffic, and a binary format (PDS - Packet Data Stream) we have to use to send/receive UDP traffic (though on newer versions of mumble this is all handled via [ProtoBufs too](https://github.com/mumble-voip/mumble/blob/master/src/MumbleUDP.proto)).

Cfx opted to implement the mumble protocol itself (and for the server opted to use uMurmur) because the base implementation `murmur` (server) / `mumble` (client) are not embeddable, they rely heavily on QT to handle Mutexes, IO, etc, which is obviously not desirable. Because we implemented the protocol ourself there's quite a few things from the mumble client that Cfx's implementation was missing

Some of which was fixed by:

https://github.com/citizenfx/fivem/pull/2990
https://github.com/citizenfx/fivem/pull/3058

This should (hopefully) be the last bit of code that we haven't implemented that the mumble client has.

Nobelium-cfx (approved):
I see, thank you for the context. Looks good!

PR #3331, opened 2025-04-12
Fixes oversights in original RawKeymap PR

AvarianKnight:
### Goal of this PR
This somehow managed to slip through when I originally made this pr, this
just swaps the native declaration and ordering in code to be the right order
since when `RawKeymap` was called the ordering is `keyDown`, `keyUp`.

This fixes another oversight in the initial PR, we should always register
`m_wasTriggered` changes, even if we don't have refs, otherwise we will
have "perma stuck" keymaps if the user decides to not register one part of
the keymap

This also adds a proper error message if you provide an invalid index

This also fixes an issue with `REGISTER_RAW_KEYMAPS` where it would still supply
a ref id even if there was no function (aka when set to null) to make the reference to, and when
calling into the JS ScRT it would error because `callback` would be null.

This copies the [behavior from Lua](https://github.com/citizenfx/fivem/blob/476f550dfb5d35b53ff9db377445be76db7c28bc/data/shared/citizen/scripting/lua/scheduler.lua#L470-L480) where we wrap the call and properly return
null if the `refFunction` is not a function.

js error because function call is actually null

RedM

### Successfully tested on

**Game builds:** .. 

**Platforms:** Windows, Linux

### Checklist

### DIFF
--- a/code/components/extra-natives-five/src/InputNatives.cpp
+++ b/code/components/extra-natives-five/src/InputNatives.cpp
@@ -192,16 +192,24 @@ public:
 						continue;
 					}
 
-					if (wasPressed && keymap->m_keyDownRef && !keymap->m_wasTriggered)
+					if (wasPressed && !keymap->m_wasTriggered)
 					{
 						keymap->m_wasTriggered = true;
-						rm->CallReference<void>(keymap->m_keyDownRef->GetRef());
+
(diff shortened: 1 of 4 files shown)


prikolium-cfx (approved):
Thanks for the contribution

PR #3342, opened 2025-04-17
tweak(net): generate truly random "GUID"'s

AvarianKnight:
### Goal of this PR

This is old code that only gets used on initial client connection in cases where the peer doesn't match the same TCP peer (for situations like Cloudflare Warp or other VPN's that don't keep TCP/UDP the same)

This ID previously allowed for someone to get disconnected if they had the unfortunate coincidence of having the same tick time as another player on initial join, and while this happened *rarely*, it still happened.

This makes the GUID a "more random" number, and in situations where the GUID will collide it will drop the connecting client instead of the existing client. Since the GUID gets incremented on every join, there will be no valid collisions anyways.

FiveM, RedM, Server

### Successfully tested on
**Game builds:** .. 

**Platforms:** Windows, Linux

### Checklist

### DIFF
--- a/code/components/citizen-server-impl/include/ClientDropReasons.h
+++ b/code/components/citizen-server-impl/include/ClientDropReasons.h
@@ -1,4 +1,4 @@
-﻿#pragma once
+#pragma once
 
 namespace fx
 {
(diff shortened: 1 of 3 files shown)


dalekenium-cfx:
Simple, but effective. Approved.

PR #3349, opened 2025-04-19
fix(gta-core-five): invalid bounding volume hierarchy

### LINKED ISSUE #3155: table-ohio-emma crash
### What happened?

Is happening to some people (not all the people) when u go to a certain zone, but not always in the same zone but surrounding zone.

Crash with a table-ohio-emma code.

### Expected result

No crashing?

### Reproduction steps

The only way to reproduce this is with my server or in my server driving in a road close to bank pacific, autoexotics, cityhall in circles.

### Importancy

Crash

### Area(s)

FiveM

### Specific version(s)

FiveM 12767 b3258

### Additional information

Here the crash dump.
[CfxCrashDump_2025_02_14_18_12_55.zip](https://github.com/user-attachments/files/18803260/CfxCrashDump_2025_02_14_18_12_55.zip)

And i let u the image of the crash.
Image

DaniGP17:
### Goal of this PR
Fix a crash that occurred in the call to the `CalcClosestLeafDistance` function that is responsible for traversing the bvh to find the squared distance to the closest leaf node from the given position.
The function could crash if the bvh or bvhNode pointer was null.

Adding a check before the original call of the function.

FiveM

### Successfully tested on
**Game builds:** 1604, 3095, 3407

**Platforms:** Windows, Linux

### Checklist

### Fixes issues
fixes #3155

### DIFF
--- a/code/components/gta-core-five/src/CrashFixes.SelfCollision.cpp
+++ b/code/components/gta-core-five/src/CrashFixes.SelfCollision.cpp
@@ -41,6 +41,25 @@ namespace rage {
 		uint16_t m_MaxNumBounds;
 		uint16_t m_NumBounds;
 	};
+
+	struct phOptimizedBvhNode
+	{
+		__int16 m_AABBMin[3];
+		__int16 m_AABBMax[3];
+		unsigned __int16 m_NodeData;
+		unsigned __int8 m_PolygonCount;
+	};
+
+	struct phOptimizedBvh
+	{
+		phOptimizedBvhNode* bvhNode;
+		char pad[120];
+	};
+
+	struct ScalarV
+	{
+		DirectX::XMVECTOR v;
+	};
 }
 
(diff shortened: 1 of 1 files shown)


code/components/gta-core-five/src/CrashFixes.SelfCollision.cpp
```diff
@@ -41,6 +41,24 @@ namespace rage {
...
+		__int16 m_AABBMin[3];
+		__int16 m_AABBMax[3];
+		unsigned __int16 m_NodeData;
+		unsigned __int8 m_PolygonCount;
+	};
+
+	struct phOptimizedBvh
+	{
+		phOptimizedBvhNode* bvhNode;
+	};
```
Nobelium-cfx, line 56: Nit: this structure is incomplete. No need to fill it all, but lets add a padding for the remaining fields size (i.e. `char pad[120]`);

DaniGP17: Added the padding as suggested, thanks!

PR #3353, opened 2025-04-21
tweak(native-decls): update documentation for trains

AvarianKnight:
### Goal of this PR
Cleanup the wording in the native declaration to make it less confusing, and add information about how this works with trains

> **NOTE**: This native doesn't guarantee the persistence of the entity.

Is confusing wording, replaced with what is actually defined behavior (that the server will not delete the entity in its normal garbage collection)

Server

### Successfully tested on
N/A

### Checklist

- [N/A] Code compiles and has been tested successfully.

### DIFF
--- a/ext/native-decls/SetEntityOrphanMode.md
+++ b/ext/native-decls/SetEntityOrphanMode.md
@@ -23,9 +23,11 @@ enum EntityOrphanMode {
 }
 ```
 
-Sets what happens when the entity is orphaned and no longer has its original owner.
+Sets what the server will do when the entity no longer has its original owner. By default the server will cleanup entities that it considers "no longer relevant".
 
-**NOTE**: This native doesn't guarantee the persistence of the entity.
+When used on trains, this native will recursively call onto all attached carriages.
+
+**NOTE**: When used with `KeepEntity` (2) this native only guarantees that the ***server*** will not delete the entity, client requests to delete the entity will still work perfectly fine.
 
 ## Parameters
 * **entity**: The entity to set the orphan mode of


Nobelium-cfx (approved):

ext/native-decls/SetEntityOrphanMode.md
```diff
@@ -23,9 +23,11 @@ enum EntityOrphanMode {
 }
 ```
 
-Sets what happens when the entity is orphaned and no longer has its original owner.
+Sets what the server will do when the entity no longer has its original owner. By default the server will cleanup entities that it consideres "no longer relevant".
```
Mathu-lmn, line 26: nit: *considers

AvarianKnight: Nice catch, fixed.

ext/native-decls/SetEntityOrphanMode.md
```diff
@@ -23,9 +23,11 @@ enum EntityOrphanMode {
 }
 ```
 
-Sets what happens when the entity is orphaned and no longer has its original owner.
+Sets what the server will do when the entity no longer has its original owner. By default the server will cleanup entities that it considers "no longer relevant".
 
-**NOTE**: This native doesn't guarantee the persistence of the entity.
+When used on trains, this native will recursively call onto all attached carriages.
+
+**NOTE**: When used with `KeepEntity` (2) this native only guarantees that the ***server*** will not delete the entity, client requests to delete the entity will still work perfectly fine.
```
csskroubledev, line 30: if entity is created serverside, won't `sv_protectServerEntities` convar prevent the client from deleting the entity?

If yes, this note may cause some misunderstanding.

Ehbw: > if entity is created serverside, won't `sv_protectServerEntities` convar prevent the client from deleting the entity?
> 
> If yes, this note may cause some misunderstanding.

At this moment trains are unable to be created on the server.

csskroubledev: GitHub is little funky with the comment creation, but I meant the line `30` exclusively.

PR #3361, opened 2025-04-26
fix(core-rdr3): fix crash when Blip limit is reached RedM

outsider31000:
### Goal of this PR

this pr fixes a crash, when blip limit has been reached `920` and players use the set waipoint in the big map
it causes them to crash.

image

image

thanks to @Ktos93

Logs in the f8 console when blip limit has been reached and blocks the player  from using the waypoint, which prevents the crash
the log will help players report to the developers so they become aware of the issues.

RedM

### Successfully tested on

**Game builds:** .. 
1491
**Platforms:** Windows, Linux
windows

### Checklist

### DIFF
--- a/code/components/gta-core-rdr3/src/PatchBlipLimitCrash.cpp
+++ b/code/components/gta-core-rdr3/src/PatchBlipLimitCrash.cpp
@@ -1,0 +1,25 @@
+#include <StdInc.h>
+#include <Hooking.h>
+#include <MinHook.h>
+#include <Pool.h>
+
+static void* (*g_origCreateWaypoint)(void* waypointController, uint32_t* blipHash);
+static void* CreateWaypoint(void* waypointController, uint32_t* blipHash)
+{
+	static auto pool = rage::GetPoolBase("fwuiBlip");
+
+	if (pool->GetCountDirect() >= pool->GetSize())
+	{
+		trace("Prevented crash: Blip pool 'fwuiBlip' is full (%d out of %d slots used)\n", pool->GetCountDirect(), pool->GetSize());
+		return 0;
+	}
+
+	return g_origCreateWaypoint(waypointController, blipHash);
+}
+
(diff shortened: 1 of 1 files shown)


d22tny:
Aren't they aware of the issue since they're crashing ?
I thought a bit about this problem when i first discovered it and i think that reserving one blip for the waypoint would be a smoother solution. Block blip creation at 919 and 920 is always reserved for the waypoint.

d22tny:
> I think that blocking blips is hidding the issue. either way would work I guess.

Actually it would be a good fix no ? Only limiting the blips to -1 to it's actual limit and never getting a crash and always being able to use the waypoint i see it as a big win.

Anyways for me it happens only in development when i don't remove blips on resource stop, a lot of the crashes are from me i guess :))

And i guess by this time that every dev making resources is aware of this issue and handles blips correctly ? Maybe, idk.

outsider31000:
the issue is not on the blips is on the waypoint you crash by using it (there might be other ways).  so we block the waypoint from being used when theres too many blips.

d22tny:
Yes, exactly and if you limit blips to 919, then 920 is always free for creating the waypoint, this way you don't get a crash and you can use the waypoint.

outsider31000:
I udnerstand what you are saying , but we patch where the issue lies
the patch is for waypoint crash.

outsider31000:
the problem is that waypoint creates a blip and then makes it crash, unlike the natives that wont let it create more, but adding a log and preventing using the waypoint, it tells developers where the issue is so they can fix it.

I personally rather know there is an issue, than removing one slot and the issue still there with "invisible blips" and wonder why they dont show

outsider31000:
if the pool for blips is not hardcoded I will ask Cfx to add it to the allowed poolnames so servers can increase them, and this patch will still be a valid fix for the crash.

prikolium-cfx (approved):

PR #3365, opened 2025-04-28
feat(extra-natives/five): add support for configurable outline render technique

Yum1x:
### Goal of this PR

Allow developers to specify a custom shader technique group for entity outlines via a new native (`SET_ENTITY_DRAW_OUTLINE_RENDER_TECHNIQUE`). This expands the flexibility of the outline rendering system, which was previously limited to the unlit technique, and addresses use cases beyond the intended FxDK environment, as some developers rely on this functionality in their own codebases.
image

> Note: While not critical to FxDK usage, this native aims to improve flexibility for developers already relying on outline rendering in their projects.

Introduces the `SET_ENTITY_DRAW_OUTLINE_RENDER_TECHNIQUE` native, allowing scripts to specify a custom shader technique group for entity outlines. The implementation updates the outline rendering system to use the provided technique group, with a fallback to the default "unlit" technique if none is specified. This expands the flexibility of outline rendering for both FxDK and community use cases.

Natives, FxDK

### Successfully tested on

Game builds: 1604, 3095, 3258, 3407

Platforms: Windows

### Checklist

### DIFF
--- a/code/components/extra-natives-five/src/GamePrimitives_Outlines.cpp
+++ b/code/components/extra-natives-five/src/GamePrimitives_Outlines.cpp
@@ -35,6 +35,10 @@ static int* g_currentDrawBucket;
 
 static CRGBA outlineColor{ 255, 0, 255, 255 };
 
+constexpr const char* DEFAULT_SHADER_TECHNIQUE_GROUP = "unlit";
+static std::string g_shaderTechniqueGroupId = DEFAULT_SHADER_TECHNIQUE_GROUP;
+
+
 class OutlineRenderer
 {
 public:
@@ -432,7 +436,7 @@ static InitFunction initFunctionBuffers([]()
 				SetBlendState(GetStockStateIdentifier(BlendStateDefault));
 
 				last = *currentShader;
-				*currentShader = _getTechniqueDrawName("unlit");
+				*currentShader = _getTechniqueDrawName(g_shaderTechniqueGroupId.c_str());
 
 				// draw bucket 0 pls, not 1
 				// #TODO: set via DC?
@@ -602,6 +606,21 @@ static InitFunction initFunctionScriptBind([]()
 				outlineEntities.erase(it);
 			}
 		}
+	});
+
+	fx::ScriptEngine::RegisterNativeHandler("SET_ENTITY_DRAW_OUTLINE_RENDER_TECHNIQUE", [](fx::ScriptContext& context)
+	{
+		std::string techniqueGroupId = context.GetArgument<const char*>(0);
+		if (techniqueGroupId.empty())
+		{
+			return;
+		}
+		g_shaderTechniqueGroupId = techniqueGroupId;
+	});
+	
+	fx::ScriptEngine::RegisterNativeHandler("RESET_ENTITY_DRAW_OUTLINE_RENDER_TECHNIQUE", [](fx::ScriptContext& context)
+	{
+		g_shaderTechniqueGroupId = DEFAULT_SHADER_TECHNIQUE_GROUP;
 	});
 
 	fx::ScriptEngine::RegisterNativeHandler("SET_ENTITY_MATRIX", [](fx::ScriptContext& context)
--- a/ext/native-decls/sdk/SetEntityDrawOutlineRenderTechnique.md
+++ b/ext/native-decls/sdk/SetEntityDrawOutlineRenderTechnique.md
@@ -1,0 +1,64 @@
+---
+ns: CFX
+apiset: client
+game: gta5
+---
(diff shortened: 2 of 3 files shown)


prikolium-cfx:
Could you please revert all formatting changes so we can see only real changes

code/components/extra-natives-five/src/GamePrimitives_Outlines.cpp
```diff
@@ -34,6 +34,10 @@ static int* currentShader;
 static int* g_currentDrawBucket;
 
 static CRGBA outlineColor{ 255, 0, 255, 255 };
+
+#define DEFAULT_SHADER_TECHNIQUE_GROUP "unlit"
```
prikolium-cfx, line 38: Let's use `constexpr const char* DEFAULT_SHADER_TECHNIQUE_GROUP = "unlit";`

code/components/extra-natives-five/src/GamePrimitives_Outlines.cpp
```diff
@@ -412,6 +416,9 @@ static InitFunction initFunctionBuffers([]()
 			return;
 		}
 
+
+		static const char* chTechniqueGroupId = g_shaderTechniqueGroupId.c_str();
```
prikolium-cfx, line 420: `static` here means that this variable will be initialized by global variable only once and it will be impossible to change it after.

code/components/extra-natives-five/src/GamePrimitives_Outlines.cpp
```diff
@@ -604,6 +611,12 @@ static InitFunction initFunctionScriptBind([]()
 		}
 	});
 
+	fx::ScriptEngine::RegisterNativeHandler("SET_ENTITY_DRAW_OUTLINE_RENDER_TECHNIQUE", [](fx::ScriptContext& context)
+	{
+		std::string techniqueGroupId = context.GetArgument<const char*>(0);
+		g_shaderTechniqueGroupId = techniqueGroupId.empty() ? DEFAULT_SHADER_TECHNIQUE_GROUP : techniqueGroupId;
```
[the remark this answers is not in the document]
prikolium-cfx: And the behavior with using empty string to return to default behavior looks a bit inconsistent. Usually in game natives we are using additional native `RESET_SOMETHING` for the parity with SET command.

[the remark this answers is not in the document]
Yum1x: invalid technique group causes the game code use the "default" technique as fallback so crashes don't occurs

ext/native-decls/sdk/SetEntityDrawOutlineRenderTechnique.md
```diff
@@ -0,0 +1,15 @@
...
+ns: CFX
+apiset: client
+game: gta5
+---
+## SET_ENTITY_DRAW_OUTLINE_RENDER_TECHNIQUE
+
+```c
+void SET_ENTITY_DRAW_OUTLINE_RENDER_TECHNIQUE(char* techniqueGroup);
+```
+Sets the render technique for drawing an entity's outline. This function allows you to specify a technique group name to control how the entity's outline is rendered in the game.
```
prikolium-cfx, line 11: As a last thing let's add known techniques to documentation, like `default` and `unlit`

prikolium-cfx (approved):

prikolium-cfx:
I've marked this PR as ready-to-merge, but looks like it needs manual testing as some part of code called in different thread and it can cause issues when setting global var.

Yum1x:
@prikolium-cfx Do you have any solution in mind? I'm considering mutexes for thread safety, wdyt?

code/components/extra-natives-five/src/GamePrimitives_Outlines.cpp
```diff
@@ -604,6 +608,21 @@ static InitFunction initFunctionScriptBind([]()
 		}
 	});
 
+	fx::ScriptEngine::RegisterNativeHandler("SET_ENTITY_DRAW_OUTLINE_RENDER_TECHNIQUE", [](fx::ScriptContext& context)
+	{
+		std::string techniqueGroupId = context.GetArgument<const char*>(0);
```
AvarianKnight, line 613:
```suggestion
		std::string techniqueGroupId = context.CheckArgument<const char*>(0);
```
Using `CheckArgument` here will make this error if index `0` is null

Yum1x: Good point! I thought the type checking in *natives* would handle that. I'll do some tests and fix it in upcoming commits

minsbat:
Bump to not stale

PR #3372, opened 2025-04-29
feat(state)!: add compression dictionary for RedM

AvarianKnight:
### Goal of this PR
Having a compression dictionary for RedM reduces the net traffic from 18-30kb/s to 16-24kb/s in testing, depending on the amount of players in the area this could result in a lot less data sent over `svNetwork`

This is protocol breaking since our compression dict will no longer be the same one used for FiveM, this adds a protocol version file that can be shared between server/client to the bump easier if its ever done again

RedM

### Successfully tested on

**Game builds:** .. 

**Platforms:** Windows, Linux

### Checklist


AvarianKnight:
For an explanation on how this was generated:

There's a function in `CloneManagerLocal::SendUpdates` to write chunks to a file, you can enable by setting #define WRITE_BYTEHUNK true (you'll have to modify the relative path to data/cache since this was used before the cache file was moved)

Afterwards just play in the game for a little bit and it'll write all the update packets it sends out and you can use `zstd --train -r dirname outfile --maxdict=64KB` to generate the raw dictionary file

After that you can use this python script to convert the dictionary file to a header file of bytes

[convert_dict_to_header.zip](https://github.com/user-attachments/files/19979886/convert_dict_to_header.zip)

code/components/citizen-server-impl/src/InitConnectMethod.cpp
```diff
@@ -564,7 +565,7 @@ static InitFunction initFunction([]()
 			auto gameBuildField = (gameBuildIt != postMap.end()) ? gameBuildIt->second : "0";
 			auto gameName = (gameNameIt != postMap.end()) ? gameNameIt->second : "";
 
-			if (protocol < 12)
+			if (protocol < kNetworkProtocolVersion)
```
FabianTerhorst, line 568: Isn't this breaking the compatibility when the current clients still send 12, because of the NETWORK_PROTOCOL macro?

AvarianKnight: Yes, I put the wrong protocol number here, will fix it when I get home this should be 12 and 13

AvarianKnight:
It should be noted that it would be best to merge this with the object I'd patch when it's finished (and train in that patch too) since it is also protocol breaking, it would be better to break all at once than to do two separate breakages.

FabianTerhorst (approved):
Changes itself are looking good to me. But backward compatibility needs to be evaluated independently.

PR #3249, opened 2025-03-15
feat(gamestate/server): sv_protectServerEntities

Ehbw:
### Goal of this PR

Prevent clients from being able to delete server-setter vehicles if ``sv_protectServerEntities`` is enabled. While also minimising the impact on the client that attempts the deletion.

### How is this PR achieving the goal

By blocking clone removal on the server side if the entity the client is attempting to delete is owned by a server-script. Checks on Functions related to delete Peds, Vehicles and Objects are hooked to ensure that these entities cannot be deleted by game code/natives. This prevents desync cases where the entity owner has deleted the entity on their client. 

### This PR applies to the following area(s)

FiveM, Server

### Successfully tested on

A test resources used to ensure the new changes worked and previous behaviour remained the same is provided with more details being found in server.lua: 
[entity-protection.zip](https://github.com/user-attachments/files/19267120/entity-protection.zip)

**Game builds:**  1604, 3258, 3407

**Platforms:** Windows

### Checklist

- [x] Code compiles and has been tested successfully.
- [x] Code explains itself well and/or is documented.
- [x] My commit message explains what the changes do and what they are for.
- [x] No extra compilation warnings are added by these changes.

### Fixes issues
resolves #962

PR #3387, opened 2025-05-09
fix(net/five): factory nullptr check

Ehbw:
### Goal of this PR

Fixes an oversight introduced in #3249 

Check if ped/vehicle/object is a nullptr before trying to call GetNetObject, matching the behaviour of the original functions.

FiveM

### Successfully tested on

**Game builds:** .. 

**Platforms:** Windows

### Checklist

### DIFF
--- a/code/components/gta-net-five/src/NetServerObject.cpp
+++ b/code/components/gta-net-five/src/NetServerObject.cpp
@@ -43,24 +43,41 @@ static bool CanEntityBeDeleted(rage::netObject* netObject)
 static void (*g_origCVehicleFactoryDestroy)(void*, CVehicle*, bool);
 static void CVehicleFactory_Destroy(void* self, CVehicle* vehicle, bool unk)
 {
+	if (!vehicle)
+	{
+		return;
+	}
+
 	if (!icgi->OneSyncEnabled || !vehicle->GetNetObject() || CanEntityBeDeleted((rage::netObject*)vehicle->GetNetObject()))
 	{
 		g_origCVehicleFactoryDestroy(self, vehicle, unk);
 	}
 }
 
-static void (*g_origCPedFactoryDestroy)(void*, fwEntity*, bool);
-static void CPedFactory_Destroy(void* self, fwEntity* ped, bool unk)
+static bool (*g_origCPedFactoryDestroy)(void*, fwEntity*, bool);
+static bool CPedFactory_Destroy(void* self, fwEntity* ped, bool unk)
 {
+	if (!ped)
+	{
+		return false;
+	}
+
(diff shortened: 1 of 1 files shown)


code/components/gta-net-five/src/NetServerObject.cpp
```diff
@@ -52,6 +57,12 @@ static void CVehicleFactory_Destroy(void* self, CVehicle* vehicle, bool unk)
 static void (*g_origCPedFactoryDestroy)(void*, fwEntity*, bool);
 static void CPedFactory_Destroy(void* self, fwEntity* ped, bool unk)
```
Nobelium-cfx, line 58: Original function also seems to return char instead of void. 0 if entity was not deleted, 1 if it was. Not sure if it actually affect anything, but lets fix.

Vehicle and object functions are ok tho - they do return void.

Nobelium-cfx (approved):
LGTM overall.

PS: It seems that only ped function does the not null check. Other original functions will crash if nullptr is given. But lets keep all added checks - maybe it will make things a bit more stable.

code/components/gta-net-five/src/NetServerObject.cpp
Ehbw: Fixed

Nobelium-cfx: You forgot to return value if `(!icgi->OneSyncEnabled || !ped->GetNetObject() || CanEntityBeDeleted((rage::netObject*)ped->GetNetObject()))` is false

Ehbw: Made it return false, as that does seem to make the most sense as the ped wouldn't get removed

code/components/gta-net-five/src/NetServerObject.cpp
```diff
@@ -43,24 +43,41 @@ static bool CanEntityBeDeleted(rage::netObject* netObject)
 static void (*g_origCVehicleFactoryDestroy)(void*, CVehicle*, bool);
 static void CVehicleFactory_Destroy(void* self, CVehicle* vehicle, bool unk)
 {
+	if (!vehicle)
```
JOVAN855, line 46: static void CVehicleFactory_Destroy(void* self, CVehicle* vehicle

JOVAN855: static void CVehicleFactory_Destroy(void* self, CVehicle* vehicle

JOVAN855 (commented):
static void CVehicleFactory_Destroy(void* self, CVehicle* vehicle

PR #3392, opened 2025-05-11
feat(core/five): patch script ptfx limits

Ekinoxx0:
### Goal of this PR

While trying to create a spell system in game, i managed to hit the script ptfx very easily with other players.
I tried to cast ~10 spells in 20 seconds with 20 other players around me, making a total of 200 ptfx.
This made a lot of ptfx created with `StartParticleFxLoopedAtCoord` invisible, because the game couldn't attribute script id to all my requests.

By overriding the init function we have access to multiple limits the game hardcode, the specific "128 ptfx" limit can be changed to 256 very safely, other limits could be changed later but need more research.

FiveM

### Successfully tested on

**Game builds:** 1604 / 2802 / 3258 / 3407

**Platforms:** Windows

Script used to test :
```
CreateThread(function()
    RequestNamedPtfxAsset('core')
    local coords = vector3(-1050.7, 458.62, 185.3)

    local i = 0
    for x = 1, 32 do
        for y = 1, 32 do
            UseParticleFxAssetNextCall('core')
            local particle = StartParticleFxLoopedAtCoord('fire_wrecked_plane_cockpit', coords.x + x * 2, coords.y + y * 2, coords.z, 0.0, 0.0, 0.0, 1.0, false, false, false, false)
            if particle == 0 then print('BROKE AT', i) break end
            i += 1
        end
    end
end)
```
image

### Checklist

### Fixes issues

None

Special thanks to @Mathu-lmn for the help testing and reversing things around this feature.
Also thanks to @tens0rfl0w for the tip around hook::trampoline

### DIFF
--- a/code/components/gta-core-five/PtfxScriptLimit.cpp
+++ b/code/components/gta-core-five/PtfxScriptLimit.cpp
@@ -1,0 +1,32 @@
+/*
+ * This file is part of the CitizenFX project - http://citizen.re/
+ *
+ * See LICENSE and MENTIONS in the root of the source tree for information
+ * regarding licensing.
+ */
+
+#include <StdInc.h>
+#include <Hooking.h>
+#include <Hooking.Stubs.h>
+
+/*
+ * Override the value (5th parameter) used in the "ptfx manager init" function to create script ptfx array
+ * On b1604 to b3407 default value is 128
+ *
+ * The game keep track of script ptfxs in an atArray defined as :
+ * struct scriptPtfx {
+ *     int a;
+ *     bool b;
+ *     bool c;
+ *     char padding[2];
+ * };
+ *
+ * This was tested with a limit up to 1024 without having any obvious bad side effects, but the game doesn't render some fx above ~384?
+ * This seems related to PtFxSortedEntity pool, but i didn't manage to resolve the rendering issue
+ * I also tried making other parameters of this function to bigger values but it didn't change anything
+ */
+
+static HookFunction hookFunction([]
+{
+  hook::put<int32_t>(hook::get_pattern("C7 44 24 ? ? ? ? ? E8 ? ? ? ? 48 8B CB E8 ? ? ? ? 48 8B CB E8 ? ? ? ? 48 8B CB", 4), 256);
+});


code/components/gta-core-five/PtfxScriptLimit.cpp
```diff
@@ -0,0 +1,46 @@
...
+ * struct scriptPtfx {
+ *     int a;
+ *     bool b;
+ *     bool c;
+ *     char padding[2];
+ * };
+ * So making it bigger by 128 shouldn't have any meaningful effect.
+ *
+ * This was tested with a limit up to 1024 without having any obvious bad side effects, but the game doesn't render some fx above ~384
+ * This seems related to PtFxSortedEntity pool, but i didn't manage to resolve the rendering issue
```
DaniGP17, line 34: For the issue that the game doesn't render above 384 fx did you try to increase the 2 arg of the Init method? Also if the issue is the PtFxSortedEntity pool I think that can be increased using the [increase pool size](https://docs.fivem.net/docs/server-manual/server-commands/#increase_pool_size-poolname-increase) convar but may be this has to be allowed by cfx.

Mathu-lmn: Hey, we did try to increase all params to like 5*1024 for the sake of testing but the draw limit didn't change unfortunately 
For the pool, it isn't listed in the pool monitor so I'm not sure it can be increased like other pools

Mathu-lmn:
> Lets maybe use something like this:
> 
> ```c++
> hook::put<int32_t>(hook::get_pattern("C7 44 24 ? ? ? ? ? E8 ? ? ? ? 48 8B CB E8 ? ? ? ? 48 8B CB E8 ? ? ? ? 48 8B CB", 4), 256);
> ```
> 
> so we don't need to use a trampoline here.

Thanks we'll do that asap !

code/components/gta-core-five/PtfxScriptLimit.cpp
Ekinoxx0: Hello, just tried to increase pool size with gameconfig.xml modification and also with a custom "increase pool size" modification, it didn't solve the rendering issue

Here is a video of it https://img.flashfa.fr/file/eki/282e1231-d0b2-4621-b659-9d5af47736a2.mp4

It seems the limit is linked to the number of particle rules, not the particles themself, because we can see either the smoke or the flame by themself, and not other particle rules.

This was a fun experiment, but either way we don't really need to display 1024 particles i the same time for now i suppose. The 128 increase seems reasonable.

prikolium-cfx (approved):
Looks good. Thanks!

PR #3402, opened 2025-05-15
feat(graphics/rdr3): add crashometry for vulkan

PappaNiels:
### Goal of this PR

This pr will add crashometry of the resources that are used by the Vulkan API. The data that I added are the following: GPU name, Vulkan APIversion, and driver version. We want to have this data, to diagnose the Vulkan crashes more precisely. To add to that, I also expanded the error handling in glue. Now, instead of having an assert and not having no data, we display the VkResult, so we know what went wrong. 
The validation layers could also be enabled, but it when it is enabled in the instance, it fails and falls back to d3d12. Matter of fact, you need the Vulkan SDK installed, hence the removal for now.

This pr adds more information to try diagnose crashes, and hopefully guide to a faster solution.

RedM

### Successfully tested on

**Game builds:** 1491  

**Platforms:** Windows

### Checklist

### Fixes issues

NA

Thanks @Ehbw for the patterns.

### DIFF
--- a/code/components/rage-graphics-rdr3/src/RenderHooks.cpp
+++ b/code/components/rage-graphics-rdr3/src/RenderHooks.cpp
@@ -7,7 +7,12 @@
 
 #include <CoreConsole.h>
 
+#include <VulkanHelper.h>
+#include <CrossBuildRuntime.h>
+
 #pragma comment(lib, "vulkan-1.lib")
+
+static bool g_enableVulkanValidation = false;
 
 // Function to print the output of the validation layers
 VKAPI_ATTR VkBool32 VKAPI_CALL DebugMessageCallback(VkDebugUtilsMessageSeverityFlagBitsEXT messageSeverity, VkDebugUtilsMessageTypeFlagsEXT messageType, const VkDebugUtilsMessengerCallbackDataEXT* pCallbackData, void* pUserData)
@@ -31,11 +36,11 @@ VKAPI_ATTR VkBool32 VKAPI_CALL DebugMessageCallback(VkDebugUtilsMessageSeverityF
 		prefix = "ERROR: ";
 	}
 
-	trace("%s [%i][%s]: %s\n", prefix, pCallbackData->messageIdNumber, pCallbackData->pMessageIdName, pCallbackData->pMessage);
+	trace("%s[%i][%s]: %s\n", prefix, pCallbackData->messageIdNumber, pCallbackData->pMessageIdName, pCallbackData->pMessage);
 	return VK_FALSE;
 }
 
-static inline std::string GetGPUType(VkPhysicalDeviceType type)
+static inline std::string_view GetGPUType(VkPhysicalDeviceType type)
 {
 	switch (type)
 	{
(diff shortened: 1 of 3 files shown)


PappaNiels:
Example data that it will spit out in the console:
image

code/components/rage-graphics-rdr3/src/RenderHooks.cpp
```diff
@@ -0,0 +1,211 @@
...
+			return false;
+	}
+}
+
+static VkResult __stdcall vkCreateInstanceHook(VkInstanceCreateInfo* pCreateInfo, VkAllocationCallbacks* pAllocator, VkInstance* pInstance)
+{
+	// the validation layers should be disabled when playing, but enabled when encountering a crash
+	// if we disable these now, we fallback to dx12 for some reason
+#if 0
+	std::vector<const char*> originalLayers(pCreateInfo->enabledLayerCount + 1);
```
prikolium-cfx, line 79: Can this be putted under compilation flag or ConVar if that can be ever needed for end-user?

code/components/rage-graphics-rdr3/src/RenderHooks.cpp
```diff
@@ -0,0 +1,211 @@
...
+		case VK_PHYSICAL_DEVICE_TYPE_CPU:
+		default:
+			return false;
+	}
+}
+
+static VkResult __stdcall vkCreateInstanceHook(VkInstanceCreateInfo* pCreateInfo, VkAllocationCallbacks* pAllocator, VkInstance* pInstance)
+{
+	// the validation layers should be disabled when playing, but enabled when encountering a crash
+	// if we disable these now, we fallback to dx12 for some reason
```
prikolium-cfx, line 77: What does it mean? I think it will fallback to DirectX12 if VkResult is not VK_SUCCESS, won't it?

code/components/rage-graphics-rdr3/src/RenderHooks.cpp
```diff
@@ -0,0 +1,211 @@
...
+		trace("Vulkan Device Layer: %s", originalLayers[i]);
+	}
+
+	originalLayers.push_back("VK_LAYER_KHRONOS_validation");
+
+	pCreateInfo->enabledLayerCount = static_cast<uint32_t>(originalLayers.size());
+	pCreateInfo->ppEnabledLayerNames = originalLayers.data();
+#endif
+
+	static auto _ = ([&physicalDevice]
```
prikolium-cfx, line 144: Could you please make it more obvious, like move to external function and use `initialized` flag or so.

code/components/rage-graphics-rdr3/src/RenderHooks.cpp
PappaNiels: That was the idea as well. Sadly the validation layers will fail the instance creation as of now. I am debugging what is going on right now,

code/components/rage-graphics-rdr3/src/RenderHooks.cpp
PappaNiels: I was a bit tired yesterday I see. What I meant (as mentioned before), the validation layers are disabled because it defaults back to DX12, due to not being able to create an Vulkan instance.

code/components/rage-graphics-rdr3/src/RenderHooks.cpp
PappaNiels: On it! 🫡

PappaNiels:
I have updated the pr with the requested changes and I removed the unclear comments. Now, if you set the variable `EnableVulkanValidation` to 1 in the CitizenFX.ini file, you will enable the validation layers. 

The only must is the Vulkan SDK then, which should be documented somewhere where possible, If someone will try to use the validation layers without the SDK, it will mention it in the logs. But a heads up would be better. 

If there are any other concerns, feel free to let me know.

prikolium-cfx (approved):

prikolium-cfx:
Thank you for the contribution

PR #3404, opened 2025-05-17
feat(server): add external kvp natives

outsider31000:
### Goal of this PR

Add external KVP natives to the server, matching the client-side implementation. This is useful for enabling communication between resources on the server.

Aligns the server with client KVP natives by implementing the same functionality server-side.

Server

### Successfully tested on

**Game builds:** .. 

**Platforms:** Windows, Linux
windows

### Checklist

### DIFF
--- a/code/components/citizen-server-impl/src/ServerKVScriptFunctions.cpp
+++ b/code/components/citizen-server-impl/src/ServerKVScriptFunctions.cpp
@@ -204,6 +204,21 @@ struct SerializeValue<RawType>
 };
 
 template<typename T>
+static void GetResourceKvp_Impl(const std::string& key, fx::ScriptContext& context)
+{
+	auto db = EnsureDatabase();
+
+	std::string value;
+	if (db->Get(rocksdb::ReadOptions{}, key, &value).IsNotFound())
+	{
+		context.SetResult<const char*>(nullptr);
+		return;
+	}
+
+	SerializeValue<T>::Deserialize(context, value);
+}
+
+template<typename T>
 static void GetResourceKvp(fx::ScriptContext& context)
 {
 	auto db = EnsureDatabase();
@@ -217,6 +232,14 @@ static void GetResourceKvp(fx::ScriptContext& context)
 	}
 
 	SerializeValue<T>::Deserialize(context, value);
+}
+
+template<typename T>
+static void GetExternalKvp(fx::ScriptContext& context)
+{
+	auto key = FormatKey(context.CheckArgument<const char*>(1), context.CheckArgument<const char*>(0));
+
+	return GetResourceKvp_Impl<T>(key, context);
 }
 
 #include <memory>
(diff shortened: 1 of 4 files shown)


FabianTerhorst:
I'd prefer to have the key value db isolated on the server. Maybe there is a better way to have specific keys available between different resources. But right now best way is to only have one resource that writes that specific key.

outsider31000:
The implementation is the same as the client though.
Kvp is also isolated per resource so it shouldn't matter the key being used?,  Unless I'm not seeing the bigger picture here.

Without this I need to trigger events with callbacks to another resource that people might not have and the callback is never returned.  Same happens with exports.

This change simplifies it. These are getters don't see what harm can it do.

Thanks for the review ofc.

FabianTerhorst:
> The implementation is the same as the client though. Kvp is also isolated per resource so it shouldn't matter the key being used?, Unless I'm not seeing the bigger picture here.
> 
> Without this I need to trigger events with callbacks to another resource that people might not have and the callback is never returned. Same happens with exports.
> 
> This change simplifies it. These are getters don't see what harm can it do.
> 
> Thanks for the review ofc.

With isolation I mean that no other resource can read the key from another resource on the server. Because the key is prefixed with the resource name this is not possible currently with the other kv getters. With your change a resource could not know which other resources access the data. When a resource stores sensitive data inside the key value db this data could be grabbed from any resource.

theskiratta:
I'm a bit confused by your statement of  `"Without this I need to trigger events with callbacks to another resource that people might not have and the callback is never returned. Same happens with exports.".`
1.: The same resources should be running on the server and all clients?
2,; Following your assumption of inconsistency of resources, what difference would it make whether you do this through a callback/export or through KVP? If the resource isn't there, it isn't there, so I don't see how this would change an outcome?

outsider31000:
trigger an event that doesnt exist and check when that callback is returned you will have your answer.

outsider31000:
exports will also error when they dont exist.
using external kvp , it would not need to wait for no callback, it returns always something. 
it simplifies its usage and aligns it with the client natives that do the same.

Not adding this because we dont want to allow other scripts getting data from other scripts its not a valid reason in my opinion, because its no different saving to a mysql database. I would then save the same very data to a database and fecth it anyways from any resource, so why should KVP be any different...

WLVF:
> exports will also error when they dont exist. using external kvp , it would not need to wait for no callback, it returns always something. it simplifies its usage and aligns it with the client natives that do the same.
> 
> Not adding this because we dont want to allow other scripts getting data from other scripts its not a valid reason in my opinion, because its no different saving to a mysql database. I would then save the same very data to a database and fecth it anyways from any resource, so why should KVP be any different...

Could be getting a bit off the trail here, but if you're having concerns with erroring exports on missing resources, couldn't using a pcall in your code fix your issue? And with events, some form of timeout function with a return value?

prikolium-cfx:
Thanks for your efforts.
This functionality is fundamentally incompatible with the server, since sensitive information may be stored on the server. If you need to exchange information between resources, use function export or events.

PR #3415, opened 2025-05-22
fix(voip-server-mumble): disallow external mumble connections

DaniGP17:
### Goal of this PR
Block external connections (outside the FiveM client) to the Mumble server by default. Malicious users connect to this server and are able to play audio and speak to players on the server without even being inside the fivem server.
I'd like to get a feature branch made based on this change so I can test it thoroughly on a server with many people.

This problem was previously addressed in PR #3264 by verifying that the IP address of incoming connections to the Mumble server was assigned to a player. However, I noticed an issue with this approach: servers using proxies tend to recognize all player-related incoming connections as coming from the same IP(the proxy’s IP). That PR was reverted because wasn't allowing connections from any ip, when investigating I saw that `address.GetHost()` returns the IP masked as ipv6 and the client registry store the addresses as ipv4, If I remember correctly, even though I modified that so that it wouldn't mask it to IPv6, it still didn't work for another reason, but I preferred to choose this solution because I find it more difficult to circumvent if using a proxy.

Although I don't have deep knowledge of proxies or networking and may be what I'm talking about don't make full sense, I discovered that it's possible to bypass this protection. By attempting to connect to the FiveM server (even if the connection is rejected due to a whitelist), and then connecting to the Mumble server, the Mumble server still sees the connection as originating from the proxy IP. 

First of all, in the authentication message I have added a check for the `os_version` and the `release` that the external client sends, the fivem client sends [these](https://github.com/citizenfx/fivem/blob/master/code/components/voip-mumble/src/MumbleClient.cpp#L1113-L1114).
The username of mumble that comes from fivem are structured like this `[id] username`, I've extracted the id to check if the player is in the server and check that the player only have one connection to the mumble server.

Server

### Checklist


MathiAs2Pique:
Hello,

I don't really get the point of checking the os version / release, as these two can easily be crafted with a python script for example. The Id check is good IMHO, but actually it's too easy to bypass it, by just naming with the right pattern + unused ID. If you want to achieve this, I guess it would be better to check for in-game player IDs rather than in-mumble IDs.

Gogsi:
> If you want to achieve this, I guess it would be better to check for in-game player IDs rather than in-mumble IDs.

Oh I didn't realize it doesn't actually check if that ID is in-game. I agree that should be done. It will only leave a small vector where you could possibly spam random IDs in hope that you join Mumble just as a player joins, but that seems unlikely.

DaniGP17:
> I don't really get the point of checking the os version / release, as these two can easily be crafted with a python script for example. The Id check is good IMHO, but actually it's too easy to bypass it, by just naming with the right pattern + unused ID. If you want to achieve this, I guess it would be better to check for in-game player IDs rather than in-mumble IDs.

Regarding checking os_version and release, I think it's best to try to make it as difficult as possible for people trying to do this. In my case, people entering the Mumble server used a simple client that didn't allow changing that data. For some reason, I haven't checked that the player ID is connected to the server (in my head, I thought I had already done so). I'll do it this afternoon sorry :sweat_smile:

DaniGP17:
> > If you want to achieve this, I guess it would be better to check for in-game player IDs rather than in-mumble IDs.
> 
> Oh I didn't realize it doesn't actually check if that ID is in-game. I agree that should be done. It will only leave a small vector where you could possibly spam random IDs in hopes that you join Mumble just as a player joins with the same ID, but that seems unlikely.

Yes, I just realized that lol

prikolium-cfx:
> Hi @prikolium-cfx, can we get another update in the feature branch? This should be the last one if all goes well. Thank you

Should have effect in ~20 minutes

code/components/voip-server-mumble/src/client.cpp
```diff
@@ -455,6 +455,23 @@ void Client_free(client_t *client)
...
 }
 
+int Client_getPlayerId(client_t* client, char* username)
+{
+	int playerId = -1;
+	if(!username)
+	{
+		return -1;
+	}
+	if (strncmp(username, "[", 1) == 0) {
```
authclient, line 465: Single char check can probably be replaced with `if (username[0] == '[') {`
Also there is no reason to use c only code in a cpp project.
Convert the username to a string view first for this.
It would make sense to verify first if the ] exists as well before trying to convert it.
```cpp
if (endPos != std::string::npos) {
  size_t endPos = username.find(']', 1);
  if (username.size()  - endPos - 2/*[ and ]*/ > 0) {
    std::string_view numberView = username.substr(1, endPos - 1);
    auto [ptr, ec] = std::from_chars(numberView.data(), numberView.data() + numberView.size(), playerId);
    if (ec != std::errc()) {
      playerId = -1;
    }
  }
}

```

For ´std::from_chars´ you can include ´<charconv>´

code/components/voip-server-mumble/src/sharedmemory_struct.h
```diff
@@ -21,7 +21,7 @@ typedef struct
 typedef struct
 {
   int shmtotal_size, shmclient_size;
-  int clientcount, server_max_clients;
+  int clientcount, server_MUMBLE_MAX_CLIENTS;
```
FabianTerhorst, line 24: This rename doesn't seems to be intended?

code/components/voip-server-mumble/src/server.h
```diff
@@ -31,8 +31,16 @@
...
+#include <unordered_set>
+
+#include "ClientRegistry.h"
+#include "console/Console.VariableHelpers.h"
 
 void Server_run();
 void Server_shutdown();
 
+extern std::shared_ptr<ConVar<bool>> mumble_allowExternalConnections;
+inline std::unordered_set<int> g_clientIds;
```
FabianTerhorst, line 43: Is the g_clientIds modified from two different threads? The close callback and the io worker that is used to call the mumble message handler. In that case better add a lock guard.

DaniGP17: Good catch, already fixed!

FabianTerhorst: Looks good now, thanks.

FabianTerhorst (approved):
Thanks a lot for the contribution.

Identity-labs:
Should be nice if disallow is active to have hosts whitelist to bypass this cond

DaniGP17:
> Should be nice if disallow is active to have hosts whitelist to bypass this cond

Oh yeah, didn't think about it :sweat_smile:

Sasino97:
This PR broke our system for moderators to monitor voice chats while not being in-game, I guess at least there should be a way for us to still connect to the mumble server if we had a key or something?

DaniGP17:
> This PR broke our system for moderators to monitor voice chats while not being in-game, I guess at least there should be a way for us to still connect to the mumble server if we had a key or something?

Sorry for that, until I or someone add a key for that you can enable this convar `mumble_allowExternalConnections`

Sasino97:
> > This PR broke our system for moderators to monitor voice chats while not being in-game, I guess at least there should be a way for us to still connect to the mumble server if we had a key or something?
> 
> Sorry for that, until I or someone add a key for that you can enable this convar `mumble_allowExternalConnections`

Thank you, this will do 🙏

S4CKY585:
Instead of making a key, perhaps just IP whitelisting **could** be simple for you to make yet protective enough to stop other malicious users from joining and trolling.

but besides that this is great, thanks

DaniGP17:
> Instead of making a key, perhaps just IP whitelisting **could** be simple for you to make yet protective enough to stop other malicious users from joining and trolling.

The issue is with servers using reverse proxies, that all connecting ips are the same so it's difficult to filter by that. I think this was the best solution for all servers.

WoenselRP:
Hello

Perhaps it would be a good idea to allow clients to join, even if they are external if they present the right ADMIN_PASSPHRASE

It seems this is unused by normal clients anyways. I think it would be a good addition for those who want create out-game moderation tools.

If there is intrest I could propose a quick PR

S4CKY585:
I am interested in something like that.

DaniGP17:
> If there is intrest I could propose a quick PR

If you want you can create yeah

gewrutv:
@DaniGP17 How does this work exactly when you use another FXserver as the external mumble server. To get users connected we had to set mumble_allowExternalConnections to true on the FXserver serving as the mumble server.

DaniGP17:
> @DaniGP17 How does this work exactly when you use another FXserver as the external mumble server. To get users connected we had to set mumble_allowExternalConnections to true on the FXserver serving as the mumble server.

Never used a external mumble server but I think my PR don't affect that because the external mumble is the one that handle the connections not the fxserver in that case.

gewrutv:
> > @DaniGP17 How does this work exactly when you use another FXserver as the external mumble server. To get users connected we had to set mumble_allowExternalConnections to true on the FXserver serving as the mumble server.
> 
> Never used a external mumble server but I think my PR don't affect that because the external mumble is the one that handle the connections not the fxserver in that case.

What I see happen is that the users are not connected to the server so they are all counted as not being "authenticated". The only work around I have seen work successfully is "mumble_allowExternalConnections true" on the FXserver serving as the mumble server. I believe someone was working on a PR to allow external connections if they present the right key. I guess I will have to see how that progresses.

Issue #3401, opened 2025-05-15
Physics Validation Error Vehicle YFT's - Invalid Octant Data (OctantVertCounts <=0)


ThatGuyWookie:
### What happened?

![Image](https://github.com/user-attachments/assets/b3984664-38cc-4e72-8811-e40c80d21897)

Our server owner noticed that a handful of vehicle yft's are now failing physics validation. We haven't added or updated our vehicles in 2 months but in the last few days they said they keep seeing these errors out of the blue. We haven't updated our game build or artifact either, only some scripts have been added.

I've never seen octant data (OctantVertCounts <=0) either, only polyedge in errors like this. The exact same vehicle pack is on our dev server and not spitting these errors, however, the dev server is the same build and artifact but has around 40 less resources.

### Expected result

No error at all.

### Reproduction steps

-> One of these vehicles is spawned into the server, either by trainer or being taken out of a garage.

-> Error immediately displays in console.

### Importancy

Unknown

### Area(s)

FiveM, FXServer

### Specific version(s)

Build 3258, Artifact 13943

Ehbw:
It is a new warning message, introduced in https://github.com/citizenfx/fivem/commit/96ef4ed615fda2fa0cd946e832e3b18db2a2c73c your assets have bad octants data.

ook3D:
to fix it, using the latest version of codewalker, export the .yft files as XML and re import them

radium-cfx:
Leaving this open a bit longer to gather additional feedback.

If you're still encountering these error messages after updating to a newer version of the exporter you used (e.g. CodeWalker), please attach the affected `.yft` files so we can investigate further.

nearofsize:
How does this harm performance?

ThatGuyWookie:
> to fix it, using the latest version of codewalker, export the .yft files as XML and re import them

I will give this a shot! I've been a tad busy with work, apologies for the delayed response!

Sasino97:
> It is a new warning message, introduced in [96ef4ed](https://github.com/citizenfx/fivem/commit/96ef4ed615fda2fa0cd946e832e3b18db2a2c73c) your assets have bad octants data.

A bit strange that more than half of my cars have this issue. Also why would this be spammed as a warning if nothing bad actually happens (even though the data is malformed)?

Gogsi:
> Also why would this be spammed as a warning if nothing bad actually happens

Prior to the fix in the same commit that added the message, such assets would cause crashes but only on game build 3407

[...]

PR #3422, opened 2025-05-28
tweak(gta-streaming-five): mitigate octants crash only on builds > 3407

d22tny:
### Goal of this PR
This PR aims to use the patch for the crash related to invalid octants that occurs only on 3407 game build. On older builds, this issue does not cause a crash, so no crash mitigation is needed.

To help server owners and developers diagnose the issue without overwhelming regular players, a new ConVar named enableOctantCheckWarning has been introduced. When enabled, this ConVar will log warnings when invalid octants are detected. It is disabled by default, meaning no warnings will be shown unless explicitly opted into.

This change ensures stability on newer builds while maintaining a clean experience for end users unless debugging is required.
...

This PR introduces a conditional check on the game build and only applies the crash mitigation patch if the game build is 3407 or higher, where the crash is known to occur.

Additionally, a new ConVar named enableOctantCheckWarning has been added. When set to true, it enables logging of warnings when invalid octants are detected. This is primarily intended for developers and server owners who wish to diagnose those issues. The ConVar defaults to false, so regular players will not see any warnings unless it is manually enabled.
...

FiveM Client

...

### Successfully tested on
**Game builds:** .. 
3258, 3407
**Platforms:** Windows, Linux

### Checklist

### Fixes issues
https://github.com/citizenfx/fivem/issues/3401


ook3D:
> Hello, sorry for jumping in.
> 
> My server is running game build 3095, and ever since the patch was applied, there has been a noticeable reduction in crash reports. This suggests that the benefits may extend beyond build 3407 to other game builds as well.
> 
> Based on that, I would propose the following:
> - A version check based on the game build does not appear to be necessary; a conditional check solely on enableOctantCheckWarning should suffice.
> - The default value for enableOctantCheckWarning can remain true, as before.

Yea, I'm not sure where this assumption it doesn't exist prior to 3407, but if a vehicle does not have octants and collides with something it will crash

d22tny:
So from what i understood, on 3407 having a vehicle model with this problem would crash instantly, but on other game builds it would, that's where i assumed it's not needed on other game builds, since it impacts performance. So what you say is that on other game builds it `maybe can` crash sometimes ? If that's the case i'll remove the game build check and only leave the convar.

imo the end user shouldn't see spamming assets going on like this, that's why we need a convar like this, and i think it should default to false, if a server owner/dev wants to optimize his runtime performance, loading times etc, he can enable the warnings and probably delete those cars, or find good ones, because most assets sellers don't really care about performance.

ook3D:
> So from what i understood, on 3407 having a vehicle model with this problem would crash instantly, but on other game builds it wouldn't, that's where i assumed it's not needed on other game builds, since it impacts performance. So what you say is that on other game builds it `maybe can` crash sometimes ? If that's the case i'll remove the game build check and only leave the convar.
> 
> imo the end user shouldn't see spamming assets going on like this, that's why we need a convar, and i think it should default to false, if a server owner/dev wants to optimize his runtime performance, loading times etc, he can enable the warnings and probably delete those cars, or find good ones, because most assets sellers don't really care about performance.

Yea I agree, should be an option to disable these, no reason for the average user to be shown these

d22tny:
@prikolium-cfx please tell me if this change is all right and you agree with it, and if i should make a more generic ConVar ( showAssetValidationWarnings ) and include both octants and physics validations prints in it.

FlawwsX:
I feel like these warning messages should only show when on Canary or Developer mode. Plus, this convar should be true by default, in my opinion, or it can be added to the F8 menu as a toggle. It’s getting awfully difficult to remember all these commands and convars 😭😭

d22tny:
> I feel like these warning messages should only show when on Canary or Developer mode. Plus, this convar should be true by default, in my opinion, or it can be added to the F8 menu as a toggle. It’s getting awfully difficult to remember all these commands and convars 😭😭

Should be true means - You shouldn't see the warnings on default behaviour or you should see them ?

DaniGP17:
> Should be true means - You shouldn't see the warnings on default behaviour or you should see them ?

I think that by default the value of the convar should be true (that is, warnings should be shown by default) because if someone has a crash related to this, I'm pretty sure they won't remember to enable this convar.

Gogsi:
I think this can be made like some other logs - on the Release channel, it's only shown in the log file (for crash diagnostics) and not in the console. On Beta/Canary, it's also shown in the console for easy debugging. I'm not sure how this is done though 😄

d22tny:
I've collided 2 vehicles with the error, nothing happened.

PR #3427, opened 2025-06-01
feat(extra-natives-five): allow manipulation of ped reaction to sirens

nikez:
### Goal of this PR

Allow users to control how pedestrian vehicles react to sirens.

By adding two new natives.

FiveM

### Successfully tested on

**Game builds:** 2802

**Platforms:** Windows, Linux

### Checklist


prikolium-cfx (approved):

pnwparksfan:
@nikez I would like to suggest revisiting this change to make it work on a per-vehicle basis. See how `g_skipRepairVehicles` / `SET_VEHICLE_AUTO_REPAIR_DISABLED` works as a guide. Since `g_origReactToSirens` already has a `CVehicle` instance as an argument, it should be trivial to check whether the particular vehicle is contained within a set of IDs, and to create a native that adds/removes from this set. This would allow much more granular control over which vehicles cause ambient drivers to pull over when their sirens are on.

PR #3429, opened 2025-06-02
fix(lua): fix file write always truncating to zero

### LINKED ISSUE #3428: Log files truncated with each new line written
### What happened?

https://github.com/citizenfx/fivem/commit/ce8a6d0aff9ec8b457c43d9c13af05dea9b3d360#diff-384d2170a25c048a23f01e2f5b3955dee9fe62fe078a9ccbdfb060fffc236442

With presumably this change, in server build 15488, our text log files are now not able to be written to as the file is truncated with each new write.  using code similar to below

  file = io.open('@vrp/' .. GlobalState.playerLogFileName, "a")
  if file ~= nil then
    file:write(json.encode(line), "\n")
    file:close()
  end

### Expected result

Allow us to write to our logs...

### Reproduction steps

  file = io.open('@vrp/' .. GlobalState.playerLogFileName, "a")
  if file ~= nil then
    file:write(json.encode(line), "\n")
    file:close()
  end

### Importancy

Unknown

### Area(s)

ScRT: Lua

### Specific version(s)

Fivem Server 15488

### Additional information

_No response_

Mathu-lmn:

Added logic to truncate the file to the current position only if the new position is smaller than the original length, ensuring unnecessary truncation is avoided especially before writing, which broke appending to files

Lua io lib, FxServer

### Successfully tested on

**Platform:** Windows
Used these commands and got the expected result for each : 
```lua
RegisterCommand('test1', function()
    local file = io.open('@ava/test.txt', 'w+')
    if file then
        file:write('Test')
        file:write('te')
        file:close()
    end
end)

RegisterCommand('test2', function()
    local file = io.open('@ava/test.txt', 'a+')
    if file then
        file:write('Test')
        file:write('te')
        file:close()
    end
end)
```

### Checklist

### Fixes issues
Fixes #3428

### DIFF
--- a/code/components/citizen-scripting-lua/src/LuaIO.cpp
+++ b/code/components/citizen-scripting-lua/src/LuaIO.cpp
@@ -844,7 +844,8 @@ int LuaIOFileWrite(lua_State* L)
 	const fwRefContainer<vfs::Stream> f = LuaIOToFile(L);
 
 	auto device = f->GetDevice();
-	device->Truncate(f->GetHandle(), 0);
+	auto handle = f->GetHandle();
+	size_t originalLen = f->GetLength();
 
 	// push file at the stack top (to be returned)
 	lua_pushvalue(L, 1);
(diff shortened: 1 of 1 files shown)


radium-cfx (approved):
Thanks for the quick fix, works as intended now!🥳

Issue #1193, opened 2021-12-18
RedM crashes with default config on Intel/NVIDIA hybrid GPU setups


blattersturm:
From support chats:

![image](https://user-images.githubusercontent.com/24576130/146641233-966866cd-d4d0-4a43-b809-47758bffac55.png)

It might be the case that Vulkan stuff needs an equivalent of the IDXGIAdapter selection-by-high-performance thing, _if_ that even exists in the Vulkan API (or if not, a way to get a Vulkan device selection from a DXGI device - using the LUID maybe?).

This may also affect 'pure' Intel setups.

gottfriedleibniz:
Disquse, is this still an issue?

Disquse:
@gottfriedleibniz, yes, it's still a thing.

PR #3433, opened 2025-06-02
fix(graphics/rdr3): force dedicated gpu

PappaNiels:
### Goal of this PR
This pr will (re)try to select a dedicated GPU for Vulkan if possible. We found that people were crashing due to a lack of memory, and (some of) those people were using their IGPU rather then their dedicated GPU. The dedicated GPU has more memory, thus this crash will be resolved (hopefully) (and the game would be ran on the IGPU, which would not be a nice experience). 

Since I was not able to reproduce this issue on my own laptop with a hybrid GPU setup, it would be a better idea to have a test branch.

We use the hook that was there for `vkCreateDevice()` and we check if the GPU is a dedicated GPU. If it is not, it will try to select a dedicated GPU from the GPUs that are available. It checks if the extensions are supported and then uses it to create a virtual device. If there are none, the user will get a message that it is using an IGPU, and the initial GPU will be used, that was passed by RAGE.

RedM

### Successfully tested on
**Game builds:** 1491

**Platforms:** Windows

### Checklist

### Fixes issues

Possible fix for [this issue](https://github.com/citizenfx/fivem/issues/1193)


PappaNiels:
Hi! Although it is great to see this being merged, I have not seen many cases from my perspective that it has improved much cases. Did you guys test this and get the desired results? Since I do not have access to that, that might be the case, but I would live to know if my contributions are indeed fixing the issue :). Thanks!

Ehbw:
> Hi! Although it is great to see this being merged, I have not seen many cases from my perspective that it has improved much cases. Did you guys test this and get the desired results? Since I do not have access to that, that might be the case, but I would live to know if my contributions are indeed fixing the issue :). Thanks!

It's hard to fully confirm if the linked issue was resolved with this patch given how it was dependent on certain system configurations (some intel systems with IGPUs and likely some specific driver/intel software) and we weren't able to find the right effected users to try out the patch. However forcing a dedicated GPU for Vulkan (rather then letting it possibly choose the IGPU) should resolve some variations of the no-ui issue with NUI and with it being on canary hopefully the users being effected by the linked issue can provide any sort of feedback on if this resolved/improved the situation.

The error message changes are also important as it can allow for some other vulkan issues (that could have also be caused by an IGPU being chosen over the dedicated GPU) to be properly investigated.

PR #3446, opened 2025-06-07
feat(conhost-v2): add log filter to console

DaniGP17:
### Goal of this PR
Add a text input field to allow users to filter the displayed logs in the console. This improves usability by enabling users to search for specific log entries or categories quickly.

Whenever I've wanted to search for specific logs, I've had to open the log in a text editor, and this makes it much more convenient to search for something specific throughout the console.

https://github.com/user-attachments/assets/cd9b5a6b-08d9-4af7-b216-b6b965127b4b

Created a FilterBuf buffer to store the filter text inputted by the user, then modified the log rendering loop to check if the FilterBuf is empty, if not, display only log entries or categories that contain the filter text.

FiveM

### Successfully tested on
**Platforms:** Window

### Checklist


tens0rfl0w:
This needs some adjustments to not break the text selection feature when filtering is active. Item rendering should still use the ImGUI clipper while filtering. Nice change though!😄

DaniGP17:
> This needs some adjustments to not break the text selection feature when filtering is active. Item rendering should still use the ImGUI clipper while filtering. Nice change though!😄

I'll check that, thanks for pointing that out

PR #3459, opened 2025-06-18
feat(gamestate/server): Add IS_OBJECT_A_PICKUP

Ekinoxx0:
### Goal of this PR

I'm currently dealing with an issue related to pickup creation exploits on my server.
I want to be able to detect and prevent the creation of pickups reliably.

At the moment, many servers work around this by using `GetEntityModel(entity) == 0` during `entityCreating`, but this approach is fragile and likely to break once the CPickupCreationDataNode is parsed properly (as introduced in PR https://github.com/citizenfx/fivem/pull/2940).
Additionally, we still can't access the raw entity type until PR https://github.com/citizenfx/fivem/pull/2945 is merged.

This PR introduces the `IS_OBJECT_A_PICKUP` native, matching the existing client-side counterpart.
Ideally, I’d prefer the other PR (https://github.com/citizenfx/fivem/pull/2945) to be accepted in order to unify this logic under a single getter, rather than relying on multiple `IS_..._A_...` checks.
However, since that PR has been stalled since November, this solution offers a more straightforward and less controversial path forward.

```lua
AddEventHandler('entityCreating', function(entity)
	if IsObjectAPickup(entity) then
		CancelEvent()
	end
end)
```

Server

### Successfully tested on

**Game builds:** All builds

**Platforms:** Windows, Linux

### Checklist

### Fixes issues

Important note:
This PR is useless if https://github.com/citizenfx/fivem/pull/2945 is accepted.

### DIFF
--- a/code/components/citizen-server-impl/src/state/ServerGameState_Scripting.cpp
+++ b/code/components/citizen-server-impl/src/state/ServerGameState_Scripting.cpp
@@ -1416,6 +1416,11 @@ static void Init()
 		return uint32_t(node ? node->curWeapon : 0);
 	}));
 
+	fx::ScriptEngine::RegisterNativeHandler("IS_OBJECT_A_PICKUP", makeEntityFunction([](fx::ScriptContext& context, const fx::sync::SyncEntityPtr& entity)
+    {
+        return entity->type == fx::sync::NetObjEntityType::Pickup || entity->type == fx::sync::NetObjEntityType::PickupPlacement;
+    }));
+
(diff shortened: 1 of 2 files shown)


AvarianKnight:
I was going to say "why not just add a native that returns the raw version" just to realize that there is already a pr that returns the raw version :P

Ekinoxx0:
Just modified the check so it's `PickupPlacement` or `Pickup`, rather than just `Pickup`

Ekinoxx0:
Closing as the raw type native (PR 2945) is now ready to merge

PR #3465, opened 2025-06-20
tweak(core/five): Make ped components validation deactivatable

BuddiesTV:
### Goal of this PR
Setting null drawables/invalid variations is sometimes used in scripts that make screenshots of clothing pieces, for example. With the validation in place, the head(or other components) cannot be invisible anymore, which breaks these screenshots, so making this validation deactivatable is crucial for these scripts to work. It also never crashed the game for me before the validation.

Adds the `game_disablePedComponentsValidation` convar. (The default is false)

FiveM

### Successfully tested on
Game builds: 3258
Platforms: Windows

### Checklist

### Fixes issues
N/A

### DIFF
--- a/code/components/gta-core-five/src/CrashFixes.ValidatePedComponents.cpp
+++ b/code/components/gta-core-five/src/CrashFixes.ValidatePedComponents.cpp
@@ -1,8 +1,10 @@
 #include "StdInc.h"
 
-#include "console/Console.Base.h"
+#include "CoreConsole.h"
 #include "Hooking.Patterns.h"
 #include "Hooking.Stubs.h"
+
+static bool g_disablePedComponentsValidation;
 
 enum ePedVarComp
 {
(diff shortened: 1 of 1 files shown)


tens0rfl0w:
This is not an option either as this data is replicated over the network. What can be done is to implement a native that allows to overwrite ped components locally.

If this would be the approach to go with I would only allow this ConVar to be set with `developer` mode set to `true` and not  as replicated and rather set per client. With this a ConVar would be acceptable.

Yum1x:
> What can be done is to implement a native that allows to overwrite ped components locally.

A similar behavior of [SET_PED_RESET_FLAG](https://docs.fivem.net/natives/?_0xC1E8A365BF3B29F2) with 166<sup><sub>(PRF_MakeHeadInvisble)</sup></sub> flag would be nice.

PR #3468, opened 2025-06-21
feat(gta-streaming-five): add convar to allow gen9 assets

DaniGP17:
### Goal of this PR
There are currently a file called `gen9_exclusive_assets_peds.meta` that define which clothes drawables are considered gen9 files to avoid loading them on PC (legacy version). I haven't found any file mounters to replace this file, so I don't think there's another way to do this.
Most of the gen9 files are empty placeholders, but the meta files for these assets (textures, first_person, pedalternatevariations, etc.) still exist, so it would be very useful to be able to export the enhanced files and use the same slots in legacy.

I would like to thank [Celest](https://forum.cfx.re/u/celestiamech/summary) because she explained the whole problem to me in great detail and provided me with a script to reproduce the problem and check the final solution.

More info about the issue in this [post](https://forum.cfx.re/t/g9ec-clothes-not-working-on-gta5-legacy/5334534).

Creating a convar `game_AllowGen9Assets` and if is true return false in the Gen9 exclusive hooked functions.

FiveM

### Successfully tested on
**Game builds:** 2699, 3407
**Platforms:** Windows

### Checklist

### DIFF
--- a/code/components/gta-streaming-five/src/LoadStreamingFile.cpp
+++ b/code/components/gta-streaming-five/src/LoadStreamingFile.cpp
@@ -3474,6 +3474,35 @@ static void CleanupStreaming()
 	g_unloadingCfx = false;
 }
 
+// By default, Gen9 assets are placeholders in PC(legacy version) and are skipped in SetVariation and SetModIndex
+// With this we can allow servers to manually those assets exported from enhanced
+static bool g_allowGen9Assets = false;
+
+static bool (*g_origIsGen9ExclusivePedDrawable)(void* ped, __int32 component, unsigned int drawableIdx);
+static bool IsGen9ExclusivePedDrawable(void* ped, __int32 component, unsigned int drawableIdx)
+{
+	if(g_allowGen9Assets)
+	{
+		return false;
+	}
+	return g_origIsGen9ExclusivePedDrawable(ped, component, drawableIdx);
+}
+
+static bool (*g_origIsGen9ExclusiveVehicleMod)(void* vehicle, int modSlot, unsigned __int8 modIndex);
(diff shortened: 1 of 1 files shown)


Celestiamech:
I would like to add that this will not only make Gen9 clothing work, it will also make all vehicle modifications also marked as exclusive Gen9 work, and although for visual modifications you need to load the 3d model separately, engine modifications and other technical modifications for certain vehicles will work immediately, the cars in question would be those that have the HAO Upgrade available in GTA Online (Only PS5, Xbox S/X or PC Enhanced)

Rm1998:
Why not just wait the 10 years that it'll take CFX to release support for enhanced on FiveM? This will likely only cause issues with stuff that is not meant to be loaded on legacy

ook3D:
> Why not just wait the 10 years that it'll take CFX to release support for enhanced on FiveM? This will likely only cause issues with stuff that is not meant to be loaded on legacy

? These are legacy files

Rm1998:
> > Why not just wait the 10 years that it'll take CFX to release support for enhanced on FiveM? This will likely only cause issues with stuff that is not meant to be loaded on legacy
> 
> ? These are legacy files

You're right. I misread the message - My apologies

prikolium-cfx:
Thank you for your efforts, but we can't accept this because we want to preserve Gen9 exclusivity for these assets.

DaniGP17:
Yeah np, I guess when fivem get the enhanced version we will be able to use those assets without restriction. @Celestiamech

DaniGP17:
And wouldn't this PR be ok if the approach were different? There are many Gen9 slots that are empty and occupying ymt slots, so we can use our custom clothing in those slots. The name could be something like game_unlockGen9Slots. Because by default even if you unlock them you can't just use Gen9 clothes because they are placeholders.

PR #3469, opened 2025-06-22
feat(mumble): add `GET/SET_PLAYER_VOICE_VOLUME_FROM_SERVER_ID`

poco8537:
### Goal of this PR

Allows adjusting the player's voice volume

This PR introduces two new natives, `GET/SET_PLAYER_VOICE_VOLUME_FROM_SERVER_ID`, which allow retrieving and setting a player's voice volume by their server ID. The voice output volume is adjusted according to the set value.

FiveM, RedM

### Successfully tested on

**Game builds:** 3095

**Platforms:** Windows

### Checklist


shkssj:
What’s the difference from `MumbleSetVolumeOverrideByServerId`?

poco8537:
> What’s the difference from `MumbleSetVolumeOverrideByServerId`?

I just learned about that native for the first time. According to its description, 'This will also bypass 3D audio and distance calculations.' But the native I implemented does not bypass them.

Yum1x:
> But the native I implemented does not bypass them.

It seems this key difference is not mentioned in the PR description. To help other developers, perhaps the documentation could be updated to include this detail. It might also be a good idea to reconsider the name to better reflect its purpose and prevent confusion with the existing [MumbleSetVolumeOverrideByServerId](https://docs.fivem.net/natives/?_0xCE8E25B4)

PR #3471, opened 2025-06-22
feat(game/five): save pure level

someone005:
### Goal of this PR
This PR saves players time when connecting to servers with `sv_pureLevel` > 0. Currently, joining such servers requires a game restart every time, and this change eliminates that need.

Save pure level to `CitizenFX.ini` to avoid game restart when joining pure servers

FiveM

### Successfully tested on

**Platforms:** Windows

### Checklist


code/client/launcher/CrossBuildLaunch.cpp
```diff
@@ -77,5 +77,13 @@ void XBR_EarlySelect()
 	{
 		wcscat(state->initCommandLine, va(L" -b%d", retainedBuild));
 	}
+
+	auto retainedPureLevel = GetPrivateProfileInt(L"Game", L"SavedPureLevel", 0, fpath.c_str());
```
prikolium-cfx, line 81: Just "PureLevel" should be better.

prikolium-cfx (changes requested):
Thank you for the contribution, but as additional challenge I want to suggest you to also implement logic to pass players with higher pure level on server. Currently pure level should be strictly equal to server pure level, but server do not lose anything if joining player has pure2 and server is on pure0.

someone005:
Gonna do when I get home 👍

Gogsi:
I'm a bit worried about how this may come off as confusing for players that switch between pure and non-pure servers. Imagine the following scenario:

- You install Reshade
- You join a pure mode server, the game restarts, `PureLevel=1/2` gets saved in the ini
- You quit the game and start it again, it loads in pure mode.
- You now join a non-pure mode (0) server but your mods don't work. Nevertheless, it saves `PureLevel=0` in the ini, because it reads it from the server configuration.
- You restart your game, it loads in pure level 0, and you join the second server again. Now suddenly your mods work.

Feels confusing that after joining a pure mode server, your first join on a non-pure mode server will have mods disabled for no reason, but subsequent joins will work fine.

someone005:
Hmmm, we can do something like this: when player's connecting to server without changing game build and user pure level >= server pure level we would ask him if he would like to change the pure mode. If no, he will continue with current one, if yes, game will restart. When player has to change game build, we will change the pure level automatically.

prikolium-cfx:
> Hmmm, we can do something like this: when player's connecting to server without changing game build and user pure level >= server pure level we would ask him if he would like to change the pure mode. If no, he will continue with current one, if yes, game will restart. When player has to change game build, we will change the pure level automatically.

I think it makes sense to ask players if they want to lower their puremode level with automatic skip after 5 seconds and load with puremode

someone005:
@prikolium-cfx so i code your suggestion up, but i occured an error. may it be related to changes i did in my code or it's sth else?
image

someone005:
@prikolium-cfx can you review and test latest commit?

alberttheprince:
This would be a major benefit to the community if we could get this in - big blocker for small/medium servers to get people to join is that players are constantly needing to restart to change pure mode level. 🙏

ook3D:
> This would be a major benefit to the community if we could get this in - big blocker for small/medium servers to get people to join is that players are constantly needing to restart to change pure mode level. 🙏

i dont really see how it matters the size of the community, majority of servers use pure mode its not exclusive to any size of servers

alberttheprince:
> > This would be a major benefit to the community if we could get this in - big blocker for small/medium servers to get people to join is that players are constantly needing to restart to change pure mode level. 🙏
> 
> i dont really see how it matters the size of the community, majority of servers use pure mode its not exclusive to any size of servers

users are less willing to restart every play session for a smaller pop-server, large servers use pure mode because user's are invested/play exclusively on that server.

It's the reason the majority of servers _don't_ use puremode.

prdatur:
any chance to get this reviewed? we can not explain to our players "you have always restart the game twice", players would just not play on our server

alberttheprince:
> any chance to get this reviewed? we can not explain to our players "you have always restart the game twice", players would just not play on our server

I wouldn't waste your time with pure mode atm, there's a way to circumvent pure mode, so it's not really useful.

dougcostadev:
I think the core issue here is that `pureLevel` is currently tightly coupled to the game boot lifecycle, which forces a restart even in scenarios where it is not strictly required.

Instead of only persisting `pureLevel` in `CitizenFX.ini`, a more robust approach would be to split the concept into two distinct states:

* `desiredPureLevel` (server-driven)
* `activePureLevel` (runtime-applied)

This would enable the introduction of a transition layer with clearer behavior:

* **Upgrade (player < server):** still requires a restart for security reasons
* **Downgrade (player > server):** can be handled at runtime

For downgrades, instead of forcing a restart, the client could:

* disable external graphics hooks (ReShade, ASI, etc.)
* revalidate loaded modules
* enforce a restricted runtime environment equivalent to the lower pure level

This approach avoids the UX issues mentioned earlier (such as inconsistent mod behavior) while still maintaining the necessary security guarantees.

Additionally, persisting `pureLevel` globally in `CitizenFX.ini` may introduce cross-server inconsistencies. A per-server cache (for example, keyed by endpoint or server ID) would likely provide a more predictable and consistent experience.

Overall, this would address:

* unnecessary restarts
* inconsistent mod states
* and avoid weakening the pure system

Gogsi:
> Downgrade (player > server): can be handled at runtime

The issue with that approach is that it will likely lead to other issues. Some files get mounted early on, before you get to the server connecting screen. Which means that if you then see the server is pure mode 0 and allow the modded RPFs to load, they will "load" but some files will have no effect on the game. This can end up resulting in issues with half-loaded graphics mods causing visual glitches and other such issues.

The currently implemented approach is probably required for this to work correctly 100% of the time:
- Upgrade (player pure level < server pure level): Requires a restart to "unload" any mods
- Downgrade (player pure level > server pure level): Prompt the player about restarting the game to load their mods. If they don't respond, continue without mods. Might be nice to have a list of files or something here?

Firecul:
> > Downgrade (player > server): can be handled at runtime
> 
> The issue with that approach is that it will likely lead to other issues. Some files get mounted early on, before you get to the server connecting screen. Which means that if you then see the server is pure mode 0 and allow the modded RPFs to load, they will "load" but some files will have no effect on the game. This can end up resulting in issues with half-loaded graphics mods causing visual glitches and other such issues.
> 
> The currently implemented approach is probably required for this to work correctly 100% of the time:
> 
>     * Upgrade (player pure level < server pure level): Requires a restart to "unload" any mods
> 
>     * Downgrade (player pure level > server pure level): Prompt the player about restarting the game to load their mods. If they don't respond, continue without mods. Might be nice to have a list of files or something here?

I definitely feel this would be the most seamless user side experience.

PR #3476, opened 2025-06-24
fix(citizen-server-impl): avoid crash on empty headers in writeHead

### LINKED ISSUE #3464: Server crashes after passing an empty table to SetHTTPHandler's res.writeHead
### What happened?

The title describes it perfectly, there's nothing much to it.

### Expected result

It shouldn't crash

### Reproduction steps

```lua
SetHttpHandler(function(req, res)
    res.writeHead(418, {});
    res.send("im ded gn");
end);
```

### Importancy

Crash

### Area(s)

ScRT: Lua

### Specific version(s)

Server b16085/Win

### Additional information

nope

DaniGP17:
### Goal of this PR
Prevent the server from crashing when res.writeHead is called with an empty header table.

By adding a type check to ensure that the second argument of the unpacked message is a valid msgpack::type::MAP before attempting to iterate or cast it.

Server

### Successfully tested on
**Platforms:** Windows

### Checklist

### Fixes issues
fixes #3464

### DIFF
--- a/code/components/citizen-server-impl/src/ResourceHttpHandler.cpp
+++ b/code/components/citizen-server-impl/src/ResourceHttpHandler.cpp
@@ -225,7 +225,7 @@ public:
 			{
 				auto state = unpacked.get().as<std::vector<msgpack::object>>();
 
-				if (state.empty())
+				if (state.empty() || state[0].is_nil())
 				{
 					response->End();
 				}


DaniGP17:
Fixed another crash when sending a nil response. Thanks to @packfile (lukxsy) for reporting
```lua
SetHttpHandler(function(req, res)
      res.send(nil)
end)
```

PR #3480, opened 2025-06-25
Scheduler.lua cleanup

Mathu-lmn:
This PR cleans the scheduler file of its unused variables (some were introduced 8 years ago and are no longer relevant), it also renames some variables to avoid confusion upon reading (see line 724/818 for examples).
Finally it also formats the file


iridium-cfx (approved):
This all looks correct, thanks!

PR #3482, opened 2025-06-25
feat(extra-natives-five): Add health config natives

grzybeek:
### Goal of this PR

Unfortunately `pedhealth.meta` cannot be streamed via DATA_FILE entries, so decided to add natives to manipulate those values. It allows to change ped model health config, add new health config, remove health config and modify values of existing health configs.

This will allow to set for example default health value that users spawn with on the server without need to use sethealth/setmaxhealth natives. Didn't test functionality of all fields, but for example dying threshold allows us to change default "100" value that our ped dies to other values.
...

Hooked directly to function that returns health config from array, if it exists in our map then return modified values instead.
...

Fivem natives

### Successfully tested on

**Game builds:**  3258

**Platforms:** Windows

### Checklist

### DIFF
--- a/code/components/extra-natives-five/src/PedExtraNatives.cpp
+++ b/code/components/extra-natives-five/src/PedExtraNatives.cpp
@@ -13,6 +13,8 @@
 #include <GameInit.h>
 #include <fxScripting.h>
 #include <Resource.h>
+#include <Hooking.Stubs.h>
+#include <unordered_set>
 
 class CPedHeadBlendData
 {
@@ -58,6 +60,9 @@ public:
 static uint64_t* _id_CPedHeadBlendData;
 static uintptr_t _baseClipsetLocation;
 static uint32_t _pedSweatOffset;
+static uint8_t _pedModelInfoOffset;
+static uint32_t _pedPersonalityIndexOffset;
+static uint8_t _pedHealthConfigHashOffset;
 static float* _motionAimingTurnTransitionThresholdMin = nullptr;
 static float* _motionAimingTurnTransitionThresholdMax = nullptr;
 
@@ -83,8 +88,10 @@ static hook::cdecl_stub<bool(void*, int, int)> _doesPedComponentDrawableExist([]
 
 struct PedPersonality
 {
-	uint32_t hash;
-	char pad[180];
+	uint32_t hash; // 0x00
+	char padding[20];
+	uint32_t healthConfigHash; // 0x18
+	char paddingAfter[156];
 };
 
 static atArray<PedPersonality>* g_pedPersonalities;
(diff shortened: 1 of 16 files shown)


code/components/extra-natives-five/src/PedExtraNatives.cpp
```diff
@@ -130,12 +136,92 @@ static void SetWetClothPinRadiusScale(float scale)
...
+	HashString("Fish"),
+	HashString("Rat"),
+	HashString("Shark"),
+	HashString("Humpback"),
+};
+
+using GetHealthInfoFn = const CHealthConfigInfo*(__fastcall*)(uint32_t);
+static GetHealthInfoFn g_origGetHealthInfo = nullptr;
+
+static const CHealthConfigInfo* __fastcall GetHealthInfo(uint32_t pedPtr)
```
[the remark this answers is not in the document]
prikolium-cfx: __fastcall is not needed as well because it's default calling convention on win64

code/components/extra-natives-five/src/PedExtraNatives.cpp
```diff
@@ -130,12 +136,92 @@ static void SetWetClothPinRadiusScale(float scale)
...
+	float dyingHealthThreshold;
+	float hurtHealthThreshold;
+	float dogTakedownThreshold;
+	float writheFromBulletDamageThreshold;
+	bool meleeCardinalFatalAttackCheck;
+	bool invincible;
+};
+
+
+static std::map<uint32_t, CHealthConfigInfo> g_healthConfigs;
```
prikolium-cfx, line 157: We don't care about order here, so we can use `unordered_map` and `unordered_set`

prikolium-cfx (changes requested):

ext/native-decls/GetPedModelHealthConfig.md
```diff
@@ -0,0 +1,21 @@
+---
+ns: CFX
+apiset: client
+game: gta5
+---
+## GET_PED_MODEL_HEALTH_CONFIG
+
+```c
+void GET_PED_MODEL_HEALTH_CONFIG(Hash modelHash);
```
ivanzaida, line 9: according to `context.SetResult<uint32_t>(healthHash);`, the result inst void

code/components/extra-natives-five/src/PedExtraNatives.cpp
```diff
@@ -305,6 +392,27 @@ static HookFunction initFunction([]()
 			}
 		}
 
+		for (auto& [pedModel, savedHealthConfig] : undoHealthConfigs)
+		{
+			auto archetype = GetPedModel(pedModel);
+
+			if (archetype)
+			{
+				auto personalityIndex = *(uint16_t*)((char*)archetype + _pedPersonalityIndexOffset);
```
prikolium-cfx, line 401: And here

PR #3409, opened 2025-05-19
Fix PR #2546 and Add Delegate Serialization Support

manups4e:
### Goal of this PR
Re-enable (De)Serialization using the new custom msgpack provided in https://github.com/citizenfx/fivem/pull/2546

### How is this PR achieving the goal
This PR adds back all Thorium's modifications and some extras to be fully compatible with the new MsgPack (de)serialization.

## Warning
- https://github.com/citizenfx/msgpack-cs/pull/2 needs to be merged before applying this PR for it to be working on production.

### Successfully tested on
FXServer and FiveM client

**Game builds:** N/A

**Platforms:** Windows

### Checklist

- [x] Code compiles and has been tested successfully.
- [x] Code explains itself well and/or is documented.
- [x] My commit message explains what the changes do and what they are for.
- [x] No extra compilation warnings are added by these changes.

### Fixes issues
`fixes #2653`, `fixes #2613`, `fixes #2590`, `fixes #2582`, `fixes #2571`,

PR #3483, opened 2025-06-26
tweak(exports): wrap non-table returns in table

BuddiesTV:
### Goal of this PR
If a async C# export (e.g., `async Coroutine<string>) returns a non-table value via __cfx_async_retval, the scheduler would return nil when trying to table.unpack the result. This change wraps the result in a table if it isn't one already, ensuring consistent handling and preventing unexpected behaviors.

This would then allow to have async C# exports that return a value without allocating a new array.

Check if the result is a table, if not wrap it in one.

ScRT: Lua

### Successfully tested on
**Platforms:** Windows, Linux

### Checklist

### DIFF
--- a/data/shared/citizen/scripting/lua/scheduler.lua
+++ b/data/shared/citizen/scripting/lua/scheduler.lua
@@ -599,7 +599,7 @@ funcref_mt = msgpack.extend({
 
 			rvs[1].__cfx_async_retval(function(r, e)
 				if r then
-					p:resolve(r)
+                			p:resolve(type(r) == "table" and r or { r })
 				elseif e then
 					p:reject(e)
 				end


manups4e:
in my monoV2 PR i had the same bug and i solved it by doing:

```lua
p:resolve(table.pack(r))
```

- `type(r) == "table" and r or { r }` is simpler but does not preserve nil values and lacks the count n, which can cause bugs when nil is present.

check https://github.com/citizenfx/fivem/pull/3409/commits/0af1de56d6e1c00b34f5d0b3875907df38e0ce43

Edit: in my pr if you sent a whole class with `Indexed` layout, your solution would return only the first parameter of the resulting array, while my approach somehow returns the whole array

- Before (your PR)
image

- After (my PR)
image

when sending a struct with indexed lyout

```c#
	[MsgPackSerializable(Layout.Indexed)]
	public struct TestClass
	{
	 	[Index(0)] public int Test1 { get; set; }
		[Index(1)] public string Test2 { get; set; } = "";
		[Index(2)] public myEnum Test3 { get; set; } = myEnum.One | myEnum.Two;

		public TestClass() { }
		public TestClass(int test1, string test2, myEnum @enum)
		{
			Test1 = test1;
			Test2 = test2;
			Test3 = @enum;
		}
		public TestClass(int test1)
		{
			Test1 = test1;
			Test2 = "ciaooo";
			Test3 = myEnum.One | myEnum.Two;
		}

		public override string ToString()
		{
			return $"TestClass: {Test1} - {Test2} - {Test3}";
		}
	}
```

BuddiesTV:
Resolved as manu included this in his mono pr.
https://github.com/citizenfx/fivem/pull/3409 / https://github.com/citizenfx/fivem/commit/0af1de56d6e1c00b34f5d0b3875907df38e0ce43

PR #3488, opened 2025-06-26
feat(server/gamestate): add a argument to onPlayerDropped

spacevx:
Add an argument to onPlayerDropped to differentiate when the player really disconnected from the server or just left the onesync range

In my case it's useful, because I want to write a 3D text when someone is disconnecting from the server, but it bothers me to loop through each player server-side and find nearest players to send them a client-side event.

This code wasn't tested yet. I need to build FiveM Client and test

```lua
AddEventHandler('onPlayerDropped', function(playerId, playerName, slotId, isServerDisconnect)
    if (isServerDisconnect) then
        print(playerName .. " dropped from the server")
    else
        print(playerName .. " left your onesync range")
    end
end)
```

### DIFF
--- a/code/components/citizen-server-impl/src/GameServer.cpp
+++ b/code/components/citizen-server-impl/src/GameServer.cpp
@@ -1171,7 +1171,7 @@ namespace fx
 			if (!fx::IsBigMode() && isFinal)
 			{
 				// send every player information about the dropping client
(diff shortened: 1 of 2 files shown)


csskroubledev:
https://docs.fivem.net/docs/scripting-reference/events/list/playerdropped/

spacevx:
We don't talk about the same event....

I'm talking about onPlayerDropped (who is a client side event)
not playerDropped who is a server side event

PR #3148, opened 2025-02-11
feat(extra-natives): add REGISTER_RAW_KEYMAP and REMAP_RAW_KEY_MAP

AvarianKnight:
### Goal of this PR
This is a continuation of https://github.com/citizenfx/fivem/pull/3147

This adds a way for users to have listeners for raw keys with `REGISTER_RAW_KEYMAP` and allow the user to rebind them with `REMAP_RAW_KEYMAP`, this mainly an alternative for [REGISTER_KEY_MAPPING](https://docs.fivem.net/natives/?_0xD7664FD1) until its implemented for RedM.

### This PR applies to the following area(s)
FiveM (not tested), RedM

### Successfully tested on

**Game builds:** b1491

**Platforms:** Windows, Linux

### Checklist

- [x] Code compiles and has been tested successfully.
- [x] Code explains itself well and/or is documented.
- [x] My commit message explains what the changes do and what they are for.
- [x] No extra compilation warnings are added by these changes.

PR #3489, opened 2025-06-26
feat(extra-natives-five): make REGISTER_RAW_KEYMAP and REMAP_RAW_KEYMAP available for FiveM.

CENSOR37:
### Goal of this PR

This is to remove RDR guards from raw keymap natives in this implementation https://github.com/citizenfx/fivem/pull/3148 a huge thanks to @AvarianKnight.

`REGISTER_KEY_MAPPING` serves a different purpose than raw keymaps. Keybindings registered this way are user-editable in the settings menu, which can lead to user errors (e.g., accidentally unbinding or changing keys). It is not so easy to reset it back to the default. Furthermore, when multiple resources (e.g., RESOURCE_A, RESOURCE_B, RESOURCE_C) use the same key (such as "E" for interaction), managing these bindings can become inconsistent and difficult for users.

Raw keymaps provide a more straightforward and manageable approach for these scenarios, allowing resources to define and use bindings without conflicting with user-managed settings.

FiveM

### Successfully tested on

**Game builds:** 3258

**Platforms:** Windows

### Checklist


outsider31000:
`REGISTER_RAW_KEYMAP` does not work the same for every client, it all depends on your keyboard, os and language.
so you will face bigger problems than allowing them to change keybinds with `REGISTER_KEY_MAPPING`

`REGISTER_KEY_MAPPING` is better, always. it was only added for RedM because of it RedM doesnt support it yet.

CENSOR37:
Does it also apply to IS_RAW_KEY_PRESSED, IS_RAW_KEY_RELEASED?

I have multiple resources on the server that use this native to detect key events with ~1000 players online at peak time, and haven't received any reports. If so, does this mean the raw key isn't meant to be used in FiveM at all?

outsider31000:
I'm only telling you that you will probably face issues depending on the cases I mentioned.
and yes it applies to those natives they are for virtual raw keys, you might not have issues but this is for other servers to and they might not be aware just like you weren't.
it's prone to fail and no one will know why, while `REGISTER_KEY_MAPPING` won't fail cause they can even change if it does.

it's not up to me, i'm just saying.

CENSOR37:
No worries, I also want to make sure this doesn't cause any problems later on. Thanks for the input.

prikolium-cfx:
Raw keymappings limited to RedM only because I've asked AvarianKnight to do it. I don't plan to change my opinion and add more confusing APIs to FiveM.

PR #3492, opened 2025-06-28
chore: name hardcoded natives

PulseHabour:
### Goal of this PR

Replaces hardcoded native id with their name (includes community names).

Replaces hardcoded native ids to a NativeIdentifiers enum.
Removes native name comments.

FiveM, RedM, FXServer

### Successfully tested on

**Platforms:** Windows

### Checklist


PulseHabour:
If possible it would be beneficial instead of community names, if we could use the official names, It would be for:
15 GTAV natives
1 RDR3 natives

Also since slot of enums are duplicates, possible for a shared enum file, or maybe hook into other definitions for natives

PR #3502, opened 2025-06-30
fix(scripting): throw error for function references

### LINKED ISSUE #3501: `pcall` does not catch errors for function references
### What happened?

When using `pcall` with function references (e.g. functions passed to exports), it will return `false` as the first "success" value, but the error would still be printed out in the console, and would not be returned as the second "error" return value.

### Expected result

`pcall` should have caught the error and returned it as the second return value.

### Reproduction steps

1. Set up an export that accepts a function and executes it at some point using `pcall`. For example:
   ```lua
   exports('test', function(func)
       local success, error = pcall(func)
       print('success', success)
       print('error', error)
   end)
   ```
2. Use the export with a function that throws an error. For example:
   ```lua
   exports.myresource:test(function()
       error('hello world')
   end)
   ```
3. See unexpected error in console, with the `success` value being `false`, however the `error` value being `nil`:
   ```
   [         script:test] SCRIPT ERROR: @test/server.lua:10: hello world
   [         script:test] > ref (@test/server.lua:10)
   [         script:test] > ref (@myresource/shared/tests.lua:26)
   [         script:test] > fn (@test/server.lua:9)
   [   script:myresource] success  false
   [   script:myresource] error    nil
   ```

### Importancy

Slight inconvenience

### Area(s)

ScRT: Lua

### Specific version(s)

FXServer 14033 on Windows

### Additional information

_No response_

D4isDAVID:
### Goal of this PR

Return the result error when function references throw them.
This both fixes `pcall` not being able to catch and prevent errors from being printed, but also improves the error message(s) when the function reference isn't wrapped in a `pcall`.

#### Before
With pcall:
image
Without pcall:
image

#### After
With pcall:
image
Without pcall:
image

Error handling for function references is done in the scheduler, so when an error is detected, simply throw it rather than printing it and throwing `nil`.

ScRT: Lua

### Successfully tested on

**Platforms:** Windows, Linux

### Checklist

### Fixes issues

Fixes #3501

### DIFF
--- a/code/client/clrcore-v2/Debug.cs
+++ b/code/client/clrcore-v2/Debug.cs
@@ -61,9 +61,13 @@ namespace CitizenFX.Core
 
 		internal static void PrintError(Exception what, string where = null)
 		{
+			WriteLine(CreateError(what, where));
+		}
+
+		internal static string CreateError(Exception what, string where = null)
+		{
 			where = where != null ? " in " + where : "";
-			WriteLine($"^1SCRIPT ERROR{where}: {what.GetType().FullName}: {what.Message}^7");
-			WriteLine(what.StackTrace.ToString());
+			return $"^1SCRIPT ERROR{where}: {what.GetType().FullName}: {what.Message}^7\n" + what.StackTrace.ToString();
 		}
 
 		/*[SecuritySafeCritical]
--- a/code/client/clrcore-v2/Interop/LocalFunction.cs
+++ b/code/client/clrcore-v2/Interop/LocalFunction.cs
@@ -44,8 +44,21 @@ namespace CitizenFX.Core
 		[SecuritySafeCritical]
 		internal unsafe Coroutine<object> Invoke(object[] args)
 		{
-			object[] returnData = MsgPackDeserializer.DeserializeArray(ScriptInterface.InvokeFunctionReference(m_reference, new InPacket(args).value));
-			if (returnData != null && returnData.Length > 0)
+			object[] values = MsgPackDeserializer.DeserializeArray(ScriptInterface.InvokeFunctionReference(m_reference, new InPacket(args).value));
+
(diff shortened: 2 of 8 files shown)


D4isDAVID:
Might look into the other ScRTs as the same issue might be there as well. I'll probably open separate PRs for those in that case

iridium-cfx:
Will this work correctly when interacting with other languages (JS/C#)?
I think you would also need to update the other languages to support returning the error data in the same way.

D4isDAVID:
Yeah, this PR currently only fixes exports created with Lua. If other languages use Lua exports it will be fine, but the other languages still have broken exports. I'm on/off on making a fix for the other languages because the underlying code for those seems a bit weirder than the Lua scheduler.

iridium-cfx:
> Yeah, this PR currently only fixes exports created with Lua. If other languages use Lua exports it will be fine, but the other languages still have broken exports. I'm on/off on making a fix for the other languages because the underlying code for those seems a bit weirder than the Lua scheduler.

Are you sure they will work correctly? Because other runtimes will still be expecting the msgpack'd result to only contain the result, not the bool/value pair.

D4isDAVID:
Ah, I get what you're saying. I'll check later today, but now that I think about it it probably won't work without the fix being in the other langs.

D4isDAVID:
Ok, I think I figured out the fix for JS, next up is C# and then I'll update the PR

D4isDAVID:
@iridium-cfx PR has been updated with a fix for the rest of the languages.

Scorpion7162:
This will help with debugging a lot, big upvote for this 👊

D4isDAVID:
Oops....

D4isDAVID:
Converted to draft as this currently only works when the top-level code isn't in a function reference.
Needs more work as it won't show errors in certain contexts such as when running commands.

PR #3503, opened 2025-07-01
feat(scrt): disable mono v2 date restrictions

csskroubledev:
### Goal of this PR
Unlock mono v2 fully without date restrictions, as currently all resources stopped working because of the check.

...

Date restrictions were removed, old values kept available to use with one new added `"true"` to prevent confusions.

...

ScRT: C# (Mono V2)

...

### Successfully tested on

**Game builds:** ..

**Platforms:** Windows

### Checklist

### Fixes issues
-

### DIFF
--- a/code/components/citizen-scripting-mono-v2/src/MonoScriptRuntime.cpp
+++ b/code/components/citizen-scripting-mono-v2/src/MonoScriptRuntime.cpp
@@ -263,12 +263,10 @@ int MonoScriptRuntime::HandlesFile(char* filename, IScriptHostWithResourceData* 
 		return false;
 	}
 
-	// last supported date for this pilot of mono_rt2, in UTC
-	constexpr int maxYear = 2025, maxMonth = 6, maxDay = 30;
-
 	// Allowed values for mono_rt2
 	constexpr std::string_view allowedValues[] = {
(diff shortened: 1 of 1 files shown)


prikolium-cfx:
We are working on ways to transition from mono to a more modern runtime and that's why Mono v2 will not out of experimental.
It's not recommended to use Mono v2 runtime on production servers.

csskroubledev:
> We are working on ways to transition from mono to a more modern runtime and that's why Mono v2 will not out of experimental. It's not recommended to use Mono v2 runtime on production servers.

As a meantime to the new runtime not appearing, could we atleast postpone the expiration or change the required to text to something in orders of: "I acknowledge that this runtime is not production-ready and may not be stable."

prikolium-cfx:
> > We are working on ways to transition from mono to a more modern runtime and that's why Mono v2 will not out of experimental. It's not recommended to use Mono v2 runtime on production servers.
> 
> As a meantime to the new runtime not appearing, could we atleast postpone the expiration or change the required to text to something in orders of: "I acknowledge that this runtime is not production-ready and may not be stable."

I already extended experimental to 2025-12-31. That should be ok for now.

PR #3506, opened 2025-07-01
feat(extra-natives-five): Add natives to modify fall damage

grzybeek:
### Goal of this PR

Add getters and setters to modify fall damage. There is value for fatal height from which peds will always die and multiplier of damage if ped is falling from lower, both need to be modifiable.

I had to add overload method for SetLocation in GameValueStub, the original method assumed the relative offset was always at the instruction start and did not account for instruction size, causing crashes in some cases.

Other than that its just basic hooking to global float values, had to use GameValueStub because complier optimization makes that they were used in other places as well.

Fivem client natives

### Successfully tested on

**Game builds:** ..  3258 tested, 2699 checked if patterns exist

**Platforms:** Windows

### Checklist

### DIFF
--- a/code/components/extra-natives-five/include/GameValueStub.h
+++ b/code/components/extra-natives-five/include/GameValueStub.h
@@ -40,4 +40,13 @@ struct GameValueStub
 	{
 		hook::put<int32_t>(location, (intptr_t)m_address - (intptr_t)location - offset);
 	}
+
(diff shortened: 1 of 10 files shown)


Paneedah:
Will be awesome once this is merged. Really need it for a system I'm working on.

prikolium-cfx (approved):

PR #3523, opened 2025-07-09
fix(net-event-doc): fix typedoc not recognizing int

Mathu-lmn:
Main code changes are just renaming the comments event parameter types, but line endings weren't consistent so this fixes it as well ig 
image


Mathu-lmn:
This fixes event-doc-gen throwing errors :
image

PR #3524, opened 2025-07-09
fix(net-event-doc): fix typedoc not recognizing int

Mathu-lmn:
This PR fixes the errors when trying to generate the event docs : 
image


prikolium-cfx (approved):

prikolium-cfx:
Thank you for the PR

PR #3529, opened 2025-07-10
feat(citizen-server-impl): add state bag name in console log drops

### LINKED ISSUE #3528: The most retarded error ever about statebag overflow
### What happened?

[ sbag-update-dropped] Client xxxx sent too many state bag updates and had their updates dropped.
[ sbag-update-dropped] If you believe this to be a mistake please increase your rateLimiter_stateBag_rate and rateLimiter_stateBag_burst. You can do this with `set rateLimiter_stateBag_rate [new value]`. The default rate limit is 75 and burst limit is 125
[ sbag-update-dropped] You can disable this warning with `con_addChannelFilter sbag-update-dropped drop` if you think you have this properly set up.

Who thought of this bullshit of an error message?
It doesnt have any information about the issue WHATSOEVER, nor does it give debugging information.
It just says to increase the max limit. I wanna know which statebags caused this.

### Expected result

Some fucking info to work with

### Reproduction steps

Enjoy getting an error with no information whatsoever.

### Importancy

Slight inconvenience

### Area(s)

FiveM

### Specific version(s)

FiveM

### Additional information

_No response_

DaniGP17:
### Goal of this PR
Added the state bag name to the logs when a player is rate limited or kicked due to state bag overflow.

Extracting the state bag name during early processing of StateBag and StateBagV2 packets.

Server

### Successfully tested on
**I only tested this in StateBagV2**
**Platforms:** Windows, Linux

### Checklist

### Fixes issues
fixes #3528 

I would like someone to give their opinion on the implementation of this for `StateBag`"v1" because I think I am doing it wrong in that case.

### DIFF
--- a/code/components/citizen-server-impl/src/packethandlers/StateBagPacketHandler.cpp
+++ b/code/components/citizen-server-impl/src/packethandlers/StateBagPacketHandler.cpp
@@ -71,15 +71,15 @@ void StateBagPacketHandler::HandleStateBagMessage(fx::ServerInstanceBase* instan
 
 	uint32_t slotId = client->GetSlotId();
 
-	std::string bagNameOnFailure;
-	stateBagComponent->HandlePacket(slotId, clientStateBag.data, &bagNameOnFailure);
-	std::string safeBagName = bagNameOnFailure.empty() ? "unknown" : bagNameOnFailure;
-
 	const bool hitRateLimit = !stateBagRateLimiter->Consume(netId);
 	const bool hitFloodRateLimit = !stateBagRateFloodLimiter->Consume(netId);
 
 	if (hitRateLimit)
 	{
+		std::string tempBagName;
+		stateBagComponent->HandlePacket(slotId, clientStateBag.data, &tempBagName);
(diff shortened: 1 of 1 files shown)


Gogsi:
Would it make more sense to keep the "happy path" HandlePacket after the rate limit, but do it once more the first time the rate limit is hit to get the statebag name? Otherwise malicious clients can spam state bag updates and they'll still be handled, as long as they don't exceed the flood limit

DaniGP17:
Good catch, should be good now

prikolium-cfx:
Thank you for the contribution, but I don't think that this information can be really useful as it doesn't detect the real issue and show only dropped statebag. I think something like statebags stats debug tool would be much more useful. For example we can show there that 700 statebags with name "petHealth" was sent in last rate limiting period and so.

DaniGP17:
Yeah, I though some time ago about creating a debug window in imgui, not sure if that's what you are talking about.

prikolium-cfx:
> Yeah, I though some time ago about creating a debug window in imgui, not sure if that's what you are talking about.

Yes, that's what I mean

PR #3533, opened 2025-07-15
tweak(server-impl): optimize the path for statebags

AvarianKnight:
### Goal of this PR
Reduce contention around accessing state bags.

Removes the `shared_lock`, which in this case isn't really needed, as this should only ever get written to *once* and then only read afterwards.

This should all be safe since the initial creation of a state bag by the ScRT is always guarded with a `HasStateBag` check.

Server

### Successfully tested on

**Game builds:** .. 

**Platforms:** Windows

### Checklist

### DIFF
--- a/code/components/citizen-server-impl/src/state/ServerGameState.cpp
+++ b/code/components/citizen-server-impl/src/state/ServerGameState.cpp
@@ -2735,9 +2735,10 @@ bool ServerGameState::SetEntityStateBag(uint8_t playerId, uint16_t objectId, std
 {
 	if (auto entity = GetEntity(0, objectId))
 	{
-		if (entity->GetStateBag())
+		if (entity->HasStateBag())
 		{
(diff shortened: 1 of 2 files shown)


code/components/citizen-server-impl/src/state/ServerGameState.cpp
```diff
@@ -2735,10 +2735,11 @@ bool ServerGameState::SetEntityStateBag(uint8_t playerId, uint16_t objectId, std
 {
 	if (auto entity = GetEntity(0, objectId))
 	{
-		if (entity->GetStateBag())
+		if (entity->HasStateBag())
```
FabianTerhorst, line 2738: Maybe add a method that uses compare_exchange internally to set the boolean to true if the expected state is false. This way the atomic is using just one operation and it seems to fit both into this code and into the ENSURE_ENTITY_STATE_BAG logic as well.

PR #3534, opened 2025-07-15
Fix incorrect parameter order for DrawCorona

cm8263:
### Goal of this PR
Fix incorrect parameter order in docs for DrawCorona native.

Corrects the documentation.

Documention

### Successfully tested on
N/A
### Checklist

- [ N/A ] Code compiles and has been tested successfully.
- [ Y ] Code explains itself well and/or is documented.
- [ Y ] My commit message explains what the changes do and what they are for.
- [ N/A ] No extra compilation warnings are added by these changes.

### Fixes issues
Docs listed order as RGB then Alpha, when it is actually Alpha then RGB.


cm8263:
Not directly relevant, but also worth noting: there is a `flags` param for this native, and the docs contain an `eCoronaFlags` codeblock listing what the flags are; however, on the docs website, this `eCoronaFlags` codeblock is absent.

My best guess would be because the codeblock defines the language as `cpp`, and the docs website filters by Lua, C#, and JS, thus excluding CPP?

utkuali:
@Gogsi sorry, haven't seen it.

PR #3536, opened 2025-07-15
fix(client): fix incorrect parameters for `GetVehicleXenonLightsCustomColor`

koketoo:
### Goal of this PR

Fix `GetVehicleXenonLightsCustomColor` parameters.

...

Removing the red, green and blue parameters
...

Naitves
...

### Successfully tested on

**Game builds:** 
3407

**Platforms:** 
Windows

### Checklist


AvarianKnight:
These parameters can't be removed, they are *references* just because the runtime you're using doesn't treat them like it doesn't mean they aren't.

PR #3477, opened 2025-06-24
feat(net/rdr3): physical player extension

Ehbw:
### Goal of this PR

Introduce support for RedM to support up 128 physical players inside a players scope. From the RAGE limit of 32, this PR is still very experimental and still needs extra work done, but is in a state where it can be tested on a feature branch with 32+ clients.

### How is this PR achieving the goal

(still work in progress, will provide detailed explanation when finished)

### This PR applies to the following area(s)

RedM

### Successfully tested on

**Game builds:**  1491.50

**Platforms:** Windows, Linux

### Checklist

- [ ] Code compiles and has been tested successfully.
- [x] Code explains itself well and/or is documented.
- [ ] My commit message explains what the changes do and what they are for.
- [x] No extra compilation warnings are added by these changes.

### Fixes issues
closes #2932

PR #3542, opened 2025-07-18
feat(core/rdr3): support for increased ped pool

Ehbw:
### Goal of this PR

Allow for pool-state-sizes changes to Ped related pools to work in RedM. Currently Ped pools are partially hardcoded in game to 150/160 and 110 for networked horses, ped and animals.

Patching a check for "Peds" pool size, that causes the game to exit early if it exceeds 160. Patch several values related to max peds to account for pool-sizes-state increase requests for "CNetObjPed"

RedM

### Successfully tested on

**Game builds:** 1491

**Platforms:** Windows

### Checklist


Nobelium-cfx:
Thank you for the contribution! Added `CNetObjPedBase` to `increase_pool_size` allow list. It is 110 for now (so 260/270/220 in total), but we can increase it further if needed.

Ehbw:
I've only tested with an increase of 50, but I can test again with 110 just to ensure there's no issues that arise before this ends up on canary.

Ehbw:
Setting CNetObjPedBase to 110 will lead to allocation failure when attempting to allocate CNetObjPedBase, CNetBlenderPed and CPedSyncData, I'll take my allocation patch (Increases network heap from 26mb to 35mb for now should be more then enough for this patch and future net pool increases) from #3477 and add it here and it

Nobelium-cfx:
> Setting CNetObjPedBase to 110 will lead to allocation failure when attempting to allocate CNetObjPedBase, CNetBlenderPed and CPedSyncData, I'll take my allocation patch (Increases network heap from 26mb to 35mb for now should be more then enough for this patch and future net pool increases) from #3477 and add it here and it

If 110 is troublesome to make work - we can always just limit it back to 50. There is no particular reason for it to be 110. Just at some point people were asking for pretty large increases to the pool, so I decided to set limit higher.

But if you think its easy to support with the patch - lets do that.

Ehbw:
> > Setting CNetObjPedBase to 110 will lead to allocation failure when attempting to allocate CNetObjPedBase, CNetBlenderPed and CPedSyncData, I'll take my allocation patch (Increases network heap from 26mb to 35mb for now should be more then enough for this patch and future net pool increases) from #3477 and add it here and it
> 
> If 110 is troublesome to make work - we can always just limit it back to 50. There is no particular reason for it to be 110. Just at some point people were asking for pretty large increases to the pool, so I decided to set limit higher.
> 
> But if you think its easy to support with the patch - lets do that.

I've pushed another commit which resolves the issue and allows for CNetObjPedBase to achieve 110 increase and should give more room for other potential net pool expansions if they are needed.

Nobelium-cfx (approved):

outsider31000:
yeah with these changes we can keep it at 110 increase, shouldnt need any more than this, at least for now. well see in the future.
thanks for the review.

PR #3547, opened 2025-07-21
fix(core/rdr3): crash when targetting invalid ped/player

Ehbw:
### Goal of this PR

Mitigates crash related to nullptr rage::netObject deref from weaponDamageEvent. When a server cannot allocate additional network Ids (e.g. hit the current 8000 network ID limit) or local players cannot register this entity due to network pools being full.

Check if the target is a nullptr at this point in weaponDamageEvent (which should never happen) and return 0 (fail) instead of continuing where it would crash 

RedM

### Successfully tested on

**Game builds:**  1491

**Platforms:** Windows, Linux

### Checklist

### Fixes issues

RDR2_b1491.exe!sub_1423F081C (0x386)

### DIFF
--- a/code/components/gta-core-rdr3/src/PatchWeaponTargetCrash.cpp
+++ b/code/components/gta-core-rdr3/src/PatchWeaponTargetCrash.cpp
@@ -1,0 +1,54 @@
+#include <StdInc.h>
+
+#include <Hooking.h>
+
+//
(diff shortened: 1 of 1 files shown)


outsider31000:
image
 some metrics for consideration. its not tied to all servers btw, its more than this

PR #3548, opened 2025-07-21
fix(gta-core-five): validate river entity archetype

DaniGP17:
### Goal of this PR
Fix a crash caused by river entities with null archetype.

Validating the archetype that is located in 0x20 from the river entity(CEntity).

FiveM

### Successfully tested on
**Game builds:** 3258
**Platforms:** Windows

### Checklist

### Fixes issues
Crash reported by @ook3D

### DIFF
--- a/code/components/gta-core-five/src/CrashFixes.RiverEntities.cpp
+++ b/code/components/gta-core-five/src/CrashFixes.RiverEntities.cpp
@@ -2,15 +2,15 @@
 #include <Hooking.h>
 #include <Hooking.Stubs.h>
 
-static void* (*g_orig_River__GetRiverEntity)(int RiverEntity);
+static hook::FlexStruct* (*g_orig_River__GetRiverEntity)(int RiverEntity);
 
 static void (*g_orig_GetBoundingBoxZForRiverEntity)(int RiverEntityIndex, float* MinZ, float* MaxZ);
 static void GetBoundingBoxZForRiverEntity(int RiverEntityIndex, float* MinZ, float* MaxZ)
 {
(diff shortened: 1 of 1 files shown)


code/components/gta-core-five/src/CrashFixes.RiverEntities.cpp
```diff
@@ -8,7 +8,7 @@ static void (*g_orig_GetBoundingBoxZForRiverEntity)(int RiverEntityIndex, float*
 static void GetBoundingBoxZForRiverEntity(int RiverEntityIndex, float* MinZ, float* MaxZ)
 {
 	void* entity = g_orig_River__GetRiverEntity(RiverEntityIndex);
-	if (!entity)
+	if (!entity || !*(void**)((uintptr_t)entity + 0x20))
```
colistro123, line 11: Haven't fully looked at the file. but maybe for readability purposes cast as an `fwEntity*` and then get the archetype?

i.e.
```c
static fwEntity* (*g_orig_River__GetRiverEntity)(int RiverEntity);
```

In code:

```c
fwEntity* entity = g_orig_River__GetRiverEntity(RiverEntityIndex);
if (!entity || !entity->GetArchetype()) return;
```

Or might just add a comment stating said offset is the archetype 😊

DaniGP17: Thanks for the review, I've added a comment indicating that it is the archetype, because `fwEntity` was defined in `gta-streaming-five`.

prikolium-cfx (approved):

PR #3551, opened 2025-07-23
tweak(server-impl/sbag): reduce the amount of locking for adding routing targets

AvarianKnight:
This currently uses a decently amount of the chunk of time in ETW traces because this gets called per frame, this adds another bool to the SyncData and makes sure we only add the routing target once instead of constantly spamming the write lock.

### Goal of this PR
Reduce the locking done for routing targets

Only adding routing targets once.

Server

### Successfully tested on
**Game builds:** .. 

**Platforms:** Windows, Linux

### Checklist


AvarianKnight:
Before:

image

After:

image

FabianTerhorst (approved):

PR #3565, opened 2025-08-01
fix(gta-core-five): overriding ladder registers causing bad collision checks

DaniGP17:
### Goal of this PR
Fix a issue with ladders that detect that some physical object is colliding in the test made by ladder logic, this only seems to happen in a custom map of NoPixel but very sure that can happen in ladders with same environment(cover above the end of ladder)
Clip: https://streamable.com/52f0it

I've checked and I was overwriting some registers that was used after, so here I'm restoring original rax value and r11 that is only used at the end for writing.

FiveM

### Successfully tested on
Can't test cause is in a custom map, can I get a feature branch to ask people with this issue to test if its working?
**Game builds:** .. 
**Platforms:** Windows

### Checklist

### Fixes issues
Not public issue but reported by @d22tny

### DIFF
--- a/code/components/gta-core-five/src/CrashFixes.Ladder.cpp
+++ b/code/components/gta-core-five/src/CrashFixes.Ladder.cpp
@@ -37,12 +37,21 @@ static HookFunction hookFunction([]()
 		
 		virtual void InternalMain() override
 		{
+			// Restore original rax value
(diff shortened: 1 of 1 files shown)


prikolium-cfx (approved):

d22tny:
tested on the specific branch an it works as expected! thanks.

PR #3566, opened 2025-08-01
feat(citizen-server-impl): add regex filters for console listener

utopi160:
### Goal of this PR
Add the ability to apply regex filters (channel and message) on the native `REGISTER_CONSOLE_LISTENER`. 

By retrieving filters (optional) and searching for a match when a new log arrives.

FiveM

### Successfully tested on

**Game builds:** .. 
**Platforms:** Windows, Linux

### Checklist

### Fixes issues

### DIFF
--- a/code/components/citizen-server-impl/src/ServerBufferReplay.cpp
+++ b/code/components/citizen-server-impl/src/ServerBufferReplay.cpp
@@ -14,37 +14,51 @@
 
 #include <sstream>
 
+#include <regex>
+
 static std::shared_mutex g_consoleBufferMutex;
 static boost::circular_buffer<std::string> g_consoleBuffer(1500);
 
+struct ConsoleListenerEntry
+{
+	std::string handlerRef;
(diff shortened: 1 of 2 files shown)


DaniGP17:
I think this is going to break existing scripts that use that function.

utopi160:
> I think this is going to break existing scripts that use that function.

Why is that? Arguments are optional.

DaniGP17:
> > I think this is going to break existing scripts that use that function.
> 
> 
> 
> Why is that? Arguments are optional.

Not in C#

utopi160:
> > > I think this is going to break existing scripts that use that function.
> > 
> > 
> > Why is that? Arguments are optional.
> 
> Not in C#

This change makes sense to me. How can I make it compatible with existing scripts?

DaniGP17:
> This change makes sense to me. How can I make it compatible with existing scripts?

May be creating another native for that.

PR #3577, opened 2025-08-09
feat(server): add command to reload resource client file

poco8537:
### Goal of this PR

This PR adds a command to reload a specific resource’s client files from the file server. Using this command ensures that subsequent players (or for streamed assets, when fetched later) will load the updated files.

The key difference from `ensure/restart` is that it applies to players who join afterwards rather than affecting current players immediately.

Adds the `reloadclientfile` command to reload a specific resource’s client files from the file server.

Server

### Successfully tested on

**Game builds:** N/A

**Platforms:** Windows

### Checklist

### Fixes issues

N/A

### DIFF
--- a/code/components/citizen-resources-core/include/Resource.h
+++ b/code/components/citizen-resources-core/include/Resource.h
@@ -93,6 +93,11 @@ public:
 	virtual bool Start() = 0;
 
 	//
+	// Reloads client files for the resource.
+	//
+	virtual bool ClientReloadFile() = 0;
+
+	//
 	// Stops the resource.
 	//
 	virtual bool Stop() = 0;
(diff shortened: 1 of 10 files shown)


tjsdn5956:
Good idea

Firecul:
Why would this be better than using ensure/restart? I would have thought that having some clients having different code to others may cause more issues?  
I'm not a proper developer so genuinely interested on what this would help.

V3SC:
> Why would this be better than using ensure/restart? I would have thought that having some clients having different code to others may cause more issues? I'm not a proper developer so genuinely interested on what this would help.


If there are 500 players on a server and you restart a script, it will cause a lot of network traffic. Also, if you update files such as weapon files, there is a high chance that the game will crash.

This is one of the most useful PRs I've seen so far. Thank you.

Gravxd:
Would definitely love to see some movement on this PR - i think if implemented correctly would be a great tool for quick fixing bugs and issues if needed without having to disrupt the entire the server via restart.

PR #3582, opened 2025-08-12
feat(gta-streaming-five): increase fwPortalSceneGraphNode pool size to 800

DaniGP17:
### Goal of this PR
The increase of `fwPortalSceneGraphNode` was suggested in the [forum](https://forum.cfx.re/t/expansion-of-fwportalscenegraphnode-pool/5302881).
Recently it was attempted to add the pool `fwPortalSceneGraphNode` to the gameconfig(#2695) but as prikolium said they are allocated from another precomputed heap and should be looked for in `fwSceneGraph::InitPools`.

This was already done by blattersturm [here](https://github.com/citizenfx/fivem/commit/ba52c98749a83d901af4dd4f5f415fc71d4611d0) but 2 days later they [reverted](https://github.com/citizenfx/fivem/commit/85be134d37afa561e669a9049c52d162bcf550e8) his change due to `Potential cause for 'missing doors after a while'.`

The same issue happen with my first [attempt](https://github.com/citizenfx/fivem/pull/3136) of increase.
After looking it a bit more deeper the issue was because the `AssignScratchBuffer` function that allocates and organizes a scratch buffer into aligned sections for scene graph traversal(flags, screen quads, visited bitsets, etc), my first attempt didn't increase that allocation size, which led to the same issue.

Around 2 months ago I noticed that and tried to patch, when testing again the issue was still happening, the week ago I find some missing stuff that needs patching related to screen quad storage and with that everything looks to work for now in my tests.
Also I notice that there's a var that store a guard buffer to make sure the scratch buffer doesn't get corrupted during scene graph traversal and is temporarily added in the SceneGraphPoolsDebug file for testing.

If this works in FiveM, I will try to implement it in RedM as well, because several months ago I saw that it was also a problem.

The patch is almost identical to my previous pull request, but it fixes the screen quad storage used during visibility culling and portal traversal. More information about the patch can be found in my previous [pull request](https://github.com/citizenfx/fivem/pull/3136).

FiveM

### Successfully tested on
**Game builds:** 3258
**Platforms:** Windows

### Checklist

### DIFF
--- a/code/components/conhost-v2/src/DevGui.cpp
+++ b/code/components/conhost-v2/src/DevGui.cpp
@@ -294,6 +294,7 @@ devgui_convar "Tools/Streaming/Streaming Stats" strdbg
 devgui_convar "Tools/Streaming/Streaming List" strlist
 devgui_convar "Tools/Streaming/Pool Monitor" net_showPools
 devgui_convar "Tools/Streaming/pgRawStreamer assets" net_pgStats
+devgui_convar "Tools/Streaming/Scene Graph Pools" sceneGraphPools
 devgui_convar "Tools/Network/OneSync/Network Object Viewer" netobjviewer
 devgui_convar "Tools/Network/OneSync/Network SyncLog" netobjviewer_syncLog
 devgui_convar "Tools/Network/OneSync/Network Time" net_showTime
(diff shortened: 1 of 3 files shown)


ook3D:
any update on this PR?

ook3D:
ive tested the patch myself locally, and did not run into any issues, and the issue with doors disappearing from dani's previous attempt was also fixed.

nick-strohm:
I've tested this also with our team. Seems to work fine. Can you request a review, @DaniGP17?

PR #3590, opened 2025-08-14
feat(extra-natives-five): add GET_ENTITIES_IN_RADIUS for FiveM

william-des:
### Goal of this PR

Implement the GET_ENTITIES_IN_RADIUS native for FiveM, which was added for RedM by Ktos93 in #3283. This provides feature parity between FiveM, RedM and Server, and offers a more efficient method for retrieving nearby entities.

Add an extra native for fivem, with same behavior as the RedM one.

FiveM

### Successfully tested on

**Game builds:** 1604, 3258

### Checklist

### DIFF
--- a/code/components/extra-natives-five/src/PoolTraversalNatives.cpp
+++ b/code/components/extra-natives-five/src/PoolTraversalNatives.cpp
@@ -265,6 +265,90 @@ static void CloseFindHandler(fx::ScriptContext& context)
 	}
 }
 
+template<typename TTraits>
+static void GetEntitiesInRadiusForPool(fx::ScriptContext& context, 
+    float checkX, float checkY, float checkZ, float radius, 
+    bool sortOutput, const std::unordered_set<int>& modelSet,
+    std::vector<int>& entityList)
+{
+    float squaredMaxDistance = radius * radius;
+    auto pool = TTraits::GetPool();
+    
+    if (sortOutput)
+    {
+        std::vector<std::pair<float, int>> entities;
+        
+        for (int i = 0; i < pool->GetSize(); i++)
+        {
+            auto* entity = pool->GetAt(i);
+            if (!entity)
+                continue;
+                
+            auto position = entity->GetPosition();
+            float dx = position.x - checkX;
+            float dy = position.y - checkY;
+            float dz = position.z - checkZ;
+            float distSq = dx * dx + dy * dy + dz * dz;
+            
+            if (distSq >= squaredMaxDistance)
+                continue;
+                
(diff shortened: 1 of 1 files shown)


yannbcf:
It would be better to handle the sorting logic within the inner templated call. This way, if no sorting is required, we avoid creating and populating two separate vectors unnecessarily

xalva98:
I think this native will be useful and reduce the amount of code you'll have to write.
The native documentation is missing and should be added. And a small change I think makes sense in this case

```lua
local nearbyVehicles = GetEntitiesInRadius(x,y,z, 20.0, "CVehicle")
```

because that would make it more consistent with how you're calling [GetGamePool ](https://docs.fivem.net/natives/?_0x2B9D4F50) for example.
So using the pool name instead of numbers makes sense (or allow both but iirc this is not recommended).

william-des:
I think this would imply using a different name for the native in order to have a distinct signature than the RedM and server version, due to how the native declaration system works. I don’t really have an opinion on this myself, let’s see what the reviewer thinks.

Yum1x:
> I think this native will be useful and reduce the amount of code you'll have to write. The native documentation is missing and should be added. And a small change I think makes sense in this case
> 
> ```lua
> local nearbyVehicles = GetEntitiesInRadius(x,y,z, 20.0, "CVehicle")
> ```
> 
> because that would make it more consistent with how you're calling [GetGamePool ](https://docs.fivem.net/natives/?_0x2B9D4F50) for example. So using the pool name instead of numbers makes sense (or allow both but iirc this is not recommended).

While using pool names as parameters for entity types is indeed more developer-friendly, the current ecosystem doesn't easily support it. The author's suggested approach lets us use the return of [GET_ENTITY_TYPE](https://docs.fivem.net/natives/?_0x8ACD366038D14505) a parameter.

CeebDev:
Yet no update on this ?

Drift91:
Is there a reason this native is [documented for FiveM](https://docs.fivem.net/natives/?_0xDFFBA12F) when it hasn't even been implemented yet? I just wasted about an hour trying to figure out why it would return nothing but `nil` no matter what values I input. Also the fact that it returns anything at all is a bit silly when it's essentially non-existent on FiveM and should give the standard error from trying to call an undefined function.

That being said, it'll be nice when this is merged as I've seen numerous scripts that could use such a native instead of iterating through `GetGamePool('CObject')` and the like. 👍

Gogsi:
The FiveM documentation is currently (and always) been shared between FiveM and RedM. This is why RedM-specific natives also show up there. Hopefully this changes one day :D

PR #3606, opened 2025-08-22
feat(citizen-resource-gta): add resource binds natives

DaniGP17:
### Goal of this PR
Currently, the binding systems used by Fivem scripts are based on RegisterCommand and RegisterKeyMapping. This solution would allow you to retrieve and modify existing bindings without having to create new ones. The video shows a menu using these new native options for editing existing bindings.

I'm currently marking this PR as draft because I'd like to get feedback on what people think and whether or not this is a good idea. I'm also considering changing the result of `GET_BIND_INFO_FROM_COMMAND` so that it returns multiple pieces of data instead of just one object containing the data.

https://github.com/user-attachments/assets/f95ff46a-4585-40bd-8f25-06bbd1c776f2

Three new native functions `GET_BIND_INFO_FROM_COMMAND`, `SET_BIND_KEY_FOR_COMMAND`, `UNBIND_KEY_FOR_COMMAND` have been added that allow modifying binds at runtime from scripts.

FiveM, Natives

### Successfully tested on

**Game builds:** 3258
**Platforms:** Windows, Linux

### Checklist

### DIFF
--- a/code/components/citizen-resources-gta/src/GameInputFunctions.cpp
+++ b/code/components/citizen-resources-gta/src/GameInputFunctions.cpp
@@ -4,6 +4,7 @@
 #include <GameInput.h>
 
 #include <ScriptEngine.h>
+#include <ScriptSerialization.h>
 
 #include <Resource.h>
 #include <fxScripting.h>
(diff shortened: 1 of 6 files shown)


IMCraytex:
I know I’m late, but this is something I’ve been waiting for so long.
I tried building my own keybinding system by listening to rawkeypressed natives, but that only picks up keyboard inputs, not mouse buttons. Then I kind of hacked around it by mass-registering a bunch of keybinds using all the GTA key mappers, but that just turned into a mess and would’ve completely flooded users’ fivem.cfg.
This even though it seems small would be huge. And allow for a better keymapping experience overall allowing devs to write their own keybind systems and ui's for remapping.

PR #3608, opened 2025-08-23
fix(conhost-v2): text selection with filter

DaniGP17:
### Goal of this PR
Fix the issue where text selection in the console didn’t work properly while a filter was active. Before, selecting text used the original buffer indices instead of the filtered view, so the highlighted lines didn’t match what the user actually hovered.

This PR also tweaks the auto-scroll behavior when filters are added or removed, and auto-scroll is enabled, the console will now automatically scroll to the bottom.

By updating the `TextSelect` system in `CfxBigConsole` to use filtered indices instead of the original buffer indices.

For the auto-scroll, the system now tracks filter changes by comparing the current filter string with the previous frame's filter, automatically triggering a scroll to bottom when the filter changes and auto-scroll is enabled.

FiveM

### Successfully tested on
**Game builds:** 3258
**Platforms:** Windows

### Checklist

### DIFF
--- a/code/components/conhost-v2/src/ConsoleHostGui.cpp
+++ b/code/components/conhost-v2/src/ConsoleHostGui.cpp
@@ -315,6 +315,8 @@ struct CfxBigConsole : FiveMConsoleBase
 	ImVec2 PreviousWindowSize;
 
 	char FilterBuf[128];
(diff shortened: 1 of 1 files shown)


prikolium-cfx:
Thank you for this contribution

Issue #3600, opened 2025-08-20
Faster keybinds in imgui with high FPS


HeyyCzer:
### What happened?

When you have very high FPS (I tested with NVIDIA Smooth Motion at 180fps), when using keys (for example, up and down arrow) on F8 they are pressed repeatedly in a much shorter space of time than when you have lower FPS.

Another example, during loading screens when the FPS are very high if you have a large history of commands in your console and press the up arrow just once, you can go through almost all of that history and get to the beginning.

### Expected result

Have the same delay between key repeats

### Reproduction steps

1. Have high FPS count
2. Type sometimes in console and submit like commands
3. Press arrow up to get the last executed command
4. Done

### Importancy

Slight inconvenience

### Area(s)

FiveM

### Specific version(s)

FiveM 18423 / Windows

lemanschik:
Does not even sound like a bug that a design choice of them i am new to this community but i am old in the code-oss and v8 community so this sounds as trival as adding a debounce function.

hdhddhhsbsj-stack:
صشيشيصشيص

PR #3627, opened 2025-09-09
fix(conhost-v2): console key repeat delay/rate

jxnnyo:
### Goal of this PR

Fix inconsistent key repeat behaviour in the F8 console, where high FPS causes keys (e.g. arrow up/down) to repeat much faster than intended. This ensures consistent key repeat timing regardless of framerate.

By explicitly setting `io.KeyRepeatDelay` and `io.KeyRepeatRate`:

```cpp
io.KeyRepeatDelay = 0.250f;  // 250ms initial delay before repeat starts
io.KeyRepeatRate  = 0.050f;  // 50ms between repeats (20 repeats per second)
```

This provides a stable, framerate-independent key repeat.

### Successfully tested on

**Game builds:** 19032
**Platforms:** Windows

### Checklist

### Fixes issues

Fixes high FPS causing extremely fast key repeats in the F8 console (e.g. pressing the up arrow can skip through nearly the entire command history during loading screens).

**Related Issue**: #3600

### DIFF
--- a/code/components/conhost-v2/src/ConsoleHostImpl.cpp
+++ b/code/components/conhost-v2/src/ConsoleHostImpl.cpp
@@ -661,6 +661,9 @@ static HookFunction initFunction([]()
 	io.ConfigWindowsResizeFromEdges = true;
 	io.ConfigWindowsMoveFromTitleBarOnly = true;
 
+	io.KeyRepeatDelay = 0.250f;  // 250ms initial delay before repeat starts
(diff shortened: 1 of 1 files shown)


lonebadgerCfx:
Key input in 2025 should be fixed. Please this would be more than easy to fix.

jxnnyo:
> Key input in 2025 should be fixed. Please this would be more than easy to fix.

???

PR #3638, opened 2025-09-20
feat(core/rdr3): extended ped deletion circular buffer

Ehbw:
### Goal of this PR

With #3542 the max amounts of CNetObjPedBase and Peds was increased from 110/150 up to a max of 220/260 along with most pools known to be directly tied to Peds at the time, however since the patch has been merged, a few extra pools have been found to be directly related to Peds (although in testing some were tied to certain ped types such as animals, or were generally inconsistent whether they needed to be allocated or not), this PR includes 3 of them. With ``CPedAvoidanceComponent`` being intentionally omitted as it should be increased alongside ``CAvoidanceComponent`` if required.

This patch also resolves an hardcoded limit of 160 peds in a Circular Buffer related to entity deletion. In cases where 160+ entities are deleted/queued up for deletion in a short time window (from a couple of frames to a couple of seconds) would lead to ``Rage Error: 0x48B8ACBC:1758`` crashes. This patch resizes the Circular Buffer to a new max size of 270 to account for the new potential max size.

Add ``CEmotionalLocoHelper``, ``CCrimeObserver``, ``CPedTargetingComponent``, ``CAnimalGroupMember`` to pool increases related to ``CNetObjPedBase`` to resolve several instances of ``Rage Error: 0x496AC5DF:961``

Allocate a larger block of memory for the ped deletion circular buffer to use. Along with replacing all references and offsets from the old ped buffer in order for it to take advantage of the new size.

RedM

### Successfully tested on

**Game builds:** 1491

**Platforms:** Windows

Test snippet used to test changes:

```lua
RegisterCommand("createNPC", function()
	local model = `MP_U_M_M_PROTECT_STRAWBERRY`
	RequestModel(model, false)
	repeat Wait(0) until HasModelLoaded(model)
	local coords = GetEntityCoords(PlayerPedId(), false, true)

	local c = 0
	local peds = {}

	for i=0, 220 do
		local ped = CreatePed(model, coords.x, coords.y, coords.z + 5.0, 0.0, true, true, true, true)
		if ped ~= 0 then
			peds[c] = ped
			c = c + 1
			SetRandomOutfitVariation(ped, true)
		end
		Wait(300)
	end
	print(c)

	Wait(1000)

	for i=0, 220 do
		if peds[i] then
			DeleteEntity(peds[i])
		end
	end
end, false)
```

Without patch: Crash past 150 for ``CEmotionalLocoHelper`` or ``CCrimeObserver`` pools being full. If these don't get allocated (e.g. using different peds). Crash with ``Rage Error: 0x48B8ACBC:1758`` from trying to delete more then 160 peds.

With patch: Don't crash when spawning up to 200+ peds. Also don't crash when deleting all the peds created. even if the number exceeds the previous 160 limit.

### Checklist

### Fixes issues

Rage Error: 0x48B8ACBC:1758

### DIFF
--- a/code/components/gta-core-rdr3/src/PatchIncreaseMaxPed.cpp
+++ b/code/components/gta-core-rdr3/src/PatchIncreaseMaxPed.cpp
@@ -11,6 +11,25 @@
 // This patch resolves this by improving logic to account for the increase size of "CNetObjPedBase" pool and removing a ped pool size check that would otherwise fatally error.
 //
 
+template<int EntitySize = 160>
+struct EntityCircularBuffer
+{
+	void* entity[EntitySize];
+
+	uint32_t cursorPos;
+	uint32_t baseOffset;
+	uint32_t totalEntries;
+};
+
+template<int instrLen = 7, int instrOffset = 3>
+static void PatchRelativeLocation(uintptr_t address, uintptr_t newLocation)
+{
+	uint8_t* instructions = reinterpret_cast<uint8_t*>(address);
+	uintptr_t instrNext = reinterpret_cast<uintptr_t>(instructions) + instrLen;
+	int32_t newOffset = (int32_t)((uintptr_t)newLocation - instrNext);
+	hook::put<int32_t>(instructions + instrOffset, newOffset);
+}
+
 static HookFunction hookFunction([]()
 {
 	constexpr size_t kDefaultMaxPeds = 110;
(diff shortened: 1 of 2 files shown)


ChuckFinley:
@prikolium-cfx this is more important than any of the other pull requests currently, stability of any large server is completely ruined by this crash code. Please help us out here and review this so we can get this pushed. 

Thanks you!

DaNeo61:
@prikolium-cfx please merge this for more performance on redm servers, we realy need it....

prikolium-cfx:
Merged

PR #3668, opened 2025-10-09
tweak(native/codegen): only check that we're not server before filtering

AvarianKnight:
### Goal of this PR
Currently if we set server natives to be for specific games (like `gta5` or `rdr3`) it will be filtered out of server codegen because the default game value is 'gta5' when doing server codegen.

On the server all server natives exist at all times, so we just adjust the native filtering to respect this.

This also cleans up the return descriptions of the new natives.

Server

### Successfully tested on
**Platforms:** Windows

### Checklist

### DIFF
--- a/ext/native-decls/GetMount.md
+++ b/ext/native-decls/GetMount.md
@@ -6,10 +6,11 @@ game: rdr3
 ## GET_MOUNT
 
 ```c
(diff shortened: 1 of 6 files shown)


prikolium-cfx (approved):

prikolium-cfx:
Thank you for your efforts!

PR #3693, opened 2025-10-26
feat(server-impl): add convar to hide file size warnings

ook3D:
### Goal of this PR

added a server convar to disable file size warnings from the server console.

adding a new convar

Server

### Successfully tested on

**Game builds:** .. 
3258
**Platforms:** Windows, Linux
Windows

### Checklist

### DIFF
--- a/code/components/citizen-server-impl/src/ResourceStreamComponent.cpp
+++ b/code/components/citizen-server-impl/src/ResourceStreamComponent.cpp
@@ -418,6 +418,7 @@ namespace fx
 		return (uint32_t)size;
 	}
 
+	static ConVar<bool> sv_hideSizeWarnings("sv_hideSizeWarnings", ConVar_ServerInfo, false);
 	void ResourceStreamComponent::ValidateSize(std::string_view name, uint32_t physSize, uint32_t virtSize, int* numWarnings)
 	{
 		auto checkSize = [this, &name, numWarnings](uint32_t size, std::string_view why)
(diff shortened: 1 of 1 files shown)


ook3D:
This convar is useful for production environments, as you don't really need to see these console logs, and should only really be there for development / debugging purposes.

AvarianKnight:
I don't agree, I think you give too much credit that some servers don't just drag and drop into production.

These warnings are meant to be annoying since you *should* fix them.

ook3D:
> I don't agree, I think you give too much credit that some servers don't just drag and drop into production.
> 
> These warnings are meant to be annoying since you _should_ fix them.

I mean, its an opt-in convar, thats on them if they choose to do so

colistro123:
Perhaps there should be an ImGui UI to put asset warnings in a list, rather than spamming the log. Or if assets are displayed anywhere, just show that they're oversized if so.

Scullyy:
> Perhaps there should be an ImGui UI to put asset warnings in a list, rather than spamming the log. Or if assets are displayed anywhere, just show that they're oversized if so.

Most wouldn’t check the interface to see the warnings, and a lot of servers unfortunately ignore them already even with the “spam”. Just remove the problematic assets or resolve the issue so the warnings go away and nothing will be spammed, even then it only happens on startup anyways.

FabianTerhorst:
I agree with @AvarianKnight. There is a reason for this warning. And when the server doesn't want to have them in production, then they should fix the asset.

Gogsi:
I think the warning should stay but the size might need to be bumped a bit. There are assets in new GTA Online DLCs that have textures large enough to trigger the warnings if you stream them manually. 

image-30

ook3D:
> I think the warning should stay but the size might need to be bumped a bit. There are assets in new GTA Online DLCs that have textures large enough to trigger the warnings if you stream them manually.
> image-30

i think this is the best option here, best of both

Scullyy:
> I think the warning should stay but the size might need to be bumped a bit. There are assets in new GTA Online DLCs that have textures large enough to trigger the warnings if you stream them manually.
> 
> image-30

Might be best to just remove the blue and yellow warnings and just keep the red since it has a limit of 64 and that vanilla asset is at 54. Or at least remove the blue and make yellow 64 and red 96 or something.

ook3D:
closing in favour of a new approach discussed above

PR #3704, opened 2025-11-05
docs: Deprecate SET_MANUAL_SHUTDOWN_LOADING_SCREEN_NUI

LokiLeiche:
### Goal of this PR
This native was removed years ago and now just throws a deprecation warning referencing to the new way of achieving what this native used to do and this should be written in the docs.
The source code that manages this is [/code/components/loading-screens-five/src/LoadingScreens.cpp lines 251-256](https://github.com/citizenfx/fivem/blob/0b6d5de3902cda6dd91be0c489d9b7243e554bb1/code/components/loading-screens-five/src/LoadingScreens.cpp#L251).

Updated description and reference to alternative solution.

Native Reference documentation


LokiLeiche:
Also refer to: https://github.com/citizenfx/fivem-docs/pull/564

AvarianKnight:
LGTM

PR #3715, opened 2025-11-13
feat(gta-net-five): increase vehicle orientation blend threshold

DaniGP17:
### Goal of this PR
Improve remote vehicle smoothness under OneSync by increasing the minimum orientation delta required to trigger orientation blending.
The default threshold was tuned for high tickrate peer-to-peer networking, causing excessive angle corrections and visible snapping in OneSync’s lower-tick environment.

This PR locates the vehicle blender data and adjusts its m_MinOrientationDiffToStartBlend value to 0.08(~4.58 degrees).

FiveM

### Successfully tested on
**Game builds:** 3095

**Platforms:** Windows, Linux

### Checklist

### DIFF
--- a/code/components/gta-net-five/src/netBlender.cpp
+++ b/code/components/gta-net-five/src/netBlender.cpp
@@ -1,7 +1,9 @@
 #include <StdInc.h>
 #include <Hooking.h>
-
 #include <netBlender.h>
+#include <CoreConsole.h>
+#include "nutsnbolts.h"
+
 
 static hook::cdecl_stub<void(rage::netBlender*, uint32_t timeStamp)> netBlender_SetTimestamp([]()
 {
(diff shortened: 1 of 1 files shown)


radium-cfx:
Can we remove the per-frame check and add a change callback instead?

```cpp
static float* orientationDiffPtr;

static void MinBlendDiffChanged(internal::ConsoleVariableEntry<bool>* var)
{
    if (var->GetRawValue())
    {
        *orientationDiffPtr = 0.08f; // ~4.58 degrees
    }
    else
    {
        *orientationDiffPtr = 0.05f; // ~2.86 degrees, original
    }
}

(in the lambda of the hook function):
static ConVar<bool> enableAngleCorrectionDelta("game_forceLowerAngleCorrectionDelta", ConVar_Replicated, true, &MinBlendDiffChanged);
...
*orientationDiffPtr = 0.08f; // set once on init
```

DaniGP17:
Thanks :smile:, now should be good.

code/components/gta-net-five/src/netBlender.cpp
```diff
@@ -15,3 +17,58 @@ void netBlender::SetTimestamp(uint32_t timestamp)
...
+		);
+
+		if (enableAngleCorrectionDelta.GetValue())
+		{
+			*orientationDiffPtr = kRaisedMinOrientationDelta;
+		}
+		else
+		{
+			*orientationDiffPtr = kDefaultMinOrientationDelta;
+		}
```
radium-cfx: In this stage the ConVar will always be true so we can just set it to our patched value and don't need to check its value.

code/components/gta-net-five/src/netBlender.cpp
```diff
@@ -15,3 +17,58 @@ void netBlender::SetTimestamp(uint32_t timestamp)
...
+	}
+});
```
radium-cfx: Nit: Missing new line EOF.

code/components/gta-net-five/src/netBlender.cpp
```diff
@@ -15,3 +17,58 @@ void netBlender::SetTimestamp(uint32_t timestamp)
...
+
+static float* orientationDiffPtr = nullptr;
+
+static constexpr float kDefaultMinOrientationDelta = 0.05f;   // ~2.86 degrees (original)
+static constexpr float kRaisedMinOrientationDelta  = 0.08f;   // ~4.58 degrees (patched)
+
+static void MinBlendDiffChanged(internal::ConsoleVariableEntry<bool>* var)
+{
+	if (!orientationDiffPtr)
+		return;
```
radium-cfx: Lets use braces here as well.

radium-cfx (changes requested):
Some small changes, otherwise it looks good!

Please also squash all commits once, thanks!😄

DaniGP17:
Done, thanks.

Legacy-TacticalGamingInteractive:
@DaniGP17 love you for this

PR #3719, opened 2025-11-15
tweak(citizen-devtools): remove slow system performance warning

DaniGP17:
### Goal of this PR
Remove the slow performance warning, since it provides no actionable information for users and can appear even on high-end systems, leading to confusion rather than helping diagnose real issues.

Remove the warn block

FiveM, RedM

### Successfully tested on
**Game builds:** none
**Platforms:** Windows, Linux

### Checklist

### Fixes issues
fixes https://forum.cfx.re/t/suggestion-allow-toggle-remove-slow-system-performance-warning/5356888

### DIFF
--- a/code/components/citizen-devtools/src/ResourceTimeWarnings.cpp
+++ b/code/components/citizen-devtools/src/ResourceTimeWarnings.cpp
@@ -139,8 +139,6 @@ constexpr auto tuple_slice(Cont&& t)
 	std::make_index_sequence<I2 - I1>{});
 }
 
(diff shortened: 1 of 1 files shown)


radium-cfx:
When removing this, we can also get rid of the logic around `lastHitch` and `currentTotal` and do some further cleanup there.

DaniGP17:
Thanks, I've removed `lastHitch` and `currentTotal`, let me know what else do we need to clean 🙂

PR #3725, opened 2025-11-18
fix(citizen-server-impl): validate onesync range when reassigning entity ownership

DaniGP17:
### Goal of this PR
Fixes a bug where vehicle entities become stuck/frozen after players change routing buckets or disconnect. The issue occurred when entity ownership was incorrectly transferred to players outside of OneSync range, causing entities to be uninteractable until the assigned owner (who is too far away) teleports back to the entity.

Validate that the candidate player is in the same routing bucket and is in the culling radius of the entity.

Server

### Successfully tested on
I couldn't test this because the user that report this on the forum said that to reproduce the issue I need 20-30 players.

### Checklist

### Fixes issues
https://forum.cfx.re/t/server-side-vehicles-being-frozen-at-random/

### DIFF
--- a/code/components/citizen-server-impl/src/state/ServerGameState.cpp
+++ b/code/components/citizen-server-impl/src/state/ServerGameState.cpp
@@ -2137,7 +2137,7 @@ void ServerGameState::UpdateEntities()
 		else
 		{
 			// workaround: force-migrate a stuck entity
-			if ((time - entity->lastReceivedAt) > 10s)
+			if ((time - entity->lastReceivedAt) > 5s)
 			{
 				if (g_oneSyncForceMigration->GetValue() || fx::IsBigMode())
 				{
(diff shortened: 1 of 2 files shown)


FabianTerhorst (approved):
Thanks a lot for your contribution.

lLukDreams:
this would explain some of the issues people are experiencing when changing routing buckets, great work!

C4LLZ:
This is great work

AvarianKnight:
I'm a bit confused on how this actually fixes ownership (at least in the case of routing buckets), the logic is already handled via `relevantTo` 

See:
https://github.com/citizenfx/fivem/blob/96dec18b3b375cfa2cfa17f867aff052daa828e9/code/components/citizen-server-impl/src/state/ServerGameState.cpp#L1200-L1206

FabianTerhorst (changes requested):
After further investigation this might not be the best fix for this issue. Also entities that would be stuck in this would be fixed automatically after 10 seconds.

DaniGP17:
I've made changes in case anyone wants to review it.

alberttheprince:
Would be a big thing if this could get added if working, been a long standing issue 🙏

PR #3730, opened 2025-12-01
tweak(server/gamestate): CreatePed RPC & CreatePedServerSetter

goncalobsccosta:
### Goal of this PR

Allow the usage of CreatePed RPC and create CreatePedServerSetter. Currently, the documentation states that CreatePed is an RPC, but it actually is a server setter. By separating these natives, users gain finer control over which method they want to use for spawning peds.

This PR changes the ServerSetters code to separate 2 natives, the server setter and the RPC.

FiveM, Server

### Successfully tested on

**Game builds:** 3258

**Platforms:** Windows, Linux

### Checklist

### DIFF
--- a/code/components/citizen-server-impl/src/state/ServerRPC.cpp
+++ b/code/components/citizen-server-impl/src/state/ServerRPC.cpp
@@ -132,7 +132,7 @@ static InitFunction initFunction([]()
 		{
 			// deprecated by ServerSetters
 #ifdef STATE_FIVE
-			if (native->GetName() == "CREATE_PED" || native->GetName() == "CREATE_OBJECT_NO_OFFSET")
+			if (native->GetName() == "CREATE_OBJECT_NO_OFFSET")
 			{
 				continue;
 			}
--- a/code/components/citizen-server-impl/src/state/ServerSetters.cpp
+++ b/code/components/citizen-server-impl/src/state/ServerSetters.cpp
@@ -443,7 +443,7 @@ static InitFunction initFunction([]()
 			ctx.SetResult(guid);
 		});
 
-		fx::ScriptEngine::RegisterNativeHandler("CREATE_PED", [=](fx::ScriptContext& ctx)
+		fx::ScriptEngine::RegisterNativeHandler("CREATE_PED_SERVER_SETTER", [=](fx::ScriptContext& ctx)
 		{
 			uint32_t resourceHash = 0;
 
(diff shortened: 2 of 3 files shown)


AvarianKnight:
I think the ship has sailed on this one. We cannot change expected behavior, even if it's not obvious what this is doing, most users that use this know its not a RPC native, and switching it back to a RPC will lead to even more confusion.

Ehbw:
To add to Avarian's comment, this would very likely break compatibility for all existing scripts using CREATE_PED server-setters that depended on the ped creation to be done instantly and where the ped may not be assigned an owner yet as it has been since January 2021 https://github.com/citizenfx/fivem/commit/4cbf129072fd2a93e0e854def4d5057508c96ceb). It is also worth noting that it is documented that CREATE_PED and CREATE_OBJECT_NO_OFFSET RPC was [deprecated](https://github.com/citizenfx/fivem/blob/1c490ee35560b652c97a4bfd5a5852cb9f033284/code/components/citizen-server-impl/src/state/ServerRPC.cpp#L133)in favour of their equivalent server-setters.

If you really need an CREATE_PED RPC, it would make more sense to make it a new native and clearly document that it is an RPC rather then being a server-setter, Along with removing the old ``CREATE_PED`` reference here https://github.com/citizenfx/fivem/blob/1c490ee35560b652c97a4bfd5a5852cb9f033284/ext/natives/rpc_spec_natives.lua#L18 assuming the native docs pulls data from here to determine if a native should be marked as an RPC or not. But at this moment the server RPC in FiveM is not very reliable.

goncalobsccosta:
I agree with the points everyone has made. Changing the existing behavior now would definitely cause more problems than it solves, especially with so many scripts relying on the current server-setter behavior of CREATE_PED.

My only concern was that there’s currently no alternative RPC native that uses a different name everything that exists is still tied to the original native names, so it felt inconsistent. But I understand why introducing breaking changes isn’t an option.

How should I proceed?

AvarianKnight:
The answer is.. its complicated.

You can add a new name to the entity_rpc but we would need a way to link it to the `CREATE_PED` native so when doc gen happens the rpc will point to the right native, this would also be needed to actually have the client execute the right RPC native. If you try to send `CREATE_PED_RPC` to the client for it to do invokes it wont work

spacevx:
> The answer is.. its complicated.
> 
> You can add a new name to the entity_rpc but we would need a way to link it to the `CREATE_PED` native so when doc gen happens the rpc will point to the right native, this would also be needed to actually have the client execute the right RPC native. If you try to send `CREATE_PED_RPC` to the client for it to do invokes it wont work

Can we just remove the CREATE_PED RPC since it's not used anymore?

PR #3738, opened 2025-12-12
tweak(ext/cfx-ui): display EOS warning to server owners only

airbon99:
### Goal of this PR

Don't display information most users don't understand and won't act upon.

Hiding server EOS warning from players, displaying it only to server owners.

Cfx UI

### Checklist


Mathu-lmn:
Not a huge fan of it. Some server developers might not be owners but still need to see such issues.
Also this would reduce the impact of such message that is important (keeping your artifacts updated improves stability)

PR #3785, opened 2026-01-18
fix(scripting): use `onResourceStop` for exports cache clearing

D4isDAVID:
### Goal of this PR

Fix exports not working for resources when listening to `onResourceStart` and restarting.

For example, I have a resource `A` that provides an export, and resource `B` that wants to run the export *immediately* when resource `A` starts. This would work at first; the exports cache is still empty, so resource `B` can request the export from resource `A`. However, if I restart resource `A`, resource `B` still has the old export cached, which would now have an invalid function reference, as resource `A` has been stopped, resulting in an error like this:
image

This PR changes the `scheduler.lua` and `main.js` to use `onResourceStop` instead of `onClientResourceStop`/`onServerResourceStop` to clear the cache, to prevent delaying the cache clearing until after `onResourceStart`. C# doesn't seem to cache exports like this, so no changes were made there, as the error does not occur.

ScRT: Lua, ScRT: JS

### Successfully tested on

**Platforms:** Windows

### Checklist

### DIFF
--- a/data/shared/citizen/scripting/lua/scheduler.lua
+++ b/data/shared/citizen/scripting/lua/scheduler.lua
@@ -655,7 +655,8 @@ end
 
 -- Remove cache when resource stop to avoid calling unexisting exports
 local function lazyEventHandler() -- lazy initializer so we don't add an event we don't need
-	AddEventHandler(('on%sResourceStop'):format(isDuplicityVersion and 'Server' or 'Client'), function(resource)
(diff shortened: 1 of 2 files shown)


Mathu-lmn:
Nitpick but you might want to fix the title by replacing onResourceStart by onResourceStop, this got me confused at first 😄

D4isDAVID:
Ah, my bad

PR #3790, opened 2026-01-19
feat(connection): add server blacklist validation during initConnect

WooAf:
This PR introduces a validation step in the initConnect flow to check whether the target server is present in a blacklist before completing the connection handshake.


radium-cfx:
Thank you for your idea, but unfortunately, this approach will not work as intended.

PR #3795, opened 2026-01-29
fix(resource): prevent duplicate hash parameter in resource URL

Simon2122:
### Goal of this PR
This PR prevents the duplication of the `hash` query parameter in resource URLs within `CachedResourceMounter`.

Currently, if a resource URL is serialized with its hash (e.g., during Replay recording) and then re-loaded, `AddResourceEntry` blindly appends the hash again. This results in malformed URLs like:
hash demonstration

`https://cache-southside.fiveshield.co/files/SOUTHSIDE_pedmenu/resource.rpf?hash=8bd808a2c18580d08e38f8bf7caaef6e45220a4b?hash=8bd808a2c18580d08e38f8bf7caaef6e45220a4b`

I have modified the logic in `code/components/citizen-resources-client/src/CachedResourceMounter.cpp` to check if the `hash=` parameter is already present in the `remoteUrl` before appending it.

If `refUrl.find("hash=")` returns a match, the existing hash is preserved and no new parameter is appended. This ensures that the URL remains valid regardless of whether it's a fresh load or a restored state.

FiveM, RedM, Client (Resource System)

### Successfully tested on
**Game builds:** 3095 (or whichever build you tested on)

**Platforms:** Windows

### Checklist

### DIFF
--- a/code/components/citizen-resources-client/src/CachedResourceMounter.cpp
+++ b/code/components/citizen-resources-client/src/CachedResourceMounter.cpp
@@ -308,7 +308,7 @@ void CachedResourceMounter::AddResourceEntry(const std::string& resourceName, co
 
 	auto refUrl = remoteUrl;
 	
-	if (refUrl.find("hash=") == std::string::npos)
+	if (refUrl.find("?hash=") == std::string::npos)
 	{
 		refUrl += "?hash=" + referenceHash;
 	}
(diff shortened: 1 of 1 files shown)


Simon2122:
This error happens rarely, and all requests from the connected player contain two hashes.

PR #3811, opened 2026-02-05
fix(mumble): validate our samples before passing them to the jitter buffer

AvarianKnight:
If we happen to get an invalid output from opus_decoder_get_nb_samples we could poison the jitter buffer

If we don't keep our numSamples in check with our frame size then we can also cause our jitter buffer to desync, causing audio artifacting

### Goal of this PR
(hopefully) Fix the audio issue some players experience when playing RedM

Make sure we keep our jitter buffer synchronized and that we don't push frames that have errored into our jitter buffer.

FiveM, RedM

### Successfully tested on

**Game builds:** .. 

**Platforms:** Windows, Linux

### Checklist

### Fixes issues
https://github.com/AvarianKnight/pma-voice/issues/564


AvarianKnight:
I added some more fixes here, and size validations for incoming TCP packets so a server can't crash a client if they want to by sending a packet saying the length is something insane, see: https://github.com/mumble-voip/mumble/blob/30892a5efdf5cab417e4e410d797c440adfab9f1/src/Connection.cpp#L146-L150

I would like some guidance if we want to support having multiple TCP message in a single packet (this is what the mumble-server allows, but we currently don't support this). See: https://github.com/mumble-voip/mumble/blob/30892a5efdf5cab417e4e410d797c440adfab9f1/src/Connection.cpp#L146-L150

PR #3817, opened 2026-02-07
feat(gta-core-five): clamp vehicle entrypoint index

divocbn:
### Goal of this PR

Prevent a crash caused by cheaters sending invalid vehicle entry/exit point indices that are serialized and synchronized in cloned vehicle tasks.

This PR clamps the serialized target entry/exit point index for CTaskEnterVehicle and CTaskExitVehicle to the valid range defined by the vehicle model entry/exit points.

FiveM

### Successfully tested on

**Game builds:** 3258, 3570, 3717

**Platforms:** Windows

### Checklist

### DIFF
--- a/code/components/gta-core-five/src/CrashFixes.ClampVehicleEntryIndex.cpp
+++ b/code/components/gta-core-five/src/CrashFixes.ClampVehicleEntryIndex.cpp
@@ -1,0 +1,61 @@
+#include "StdInc.h"
+
+#include "Hooking.FlexStruct.h"
+#include "Hooking.Patterns.h"
+
+#include "Hooking.Stubs.h"
+
+static uint8_t g_targetEntryPointOffset = 0x0;
+static uint8_t g_syncedVehicleOffset = 0x0;
+
+static uint32_t g_modelInfoDataOffset = 0x0;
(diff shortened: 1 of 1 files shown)


st860923:
Should there be a guard for numEntryExitPoints == 0 before the clamp?
The game itself seem early-outs when the count is zero.

divocbn:
> Should there be a guard for numEntryExitPoints == 0 before the clamp? The game itself seem early-outs when the count is zero.

Kinda sure that it should be right as is, especially because 2-seat vehicles has 0 numEntryExitPoints.
Also my proof-of-concept crasher isn't working after that fix, could you provide any screens

st860923:
You're right. I traced through the downstream code and the function already guards against numEntryExitPoints == 0 by skipping the lookup loop entirely and returning false, so the value is never actually used. The fix works as intended. 👍

radium-cfx (approved):
Looks good.

Thank you!

Pedro-Lucas14:
@prikolium-cfx @radium-cfx Is there any estimate for when this fix will be available on all FiveM versions? There are many cheaters causing crashes in the game.

PR #3825, opened 2026-02-16
fix(gta-core-five): crmtComposer stack underflow in BlendN child counting

### LINKED ISSUE #3669: Car–Motorcycle Collision at High FPS causing crash
### What happened?

When **watching** a collision between a car and a motorcycle, the client crashes if running at uncapped or very high FPS. The crash does not occur when FPS is capped at lower FPS (e.g., Vsync).

https://youtu.be/aQL8JJDJ-UU -> High FPS, crash
https://youtu.be/CroEb6FTdxM -> High FPS, crash
https://youtu.be/-J4mVdvJ2zQ -> VSync/Capped FPS, no crash

This issue is reproducible not just for me, but also for my friends, we switched places to be the one spectating the collision, and whoever was watching at high FPS would crash as well.

### Expected result

The game should not crash when viewing vehicle collisions, regardless of FPS.

### Reproduction steps

1. Spawn a car and a motorcycle. (I always had a friend on the motorcycle, not sure if it's necessary)
2. Cause them to collide (side-impact usually did the job).
3. Watch the collision closely with FPS uncapped / running very high (e.g., 240+).
4. Observe that the client crashes.
5. Repeat the steps with an FPS cap (e.g., 60–75).
6. Notice that the crash does not happen.

### Importancy

Crash

### Area(s)

FiveM

### Specific version(s)

Tested on 3407 and 3528

### Additional information

Full crash dump: https://drive.google.com/file/d/1fo5tCoxrMChHCSmKq5hyiMuF5aElMtpP/view?usp=sharing

DaniGP17:
### Goal of this PR
Fix a stack underflow crash in `crmtComposer` that occurs when handling `kNodeBlendN`/`kNodeMergeN`/`kNodeAddN` nodes with specific child configurations.

The crash was caused by incorrect boolean logic when counting active children, which could result in more `PushPair()` calls than actual frames present on the composer stack. This led to an underflow access (`m_FrameStack[-1]`).

This PR corrects the boolean condition used when counting active children in the BlendN code path.
The original logic used:
```cpp
if (!child->IsDisabled() || !child->IsSilent())
```

This incorrectly counted silent children, even though silent nodes do not push frames onto the stack during traversal.

The fix changes the logic to:
```cpp
if (!child->IsDisabled() && !child->IsSilent())
```

This ensures that only children that actually contribute frames are counted, keeping the internal composer stack balanced and preventing underflow in PushPair().

FiveM

### Successfully tested on
**Game builds:** None
**Platforms:** Windows

### Checklist

### Fixes issues
fixes #3669

### DIFF
--- a/code/components/gta-core-five/src/CrashFixes.CrmtComposerStackUnderflow.cpp
+++ b/code/components/gta-core-five/src/CrashFixes.CrmtComposerStackUnderflow.cpp
@@ -1,0 +1,23 @@
+#include "StdInc.h"
+#include "Hooking.Patterns.h"
+
+static HookFunction hookFunction([]()
+{
+	// crmtComposer::Visit (kNodeBlendN / kNodeMergeN / kNodeAddN case)
+	// Root cause:
+	// The child counting logic uses an incorrect boolean condition:
+	//     if (!child->IsDisabled() || !child->IsSilent())
+	//
+	// This incorrectly counts children that are silent, even though silent
+	// nodes do NOT push a frame onto the composer stack during traversal.
+	//
+	// As a result, the computed 'count' may exceed the actual number of
+	// frames pushed onto m_FrameStack. When PushPair() is later executed,
+	// it assumes at least two frames are available and reads:
+	//     m_FrameStack[m_TopFrame - 1]
+	//
+	// If the stack is imbalanced, m_TopFrame can reach 0, causing an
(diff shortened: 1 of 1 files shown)


imLocutor:
@DaniGP17 i have a crash that u need to see, can u contact me on discord? locutor404

PR #3859, opened 2026-03-11
fix(streaming): reinitialize scaleform .gfx on resource restart

### LINKED ISSUE #1576: ``RequestScaleformMovie`` fails to execute/invoke if called again after resource restart.
**Description:**
While developing scaleforms I've come to discover that if a streamed scaleform is loaded (and although not required disposed of accordingly with ``SetScaleformMovieAsNoLongerNeeded``) then the resource responsible restarts and attempts to load the scaleform again (for example, if you've made changes and wish to test) the call to ``RequestScaleformMovie`` will fail.

I'm Unsure how long this problem has existed however the repro I've provided was tested on the latest server artefacts although that shouldn't affect anything AFAIK.

**How to use repro:**
1. Download and install repro resource [here (repro.zip).](https://github.com/citizenfx/fivem/files/9338246/repro.zip)
2. Once in server start the resource
3. Ensure the resource to which an error will be displayed in the console.

**Misc details:**
The scaleform being streamed is a random scaleform I chose from the vanilla game however this issue can observed with both vanilla or custom scaleforms.

frostfire575:
### Goal of this PR

Fix `.gfx` scaleform files failing to load after a resource restart, eventually causing a crash.

When a resource containing a `.gfx` file is restarted, `FindSlot` finds the existing streaming slot so `initGfxTexture` is skipped (it only runs for newly created slots). `CScaleformStore` retains stale internal state, causing `RequestScaleformMovie` to fail with native error `0x11fe353cf9733e6f` and eventually crash.

Two changes in `LoadStreamingFile.cpp`:

1. When a `.gfx` entry's slot already exists and is pending removal, push it onto `newGfx` so `initGfxTexture` runs again after reload.
2. Before clearing the pending-removal flag, call `ReleaseObject` on the old streaming entry so `CScaleformStore` drops its stale state, allowing a clean reinit.

Both paths are gated on `g_pendingRemovals.count(...)`, so they only fire during resource restart — normal first-load and non-`.gfx` paths are untouched.

FiveM

### Successfully tested on

**Game builds:** FiveM

**Platforms:** Windows

### Checklist

### Fixes issues

Fixes #1576

### DIFF
--- a/code/components/gta-streaming-five/src/LoadStreamingFile.cpp
+++ b/code/components/gta-streaming-five/src/LoadStreamingFile.cpp
@@ -1901,11 +1901,24 @@ static void LoadStreamingFiles(LoadType loadType)
 					newGfx.push_back({ strId, nameWithoutExt });
 				}
 			}
+			else if (ext == "gfx" && g_pendingRemovals.count({ strModule, strId }))
+			{
+				// GH-1576: slot exists from previous load but was pending removal —
+				// CScaleformStore has stale state, reinitialize it
+				newGfx.push_back({ strId, nameWithoutExt });
+			}
 #elif IS_RDR3
 			strModule->FindSlotFromHashKey(&strId, HashString(nameWithoutExt.c_str()));
 #endif
 
 			g_ourIndexes.insert(strId + strModule->baseIdx);
+			// GH-1576: for .gfx files pending removal, release the stale object first
+			if (ext == "gfx" && g_pendingRemovals.count({ strModule, strId }))
+			{
+				auto fullIdx = strId + strModule->baseIdx;
+				cstreaming->ReleaseObject(fullIdx, 0xF1);
+				cstreaming->ReleaseObject(fullIdx);
+			}
 			g_pendingRemovals.erase({ strModule, strId });
 
 			// get the raw streamer and make an entry in there


frostfire575:
Hey @prikolium-cfx @radium-cfx, can you take a look at this? It is an important one.
Tested it and works perfectly , without this fix, restarting a resource that uses a .gfx scaleform completely breaks it, REQUEST_SCALEFORM_MOVIE throws native error 0x11fe353cf9733e6f (screenshot attached). Only way to recover was a full server restart or renaming the .gfx file, which is a pain during development. 
After this fix it just works, restart resource all you want. 

image

imsnaily:
Please accept this pr lol

manups4e:
i mean.. i simply make 2 resources... the "assets" resource with the .gfx in it (never to be restarted).. and the script one that can be restarted.. it's pretty much that :)

edit: also this is basically FiveM assets handling 101... never stream in the same resource with the code if you can avoid it.. because you should never restart an assets unless specifically needed

frostfire575:
> i mean.. i simply make 2 resources... the "assets" resource with the .gfx in it (never to be restarted).. and the script one that can be restarted.. it's pretty much that :)

yeah that works as a workaround but as someone who sells scripts on Tebex, shipping an extra resource just to work around a bug isnt great for end users.
and honestly the fix is just a few lines in LoadStreamingFile -> detect pending removal, release the stale entry, reinit. done.

manups4e:
i've been selling my RoundedMinimap with 2 resources and that went pretty fine 😄 i'm not saying this is a useless PR because it's definitely helping.. but yeah.. 
image

or

image

so far this method worked pretty well

QuadrupleTurbo:
having one instead of 2 resources is much better, and being able to restart it real time for debugging would be 🔥

frostfire575:
> i've been selling my RoundedMinimap with 2 resources and that went pretty fine 😄 i'm not saying this is a useless PR because it's definitely helping.. but yeah.. image
> 
> or
> image
> 
> so far this method worked pretty well

Yes but i have multiple resource using scaleform ( so have to ship multiple asset folder )  but yeah your one is a good solution !

manups4e:
> Yes but i have multiple resource using scaleform ( so have to ship multiple asset folder ) but yeah your one is a good solution !

And i'm almost sure 80% of them are made by me 🤔 and if you want you can group all your gfx scaleforms into one unique streamed resource :3 don't worry about the number of assets in there 
image

Consider also that scaleforms are niche, they're not being prioritized like NUI / Lua runtimes sadly

manups4e:
Anyway i'm all UP for this PR and i really hope it gets merged soon!

QuadrupleTurbo:
I made a whole tool to automate the restarting of the scaleform by renaming the scaleform and reloading, it works but man it seems so stupid when we can just use this pr, @frostfire575 it does live edits too right?

frostfire575:
> I made a whole tool to automate the restarting of the scaleform by renaming the scaleform and reloading, it works but man it seems so stupid when we can just use this pr, @frostfire575 it does live edits too right?

what do you mean by live edits? can you explain a bit ?

QuadrupleTurbo:
> > I made a whole tool to automate the restarting of the scaleform by renaming the scaleform and reloading, it works but man it seems so stupid when we can just use this pr, @frostfire575 it does live edits too right?
> 
> what do you mean by live edits? can you explain a bit ?

Sure so let's say you edit a scaleform live and restart the resource, will the scaleform reload with the changes?

frostfire575:
> > > I made a whole tool to automate the restarting of the scaleform by renaming the scaleform and reloading, it works but man it seems so stupid when we can just use this pr, @frostfire575 it does live edits too right?
> > 
> > 
> > what do you mean by live edits? can you explain a bit ?
> 
> Sure so let's say you edit a scaleform live and restart the resource, will the scaleform reload with the changes?

yes since the fix forces ReleaseObject on the stale entry and reinitializes . it will pick up the updated .gfx file on restart  so yes, edit your scaleform, restart the resource, changes load in cleanly.  This is my discord : frostfire575 you can message me .

QuadrupleTurbo:
> > > > I made a whole tool to automate the restarting of the scaleform by renaming the scaleform and reloading, it works but man it seems so stupid when we can just use this pr, @frostfire575 it does live edits too right?
> > > 
> > > 
> > > what do you mean by live edits? can you explain a bit ?
> > 
> > 
> > Sure so let's say you edit a scaleform live and restart the resource, will the scaleform reload with the changes?
> 
> yes since the fix forces ReleaseObject on the stale entry and reinitializes . it will pick up the updated .gfx file on restart  so yes, edit your scaleform, restart the resource, changes load in cleanly. This is my discord : frostfire575 you can message me .

Okay this needs to be merged! I'll shoot you a msg

manups4e:
I’d like to address the unnecessary hostility outside GH following my previous comments.

My intent was never to discourage this PR or present a workaround as a replacement for a core fix. With 8 years of experience in FiveM resource management, and 5 years of experience in scaleforms development, a depth of field experience that is not exactly common in this environment, I simply shared the "Assets vs. Scripts" separation as a fundamental best practice (FiveM 101) to ensure stability in production.

Providing architectural advice to help users avoid crashes in the meantime does not negate the value of fixing the underlying engine bug. It is disappointing to be attacked for sharing technical expertise aimed at improving workflow. I fully support this PR as a step toward a more resilient platform, but I expect a professional and respectful dialogue, especially when sharing established industry-standard methods.

gtasnail:
> especially when sharing established industry-standard methods.
 
 You provided a workaround sure, i understand the usecase but this PR is more than just "crashing when restarting a scaleform" you failed to even read the PR, 
 
 This pull request from my understand of reading it is the following
 
Fixes slots and data when reloading a scaleform providing so a scaleform can be updated and reloaded while ingame fixing the issue of having to restart your game or renaming a scaleform to see that gfx be updated.
 
Your method to streaming scaleforms and restart a resource using them separately is fine sure but it's not what this PR is about.

Splitting up streaming resources is optional and it should be working but it's not and that's why your "FiveM industry-standard" is just not industry-standard.

I simply pointed this out and you took my opinions to heart and got all egotistical on me 🤷

QuadrupleTurbo:
> > especially when sharing established industry-standard methods.
> 
> You provided a workaround sure, i understand the usecase but this PR is more than just "crashing when restarting a scaleform" you failed to even read the PR,
> 
> This pull request from my understand of reading it is the following
> 
> Fixes slots and data when reloading a scaleform providing so a scaleform can be updated and reloaded while ingame fixing the issue of having to restart your game or renaming a scaleform to see that gfx be updated.
> 
> Your method to streaming scaleforms and restart a resource using them separately is fine sure but it's not what this PR is about.
> 
> Splitting up streaming resources is optional and it should be working but it's not and that's why your "FiveM industry-standard" is just not industry-standard.
> 
> I simply pointed this out and you took my opinions to heart and got all egotistical on me 🤷

I agree, we don't want workarounds, we want true fixes! yay

frostfire575:
is this PR ever going to be accepted? 🥲

frostfire575:
any news on this bug? is this PR ever going to be accepted? 🥲

manups4e:
just for knowledge, is this PR going to be considered?

radium-cfx (approved):

QuadrupleTurbo:
HUGE!!!

frostfire575:
any eta? need this fix ❤️

PR #3861, opened 2026-03-11
fix(onesync): prevent allocation of out-of-bounds network object ID 65535

### LINKED ISSUE #3849: Network IDs around 65534 have issues networking successfully
### What happened?

I've found that over a long period of time, this happens very rarely. I've caught it 4 times total. One was at 65533
and three were at 65534. For me it happens when creating an object on the server, then trying to use NetworkGetEntityIsNetworked() on the client for the netid. It always returns false. Unfortunately I can't confirm more than that, but I will update with more information when it happens again.

### Expected result

NetworkGetEntityIsNetworked() should return true

### Reproduction steps

On the server:
local e = CreateObjectNoOffset(mdl, coords, true, true, false)
local netid = NetworkGetNetworkIdFromEntity(e)

when netid is around 65534, then on the client:
(not 100% which returns false first but...)

NetworkDoesEntityExistWithNetworkId() and/or NetworkGetEntityIsNetworked() happen next, and if my suspicion is correct, NetworkGetEntityIsNetworked() actually fails after NetworkDoesEntityExistWithNetworkId() succeeds, but I have to create a test case to know.

I suspect the latter because I'm finding reports of objects with these netids meaning they exist, but return false when queried if their netid does exist, unless the id i'm getting from NetworkGetEntityFromNetworkId() is 0.

### Importancy

Slight inconvenience

### Area(s)

FiveM, OneSync

### Specific version(s)

FiveM server build 3407, linux and windows

### Additional information

If you need a test case I'm willing to write one.

gtolontop:
## Goal of this PR

Fix a bug where network object ID `65535` (`0xFFFF`) could be allocated to clients, leading to entities that can never be resolved by the server.

`GetFreeObjectIds()` used `m_objectIdsSent.size()` as the upper bound for the ID allocation loop. The bitset is sized `roundToWord(MaxObjectId)` = **65536**, but the entity lookup vector `m_entitiesById` is only sized `MaxObjectId` = **65535** (valid indices 0–65534).

This off-by-one mismatch allowed ID `65535` to be allocated in the bitset and handed out to clients. However:

1. `ProcessClonePacket` rejects `objectId == 0xFFFF` as a sentinel value (_"that's not an object ID, that's a snail!"_)
2. `GetEntity(65535)` returns empty due to its bounds check (`65535 >= 65535`)
3. Direct `m_entitiesById[65535]` accesses in unchecked paths are out-of-bounds UB

The fix caps the allocation loop at `MaxObjectId` instead of `m_objectIdsSent.size()`, ensuring all allocated IDs fall within the valid vector range (1–65534).

FiveM, OneSync

## Successfully tested on

- Code review confirms the off-by-one: bitset size (65536) > vector size (65535)
- `CreateEntityFromTree` already correctly caps at `MaxObjectId - 1` (65534)
- The fix aligns `GetFreeObjectIds` with the same upper bound

## Checklist

## Fixes issues

Fixes #3849

### DIFF
--- a/code/components/citizen-server-impl/src/state/ServerGameState.cpp
+++ b/code/components/citizen-server-impl/src/state/ServerGameState.cpp
@@ -4029,7 +4029,7 @@ void ServerGameState::GetFreeObjectIds(const fx::ClientSharedPtr& client, uint8_
 	{
 		bool hadId = false;
 
-		for (; id < m_objectIdsSent.size(); id++)
+		for (; id < static_cast<uint16_t>(MaxObjectId); id++)
 		{
 			if (!m_objectIdsSent.test(id) && !m_objectIdsUsed.test(id))
 			{


n3xuuu:
fix the commit please.

gtolontop:
I will check at my home idk why he does that

bartusek27:
> I will check at my home idk why he does that

You were building it for Linux maybe (where it won't argue these types of problems), you should be ok with casting `MaxObjectId` to the same type as `id`... so you want something like this:
`for (; id < static_cast<uint16_t>(MaxObjectId); id++)`

FabianTerhorst (approved):

gtolontop:
> > I will check at my home idk why he does that
> 
> You were building it for Linux maybe (where it won't argue these types of problems), you should be ok with casting `MaxObjectId` to the same type as `id`... so you want something like this: `for (; id < static_cast<uint16_t>(MaxObjectId); id++)`

Hey, this is a good idea you want I change this before the merge ? (@FabianTerhorst)

bartusek27:
I think it would fail the Win build if left like this. But maybe I am wrong :/

gtolontop:
> I think it would fail the Win build if left like this. But maybe I am wrong :/

Im with you haha, added ;)

bartusek27:
Ahh, maybe we need to reissue that checks? I think that build process is already fixed...

gtolontop:
> Ahh, maybe we need to reissue that checks? I think that build process is already fixed...

I don't know we will see if they merge it or if they ask another PR

PR #3866, opened 2026-03-12
fix(scripting/node): sandbox per-resource temp directory under resource path

### LINKED ISSUE #3856: Node Router get file : no device found
### What happened?

It is no longer possible to use screenshot-basic with requestClientScreenshot, for example, because I cannot access the image that is sent in POST.

### Expected result

Error: Access to this API has been restricted. Use -- allow-fs-write to manage permissions. Filesystem permission check from 'screenshot-basic' for permission fs.read on resource 'C:\Users\mathe\AppData\Local\Temp\upload_a955ca4254a78093bdb4409bB4c134c' - no device found

### Reproduction steps

Use : 
```js
exports['screenshot-basic'].requestClientScreenshot(
	source,
	{
		fileName: path.join(savePath, filename + '.png'),
		encoding: 'png',
		quality: 1.0,
	},
	async (err, fileName) => {
		let image = await imagejs.Image.load(fileName);
		const coppedImage = image.crop({ x: image.width / 4.5, width: image.height });
		image.data = coppedImage.data;
		image.width = coppedImage.width;
		image.height = coppedImage.height;
		for (let x = 0; x < image.width; x++) {
			for (let y = 0; y < image.height; y++) {
				const pixelArr = image.getPixelXY(x, y);
				const r = pixelArr[0];
				const g = pixelArr[1];
				const b = pixelArr[2];
				if (g > r + b) {
					image.setPixelXY(x, y, [255, 255, 255, 0]);
				}
			}
		}
		image.save(fileName);
	}
);```

I have already 
```cfg
add_unsafe_child_process_permission fivem-greenscreener
add_unsafe_worker_permission fivem-greenscreener

add_filesystem_permission fivem-greenscreener write fivem-greenscreener

add_unsafe_child_process_permission screenshot-basic
add_unsafe_worker_permission screenshot-basic

add_filesystem_permission screenshot-basic write screenshot-basic
```

### Importancy

Unknown

### Area(s)

FXServer

### Specific version(s)

Server 26389 Windows

### Additional information

https://github.com/citizenfx/screenshot-basic/blob/5e89d4afb3e28b490344e84b5e2b5ac8cc5d9c75/src/server/server.ts#L22

Image

gtolontop:
## Goal of this PR

Fix Node.js resources being unable to read or write files in the OS temp directory, breaking HTTP multipart file uploads via libraries like `koa-body`/`formidable`.

Instead of allowing access to the global OS temp directory, each Node.js resource now gets its own sandboxed temp folder at `<resource_path>/tmp/`.

**Key changes:**

1. **Per-resource temp path** (`NodeScriptRuntime`): the path `<resource_path>/tmp/` is computed at resource creation but the directory is **only created lazily** on the first call to `os.tmpdir()`, so resources that never use it don't get an unnecessary folder.

2. **Native binding** (`GenericFunctions.h`): a new `Citizen.getResourceTempPath()` function exposes the temp path to JS and creates the directory on first invocation via `std::filesystem::create_directories`.

3. **JS-side override** (`JavaScriptEnvironmentCode.h`): `os.tmpdir()` is patched to return the per-resource temp path, so libraries like `formidable`/`koa-body` automatically write uploads there without any user code change.

4. **Cleanup on unload** (`NodeScriptRuntime::Destroy`): the `tmp/` directory is removed when the resource stops.

This approach keeps the server fully sandboxed — no files are written outside the server directory — while remaining transparent to existing libraries that rely on `os.tmpdir()`.

FXServer, Scripting

## Successfully tested on

- Code review confirms the temp directory is created lazily (only on first `os.tmpdir()` call)
- The VFS sandbox already covers `<resource_path>/tmp/` so no extra permission bypass is needed
- Path traversal detection (`..`) still runs before any filesystem access
- Cleanup on resource unload removes the `tmp/` folder

## Checklist

## Fixes issues

Fixes #3856

### DIFF
--- a/code/components/citizen-scripting-node/include/JavaScriptEnvironmentCode.h
+++ b/code/components/citizen-scripting-node/include/JavaScriptEnvironmentCode.h
@@ -41,12 +41,13 @@ const result = m._compile(script, 'dummy-wrapper');
 
 global.require = m.exports.require;
 
-// redirect os.tmpdir() to a per-resource sandboxed temp directory
+// redirect os.tmpdir() to a per-resource sandboxed temp directory (created lazily on first call)
 const os = require('os');
-const _cfxResourceTempPath = Citizen.getResourceTempPath();
-if (_cfxResourceTempPath) {
-	os.tmpdir = () => _cfxResourceTempPath;
-}
+let _cfxTempPath;
+os.tmpdir = () => {
+	if (!_cfxTempPath) _cfxTempPath = Citizen.getResourceTempPath();
+	return _cfxTempPath;
+};
 
 // prevent process exit
 process.on('uncaughtException', () => {});
--- a/code/components/citizen-scripting-node/include/shared/GenericFunctions.h
+++ b/code/components/citizen-scripting-node/include/shared/GenericFunctions.h
@@ -263,6 +263,10 @@ namespace fx::v8shared
 		auto runtime = GetScriptRuntimeFromArgs<RuntimeType>(args);
 		const auto& tempPath = runtime->GetTempPath();
 
+		// lazily create the directory on first call
+		std::error_code ec;
+		std::filesystem::create_directories(tempPath, ec);
+
 		args.GetReturnValue().Set(v8::String::NewFromUtf8(args.GetIsolate(), tempPath.c_str(), v8::NewStringType::kNormal, tempPath.size()).ToLocalChecked());
 	}
 }
--- a/code/components/citizen-scripting-node/src/NodeScriptRuntime.cpp
+++ b/code/components/citizen-scripting-node/src/NodeScriptRuntime.cpp
@@ -355,16 +355,8 @@ result_t NodeScriptRuntime::Create(IScriptHost* host)
 
 	m_isMonitorRuntime = resourceManager->IsMonitor();
 
-	// create a per-resource temp directory inside the resource folder for sandboxed filesystem access (e.g. HTTP multipart uploads)
-	{
-		std::error_code ec;
-		auto resourceTempDir = std::filesystem::path(resource->GetPath()) / "tmp";
-		std::filesystem::create_directories(resourceTempDir, ec);
-		if (!ec)
-		{
-			m_tempDir = std::filesystem::absolute(resourceTempDir).string();
-		}
-	}
+	// per-resource temp directory path (created lazily when os.tmpdir() is called)
+	m_tempDir = std::filesystem::absolute(std::filesystem::path(resource->GetPath()) / "tmp").string();
 
 	// create our UV loop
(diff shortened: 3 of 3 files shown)


FabianTerhorst:
I think an api to redirect the paths to use a folder inside the resource folder instead of allowing the temp folder would be preferred.

gtolontop:
I will go to pc i will check if it can be better or not I keep you inform

gtolontop:
I start working on it I will publish it soon

gtolontop:
> I think an api to redirect the paths to use a folder inside the resource folder instead of allowing the temp folder would be preferred.

Hello I've reworked the approachinstead of allowing access to the entire OS temp directory, each resource now gets its own sandboxed temp folder under %TEMP%/cfx-<resourceName>/. I also patch os.tmpdir() on the JS side so libraries like formidable/koa-body automatically write there, and the folder is cleaned up when the resource unloads.

For the part about "a folder inside the resource folder" I went with the system temp dir instead since it feels more natural for temporary files (not tracked by VCS, auto-cleaned by the OS, etc.). Let me know if you'd prefer it to be under the resource path instead, happy to adjust :)

FabianTerhorst:
> > I think an api to redirect the paths to use a folder inside the resource folder instead of allowing the temp folder would be preferred.
> 
> Hello I've reworked the approachinstead of allowing access to the entire OS temp directory, each resource now gets its own sandboxed temp folder under %TEMP%/cfx-/. I also patch os.tmpdir() on the JS side so libraries like formidable/koa-body automatically write there, and the folder is cleaned up when the resource unloads.
> 
> For the part about "a folder inside the resource folder" I went with the system temp dir instead since it feels more natural for temporary files (not tracked by VCS, auto-cleaned by the OS, etc.). Let me know if you'd prefer it to be under the resource path instead, happy to adjust :)

Hi, thanks for the changes. Yes, under the resource path would be preferred. The server should not touch files outside the server directory anymore.

gtolontop:
Hello, reworked the approach again: the per-resource temp folder is now placed directly inside the resource directory (<resource_path>/tmp/) instead of under %TEMP%/cfx-<resourceName>/, so the server never touches files outside its own directory. The VFS sandbox already covers this path so no extra permission check is needed. os.tmpdir() is still patched on the JS side, and cleanup on resource unload is unchanged.

FabianTerhorst:
> Hello, reworked the approach again: the per-resource temp folder is now placed directly inside the resource directory (<resource_path>/tmp/) instead of under %TEMP%/cfx-/, so the server never touches files outside its own directory. The VFS sandbox already covers this path so no extra permission check is needed. os.tmpdir() is still patched on the JS side, and cleanup on resource unload is unchanged.

Found one minor issue when testing your change. You always create the temp folder now. Can you only create it when the tempdir function is called. So create the folder inside the method that returns the temp folder path, instead of caching the path with the arrow function.

gtolontop:
I see the problem, I will solve this just after school

PR #3882, opened 2026-03-23
fix(server/net): make sure we don't overwrite sv_proxyIPRanges in the case that we use `set sv_proxyIPRanges [value]`

AvarianKnight:
### Goal of this PR
Fixes `set sv_proxyIPRanges` overwriting the `sv_proxyIPRanges`  making it so that you can never change the `sv_proxyIPRanges` until you restart the server.

Server

### Successfully tested on

**Platforms:** Windows

### Checklist


Simon2122:
Would be nice to have this added.

PR #3884, opened 2026-03-24
fix(server): prevent potential null dereference in ClientRegistry

dougcostadev:
### Goal of this PR

Prevent a potential crash caused by dereferencing an invalid `weak_ptr` in `ClientRegistry`.

---

This PR ensures that `weakClient.lock()` is validated before being dereferenced in the `OnAssignTcpEndPoint` callback.

Since this callback is event-driven and may execute after the client has been destroyed, directly dereferencing the result of `lock()` could lead to undefined behavior or a crash.

This change aligns the implementation with other callbacks in the same file where `weak_ptr` is safely validated before use.

A possible scenario is during rapid client disconnect/reconnect cycles, where the callback executes after the client object has already been destroyed.

No functional behavior is changed.

---

Server

---

### Successfully tested on

**Game builds:** N/A

**Platforms:** Windows

---

### Checklist

---

### Fixes issues

N/A

### DIFF
--- a/code/components/citizen-server-impl/src/ClientRegistry.cpp
+++ b/code/components/citizen-server-impl/src/ClientRegistry.cpp
@@ -135,7 +135,10 @@ namespace fx
 
 		client->OnAssignTcpEndPoint.Connect([this, weakClient]()
 		{
-			m_clientsByTcpEndPoint[weakClient.lock()->GetTcpEndPoint()] = weakClient;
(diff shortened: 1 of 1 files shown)


AvarianKnight:
Can you show code where this can even theoretically happen?

This already relies on the client being valid (it's in a callback client on the client!), so this PR doesn't make very much sense to me

PR #3885, opened 2026-03-24
fix(server/state): use signed integer for weaponType in weaponDamageEvent

### LINKED ISSUE #3827: Error in weapon hashes in the event: weaponDamageEvent
### What happened?

When we debug the **weaponDamageEvent** event, when printing the supposed weapon hash (data.weaponType), one result should appear, but another appears instead.
For example, with the weapon_rpg weapon, when I shoot at another player, the hash result will be 2982836145, but if I use runcode and print(GetHashKey(‘weapon_rpg’)) the result will be -1312131151.
Now, if I debug the startProjectileEvent event and print the supposed weapon hash (data.weaponHash), the result is as expected: -1312131151.

### Expected result

Return the correct hash in the weaponDamageEvent event and globalise the hashes

### Reproduction steps

1 - Have 2 clients open 
2 - Debug the weaponDamageEvent and startProjectileEvent events

```
AddEventHandler(‘weaponDamageEvent’, function(sender, data)
    print(‘weaponDamageEvent: ’ ..data.weaponType)
end)

AddEventHandler(“startProjectileEvent”, function(sender, data)
    print(‘startProjectileEvent: ’ .. data.weaponHash)
end)
```

3 - On client 1, shoot at client 2 with an RPG or other weapon
4 - Observe and save the weapon hashes.
5 - Use the runcode and execute 

`print(GetHashKey(‘weapon_rpg’))`

6 - Observe and compare the results

Results: 

weaponDamageEvent: 2982836145 
startProjectileEvent: -1312131151
runcode: -1312131151

It can be seen that the hashes are not the same.

Image

Image

Image

Image

### Importancy

Slight inconvenience

### Area(s)

FiveM, OneSync, ScRT: Lua

### Specific version(s)

Fivem/Latest/3570 - Server/Windows/25770

### Additional information

_No response_

DebugPadawan:
Closes #3827. The weaponType in CWeaponDamageEvent was being serialized as a uint32_t, causing a signed/unsigned mismatch in Lua where hashes are expected to be signed 32-bit integers.

### DIFF
--- a/code/components/citizen-server-impl/src/state/ServerGameState.cpp
+++ b/code/components/citizen-server-impl/src/state/ServerGameState.cpp
@@ -5046,7 +5046,7 @@ struct CWeaponDamageEvent
 	}
 
 	uint8_t damageType;
-	uint32_t weaponType; // weaponHash
+	int32_t weaponType; // weaponHash
 
 	bool overrideDefaultDamage;
 	bool hitEntityWeapon;
@@ -6308,7 +6308,7 @@ struct CWeaponDamageEvent
 	}
 
 	uint8_t damageType;
-	uint32_t weaponType; // weaponHash
+	int32_t weaponType; // weaponHash
 	uint32_t f92;
 
 	bool overrideDefaultDamage;


FabianTerhorst:
This would break backward compatibility.

DebugPadawan:
I understand that changing the internal network struct breaks backwards compatibility for existing servers relying on the unsigned `weaponType` in `weaponDamageEvent`. It seems the better path is to let the user script handle the conversion (e.g. `data.weaponType & 0xFFFFFFFF` or `data.weaponType | 0` in Lua depending on use case) rather than altering the core C++ definition. Closing this PR. Thanks for the quick review!

PR #3914, opened 2026-04-09
fix(devtools-five): guard hook behind debug & removed ComputeSceneNodeIndex

n3xuuu:
### Goal of this PR

Remove unnecessary runtime debug logging from `ComputeSceneNodeIndex` to reduce console spam and avoid potential performance overhead. Debug output is retained only when `_DEBUG` builds are active.

- Removes the `trace` call.
- Wraps debug-related code with #ifdef `_DEBUG` to ensure it only runs in debug builds.

FiveM

### Successfully tested on

**Game builds:** -

**Platforms:** Windows, Linux

### Checklist

### Fixes issues

{722724C3-7AB7-4950-B946-AF4DE78186E1}

### DIFF
--- a/code/components/devtools-five/src/SceneGraphPoolsDebug.cpp
+++ b/code/components/devtools-five/src/SceneGraphPoolsDebug.cpp
@@ -261,17 +261,8 @@ static void FwBasePool(rage::fwBasePool* pool, int poolSize, uint8_t* storage, u
 	g_origFwBasePool(pool, poolSize, storage, scratchBuffer, name, classSize, flags);
 }
 
-static int16_t (*g_origComputeSceneNodeIndex)(const void* sceneNode);
-static int16_t ComputeSceneNodeIndex(const void* sceneNode)
-{
-	auto result = g_origComputeSceneNodeIndex(sceneNode);
-	trace("Scene Node Index: %d\n", result);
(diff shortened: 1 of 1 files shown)


code/components/devtools-five/src/SceneGraphPoolsDebug.cpp
DaniGP17: You can remove this trampoline and the full ComputeSceneNodeIndex function

DaniGP17, line 341: Could you set this to only add for debug build?

n3xuuu: Done. Could you take a quick look and confirm it’s okay? I’ll also squash the commits right away

DaniGP17: It looks good, if you know how to add the scene graph pools to the pool list of PoolManagement.cpp you can do, if no ignore it.

n3xuuu: I will try, but I can't promise anything 😅 In any case, I'll attempt it in another pr to avoid causing issues here.

Issue #3843, opened 2026-02-27
Outdated CEF causes backdrop-filter blur flickering in NUI


Osheun:
### What happened?

When using CSS backdrop-filter: blur() inside a NUI, the blur effect flickers continuously in-game.
The UI appears to constantly repaint, causing visible instability and making modern UI blur effects unusable.

This happens consistently across servers and different UI implementations.

### Expected result

backdrop-filter should render a stable blur effect without flickering or constant repainting. Modern CSS visual effects should behave similarly to standard Chromium behavior.

### Reproduction steps

Create a simple NUI with:

backdrop-filter: blur(10px);

Open the UI in FiveM

Observe blur flickering and unstable rendering

### Importancy

Slight inconvenience

### Area(s)

FiveM

### Specific version(s)

 Embedded CEF: cfx-m103 (Chromium 103)

### Additional information

CEF appears significantly outdated compared to current Chromium versions.
Developers cannot update or replace the embedded CEF themselves, as it is tightly integrated into FiveM.

Is there any plan to update the CEF version or address the rendering instability?

PR #3917, opened 2026-04-11
Fix NUI backdrop-filter flicker

VeinDevTtv:
### Goal of this PR

Fix or mitigate NUI `backdrop-filter: blur()` flickering in normal resource NUI.

Normal resource NUI is rendered through the root raw NUI window and CEF shared-texture/compositor path. The previous shader approach affected DUI/runtime-texture windows only, so it has been reverted.

This PR changes CEF GPU compositor defaults for normal NUI:

- `disable-gpu-driver-bug-workarounds` is no longer forced by default.
- `enable-gpu-rasterization` is no longer forced by default.
- The old behavior remains available through convars:
  - `nui_disableGpuDriverBugWorkarounds`
  - `nui_useGpuRasterization`

This keeps the risky compositor settings available for diagnostics while avoiding them by default for NUI rendering.

FiveM, NUI

### Successfully tested on

**Game builds:** Not runtime-tested locally.

**Platforms:** Windows

Validation performed:
- Traced the normal NUI and DUI render paths.
- Confirmed normal NUI does not use the DUI `quadPS` shader path.
- Restored the DUI shader behavior.
- Confirmed the PR now only changes `code/components/nui-core/src/NUIApp.cpp`.
- Ran `git diff --check origin/master...HEAD`.

### Runtime validation needed

Test a normal resource NUI repro using:

```css
backdrop-filter: blur(10px);
```

Then compare:
- default settings
- `nui_useGpuRasterization 1`
- `nui_disableGpuDriverBugWorkarounds 1`
- both enabled
- `nui_useSharedResources 0`

### Fixes issues

Related to #3843

### DIFF
--- a/code/components/nui-core/src/NUIApp.cpp
+++ b/code/components/nui-core/src/NUIApp.cpp
@@ -169,6 +169,7 @@ void NUIApp::OnContextReleased(CefRefPtr<CefBrowser> browser, CefRefPtr<CefFrame
 void NUIApp::OnBeforeCommandLineProcessing(const CefString& process_type, CefRefPtr<CefCommandLine> command_line)
 {
 	static ConVar<bool> nuiUseInProcessGpu("nui_useInProcessGpu", ConVar_Archive, true);
+	static ConVar<bool> nuiDisableGpuDriverBugWorkarounds("nui_disableGpuDriverBugWorkarounds", ConVar_Archive, false);
 
 	static std::string defaultUiUrl = "https://nui-game-internal/ui/app/index.html";
 	static ConVar<std::string> uiUrlVar("ui_url", ConVar_None, defaultUiUrl);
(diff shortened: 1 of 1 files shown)


VeinDevTtv:
@Ehbw Do you have any comment?

Ehbw:
These shaders only touch DUI Windows and will have zero impact on NUI. Not only that this logic will negatively impact DUIs default behaviour

VeinDevTtv:
Updated the branch to pivot away from the DUI shader path.

After tracing the call path, the previous `quadPS` change only affected non-raw `NUIWindow`s, which are used by DUI/runtime-texture windows. Normal resource NUI goes through the root raw NUI window and CEF shared-texture/compositor path, so that shader change would not fix #3843 and could affect DUI behavior.

The PR now restores the DUI shader behavior and changes normal NUI CEF startup defaults instead:

- `disable-gpu-driver-bug-workarounds` is no longer forced by default; it is now behind `nui_disableGpuDriverBugWorkarounds`.
- `enable-gpu-rasterization` is no longer forced by default; it is now behind `nui_useGpuRasterization`.

Local validation performed:
- `git diff --check origin/master...HEAD` passed.
- Confirmed the PR now only changes `code/components/nui-core/src/NUIApp.cpp`.

I could not run a local client build/runtime repro in this environment because PowerShell 7, a generated `CitizenMP.sln`, and `msbuild` are not available. The next runtime check should be the #3843 normal NUI `backdrop-filter: blur(10px)` repro with default settings, then toggling the two new convars and `nui_useSharedResources 0` as diagnostics.

radium-cfx:
Thanks for your PR. Unfortunately, the proposed changes do not address the reported issue and would introduce several unintended side effects, so we’re unable to merge this as-is.

Regarding your mention of "needed additional validation": we genuinely appreciate contributions aimed at improving the platform. However, please ensure that any proposed changes are properly tested and actually resolve the targeted issue before submission. The same applies when using AI-assisted solutions: these should be carefully reviewed and validated, as your current proposal is based on wrong information (provided by the used AI Agent).

Issue #2209, opened 2023-09-21
[Vulnerability] Path Traversal


thewhistledev:
https://github.com/citizenfx/fivem/blob/4a1dc4c3d0325601d55f2cc97d2c47cdc212708b/ext/sdk/resources/sdk-root/shell/src/backend/fxcode/fxcode-service.ts#L71C9-L71C37

Unsanitized input from the request URL flows into sendFile, which is used as a path. This may result in a Path Traversal vulnerability, allowing an attacker to read arbitrary files.

PR #3920, opened 2026-04-11
fix(fxdk): prevent path traversal in remote resources

VeinDevTtv:
## Summary

Fixes the FxDK `/vscode-remote-resource` path traversal reported in #2209.

The shell backend previously passed the user-controlled `path` query through `getRemoteResourcePath()` and directly into `res.sendFile()`. Since `sendFile()` accepts absolute paths, a request such as `?path=C:/Windows/win.ini` could read files outside the FxDK install.

## Changes

- Resolve and validate the FxCode resource path before serving it.
- Reject empty resource paths.
- Reject any resource path that resolves outside `sdkRootFXCode` using `path.relative()` and an absolute-path check.
- Return HTTP 400 for invalid paths instead of passing them to Express `sendFile()`.

This keeps the remote resource endpoint limited to the same SDK root already exposed by `/fxcode-static`, while blocking arbitrary absolute-path reads and `..` traversal out of the FxCode root.

## Validation

- Reviewed the original issue and local path traversal analysis.
- Verified the committed diff is limited to `ext/sdk/resources/sdk-root/shell/src/backend/fxcode/fxcode-service.ts`.
- Checked path-boundary behavior for in-root FxCode paths, sibling-prefix paths, Windows system paths, and `..` traversal.

Full package build was not run because this checkout does not have `node_modules`, and `yarn` is not installed locally.

### DIFF
--- a/ext/sdk/resources/sdk-root/shell/src/backend/fxcode/fxcode-service.ts
+++ b/ext/sdk/resources/sdk-root/shell/src/backend/fxcode/fxcode-service.ts
@@ -9,6 +9,7 @@ import { concurrently } from "utils/concurrently";
 import { ShellBackend } from 'backend/shell-backend';
 import { URL } from 'url';
 import { NotificationService } from "backend/notification/notification-service";
+import * as path from 'path';
 
 interface FXCode {
   getRootPath(): string;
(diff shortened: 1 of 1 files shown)


thewhistledev:
Request for update on this.

PR #3934, opened 2026-04-16
feat(citizen-server-impl): add prevent_restart metadata to protect resources from console commands

opastorello:
### Goal of this PR

Allow resource authors to protect their resources from being stopped, started, restarted, or re-ensured via server console commands. This is useful for critical resources such as anti-cheat systems, core framework scripts, or any resource that should remain running without interference from console access.

A new `prevent_restart` metadata key is introduced. When set in `fxmanifest.lua`, the `start`, `stop`, `restart`, and `ensure` console commands will refuse to operate on that resource and emit a warning trace instead.

The implementation follows the existing pattern used by `server_only` metadata, reading entries via `ResourceMetaDataComponent::GetEntries`.

**Usage in `fxmanifest.lua`:**
```lua
fx_version 'cerulean'
game 'gta5'

prevent_restart 'yes'
```

**Console output when blocked:**
```
^3Resource my_anticheat has prevent_restart set and cannot be managed via console commands.^7
```

Note: this only blocks console commands. Programmatic lifecycle control via resource events is unaffected.

Server

### Successfully tested on

**Platforms:** Windows

### Checklist

- [x] The code follows the existing patterns in the codebase (`server_only` metadata pattern).
- [x] No breaking changes to existing resources that do not use this metadata.

### DIFF
--- a/code/components/citizen-server-impl/src/ServerResources.cpp
+++ b/code/components/citizen-server-impl/src/ServerResources.cpp
@@ -670,6 +670,21 @@ static InitFunction initFunction([]()
 			return resources;
 		};
 
+		static auto isPreventRestart = [resman](const std::string& resourceName) -> bool
+		{
+			auto resource = resman->GetResource(resourceName);
+
+			if (!resource.GetRef())
+			{
+				return false;
+			}
+
+			auto metaData = resource->GetComponent<fx::ResourceMetaDataComponent>();
+			auto iv = metaData->GetEntries("prevent_restart");
+
+			return iv.begin() != iv.end();
+		};
+
 		static auto commandRef = instance->AddCommand("start", [=](const std::string& resourceName)
 		{
 			if (resourceName.empty())
@@ -693,6 +708,12 @@ static InitFunction initFunction([]()
 			if (!resource.GetRef())
 			{
 				trace("^3Couldn't find resource %s.^7\n", resourceName);
+				return;
+			}
+
+			if (isPreventRestart(resourceName))
+			{
+				trace("^3Resource %s has prevent_restart set and cannot be started via console commands.^7\n", resourceName);
 				return;
 			}
 
@@ -732,6 +753,12 @@ static InitFunction initFunction([]()
 				return;
 			}
 
+			if (isPreventRestart(resourceName))
+			{
+				trace("^3Resource %s has prevent_restart set and cannot be stopped via console commands.^7\n", resourceName);
+				return;
+			}
+
 			if (!resource->Stop())
 			{
 				if (resource->GetState() != fx::ResourceState::Stopped)
@@ -749,6 +776,12 @@ static InitFunction initFunction([]()
 			if (!resource.GetRef())
 			{
 				trace("^3Couldn't find resource %s.^7\n", resourceName);
+				return;
+			}
+
+			if (isPreventRestart(resourceName))
+			{
+				trace("^3Resource %s has prevent_restart set and cannot be restarted via console commands.^7\n", resourceName);
 				return;
 			}
 
@@ -789,6 +822,12 @@ static InitFunction initFunction([]()
 				if (!resource.GetRef())
 				{
 					trace("^3Couldn't find resource %s.^7\n", resourceName);
+					return false;
+				}
+
+				if (isPreventRestart(resourceName))
+				{
+					trace("^3Resource %s has prevent_restart set and cannot be managed via console commands.^7\n", resourceName);
 					return false;
 				}
 


AvarianKnight:
I'm really confused on the use case here, resources shouldn't be able to restrict their ability to be restarted.

opastorello:
> I'm really confused on the use case here, resources shouldn't be able to restrict their ability to be restarted.

Thank you for the feedback.

The use case is server operators who want to protect critical resources (anti-cheat systems, core framework scripts) from being accidentally or maliciously stopped/restarted via console access.

A common scenario: a server has multiple admin levels. High-trust admins have console access, but the server owner wants to ensure that certain resources (e.g. an anti-cheat or a session manager) cannot be taken down by any console command, even by mistake.

Currently there is no way to declare a resource as protected from console lifecycle commands. The `g_managedResources` set already does this for a hardcoded list of system resources, but there is no way for server operators to extend this protection to their own resources.

If the concern is that a resource should not control its own lifecycle restrictions, an alternative approach would be to expose this via a server.cfg convar instead:

```
sv_protectedResources "my_anticheat,core_resource"
```

This way the decision belongs entirely to the server operator, not the resource author. Happy to rework the PR in that direction if that better fits the project's design philosophy.

dug:
If bad actor has console access it's already too late useless feat

opastorello:
> If bad actor has console access it's already too late useless feat

The threat model here is not a bad actor with console access — it's a DDoS attack scenario.

**Real-world use case:**

I have clients running FiveM servers that receive DDoS attacks. When the attack hits, the server itself stays up and players remain connected — the server handles it. However, if during the attack window a resource `ensure` is triggered (either manually by an admin or by an automated script), **all players drop**. The `ensure` causes a resource restart which kicks every connected player, even though the server would have otherwise survived the attack with everyone still in.

The idea is to have a resource that polls an external API endpoint which monitors the server's IP for DDoS activity. When an attack is detected, that resource would set itself (and other critical resources) as "unrestartable" via this metadata, so that no console command — intentional or accidental — can trigger an `ensure`/`restart` during the attack window.

This is not about protecting from a compromised machine. It's about protecting player sessions from being needlessly dropped during network-layer attacks where the server is still functional.

The `sv_protectedResources` convar approach I mentioned previously could also work for this, since the resource could toggle the convar dynamically via a server command or event when the attack is detected.

Happy to discuss the best approach to implement this safely.

Mathu-lmn:
Or, hear me out, remove any resource triggering random ensure during runtime 👍🏻

opastorello:
> Or, hear me out, remove any resource triggering random ensure during runtime 👍🏻

Fair point in an ideal setup — but in practice, large servers have multiple admins, automated health check scripts, and third-party tools that can all trigger `ensure`. You can't always audit or lock down every vector, especially during an incident when people are actively trying to "fix" things.

The goal isn't to block malicious actors — it's to make critical resources resilient to well-intentioned but harmful actions during specific time windows (DDoS attack, maintenance mode, etc.). A convar-based approach like `sv_protectedResources` would keep that control entirely on the server operator side, no resource self-governance involved.

Mathu-lmn:
In practice, even big servers control what they host on their hardware or hardware they rent 😄 
Also let's stop the AI to discuss please, I feel like I'm talking to Claude rn

dug:
Resources shouldn't be restarted / stopped in a production environment. What if a resource failed to start or caused script errors with exports that depended on that resource? Also, your idea for the fix seems to me stupid. Why don't you just have an export that sets a state in whatever resources auto-restart to just not restart whenever the state is true? And do admins need direct console access if they can’t be relied on to not trigger random commands? Why not make a Discord bot or admin panel for them to conduct their administrator duties?

xalva98:
I can see why this could be useful:

Sometimes you have multiple live consoles open, production server & test server and sometimes you just start/stop ressources on the wrong browser tab so this could technically be "useful" but then you'd also have to somehow detect in what environment you are. Other than that I think this is pretty pointless.

dug:
> I can see why this could be useful:
> 
> 
> 
> Sometimes you have multiple live consoles open, production server & test server and sometimes you just start/stop ressources on the wrong browser tab so this could technically be "useful" but then you'd also have to somehow detect in what environment you are. Other than that I think this is pretty pointless.

Why don't you just hide the live production server console?

ook3D:
> > Or, hear me out, remove any resource triggering random ensure during runtime 👍🏻
> 
> Fair point in an ideal setup — but in practice, large servers have multiple admins, automated health check scripts, and third-party tools that can all trigger `ensure`. You can't always audit or lock down every vector, especially during an incident when people are actively trying to "fix" things.
> 
> The goal isn't to block malicious actors — it's to make critical resources resilient to well-intentioned but harmful actions during specific time windows (DDoS attack, maintenance mode, etc.). A convar-based approach like `sv_protectedResources` would keep that control entirely on the server operator side, no resource self-governance involved.

No serious server starts and stops resources in a live production environment, your resources are fundementally designed wrong if this is the case

opastorello:
bro i own the servers, i know what runs on them. the point is during a ddos window even a legit ensure from a monitoring script drops everyone. not a design flaw, just an edge case. chill

ook3D:
> bro i own the servers, i know what runs on them. the point is during a ddos window even a legit ensure from a monitoring script drops everyone. not a design flaw, just an edge case. chill

its a design flaw. if you properly design your resources you wouldnt need to do such a thing

opastorello:
gonna handle it locally, thanks for nothing

dug:
> bro i own the servers, i know what runs on them. the point is during a ddos window even a legit ensure from a monitoring script drops everyone. not a design flaw, just an edge case. chill

What's the use case for automatically restarting resources? Can they not be cleaned up themselves and start the logic again?

spacevx:
> I can see why this could be useful:
> 
> Sometimes you have multiple live consoles open, production server & test server and sometimes you just start/stop ressources on the wrong browser tab so this could technically be "useful" but then you'd also have to somehow detect in what environment you are. Other than that I think this is pretty pointless.

If your using txAdmin, configure your production txAdmin in white mode, and the dev one in black

taoletsgo:
bro is just opposing for the sake of opposing. Most server owners have 0 coding experience. Since js / lua sandbox has already been designed for them, I think this PR is also a way to protect these noobs 👀

opastorello:
> its a design flaw. if you properly design your resources you wouldnt need to do such a thing

the resource itself isn't doing anything wrong. the issue is external health monitors triggering ensure to keep the resource up — you can't control that when you have multiple admin levels in production

opastorello:
> What's the use case for automatically restarting resources? Can they not be cleaned up themselves and start the logic again?

external monitoring scripts that ensure the resource is running. and no, cleaning up internal state doesn't fix it because the ensure drops players during the restart anyway

opastorello:
> bro is just opposing for the sake of opposing. Most server owners have 0 coding experience. Since js / lua sandbox has already been designed for them, I think this PR is also a way to protect these noobs 👀

exactly, that's literally the point. not every server owner is a dev and even the ones who are can't audit every script and every admin action 24/7

ook3D:
if you want my honest opinion, you shouldnt be a server owner if you dont understand the code running on your server :)

opastorello:
> if you want my honest opinion, you shouldnt be a server owner if you dont understand the code running on your server :)

and by that logic half the fivem ecosystem shouldn't exist lol. txAdmin, cfx's own permission system, all of it exists because not everyone is a dev. that's literally the target audience

opastorello:
> if you want my honest opinion, you shouldnt be a server owner if you dont understand the code running on your server :)

most server owners hire devs to build their resources, they're in the business of hosting, not coding — that's literally how this ecosystem works. i came up with the idea because i wanted a resource that polls my cloudflare API to detect ddos and lock things down. tried everything to block ensure from within the resource, best i could do was auto-restart 3 seconds after being stopped — which is pointless, the damage is already done. turns out fivem just doesn't expose a way to block the command entirely. that's the gap this PR was trying to fill. not sure why a small quality-of-life addition to the codebase triggered this much pushback

FabianTerhorst:
Can't you print a error message inside the stop event of your resource to tell the user it should not be stopped?

opastorello:
> Can't you print a error message inside the stop event of your resource to tell the user it should not be stopped?

the stop event fires after the resource is already stopping, at that point players are already being dropped and the state is unloading. there's no way to cancel the stop from inside onResourceStop, the command has already committed by then. that's exactly why i opened this PR, there's no hook early enough in the lifecycle to block it

PR #3940, opened 2026-04-20
feat(extra-natives-rdr3): add weapons no auto reload

DaniGP17:
### Goal of this PR
The `SET_CLIENT_CONFIG_BOOL` native have the `WeaponsNoAutoReload ` option, but I forgot to add the logic in #3656.

RedM

### Successfully tested on
**Game builds:** 1491
**Platforms:** Windows

### Checklist

### DIFF
--- a/code/components/extra-natives-rdr3/src/WeaponExtraNatives.cpp
+++ b/code/components/extra-natives-rdr3/src/WeaponExtraNatives.cpp
@@ -1,0 +1,64 @@
+#include "StdInc.h"
+
+#include "ClientConfig.h"
+#include "Hooking.h"
+#include "Hooking.Stubs.h"
+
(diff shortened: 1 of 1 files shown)


forzayt:
meme

forzayt:
This ones good but you know we will just end up holding an empty gun that won’t fire and won’t reload automatically, feel like stucked ..... we want to pair with some other to this right ??

outsider31000:
> This ones good but you know we will just end up holding an empty gun that won’t fire and won’t reload automatically, feel like stucked ..... we want to pair with some other to this right ??

That's the point of manual reload, you must press a key to reload, and this is an optional behaviour that's why its a config bool.
by default it's still automatic reload

PR #3944, opened 2026-04-21
fix: null-dereference in local manifest itydependencies fixup

### LINKED ISSUE #3943: Null-deref in LoadManifest itypDependencies fixup path (zero-init DataFile → UnloadDataFile AV)
## Summary

Client crashes with `c0000005` inside `LoadManifest` when a streamed resource's `_manifest.ymf` contains `itypDependencies` whose names hash-collide with entries in `g_itypHashList`. The fixup loop at [`code/components/gta-streaming-five/src/LoadStreamingFile.cpp:2292-2306`](https://github.com/citizenfx/fivem/blob/master/code/components/gta-streaming-five/src/LoadStreamingFile.cpp#L2292-L2306) calls `UnloadDataFile` on a zero-initialised `CDataFileMgr::DataFile` where only the `name` field is set — the native `DLC_ITYP_REQUEST` unloader then dereferences unset fields and faults.

Symmetric site on the removal path was hardened in commit [`aa9b332`](https://github.com/citizenfx/fivem/commit/aa9b332) (`fix(streaming/five): avoid using GetEntryByName`), but this manifest-fixup site was missed.

## Crash site

```cpp
for (auto& dep : manifestChunk->itypDependencies)
{
    if (auto it = g_itypHashList.find(dep.itypName); it != g_itypHashList.end())
    {
        auto name = fmt::sprintf("dummy/%s.ityp", it->second);
        trace("Fixing manifest-required #typ dependency for %s\n", name);

        auto mounter = LookupDataFileMounter("DLC_ITYP_REQUEST");

        CDataFileMgr::DataFile entry = { 0 };   // zero-initialised
        strcpy_s(entry.name, name.c_str());     // only name is populated
        mounter->UnloadDataFile(&entry);        // native unloader derefs unset fields → AV
    }
}
```

## Crash signature

- Exception: `c0000005` (access violation) at `fivem.exe+169FFBE`
- Phase: `CFX_PSEUDO_CACHE <resource>` streaming
- Immediately preceded in `CfxCrashDump` logs by: `Fixing manifest-required #typ dependency for dummy/<name>.ityp`
- Client build seen: `master v1.0.0.25770`

## Repro shape

A streamed resource with:
1. An MLO + its interior-proxy ytyp loaded via `DLC_ITYP_REQUEST`,
2. A `_manifest.ymf` whose `itypDependencies` list an ityp name whose joaat collides with an entry already in `g_itypHashList`,
3. Approached by a client so the fixup loop runs.

We've reproduced this with a production Quasar resource (`qs_royale_restaurant`). Other Quasar MLOs with apparently the same pattern (`qs_cottage_park`, `Bilingsgate`, `smokey_motel`) do not crash — likely a specific dependency-graph shape in the native unloader's internal state. Happy to share sanitised ytyps/manifest if helpful.

## Suggested fix

Mirror the `aa9b332` defensive pattern here — either populate the `DataFile` entry from the real manifest entry before calling `UnloadDataFile`, or guard the native unloader against zero-initialised input.

## Workaround (for reference)

Downstream (SlurpNet) we ship a 0-byte `_manifest.ymf` so the `ForcedDevice.Read(16)` check at line 2240 fails and the fixup loop is skipped entirely. MLO still loads via `DLC_ITYP_REQUEST`; only interior-proxy LOD metadata is lost. See [slurpnet-fivem PR #83](https://github.com/zachyzissou/slurpnet-fivem/pull/83).

forzayt:
### Goal of this PR

fixes https://github.com/citizenfx/fivem/issues/3943  in LoadManifest while processing itypDependencies a streamed resource's manifest contained #typ entries that hash-collided 

field with DataFileType::DLC_ITYP_REQUEST before calling UnloadDataFile this field to correctly identify the resource type , 

FiveM, RedM, Streaming

### Successfully tested on

**Game builds:** 

2545+ (GTA V), 1436.28 (RDR3)

**Platforms:** 

Windows

### Checklist

### Fixes issues

https://github.com/citizenfx/fivem/issues/3943

### DIFF
--- a/code/components/gta-streaming-five/src/LoadStreamingFile.cpp
+++ b/code/components/gta-streaming-five/src/LoadStreamingFile.cpp
@@ -2300,8 +2300,15 @@ void LoadManifest(const char* tagName)
 
 				CDataFileMgr::DataFile entry = { 0 };
 				strcpy_s(entry.name, name.c_str());
+				entry.type = DataFileType::DLC_ITYP_REQUEST;
 
-				mounter->UnloadDataFile(&entry);
+				__try
+				{
+					mounter->UnloadDataFile(&entry);
+				}
+				__except (FilterUnmountOperation(entry))
+				{
+				}
 			}
 		}
 
@@ -2344,8 +2351,15 @@ void LoadManifest(const char* tagName)
 
 					CDataFileMgr::DataFile entry = { 0 };
 					strcpy_s(entry.name, name.c_str());
+					entry.type = DataFileType::DLC_ITYP_REQUEST;
 
-					mounter->UnloadDataFile(&entry);
+					__try
+					{
+						mounter->UnloadDataFile(&entry);
+					}
+					__except (FilterUnmountOperation(entry))
+					{
+					}
 				}
 			}
 		}


Ehbw:
This doesn't fix or change anything. UnloadDataFile for DLC_ITYP_REQUEST's file mounter only ever reads the name and doesn't care about the values of the other fields. and without any further information from the linked issue its hard to say if the exception handling would make any difference

forzayt:
Oh thanks for the update , lemme see is there anything more i can collect from issue

PR #3951, opened 2026-04-27
feat(gta-core-five): add more vehicle pools

### LINKED ISSUE #3883: Vehicle pool increase or more aggressive culling
### What happened?

We have persistent vehicles for all players spawning via `CreateVehicleServerSetter`.
In some regions, culling does not seem to work as expected or has too wide a range, causing the `CNetObjVehicle` pool to most likely exceed its max value of 250 vehicles.

Creating new vehicles has no effect unless another vehicle is removed from the pool. When moving through the map, certain vehicles frequently fail to render for the player. The client crashes almost every time as well.

I expected the pool to prioritize vehicles that are closer over those sitting outside the visible range, 300+ units away.
Is there a way to increase the vehicle pool size or adjust culling values to improve the experience with persistent vehicles?

### Expected result

Near vehicles have higher priority to render

### Reproduction steps

Spawn more than 250 vehicles in a row and move.

### Importancy

Slight inconvenience

### Area(s)

FiveM

### Specific version(s)

FiveM latest build, Linux

### Additional information

_No response_

Bentix-cs:
### Goal of this PR

Add more Vehicle pools to fivem for allowing server owners to adjust these pools, because on large servers a full vehicle pool can cause invisible vehicles and client crashes.

I also suggest adding the pools `CNetObjVehicle` and `CVehicleSyncData` to the https://content.cfx.re/mirrors/client/pool-size-limits/fivem.json and set the max allowed limit for `CNetObjVehicle` to 500 and for `CVehicleSyncData` to 510. 

Added the pools to the poolEntriesTable.

FiveM

### Successfully tested on

**Game builds:** 3751

**Platforms:** Windows

### Checklist

### Fixes issues

resolves #3883

### DIFF
--- a/code/components/gta-core-five/src/PoolManagement.cpp
+++ b/code/components/gta-core-five/src/PoolManagement.cpp
@@ -166,6 +166,8 @@ static const char* poolEntriesTable[] = {
 	"CTask",
 	"OcclusionPathNode",
 	"OcclusionPortalInfo",
+	"CNetObjVehicle",
+	"CVehicleSyncData",
 #include "gta_vtables.h"
 	"Decorator",
 	"StreamPed req data",


Ehbw:
These two fields are already present in [gta_vtables.h](https://github.com/citizenfx/fivem/blob/afc95ef9e1ec91e08f2d22c320ed2d8f2a4ac9a7/code/components/gta-core-five/include/gta_vtables.h#L4742).

Bentix-cs:
> These two fields are already present in [gta_vtables.h](https://github.com/citizenfx/fivem/blob/afc95ef9e1ec91e08f2d22c320ed2d8f2a4ac9a7/code/components/gta-core-five/include/gta_vtables.h#L4742).

Yes but it's currently not possible to raise the limits, even with the `set moo 31337` convar enabled.

AvarianKnight:
Aren't these pools hard coded? RedM had to do custom logic to allow for these to be increase.

see: https://github.com/citizenfx/fivem/blob/afc95ef9e1ec91e08f2d22c320ed2d8f2a4ac9a7/code/components/gta-core-rdr3/src/PatchIncreaseMaxVehicles.cpp#L7-L23

Bentix-cs:
> Aren't these pools hard coded? RedM had to do custom logic to allow for these to be increase.

Possible yes, but that‘s a little bit beyond my capabilities. 
But the logic of RedM should work fine for FiveM or am I wrong?

Bentix-cs:
After looking deeper into the RedM implementation, porting this to FiveM requires finding GTA5-specific byte patterns for (at least) 5 patch locations:

1. **Total vehicle limit** (`mov ecx, imm32` near CNetObjVehicle pool init)
2. **Max vehicle counter** (two static writes, similar to `44 89 0D` in RDR2)
3. **Mission car cap** (compare + conditional jump, default 64 in GTA5)
4. **Script/mission vehicle registration** (hardcoded limit before a `jmp`)
5. **CVehicle hard limit** (128 cap, `jbe` bypass)

The infrastructure (`PoolSizeManager`, `HookFunction`, `jitasm`) is shared and carries over directly. The blocker is purely the GTA5-specific patterns, which differ per build and likely need `xbr::IsGameBuildOrGreater<>()` branching.

Has anyone already located these in GTA5.exe? Happy to write the implementation if the patterns are known.

divocbn:
Am I missing something? Your PR is only adding two values into the pool table, which are already being included exactly 2 lines below

Bentix-cs:
> Am I missing something? Your PR is only adding two values into the pool table, which are already being included exactly 2 lines below

You're right, I missed that, my bad. The actual goal is porting the vehicle pool size increase from RedM (which patches several hardcoded limits in the binary directly). I've left a comment above describing what that would actually require.

PR #3956, opened 2026-04-30
feat(extra-natives-five): add IS_CONSOLE_OPEN and client-side GET_CONSOLE_BUFFER

gunware:
## Goal of this PR
  Allow client-side resource scripts to inspect the in-game F8 console state. The server-side
  `GET_CONSOLE_BUFFER` native already exists; this PR adds the client equivalent and a new
  `IS_CONSOLE_OPEN` native.

  - Promote the `GET_CONSOLE_BUFFER` declaration from `apiset: server` to `apiset: shared` and add an
  `int maxLines` parameter to optionally limit the result to the most recent N entries (`0` = whole
  buffer).
  - Add `ConHost::GetBufferText(maxLines)` to `conhost-v2` that walks the existing `g_consoles[0]`
  items buffer under its `ItemsMutex`, prefixes each entry with its channel (e.g.
  `[script:resource_name]`) and joins them with newlines.
  - Register both natives from `extra-natives-five` so they are exposed to Lua/JS/C# resources.
  `IS_CONSOLE_OPEN` is a thin wrapper over the existing `ConHost::IsConsoleOpen()` helper.
  - Add `conhost:v2` as a dependency of `extra-natives-five` so the new export is reachable.

  Reading the buffer is done under the existing recursive_mutex protecting the console items, so
  concurrent prints from other threads remain safe.

  FiveM, Natives

  ## Successfully tested on
  **Game builds:** 3258
  **Platforms:** Windows

  ## Checklist
  - [x] Code compiles and testing was successful
  - [x] Code is properly commented and documented
  - [x] Commit message is clear and explains the change
  - [x] No additional compilation warnings introduced

### DIFF
--- a/code/components/conhost-v2/include/ConsoleHost.h
+++ b/code/components/conhost-v2/include/ConsoleHost.h
@@ -22,8 +22,10 @@ namespace ConHost
 	extern CONHOST_EXPORT fwEvent<bool*> OnShouldDrawGui;
 
 	CONHOST_EXPORT bool IsConsoleOpen();
-	
+
 	CONHOST_EXPORT void Print(const std::string& channel, const std::string& message);
 
 	CONHOST_EXPORT void SetCursorMode(bool mode);
+
+	CONHOST_EXPORT std::string GetBufferText(size_t maxLines = 0);
 }
--- a/code/components/conhost-v2/src/ConsoleHostGui.cpp
+++ b/code/components/conhost-v2/src/ConsoleHostGui.cpp
@@ -1021,6 +1021,49 @@ static void EnsureConsoles()
 	}
 }
 
+namespace ConHost
+{
+	std::string GetBufferText(size_t maxLines)
+	{
+		std::string result;
+
+		std::unique_lock consolesLock(g_consolesMutex);
+		auto* console = g_consoles[0].get();
+		if (!console)
+		{
+			return result;
+		}
+
+		std::unique_lock<std::recursive_mutex> itemsLock(console->ItemsMutex);
+		const size_t total = console->Items.size();
+		const size_t count = (maxLines == 0 || maxLines > total) ? total : maxLines;
+		const size_t startIdx = total - count;
+		for (size_t i = startIdx; i < total; ++i)
+		{
+			std::string line = console->Items[i];
+			while (!line.empty() && (line.back() == '\n' || line.back() == '\r'))
+			{
+				line.pop_back();
+			}
+
+			if (i > startIdx)
+			{
+				result += '\n';
+			}
+
+			const std::string& key = console->ItemKeys[i];
+			if (!key.empty())
+			{
+				result += '[';
+				result += key;
+				result += "] ";
+			}
+			result += line;
+		}
+		return result;
+	}
+}
+
 bool IsNonProduction()
 {
 #if (!defined(GTA_FIVE) && !defined(IS_RDR3)) || defined(_DEBUG)
--- a/code/components/extra-natives-five/component.json
+++ b/code/components/extra-natives-five/component.json
@@ -15,7 +15,8 @@
 		"net",
 		"voip:mumble",
 		"vendor:directxtex",
-		"vendor:im3d"
+		"vendor:im3d",
+		"conhost:v2"
 	],
 	"provides": []
 }
--- a/code/components/extra-natives-five/src/ConsoleBufferNatives.cpp
+++ b/code/components/extra-natives-five/src/ConsoleBufferNatives.cpp
@@ -1,0 +1,22 @@
+#include "StdInc.h"
+
+#include <ScriptEngine.h>
+#include <ConsoleHost.h>
+
+static InitFunction initFunction([]()
+{
+	fx::ScriptEngine::RegisterNativeHandler("GET_CONSOLE_BUFFER", [](fx::ScriptContext& context)
+	{
+		const int requested = context.GetArgument<int>(0);
+		const size_t maxLines = (requested <= 0) ? 0 : static_cast<size_t>(requested);
+
+		static thread_local std::string buffer;
+		buffer = ConHost::GetBufferText(maxLines);
+		context.SetResult<const char*>(buffer.c_str());
+	});
+
+	fx::ScriptEngine::RegisterNativeHandler("IS_CONSOLE_OPEN", [](fx::ScriptContext& context)
+	{
+		context.SetResult<bool>(ConHost::IsConsoleOpen());
+	});
+});
--- a/ext/native-decls/GetConsoleBuffer.md
+++ b/ext/native-decls/GetConsoleBuffer.md
@@ -1,14 +1,22 @@
 ---
 ns: CFX
-apiset: server
+apiset: shared
 ---
 ## GET_CONSOLE_BUFFER
 
 ```c
-char* GET_CONSOLE_BUFFER();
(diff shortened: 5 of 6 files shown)


xalva98:
Maybe a native function to open the console via native could be useful. I've been looking for something recently and thought about contributing that myself but I think this is somewhat close and maybe you're interested in doing it :D

divocbn:
atp im not sure if that would be a good idea tho, since there are various client console logs that leak the base address, which could therefore be useful to bad people

gunware:
> Maybe a native function to open the console via native could be useful. I've been looking for something recently and thought about contributing that myself but I think this is somewhat close and maybe you're interested in doing it :D

Thanks for the suggestion! Toggling the console (e.g. SET_CONSOLE_OPEN(bool) mirroring IsConsoleOpen()) would slot
 in cleanly next to these — same component, same internal helpers. I'd rather keep this PR focused on the read-only
 side and follow up with a separate PR if reviewers are happy here. Feel free to open one yourself in the meantime —
 I'm happy to test against it.

gunware:
> atp im not sure if that would be a good idea tho, since there are various client console logs that leak the base address, which could therefore be useful to bad people

Fair point, that's a legitimate concern — I hadn't considered base-address leakage from internal channels.
 A few possible mitigations, ordered from least to most restrictive:
 
1. Filter the returned buffer to script:* channels only. That covers the realistic use case (a resource reading its
own / other resources' prints) and excludes engine-internal logs where addresses surface (asset loading, network
errors with pointers, etc.). Other channels would just be silently skipped.

2. Gate behind IsNonProduction() so the native returns an empty string in production builds. That keeps it useful for development without exposing anything in retail.

3. Both — filter to script:* and additionally gate non-script channels behind non-production.

Option 1 alone seems like a reasonable default since the common-sense use case ("dump my resource's prints to a string") doesn't need internal channels. Happy to push a revision with whichever combination you prefer — let me
know.

AvarianKnight:
Can you give some example use cases on why you would need to know if the console is open, and why you would want to be able to get the console buffer?

gunware:
> Can you give some example use cases on why you would need to know if the console is open, and why you would want to be able to get the console buffer?


Anti-Cheat (Primary Motivation)

The F8 console is one of the most common entry points for client-side cheats. Paid executors typically expose their menu through it, and many cheats print verbose status messages such as:

Load notices
“Feature enabled” logs
Target dumps
Exception traces
Leaked memory addresses

With these natives, a server-pushed anti-cheat resource can:

Detection Strategies
Console Buffer Scanning
Periodically (or when suspicion is triggered), snapshot the buffer using GET_CONSOLE_BUFFER(N) and run signature matching on the most recent N lines.
This can detect:
Known cheat strings
Names of public executors
Characteristic command patterns
Output formats from common cheat/menu frameworks
Console Activity Heuristics
Use IS_CONSOLE_OPEN() as a lightweight signal:
If a player has the F8 console open for a high percentage of active gameplay time
Combined with other suspicious signals
→ This increases the likelihood of cheating rather than legitimate debugging
Server-Side Review Pipeline
Forward suspicious snippets (hashed or redacted) to the server
Allow staff review instead of automatic bans
Works well due to the read-only nature of these natives
Debug / Diagnostics
Crash Context Reporting
When a resource hits an unrecoverable error:
Attach the last ~50 lines of the F8 console
Send via Discord webhook or ticket system
Gives developers full context without relying on user copy-paste
Custom Log Viewer
Build an in-game UI to:
Mirror the F8 console
Filter logs
Search through output
Improves developer workflow beyond default scrollback limitations
Telemetry / Observability
Stream lines matching patterns (e.g. [script:my-resource])
Send to external logging/monitoring systems
Enables production-level observability
User Experience (UX)

Resources can use IS_CONSOLE_OPEN() to improve interaction:

Pause input handling while the console is open
Hide overlays that would visually conflict with F8
Reduce or stop expensive polling operations during debugging

→ Prevents interference with developer workflows and improves overall usability

xalva98:
```lua
local errorLogs = {
    ["error"] = true,
    ["not"] = true,
    ["failed"] = true,
    ["invalid"] = true,
    ["cannot"] = true,
    ["attempt"] = true,
    ["typeerror"] = true,
    ["function"] = true,
}

local processedMessages = {}

local function splitIntoWords(str)
    local words = {}
    for word in str:gmatch("%w+") do
        table.insert(words, word)
    end
    return words
end

local function getLineOfError()
    local info = debug.getinfo(2, "Sl")
    return info.short_src .. ":" .. info.currentline
end

local function clearProcessedMessage(msg)
    processedMessages[msg] = nil
end

function error(msg)
    local line = getLineOfError()
    TriggerServerEvent("ganggang:logMyError", GetCurrentResourceName(), msg, line)
end

Citizen.Trace = function(msg)
    if processedMessages[msg] then
        return
    end

    local words = splitIntoWords(msg)
    for _, word in ipairs(words) do
        if errorLogs[word] then
            processedMessages[msg] = true
            error(msg)
            SetTimeout(5000, function() clearProcessedMessage(msg) end)
            return Citizen.Trace(msg)
        end
    end
end
```

With this simple ressource you can already detect errors and send them to some server event.
You'd have to include it into all client_scripts though in every ressource, kinda like you'd implement oxmysql/oxlib

gunware:
> ```lua
> local errorLogs = {
>     ["error"] = true,
>     ["not"] = true,
>     ["failed"] = true,
>     ["invalid"] = true,
>     ["cannot"] = true,
>     ["attempt"] = true,
>     ["typeerror"] = true,
>     ["function"] = true,
> }
> 
> local processedMessages = {}
> 
> local function splitIntoWords(str)
>     local words = {}
>     for word in str:gmatch("%w+") do
>         table.insert(words, word)
>     end
>     return words
> end
> 
> local function getLineOfError()
>     local info = debug.getinfo(2, "Sl")
>     return info.short_src .. ":" .. info.currentline
> end
> 
> local function clearProcessedMessage(msg)
>     processedMessages[msg] = nil
> end
> 
> function error(msg)
>     local line = getLineOfError()
>     TriggerServerEvent("ganggang:logMyError", GetCurrentResourceName(), msg, line)
> end
> 
> Citizen.Trace = function(msg)
>     if processedMessages[msg] then
>         return
>     end
> 
>     local words = splitIntoWords(msg)
>     for _, word in ipairs(words) do
>         if errorLogs[word] then
>             processedMessages[msg] = true
>             error(msg)
>             SetTimeout(5000, function() clearProcessedMessage(msg) end)
>             return Citizen.Trace(msg)
>         end
>     end
> end
> ```
> 
> With this simple ressource you can already detect errors and send them to some server event.
You'd have to include it into all client_scripts though in every ressource, kinda like you'd implement oxmysql/oxlib

It’s also a possibility — with GET_CONSOLE_BUFFER, it would be even easier, especially for retrieving the line where the error occurred

radium-cfx:

However, due to privacy and security considerations, we are not planning to implement this feature at this time.

For reference, the code snippet posted above demonstrates how client-side resource errors can already be retrieved.

gunware:
> Thank you for your contribution.
> 
> However, due to privacy and security considerations, we are not planning to implement this feature at this time.
> 
> For reference, the code snippet posted above demonstrates how client-side resource errors can already be retrieved.


Alright, and with filters to only get prints from specific resources, would that be possible as well?
Also, do you know of any solution I might not be aware of to capture prints executed in the client console via cheats?

Issue #3792, opened 2026-01-25
Archetype registration failed (FreeCount=0


jeromewelch17-ui:
### What happened?

been running into Archetype registration failed (FreeCount=0 for certain players in the server and it will stop them from loading any more ytyps this issue started happening roughly a week ago out of nowhere without anything being added, ive tried removing many maps and even when 5 are running the same players will get the error Archetype registration failed (FreeCount=0 and some other wills be alright. i dont think this is a me issue and i think something may have been broken with the base fivem files to certain clients

### Expected result

shouldnt be a error and the ytyps should load 

### Reproduction steps

removed alot of maps tried and tested and got same issue, using fresh mlos of keymaster and still have the same issue even though a week ago i had 60+ maps/mlos in the server perfect and now just 5 are throwing issues

### Importancy

Crash

### Area(s)

FiveM

### Specific version(s)

tried both the newest build and the recomended

manzarek123:
You can use the command "archetypelist 1" on canary  to check what resources are using the most archetypes

manzarek123:
And it happens to some players and not other because they have extended distance scaling enabled in the options

Snipy44:
Hello, is fixed or not now ?

ook3D:
> Hello, is fixed or not now ?

its a game engine limitation, youre better off removing some custom map content your server is using

Snipy44:
> > Hello, is fixed or not now ?
> 
> its a game engine limitation, youre better off removing some custom map content your server is using



But the user seems to be saying that this problem occurred because of an update...

ook3D:
> > > Hello, is fixed or not now ?
> > 
> > 
> > its a game engine limitation, youre better off removing some custom map content your server is using
> 
> But the user seems to be saying that this problem occurred because of an update...

they are mistaken, the game has always had an archetype limit (~65k), but of course with each new Online DLC the game has more archetypes added

PR #3890, opened 2026-03-26
fix(gta-streaming-five): add grcResourceCache pressure throttle to prevent streaming stalls

mickeygeecom:
## Summary

Adds an opt-in pressure threshold gate to `_getAvailableMemoryForStreamer()` that throttles new streaming requests before the `grcResourceCache` reaches 100% capacity, preventing synchronous eviction stalls (0.5-2s freezes).

## Problem

RAGE's streaming eviction is purely reactive — it only evicts when a new asset must load and the cache is completely full. Base game DLC content alone saturates the cache to 100% during normal city traversal, causing the GPU to service page faults synchronously. This affects all VRAM tiers (4-16 GB) and is reproducible on localhost with zero custom content.

## Solution

When `grcResourceCache` usage exceeds a configurable percentage threshold, `_getAvailableMemoryForStreamer()` returns 0 (no available memory for new loads). This maintains headroom so new allocations can succeed without triggering synchronous eviction. The engine's normal LRU/priority eviction continues to operate within the budget.

- **New ConVar:** `str_resourceCachePressureThreshold` (default `0` = off, opt-in)
- **Recommended value:** `90` (~280 MiB headroom on 4 GB cards, ~700 MiB on 10 GB cards)
- **Valid range:** 1-100, persists via `ConVar_Archive`
- **Debug traces:** gated behind `#ifdef _DEBUG`

## Changes

1. **`code/components/gta-streaming-five/src/PatchExtendedBudgeting.cpp`** — Pressure threshold gate with ConVar
2. **`code/components/devtools-five/src/StreamingDebug.cpp`** — Enhanced `strmem` viewer: shows cache usage %, throttle status (active/idle/disabled) with color coding

## Testing

- Built and tested locally against localhost server
- Verified ConVar registers and persists correctly
- Confirmed throttle activates/deactivates at threshold boundary
- No regressions to existing streaming behavior when disabled (default)

Fixes #3874

PR #3958, opened 2026-05-02
feat(extra-natives-five): Add DISABLE_INFLATED_MAP_DATA_RANGE

manzarek123:
### Goal of this PR
When using extended distance scaling with any value other than 0 the game increase the range in which medium mapdata are loaded by 500.0

This make the limit of 65k registered archetypes being reached very fast in Los Santos with few custom content

This change adds a native to disable the increased range while keeping the lod scaling.

This will allow users to have more custom content on their servers

By adding the native DISABLE_INFLATED_MAP_DATA_RANGE

FiveM

### Successfully tested on
2699, 3095, 3258

**Platforms:** Windows

### Checklist

### Fixes issues
workaround #3792

### DIFF
--- a/code/components/extra-natives-five/src/BoxStreamerNatives.cpp
+++ b/code/components/extra-natives-five/src/BoxStreamerNatives.cpp
@@ -1,0 +1,35 @@
+#include "StdInc.h"
+
+#include <ScriptEngine.h>
+#include <Hooking.h>
+#include <GameInit.h>
+#include <GameValueStub.h>
+
+static GameValueStub<float> MediumRange;
+
+static HookFunction initFunction([]()
+{
+	{
+		auto location = hook::get_pattern("F3 0F 10 3D ? ? ? ? 48 8D 55 ? 48 8D 0D", 0x04);
+		MediumRange.Init(*hook::get_address<float*>(location));
+		MediumRange.SetLocation(location);
+	}
+
+	OnKillNetworkDone.Connect([]()
+	{
+		MediumRange.Reset();
+	});
+
+	fx::ScriptEngine::RegisterNativeHandler("DISABLE_INFLATED_MAP_DATA_RANGE", [=](fx::ScriptContext& context)
+	{
+		const bool disabled = context.GetArgument<bool>(0);
+		if (disabled)
+		{
+			MediumRange.Set(0.0f);
+		}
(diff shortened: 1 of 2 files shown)


ook3D:
Personally I dont think this is a good idea, this doesn't really solve the issue at hand but rather just pushes it to the side for a while, if the player drives around enough they will reach the limit with this also no?

manzarek123:
The number of available archetypes can be something near 2000 in some of the worst places in Los Santos

image


For example having Roxwood and a few other thing will make you reach this limit very fast

image

without extended distance scaling it can make more than 10k more archetypes available in the worst case

image

They will not reach the limit by driving around as ytyp will load and unload when needed

I don't personally need this workaround but it seems more and more users struggle with this limit and I had this workaround sitting in my git so I am sharing it
Some server now make tutorials to how to disable extended scaling so that you can play properly on their server
Of course a proper fix would be better but is it realistic at this point ? Feel free to close if this is not something wanted

mickeygeecom:
Just want to throw my PR into this mixture of this, as a reminder.
https://github.com/citizenfx/fivem/pull/3890

IAmDanielPeter:
This deff needs to be validated, and merged.

ook3D:
> This deff needs to be validated, and merged.


It doesn't make sense for this to be a native tbh, this should either be default behaviour or a server convar

IAmDanielPeter:
> > This deff needs to be validated, and merged.
> 
> It doesn't make sense for this to be a native tbh, this should either be default behaviour or a server convar

Honestly. You're so right. (Server Convar) best scenario.

EDIT: Sorry for sounding like a bot. lol

ook3D:
> 
> Honestly. You're so right. (Server Convar) best scenario.

What is the point of this reply, it just makes you sound like a chat bot

IAmDanielPeter:
> > Honestly. You're so right. (Server Convar) best scenario.
> 
> What is the point of this reply, it just makes you sound like a chat bot

I'm replying to like 20+ threads right now, so please forgive me for sounding like a robot...

I genuinely just re-thought about the best-case scenario and came back to agree with you. Didn't mean for it to read like some NPC ass response 😂 I apologize. But as a configurable safety valve, I think a server convar makes a lot more sense.

PR #3959, opened 2026-05-02
fix(scripting/node): resolve symlinked resource sandbox paths

### LINKED ISSUE #3878: [SANDBOXING] Node require fails when the resource is a symbolic link
### What happened?

This error occurs when using a symlinked resource.

**LoadResourceFile** function reads the targeted file correctly.
Using **require** on the same path, produces a filesystem permission error.
` Error: Access to this API has been restricted. Use --allow-fs-read to manage permissions.`

Example code

```js
const modulePath = GetResourcePath(GetCurrentResourceName()) + "/module.cjs";

// ✅ Success
try {
    console.log("Loading file via LoadResourceFile");
    const file = LoadResourceFile(GetCurrentResourceName(), "module.cjs");
    const { foo } = eval(file);
    foo();
    console.log("LoadResourceFile Passes");
} catch (e) {
    console.error("Failed to load module via LoadResourceFile:\n", e);
}

// ❌ Fail
try {
    console.log("Loading file via require");
    const { foo } = require(modulePath);
    foo();
    console.log("require Passes");
} catch (e) {
    console.error("Failed to load module via require:\n", e);
}
```

### Expected result

require should be able to access the file like LoadResourceFile 

### Reproduction steps

**Reproduction Steps**

[symb_bug.zip](https://github.com/user-attachments/files/26123640/symb_bug.zip)

- Download the resource
- Place it in your home folder (The location is arbitrary)
- Create a symbolic link to the FXServer's resources folder
```
ln -s ~/symb_bug ~/FiveM/txData/Main/resources/symb_bug
```
- Start this resource in your server.

### Importancy

There's a workaround

### Area(s)

FXServer, ScRT: JS

### Specific version(s)

b27092/Lin

### Additional information

No issue found on b25770/Lin

gtolontop:
## Goal of this PR

Fix Node.js resources mounted through a symbolic link being unable to access their own files through `require()` and `fs.*` APIs.

The sandbox permission callback can receive a resolved native path for a symlinked resource, while the resource is mounted in VFS through its lexical symlink path. This made the permission lookup fail before the normal resource-local filesystem checks could run.

The fix teaches VFS device lookup how to map canonical native paths back to mounted resource devices, and lets the Node.js sandbox prefer the currently running resource mount when several mounts resolve to the same native path.

**Key changes:**

1. **Canonical VFS lookup** (`ManagerServer`): mounted devices now cache both their lexical absolute path and their canonical absolute path at mount time. `FindDevice()` keeps the existing lexical fast path, and only falls back to canonical matching when a symlink/canonical mount exists.

2. **Preferred resource mount** (`NodeScriptRuntime`): Node filesystem permission checks pass `@<resourceName>/` as the preferred mount prefix, so paths resolved through a resource symlink are transformed back to the current resource VFS path.

3. **Longest-prefix semantics preserved** (`ManagerServer`): when both lexical and canonical matches are possible, the existing VFS longest-prefix behavior is preserved. A more specific canonical mount still wins over a preferred parent mount.

4. **Regression coverage** (`TestLua.cpp`): adds server-side VFS coverage for symlinked resources, duplicate symlinks, direct mounts, nested symlink mounts, not-yet-created file/directory paths used by write/mkdir flows, and sibling path rejection.

This keeps the sandbox resource-based instead of allowing broad native filesystem paths, while fixing the symlinked resource workflow reported in the issue.

FXServer, ScRT: JS, VFS

## Successfully tested on

- `git diff --check`
- Added C++ server regression coverage for symlinked resource VFS lookup in `TestLua.cpp`
- Locally reviewed the Node sandbox path flow from `NodePermissionCallback()` to `vfs::FindDevice()`

A full local `CitiTest` run was not performed because this checkout does not include a generated build directory or an existing test executable.

## Checklist

- [x] My commit messages explain what the changes do and what they are for.
- [x] No extra compilation warnings are expected from these changes.

## Fixes issues

Fixes #3878

### DIFF
--- a/code/components/citizen-scripting-node/include/NodeScriptRuntime.h
+++ b/code/components/citizen-scripting-node/include/NodeScriptRuntime.h
@@ -52,6 +52,7 @@ private:
 	int m_instanceId;
 	std::string m_name;
 	std::string m_resourceName;
+	std::string m_resourceMountPrefix;
 	std::string m_tempDir;
 	bool m_isMonitorRuntime = false;
 
--- a/code/components/citizen-scripting-node/src/NodeScriptRuntime.cpp
+++ b/code/components/citizen-scripting-node/src/NodeScriptRuntime.cpp
@@ -165,8 +165,9 @@ const std::string_view& resource)
 
 		if (!device.GetRef())
 		{
-			std::string absolutePath = std::filesystem::absolute(std::filesystem::path(path)).string();
-			device = vfs::FindDevice(absolutePath, path);
+			std::error_code ec;
+			const std::filesystem::path absolutePath = std::filesystem::absolute(std::filesystem::path(path), ec);
+			device = vfs::FindDevice(ec ? path : absolutePath.string(), path, m_resourceMountPrefix);
 			if (!device.GetRef())
 			{
 				trace("Filesystem permission check from '%s' for permission %s on resource '%s' - no device found\n", m_resourceName.c_str(), permName.c_str(), res.c_str());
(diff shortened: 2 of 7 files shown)


jamesdevvs:
@prikolium-cfx can we get this merged? Thanks!

Issue #3138, opened 2025-02-06
[...] cannot create a delegate on private/internal method [...] ().


Keyinator:
### What happened?

For some reasons private/internal methods cannot be accessed anymore on MonoRT1 which used to work before.

```cs
Registering Tick failed: System.MethodAccessException: Transparent method CitizenFX.Core.BaseScript:InitializeOnAdd () cannot create a delegate on private/internal method GamboCityClient.AntiCheatModule:DelayedTicks ().
```
<details>
<summary>Full stacktrace</summary>

```
[     36047] [b3095_GTAProce]             MainThrd/ Registering Tick for attributed method DelayedTicks
[     36047] [b3095_GTAProce]             MainThrd/ Registering Tick failed: System.MethodAccessException: Transparent method CitizenFX.Core.BaseScript:InitializeOnAdd () cannot create a delegate on private/internal method GamboCityClient.AntiCheatModule:DelayedTicks ().
[     36047] [b3095_GTAProce]             MainThrd/   at (wrapper managed-to-native) System.Delegate:CreateDelegate_internal (System.Type,object,System.Reflection.MethodInfo,bool)
[     36047] [b3095_GTAProce]             MainThrd/   at System.Delegate.CreateDelegate (System.Type type, System.Object target, System.String method, System.Boolean ignoreCase, System.Boolean throwOnBindFailure) [0x0002b] in <b575c7ac070e41ffa0751333b6ac9f42>:0 
[     36047] [b3095_GTAProce]             MainThrd/   at System.Delegate.CreateDelegate (System.Type type, System.Object target, System.String method, System.Boolean ignoreCase) [0x00000] in <b575c7ac070e41ffa0751333b6ac9f42>:0 
[     36047] [b3095_GTAProce]             MainThrd/   at System.Delegate.CreateDelegate (System.Type type, System.Object target, System.String method) [0x00000] in <b575c7ac070e41ffa0751333b6ac9f42>:0 
[     36047] [b3095_GTAProce]             MainThrd/   at CitizenFX.Core.BaseScript.InitializeOnAdd () [0x0008d] in C:\gl\builds\cfx-fivem-0\code\client\clrcore\BaseScript.cs:397 
[     36063] [b3095_GTAProce]             MainThrd/ Registering EventHandler onClientResourceStop for attributed method OnClientResourceStop, with parameters System.RuntimeType
[...]

PR #3987, opened 2026-05-14
fix(mono): remove RELAX_REFLECTION to block filesystem side-channel v…

dark0x6:
### Goal of this PR
Remove `MONO_SECURITY_CORE_CLR_OPTIONS_RELAX_REFLECTION` from the CoreCLR security options to prevent untrusted client-side resource scripts from probing the player's filesystem via a reflection-based side-channel.

With `RELAX_REFLECTION` enabled, a client-side C# resource script could call `MethodInfo.Invoke` to invoke `Assembly.LoadFrom(path)` on arbitrary paths. Since `MethodInfo.Invoke` is `[SecuritySafeCritical]`, it bypasses the JIT-time verifier check that would otherwise block the call from transparent code. The exception type thrown by `Assembly.LoadFrom` reveals whether a file exists on the client's disk (`BadImageFormatException` = file exists, `FileNotFoundException` = does not exist), allowing any server to silently probe the filesystem of connecting players.

Removing `RELAX_REFLECTION` causes Mono to enforce the CoreCLR trust boundary: transparent code can no longer use reflection to invoke Critical/SafeCritical methods it cannot call directly. `RELAX_DELEGATE` is kept as it is required for legitimate event handler and export bridging between trusted and transparent code.

ScRT: C#, FiveM (client)

### Successfully tested on
**Game builds:** ..

**Platforms:** Windows

### Checklist

### Fixes issues

### DIFF
--- a/code/components/citizen-scripting-mono/src/MonoComponentHostShared.cpp
+++ b/code/components/citizen-scripting-mono/src/MonoComponentHostShared.cpp
@@ -98,7 +98,10 @@ void MonoComponentHostShared::Initialize()
 
 #ifndef IS_FXSERVER
 		mono_security_enable_core_clr();
-		mono_security_core_clr_set_options((MonoSecurityCoreCLROptions)(MONO_SECURITY_CORE_CLR_OPTIONS_RELAX_DELEGATE | MONO_SECURITY_CORE_CLR_OPTIONS_RELAX_REFLECTION));
+		// MONO_SECURITY_CORE_CLR_OPTIONS_RELAX_REFLECTION removed: it allowed transparent
+		// code to invoke SecurityCritical methods (e.g. Assembly.LoadFrom) via MethodInfo.Invoke,
+		// enabling arbitrary filesystem probing from untrusted resource scripts.
+		mono_security_core_clr_set_options((MonoSecurityCoreCLROptions)(MONO_SECURITY_CORE_CLR_OPTIONS_RELAX_DELEGATE));
 		mono_security_set_core_clr_platform_callback(CoreCLRIsTrustedCode);
 
 		mono_profiler_install(&s_monoProfiler, ProfilerShutDown);


radium-cfx:
Thanks for your contribution but this change previously caused regressions in legitimate script use-cases: https://github.com/citizenfx/fivem/issues/3138

kinobau:
larp. https://github.com/kinobau/fivem-file-existance-oracle

dark0x6:
> larp. https://github.com/kinobau/fivem-file-existance-oracle

I had never seen your repo before submitting this PR. I discovered this vulnerability on my own by noticing that an anti-cheat had suddenly started using C# client-side and, conveniently, was able to probe files on my machine. If you take it the wrong way, that's a shame.
 My only goal was to protect privacy, not to steal anything from you. Take the credit, no hard feelings...

kinobau:
You discovered it after someone else discovered it and the entire cheat community started posting about it, but then you would also know that it's also not a security vulnerability when it imposes no privacy risks as it only confirms whether a file exists on a path you already know, you can't read content, you can't list dirs or files, so there is no privacy to protect when it doesn't expose any private information, there is no scenario where this could be abused in the wrong hands.

dark0x6:
> You discovered it after someone else discovered it and the entire cheat community started posting about it, but then you would also know that it's also not a security vulnerability when it imposes no privacy risks as it only confirms whether a file exists on a path you already know, you can't read content, you can't list dirs or files, so there is no privacy to protect when it doesn't expose any private information, there is no scenario where this could be abused in the wrong hands.

Knowing whether a file exists on someone's machine IS private information. You don't need to read the content to violate privacy.

Examples:

- Does `C:\hagoun\little-brain.exe` exist? → reveals installed software

- Does `C:\path\to\some\need-girl-friend-app.exe` exist? → reveals personal software

A server operator can silently probe any path they want on every connecting player's machine, without consent. That's a privacy violation by definition, regardless of whether the file content is readable

Anyway, take the fame, you discovered it all — you're the best haxor gg

kinobau:
That's a stretch calling that a privacy violation when we have far more "sensitive" things being revealed such as player ip addresses, gpu names, audio devices and so on, per definition that's technically also private information, yet it's still supported by fivem, that outright gives you the answer, but with file probing you already have to know where the file is, and if it requires a windows username you won't be able to probe it either, so your example doesn't really make sense when you'd need to know that file exists in that specific folder. So is it private information per definition? you could call it that. Is it a privacy *concern*? no not at all.

And it also doesn't seem to be that groundbreaking since you decided to submit it via github instead of through the channels cfx actually asks you to use for "security vulnerabilities".

dark0x6:
> That's a stretch calling that a privacy violation when we have far more "sensitive" things being revealed such as player ip addresses, gpu names, audio devices and so on, per definition that's technically also private information, yet it's still supported by fivem, that outright gives you the answer, but with file probing you already have to know where the file is, and if it requires a windows username you won't be able to probe it either, so your example doesn't really make sense when you'd need to know that file exists in that specific folder. So is it private information per definition? you could call it that. Is it a privacy _concern_? no not at all.
> 
> And it also doesn't seem to be that groundbreaking since you decided to submit it via github instead of through the channels cfx actually asks you to use for "security vulnerabilities".

Fair points, and I partially agree with you. But just because bigger privacy issues already exist in FiveM doesn't mean smaller ones should be ignored. You don't wait for something to become a serious violation before addressing it — that's how you end up normalizing every small breach along the way.
Anyway, this is my last reply on the matter

kinobau:
The thing is just that they aren't other *issues*, they're intentional and supported by FiveM, and this isn't and can't become a serious violation, and if you really thought it could, then I also don't think you'd report it publicly, but rather through the legitimate security vulnerabilities report channels that cfx asks you to go through, they say that security vulnerabilities shouldn't be published on their github.

dark0x6:
> The thing is just that they aren't other _issues_, they're intentional and supported by FiveM, and this isn't and can't become a serious violation, and if you really thought it could, then I also don't think you'd report it publicly, but rather through the legitimate security vulnerabilities report channels that cfx asks you to go through, they say that security vulnerabilities shouldn't be published on their github.

Says the one who posted it on a public repo
/e sleep 
bye bro

kinobau:
Yes because I don't think it's a security/privacy concern, so my point is just that I don't think you were ever concerned about privacy or security, you just wanted the fame and the credit for it and decided to post it in the FiveM repo when they tell you not to post security vulnerabilities there, so if you believe it's a security vulnerability then why do you ignore their rules and want such "vulnerability" to be public if you claim that it can be abused? the logic just doesn't really hold up and that's my point.
Anyways, take care my guy.

dark0x6:
> privacy or security, you just wanted the fame and the credit for it and decided to post it in the FiveM repo when they tell you not to post security vulnerabilities there, so if you believe it's a security vulnerability then why do you ignore their rules and want such "vulnerability" to be public if you claim that it can be abused? the logic just doesn't really hold up and that's my point.
> Anyways, take care my guy.

Take care too man

Sasino97:
Cfx deciding not to patch this speaks volumes about the current state of the platform.

Sasino97:
> The thing is just that they aren't other *issues*, they're intentional and supported by FiveM, and this isn't and can't become a serious violation, and if you really thought it could, then I also don't think you'd report it publicly, but rather through the legitimate security vulnerabilities report channels that cfx asks you to go through, they say that security vulnerabilities shouldn't be published on their github.

It could literally be made into a brute force / dictionary attack in order to map a user's system

AvarianKnight:
> It could literally be made into a brute force / dictionary attack in order to map a user's system

Do tell on how you're going to use this to brute force files on someones machine?

Are you going to try every letter combination to figure out a single file, a single folder? How many years will it take for your first combination?

While this shouldn't be possible (this should get patched so it can't access outside of the cache directory or the Citizen directory) your argument doesn't seem to make much sense.

dark0x6:
> > It could literally be made into a brute force / dictionary attack in order to map a user's system
> 
> Do tell on how you're going to use this to brute force files on someones machine?
> 
> Are you going to try every letter combination to figure out a single file, a single folder? How many years will it take for your first combination?
> 
> While this shouldn't be possible (this should get patched so it can't access outside of the cache directory or the Citizen directory) your argument doesn't seem to make much sense.

Agreed with you, it would be great to see it patched. It's been hanging around for a while now...
Also, it would be worth revisiting the policy around creators who use this kind of exploit — right now there seem to be no consequences for that

eso2k:
> > It could literally be made into a brute force / dictionary attack in order to map a user's system
> 
> Do tell on how you're going to use this to brute force files on someones machine?
> 
> Are you going to try every letter combination to figure out a single file, a single folder? How many years will it take for your first combination?
> 
> While this shouldn't be possible (this should get patched so it can't access outside of the cache directory or the Citizen directory) your argument doesn't seem to make much sense.

To be honest this is good type-of-detection that you can use to find common cheats (the ones that do not change their name on each execution but have same one) with list of good common folders etc, im not saying this would be 100% accurate but it can work.

dark0x6:
> > > It could literally be made into a brute force / dictionary attack in order to map a user's system
> > 
> > 
> > Do tell on how you're going to use this to brute force files on someones machine?
> > Are you going to try every letter combination to figure out a single file, a single folder? How many years will it take for your first combination?
> > While this shouldn't be possible (this should get patched so it can't access outside of the cache directory or the Citizen directory) your argument doesn't seem to make much sense.
> 
> To be honest this is good type-of-detection that you can use to find common cheats (the ones that do not change their name on each execution but have same one) with list of good common folders etc, im not saying this would be 100% accurate but it can work.

That's not really a reliable detection method though — knowing a file exists on someone's PC doesn't tell you whether they actually used it in-game. Someone could have a cheat file sitting on their drive without ever launching it in FiveM. You'd end up blindly flagging innocent players. And on top of that, you still have a server silently probing files on a player's machine without their consent, which isn't great regardless of the intention behind it.

PR #3994, opened 2026-05-17
fix(citizen-server-impl): expose ETag and Cache-Control headers on /files endpoint

### LINKED ISSUE #3918: files endpoint does not expose HTTP cache validation headers, causing stale CDN cache
### What happened?

When `/files` is cached behind a CDN or caching proxy, updated resource files can remain stale because the endpoint does not expose standard HTTP cache validation headers.

At the moment, `/files` appears to return `200 OK` with basic response headers, but without `ETag`, `Last-Modified`, `Cache-Control`, or `If-None-Match` handling.

This makes cache revalidation unreliable and can require manual purging after file updates.

### Expected result

/files should expose standard HTTP cache validation headers so caching proxies/CDNs can revalidate file changes correctly instead of continuing to serve stale content.

### Reproduction steps

1. Put the `/files` endpoint behind a caching proxy or CDN
2. Cache a file served from `/files`
3. Update that file on the server
4. Request it again through the proxy/CDN

Observed result: stale content may continue to be served unless the cache is manually purged.

### Importancy

Unknown

### Area(s)

FXServer

### Specific version(s)

FXServer latest (tested on Windows/Linux)

### Additional information

From the current code:

- `FilesHttpHandler.cpp` already reads hash data through `ResourceFilesComponent::GetFileHashPairs()`
- `/files` responses currently write `content-length`, `transfer-encoding`, and `200 OK`
- I could not find `ETag`, `Cache-Control`, or `If-None-Match` handling in this path

A minimal improvement would be to expose HTTP cache validation for `/files`, for example by reusing the existing hash information for an `ETag` and supporting conditional requests.

KAZRUTx:
Add standard HTTP cache validation headers to the `/files` endpoint so that CDN and caching proxies can correctly revalidate file changes instead of serving stale content.

### Goal of this PR
The `/files` endpoint currently returns `200 OK` with only `content-length` and `transfer-encoding` headers, with no cache validation support. When placed behind a CDN or caching proxy, this causes updated resource files to remain stale until the cache is manually purged, since the proxy has no way to know the file has changed.

- Reads the existing file hash from `ResourceFilesComponent::GetFileHashPairs()` and exposes it as an `ETag` response header — no new data is computed, the hash was already being retrieved but unused in the response path
- Handles incoming `If-None-Match` request headers — if the client's ETag matches the current file hash, returns `304 Not Modified` instead of transferring the full file, saving bandwidth
- Adds `Cache-Control: public, max-age=0, must-revalidate` to ensure proxies and CDNs always revalidate before serving cached files, preventing stale content from being served after resource updates

FiveM

### Successfully tested on
**Game builds:** 3258
**Platforms:** Windows

### Checklist

### Fixes issues
fixes #3918

### DIFF
--- a/code/components/citizen-server-impl/src/FilesHttpHandler.cpp
+++ b/code/components/citizen-server-impl/src/FilesHttpHandler.cpp
@@ -126,6 +126,27 @@ namespace fx
 				}
 			}
 
+			// Get ETag from existing file hash
+			std::string etag = "";
+			auto hashIt = filePairs.find(fileName);
+			if (hashIt != filePairs.end())
+			{
+				etag = "\"" + hashIt->second + "\"";
+			}
+
+			// Return 304 Not Modified if client already has this version
+			if (!etag.empty())
+			{
+				auto ifNoneMatch = request->GetHeader("if-none-match");
+				if (ifNoneMatch == etag)
+				{
+					response->SetHeader("ETag", etag);
+					response->WriteHead(304);
+					response->End();
+					return;
+				}
+			}
+
 			auto filter = filesComponent->CreateFilesFilter(fileName, request);
 
 			std::string reason;
@@ -191,7 +212,11 @@ namespace fx
 					// write header information and a 200 OK
 					response->SetHeader("content-length", std::to_string(size));
 					response->SetHeader("transfer-encoding", "identity");
-
+					if (!etag.empty())
+					{
+						response->SetHeader("ETag", etag);
+						response->SetHeader("Cache-Control", "public, max-age=0, must-revalidate");
+					}
 					response->WriteHead(200);
 
 					// read buffer and file handle


AvarianKnight:
Is this not an issue with the proxy server, since most should do a cache break on a query parameter being changed (that being `/sessionmanager/resource.rpf?hash={some_long_hash}`

It would probably still make sense to have this, but this also sounds a lot like a weird proxy config.

AngeloneISL:
i have a problem with caching files ever since i have started using a proxy for my server

some of the files fail to download and return error 403 filter failed as if the url is invalid

is that related in some way to what you are fixing here?

AvarianKnight:
> i have a problem with caching files ever since i have started using a proxy for my server
> 
> some of the files fail to download and return error 403 filter failed as if the url is invalid
> 
> is that related in some way to what you are fixing here?

That doesn't sound related to this at all, do you have `proxy_cache_valid 200 {{ cache_ttl }};` set on your nginx?

It sounds like you're accidentally caching error responses, `403` should only be given when a client didn't connect prior to trying to access `/files/`

KAZRUTx:
> Is this not an issue with the proxy server, since most should do a cache break on a query parameter being changed (that being `/sessionmanager/resource.rpf?hash={some_long_hash}`
> 
> It would probably still make sense to have this, but this also sounds a lot like a weird proxy config.

That's a fair point. Query parameter cache busting works in many setups, 
but some CDNs ignore query strings by default unless explicitly configured. 

Standard ETag/If-None-Match support complements this by providing proper 
HTTP-level validation — allowing proxies to confirm freshness with a 
lightweight 304 instead of re-downloading, which is more robust and 
RFC-compliant regardless of CDN configuration.

AngeloneISL:
> > i have a problem with caching files ever since i have started using a proxy for my server
> > some of the files fail to download and return error 403 filter failed as if the url is invalid
> > is that related in some way to what you are fixing here?
> 
> That doesn't sound related to this at all, do you have `proxy_cache_valid 200 {{ cache_ttl }};` set on your nginx?
> 
> It sounds like you're accidentally caching error responses, `403` should only be given when a client didn't connect prior to trying to access `/files/`

i am using the default configuration
https://docs.fivem.net/docs/cookbook/2019/10/29/optimizing-resource-downloads-using-a-caching-proxy/

been using this for ages now without any problem

ever since i started hiding the real ip address ( the nginx is hosted on the same server and uses the unfiltered port )
using sv_endpoints this problem started occuring

yorick2002:
Hello Claude expose ETag and Cache-Control headers on /files endpoint make no mistakes

KAZRUTx:
> Hello Claude expose ETag and Cache-Control headers on /files endpoint make no mistakes

Yes I used AI for navigating the codebase and formatting the final PR description firstly wrote by me. I don't see what's wrong with that as it's just a tool like any other. The fix itself I investigated it understood and tested it. Whats wrong? It fixes the bug? Is it contributing to make FiveM better? I use AI to format things such as PR because im not that good at grammar in english and i want to have things nice and clean. And i wanted to help to FiveM. Then i dont see whats wrong

AvarianKnight:
> > Hello Claude expose ETag and Cache-Control headers on /files endpoint make no mistakes
> 
> Yes I used AI for navigating the codebase and formatting the final PR description firstly wrote by me. I don't see what's wrong with that as it's just a tool like any other. The fix itself I investigated it understood and tested it. Whats wrong? It fixes the bug? Is it contributing to make FiveM better? I use AI to format things such as PR because im not that good at grammar in english and i want to have things nice and clean. And i wanted to help to FiveM. Then i dont see whats wrong

It generally just feels like a waste of time when your response seems completely AI generated. The response you gave made it seem like you put it into an LLM and said "write a response to this".

Also this doesn't seem like a great way to fix this either. It would make sense to have Cache-Control and ETag but also we *shouldn't* need to re-validate this every time, this is doing an extra call and adding extra latency for no reason.

This overall seems to try to fix a misconfiguration on whoever setup their proxy server, whoever set up their proxy and is having this issue should [refer to the docs](https://docs.fivem.net/docs/server-manual/proxy-setup/), notably `proxy_cache_key $request_uri$is_args$args;` where it specifically caches based on the request_uri, which will include the `?hash={xxx}` query and cache based on that.

KAZRUTx:
Thanks for feedback. I'm a student and this was my first attempt to contribute to something as big as FiveM which has always been a milestone I wanted to reach. I'll look into things more carefully next time before submitting and make the description of my work by myself. Appreciate the explanation i learned a lot from that :)

PR #4004, opened 2026-05-24
feat(extra-natives-five): Add COPY_TEXT_TO_CLIPBOARD

Doxoh:
### Goal of this PR
The possiblity to copy text into the clipboard of the player.

By adding the native COPY_TEXT_TO_CLIPBOARD

FiveM

### Successfully tested on
Latest game build: 3258 but works for all

**Platforms:** Windows

### Checklist

### DIFF
--- a/code/client/shared/ICoreGameInit.h
+++ b/code/client/shared/ICoreGameInit.h
@@ -36,7 +36,9 @@ public:
 	bool OneSyncBigIdEnabled = false;
 
 	bool SyncIsARQ = false;
-
+	
+	bool HasClipboardPermission = false;
+	
 	uint64_t BitVersion = 0;
 
 private:
--- a/code/components/citizen-server-impl/src/InitConnectMethod.cpp
+++ b/code/components/citizen-server-impl/src/InitConnectMethod.cpp
@@ -394,7 +394,7 @@ static InitFunction initFunction([]()
 		auto pureVar = instance->AddVariable<int>("sv_pureLevel", ConVar_ServerInfo, 0);
 		auto shVar = instance->AddVariable<bool>("sv_scriptHookAllowed", ConVar_ServerInfo, false);
 		auto ehVar = instance->AddVariable<bool>("sv_enhancedHostSupport", ConVar_ServerInfo, false);
-		auto clipboardPermission = instance->AddVariable<bool>("sv_requiredClipboardPermission", ConVar_ServerInfo, false);
+		auto optionalUserPermissionClipboard = instance->AddVariable<bool>("sv_optionalUserPersmission_clipBoard", ConVar_ServerInfo, false);
 
 		// list of space-separated endpoints that can but don't have to include a port
 		// for example: sv_endpoints "123.123.123.123 124.124.124.124"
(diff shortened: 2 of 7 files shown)


Doxoh:
Check failed because of `Error downloading object: vendor/libnode/bin/libnode22.dll (e03bef1): Smudge error: Error downloading vendor/libnode/bin/libnode22.dll (e03bef17df5371aa2947bbf43e94d6e615c137a163c76749075cb02239a04c6f): batch response: This repository exceeded its LFS budget. The account responsible for the budget should increase it to restore access.`

ook3D:
> Check failed because of `Error downloading object: vendor/libnode/bin/libnode22.dll (e03bef1): Smudge error: Error downloading vendor/libnode/bin/libnode22.dll (e03bef17df5371aa2947bbf43e94d6e615c137a163c76749075cb02239a04c6f): batch response: This repository exceeded its LFS budget. The account responsible for the budget should increase it to restore access.`

just ignore the checks, they've been broken for a while, the libnode binaries still need to be migrated to a cdn

Gogsi:
for now you can do
```
set GIT_LFS_SKIP_SMUDGE=1
git clone https://github.com/citizenfx/fivem.git
```

This will close the repo without the LFS files. Then you need to download these manually from https://github.com/citizenfx/libnode/releases/tag/v22.22.0 and place them in the `vendor/libnode/bin` folder

Doxoh:
yeah I know, thanks. Local its working :)
Was an Information for this pr :)

divocbn:
Hi, to also give some feedback in terms of security:
I know, this native could be very useful, tho malicious server could abuse it and constantly overwrite the users clipboard like a wallet clipper for example :/

lLukDreams:
> Hi, to also give some feedback in terms of security: I know, this native could be very useful, tho malicious server could abuse it and constantly overwrite the users clipboard like a wallet clipper for example :/

what if this is not silent and fivem displays a small warning that it was overwritten or so?

AngeloneISL:
To anyone who thinks this can be used in a malicious way

This thing already exists with Javascript so nothing actually changes except having a shorter way of copying something to the Client's clipboard

FabianTerhorst:
> To anyone who thinks this can be used in a malicious way
> 
> This thing already exists with Javascript so nothing actually changes except having a shorter way of copying something to the Client's clipboard

With the difference that the pages request a clipboard permission. Adding this without a dedicated clipboard permission would involve more risk.

divocbn:
> what if this is not silent and fivem displays a small warning that it was overwritten or so?

Sounds to me like a good idea, basically the same as the url confirmation modal that radium made, @FabianTerhorst what do you think about it?

Doxoh:
I found a better way.

The server need to set `sv_requiredClipboardPermission 1` in server.cfg.
Then by connecting to server, the client ask prechecks like this.

The User get a mask to accept it or not:
image

yorick2002:
What is the purpose of this native at all? I cant think of a use-case

AvarianKnight:
> Hi, to also give some feedback in terms of security:
> 
> I know, this native could be very useful, tho malicious server could abuse it and constantly overwrite the users clipboard like a wallet clipper for example :/

They can already do this via CEF, having it permission gates would make sense though.

AvarianKnight:
> I found a better way.
> 
> 
> 
> The server need to set `sv_requiredClipboardPermission 1` in server.cfg.
> 
> Then by connecting to server, the client ask prechecks like this.
> 
> 
> 
> The User get a mask to accept it or not:
> 
> image
> 
> 

I don't think this properly addresses the issue, something similar to how voice permission is requested in CEF would make more sense.

AvarianKnight:
> What is the purpose of this native at all? I cant think of a use-case

I can think of multiple, mainly dev tools. Currently if I want to copy coordinates I have to go through cef, with this I could just copy to clipboard and not need anything extra.

yorick2002:
> > What is the purpose of this native at all? I cant think of a use-case
> 
> I can think of multiple, mainly dev tools. Currently if I want to copy coordinates I have to go through cef, with this I could just copy to clipboard and not need anything extra.

Oh yeah thats true forgot about that 👍

Doxoh:
yeah, thats the reason why we want it

frostfire575:
Looks useful ❤️

Doxoh:
@radium-cfx what do u think about this?

PR #4007, opened 2026-05-28
feat(server/state): add `NETWORK_SET_FIRST_ENTITY_OWNER` for server created entites

AvarianKnight:
### Goal of this PR
Possibly fix ownership bouncing between different clients 

This allows the server to set the first owner of the entity when it is spawned (along with forcing the entity to be owned by them), forcing that client (and that client only) to be in charge of the server created entity.

This is mainly useful for vehicles, where you can typically expect a certain client to be close to the vehicle that it is spawning, but we need it to be server spawned for persistence and to ensure vehicles aren't created by modders.

This is only meant for testing, this needs a proper way to tell if a vehicle, ped or object *can be* relevant to a client. Since it's currently possible that you set the entity owner and then the client never actually syncs the vehicle.

This mainly relies on `hasSynced` relevancy checks to stop this from ever migrating the vehicle until we've properly synchronized the vehicle, since before this could just bounce between multiple different clients.

FiveM

### Checklist

### DIFF
--- a/code/components/citizen-server-impl/src/state/ServerGameState_Scripting.cpp
+++ b/code/components/citizen-server-impl/src/state/ServerGameState_Scripting.cpp
@@ -242,6 +242,45 @@ static void Init()
         return retval;
     }));
 
+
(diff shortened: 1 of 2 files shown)


AngeloneISL:
😍 beautiful

Whisperisgod:
merge!

jrgrimshaw:
This is hot 🔥🔥

PR #4016, opened 2026-06-02
feat(client/graphics): add FPS limit convar

xalva98:
### Goal of this PR

Adds a native client-side FPS limit convar for FiveM.

This allows users to cap their game FPS without relying on external driver/game overlay limiters.

Use case: Some racing servers tell users to cap their FPS so it can still be a competetive environment (Higher FPS usually means more vehicle speed).
Having a simple console command is easier than telling users to change NVIDIA settings for example.

Adds the archived `game_fpsLimit` convar, defaulting to `0` / disabled.

When set to a positive value, the DirectX present path uses the existing frame-time limiter to wait until the configured frame interval has elapsed before presenting the next frame.

Values are clamped to a sane range to avoid accidentally setting unusably low or excessively high limits.

FiveM, Client, Graphics

### Successfully tested on

**Game builds:** bXXXX

**Platforms:** Windows

### Checklist

### Fixes issues

PS: Thanks to Gogsi for the hint

### DIFF
--- a/code/components/rage-graphics-five/src/RenderHooks.cpp
+++ b/code/components/rage-graphics-five/src/RenderHooks.cpp
@@ -67,6 +67,8 @@ static void* g_lastBackbufTexture;
 static bool g_useFlipModel = false;
 
 static bool g_overrideVsync;
+static uint16_t g_fpsLimit = 0;
+static ConVar<uint16_t> gameFpsLimit("game_fpsLimit", ConVar_Archive, 0, &g_fpsLimit);
 
 static void CaptureBufferOutput();
 static void CaptureInternalScreenshot();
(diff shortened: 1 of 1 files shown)


AvarianKnight:
Can you not use a uint8_t instead of an int so then you don't have to use the checks? Also iirc there's constraints you can put on convars so you don't have to do any clamping

xalva98:
> Can you not use a uint8_t instead of an int so then you don't have to use the checks? Also iirc there's constraints you can put on convars so you don't have to do any clamping

I agree with using unsigned and I'll check out the constraints 👍 However I think uint16_t might be better cause some people might actually want 255+ fps.

WLVF:
Would be huge to get this in for racing servers!

xalva98:
@AvarianKnight 
I've looked it up and it seems like we could do

```c++
gameFpsLimit.GetHelper()->SetConstraints(0, 1000);
```

But this would also mean values below 30 are possible, I don't think this is good as this could brick the game for some people so we can't really get around that clamp IMO.

AvarianKnight:
> ```c++
> SetConstraints
> ```

Why would you set it to `0` instead of `30`? The first argument is meant to be the smallest amount

xalva98:
So you can still reset it by using 0

AvarianKnight:
> So you can still reset it by using 0

Ahh, yea then there is no way getting around that.

PR #4020, opened 2026-06-05
fix(server/tcp): properly dispose of connections

### LINKED ISSUE #3931: FXServer crashes with 'Assertion failed: fd_to_send >= 0' when players connect without license identifier
## Description

FXServer crashes with a native libuv assertion failure when one or more players connect without a license identifier. This occurred twice within 23 minutes, each time immediately after the 'no license identifier' warning was logged.

After the first runtime crash, the server entered a **crash loop** — crashing 5 consecutive times during startup at the exact same point before recovering on the 6th attempt.

## Assertion

```
Assertion failed: fd_to_send >= 0 (../deps/uv/src/unix/stream.c: uv__try_write: 791)
```

## Log Evidence

```
[ citizen-server-impl] Server list query returned an error: context deadline exceeded
[ citizen-server-impl] Server list query returned an error: context deadline exceeded
[     script:qbx_core] [WARN] Player XXXXX has no license identifier, cannot check ban status
[     script:qbx_core] [WARN] Player XXXXX has no license identifier, cannot check ban status
║              STDERR║ Assertion failed: fd_to_send >= 0 (../deps/uv/src/unix/stream.c: uv__try_write: 791)
[ citizen-server-impl] network thread hitch warning: timer interval of 382 milliseconds
[ citizen-server-impl] sync thread hitch warning: timer interval of 386 milliseconds
[ citizen-server-impl] server thread hitch warning: timer interval of 388 milliseconds
FXServer crashed.
```

**Startup crash loop (same assertion, 5 times in a row after the first runtime crash):**

```
║              STDERR║ Assertion failed: fd_to_send >= 0 (../deps/uv/src/unix/stream.c: uv__try_write: 791)
FXServer crashed.
```

## Steps to Reproduce

1. Server running with ~60+ active players
2. Server list endpoint begins timing out (observable via 'Server list query returned an error: context deadline exceeded')
3. One or more players connect and receive 'Player XXXXX has no license identifier, cannot check ban status'
4. FXServer crashes immediately after

## Additional Notes

- Lua script performance was verified clean via the built-in FXServer profiler. The last recorded profile (~60 seconds before the crash) shows a max event time of 8ms — no Lua issues.
- Players without license identifiers also appeared earlier in the same session without causing a crash, suggesting a race condition or load-dependent trigger.
- The crash is not caused by any custom Lua scripts.
- Crash dumps are available if needed.

## Environment

- **OS:** Linux (Debian)

AvarianKnight:
This fixes issue #3931

I'm not entirely sure if this is the best approach (and hence I gated this to be only for Linux), but this fixes the life time issue that the TCP Server currently has for Linux.

It should be noted that this is a Linux specific bug, and the same thing cannot happen on Windows since it does direct IPC calls and is sync

On Linux this appends to a shared list and the actual write call gets done later, but because we close the client right after this its possible to hit it when the request handle is invalid and we will crash the server with

```
Assertion failed: fd_to_send >= 0 (../deps/uv/src/unix/stream.c: uv__try_write: 791)
```

Server

### Successfully tested on
**Platforms:**  Linux

Through much pain and suffering I got this to build on WSL

### Checklist

### Fixes issues
Fixes #3931

### DIFF
--- a/code/components/net-tcp-server/src/UvTcpServer.cpp
+++ b/code/components/net-tcp-server/src/UvTcpServer.cpp
@@ -207,9 +207,7 @@ void UvTcpServer::OnConnection(int status)
 #endif
 
 	static char dummyMessage[] = { 1, 2, 3, 4 };
-	m_dispatchPipes[index]->write(*clientHandle, dummyMessage, sizeof(dummyMessage));
-
-	clientHandle->close();
+	m_dispatchPipes[index]->writeAndClose(clientHandle, dummyMessage, sizeof(dummyMessage));
 }
 
 UvTcpChildServer::UvTcpChildServer(UvTcpServer* parent, const std::string& pipeName, const std::array<uint8_t, 16>& pipeMessage, int idx)


AvarianKnight:
This is only meant to be tested on Linux, there is no change here for a windows server

If anyone would like to test this PR you can use this feature build:

https://runtime.fivem.net/artifacts/fivem/build_proot_linux/feature/pull-4020/

Please report back if you have performance degradation, but this should solve the issues with Linux servers crashing.

prikolium-cfx (approved):

PR #4032, opened 2026-06-13
fix(gta-core-five): correct original bike jump patch

### LINKED ISSUE #3024: sv_replaceExeToSwitchBuilds should always be true in the future
### What happened?

After reading https://docs.fivem.net/docs/server-manual/server-commands/#sv_replaceexetoswitchbuilds-newvalue I found that in the future the value will be changed to false
`At some point the default value will be changed to false and, eventually, false will be left as the only option. We will only do that after we are certain that it doesn't affect the users.`

But many vehicle physics behave completely differently on different client exe files. For example, 60% of the tricks of BMX are lost on versions greater than 2845. If you don't know what BMX is, you can refer to https://youtu.be/P8C8El-OqNE

### Expected result

Keep the value as true in the future

### Reproduction steps

...

### Importancy

Slight inconvenience

### Area(s)

FiveM

### Specific version(s)

Server future builds

### Additional information

_No response_

DaniGP17:
### Goal of this PR
Fix `game_originalBikeJump` convar on build 3751. It was meant to restore the pre b3095 bike bunny-hop, a straight-up jump but instead the bike shot off horizontally and barely left the ground.

https://github.com/user-attachments/assets/c869d019-b104-4534-9fbe-17236cfbeac4

A better view of the issue can be also see here: https://www.youtube.com/watch?v=kfEQf1Kp5NQ

The old patch injected too early and jumped over part of the game's own code, leaving the jump's rotation axes and lever-arm uninitialized and (on 3751) overwriting the register that holds the jump force.

The patch now injects later, after the game has finished building the impulse, and only overwrites the three values of the impulse vector with (0, 0, 1) (world up).

FiveM

### Successfully tested on
**Game builds:** 3407, 3751
**Platforms:** Windows

### Checklist

### Fixes issues
fixes #3024

### DIFF
--- a/code/components/gta-core-five/src/PatchBikeJump.cpp
+++ b/code/components/gta-core-five/src/PatchBikeJump.cpp
@@ -30,86 +30,52 @@ static HookFunction hookFunction([]()
 
 	static struct : jitasm::Frontend
 	{
-		uintptr_t m_origBikeJumpBoolLoc{ 0 };
-		uintptr_t m_defaultPathLoc{ 0 };
-		uintptr_t m_continueLoc{ 0 };
-		uintptr_t m_earlyOutLoc{ 0 };
-		uint8_t m_impulseVarStackOffset{ 0 };
+		uintptr_t m_boolLoc{ 0 };
+		uintptr_t m_jmpBackLoc{ 0 };
+		uint8_t m_impulseOffX{ 0 };
 
-		float m_origJumpImpulseVec[4]{ 0.0f, 0.0f, 1.0f, 0.0f };
-		float m_minBunnyHopTime{ 0.0f };
-
-		void Init(uintptr_t origBikeJumpBoolLoc, uintptr_t defaultPathLoc, uintptr_t continueLoc,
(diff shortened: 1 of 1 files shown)


taoletsgo:
❤️bmx savior. When this PR is merged, around a dozen bmxers (a tenfold decrease since 2023 lol) will test the patch

taoletsgo:
My friends and I conducted comprehensive testing on the fix, and we unanimously agreed on the following:
- Across various scenarios (flat ground, ramps, and different rotation), the direction of the final impulse affecting the jump is correct;
- At different frame rates (30fps, 60fps, 150fps), the execution of tricks matches the physics of the original version;
- The patch works effectively across different versions (3407, 3570?/3571);

Therefore, we can reasonably confirm that the changes in the PR are on the right track and have achieved the intended results.

> As there was no placebo control group in this test, there were slight differences of opinion regarding jump height: some felt the height was the same as the original version, while others felt it was 5%–10% higher (based purely on feel). However, since frame rate and vehicle speed also influence jump height, we consider this discrepancy negligible.

dalekenium-cfx (approved):

PR #4039, opened 2026-06-17
fix(vfs/posix): various vfs fixes for Linux

AvarianKnight:
While trying to find what could possibly cause files on disk to become corrupted on Linux I started going through the VFS implementation for Linux and noticed a few issues.

This fixes a few issues here:

Read/ReadBulk/Write don't try to loop until all of their bytes are written

RenameFile returns `0` on success, while we were checking if it was successful by checking if it was `!= 0`

GetLength was still reading the buf even though there isn't valid data in it.

See:

https://man7.org/linux/man-pages/man2/write.2.html https://man7.org/linux/man-pages/man2/read.2.html
https://man7.org/linux/man-pages/man2/pread.2.html https://man7.org/linux/man-pages/man3/fstat.3p.html https://man7.org/linux/man-pages/man2/rename.2.html

Server

### Successfully tested on
**Game builds:** .. 

**Platforms:**  Linux

### Checklist

### DIFF
--- a/code/components/vfs-impl-server/src/PlatformDevice.Posix.cpp
+++ b/code/components/vfs-impl-server/src/PlatformDevice.Posix.cpp
@@ -49,7 +49,7 @@ Device::THandle LocalDevice::Create(const std::string& filename, bool createIfEx
 		flags |= O_APPEND;
 	}
 
-	int fd = open(filename.c_str(), flags, 0755);
+	int fd = open(filename.c_str(), flags, 0644);
 
 	if (fd < 0)
 	{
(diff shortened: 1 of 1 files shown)


code/components/vfs-impl-server/src/PlatformDevice.Posix.cpp
```diff
@@ -63,27 +63,90 @@ size_t LocalDevice::Read(THandle handle, void* outBuffer, size_t size)
...
 {
 	assert(handle != Device::InvalidHandle);
 
-	ssize_t bytesRead = pread(static_cast<int>(handle), outBuffer, size, static_cast<off_t>(ptr));
+	size_t totalBytesRead = 0;
 
-	return (bytesRead < 0) ? -1 : static_cast<size_t>(bytesRead);
+	while (totalBytesRead < size)
+	{
+		ssize_t bytesRead = pread(static_cast<int>(handle), outBuffer + totalBytesRead, size - totalBytesRead);
```
FabianTerhorst: Did the ptr (last arg of original method) got missing in the pread call?

```cpp
pread(static_cast<int>(handle),
                                  static_cast<char*>(outBuffer) + totalBytesRead,
                                  size - totalBytesRead,
                                  ptr + totalBytesRead);
```

code/components/vfs-impl-server/src/PlatformDevice.Posix.cpp
```diff
@@ -63,27 +63,90 @@ size_t LocalDevice::Read(THandle handle, void* outBuffer, size_t size)
...
 {
 	assert(handle != Device::InvalidHandle);
 
-	ssize_t bytesWritten = write(static_cast<int>(handle), buffer, size);
+	size_t totalBytesWritten = 0;
 
-	return (bytesWritten < 0) ? -1 : static_cast<size_t>(bytesWritten);
+	while (totalBytesWritten < size)
+	{
+		ssize_t bytesWritten = write(static_cast<int>(handle), buffer + totalWritten, size - totalBytesWritten);
```
FabianTerhorst: Where is `totalWritten` coming from?

```cpp
ssize_t bytesWritten = write(static_cast<int>(handle),
                            static_cast<const char*>(buffer) + totalBytesWritten,
                            size - totalBytesWritten);
```

code/components/vfs-impl-server/src/PlatformDevice.Posix.cpp
```diff
@@ -63,27 +63,90 @@ size_t LocalDevice::Read(THandle handle, void* outBuffer, size_t size)
...
+		if (bytesWritten < 0)
```
FabianTerhorst, line 139: This only checks for < 0, but a 0 return can also happen and then the loop should break as well.

FabianTerhorst (changes requested):

code/components/vfs-impl-server/src/PlatformDevice.Posix.cpp
AvarianKnight: Yes, this was a bad copy pasta mistake from me.

code/components/vfs-impl-server/src/PlatformDevice.Posix.cpp
AvarianKnight: Typo, Will fix here in a second.

AvarianKnight:
Question, for the `Read` calls if we get an error code should we return what we have already, or should we return `-1` since it is a failure

AvarianKnight:
Other issues here have been fixed. I will build this later today to double check that everything works.

PR #4040, opened 2026-06-19
feat(server/state): add server-side player & vehicle sync events

LeoPuntus:
### Goal of this PR

Give the **server** first-class, event-driven visibility into player and vehicle state
changes that it already receives and decodes in the sync tree, but currently discards.

For years the only way for a resource to know that a player changed model, died, equipped a
weapon, entered a vehicle, or that a vehicle's engine/doors/health changed, has been to
**poll**: iterate every player / every vehicle on a timer and diff native values each pass.
This is expensive, imprecise (you only see a change on the next poll, often seconds later),
and it scales badly with player count.

This PR replaces that pattern with **direct, precise, clean server events**: the change is
reported the exact moment the owning client's clone packet is parsed no polling, no
iteration, no client round-trip.

**Why this matters for security / anti-cheat (the main motivation):**
modern cheats spoof sync packets **straight to the server**, bypassing every client-side
check. Today the server has no hook at the point where that spoofed data lands resources
can only poll natives after the fact. These events fire from inside
`ServerGameState::ProcessClonePacket`, i.e. **at the exact entry point where the (possibly
spoofed) state reaches the server**. That lets an anti-cheat resource observe and validate
every state transition server-side, authoritatively, the instant it happens teleport
jumps via `onPlayerCoordsUpdate`, impossible velocity via `onPlayerVelocityUpdate`,
god-mode/invincibility toggles, weapon swaps, model swaps, etc. The supervision moves from
the (trusted, bypassable) client to the server, where it belongs. (The high-frequency
positional events are opt-in via convar see below so they cost nothing until an
anti-cheat enables them.)

All new code is gated behind `STATE_FIVE` (FiveM only, not RedM). The vehicle block is
additionally gated behind a master convar so server owners can opt out entirely.

#### Example: server-authoritative weapon check

A weapon-spawn cheat equips a weapon the player never legitimately obtained. With
`onPlayerWeaponEquip` the server validates the equip the instant the sync packet arrives,
against its own authoritative inventory, no client trust involved:

```lua
-- 'source' is the player; fires only on a real weapon change, not on a polling timer.
AddEventHandler('onPlayerWeaponEquip', function(oldWeapon, newWeapon)
    local src = source
    if newWeapon == 0 or newWeapon == `WEAPON_UNARMED` then return end

    -- InventoryHasWeapon = your server-side inventory, the single source of truth.
    if not InventoryHasWeapon(src, newWeapon) then
        DropPlayer(src, ('illegal weapon equip: %08X'):format(newWeapon))
    end
end)
```

**The game-changer is not that the check is now possible** (it always was) but *how* you get
there. The old way to catch this was a loop on a timer over **every** player, calling a
fistful of natives each (`GetSelectedPedWeapon` and friends) just to *maybe* notice a change
one poll interval late (typically seconds), multiplied by player count, on every pass, forever.
This PR inverts it: the server simply **waits for the event**. Zero native calls for idle
players, no iteration, no periodic scan, and the handler runs exactly once at the precise
moment the (possibly spoofed) weapon actually changed. The same inversion applies to every other event here, model swaps,
vehicle mods, health/armour, teleports, etc.

#### Real-world example: txAdmin's server-side player list

This is not hypothetical. txAdmin's own
[`sv_playerlist.lua`](https://github.com/citizenfx/txAdmin/blob/master/resource/sv_playerlist.lua)
keeps its player list fresh by **polling**: a thread loops over every online player on a
timer (dynamic interval, **1500ms to 5000ms** depending on player count) and calls roughly
six natives per player each pass to read health, vehicle and coords.

**Before (polling, simplified from txAdmin):**

```lua
CreateThread(function()
    while true do
        for _, serverId in ipairs(GetPlayers()) do
            local ped = GetPlayerPed(serverId)
            if ped and DoesEntityExist(ped) then
                cache[serverId] = {
                    health = GetPedHealthPercent(ped),
                    veh    = GetVehiclePedIsIn(ped),          -- + GetVehicleType(veh)
                    coords = GetEntityCoords(ped),
                }
            end
        end
        Wait(delay) -- 1500..5000ms, recomputed from playersOnline
    end
end)
```

**After (event-driven cache, updated only on real change):**

```lua
local cache = {}  -- authoritative, no timer, no full scan

AddEventHandler('onPlayerHealthChange', function(_, newHealth)
    local p = cache[source]; if p then p.health = newHealth end
end)
AddEventHandler('onPlayerEnterVehicle', function(vehicle) -- + onPlayerExitVehicle
    local p = cache[source]; if p then p.veh = vehicle end
end)
-- live map only: opt-in + rate-limited, fires only while the player is moving
AddEventHandler('onPlayerCoordsUpdate', function(_, _, _, x, y, z)
    local p = cache[source]; if p then p.coords = vector3(x, y, z) end
end)
```

**Why this is faster (cost model, not a benchmark).** The two approaches differ in *where*
and *how often* the work happens:

| | Polling (txAdmin today) | Event-driven |
|---|---|---|
| When work runs | on a timer (every few seconds), unconditionally | only when a value actually changes (no timer) |
| Work each time | `O(players)`: one pass over the whole player list | `O(1)`: a single cache write for the one field that changed |
| Per player, per pass | ~6 native calls (`GetPlayerPed`, `DoesEntityExist`, `GetPedHealthPercent`, `GetVehiclePedIsIn`, `GetVehicleType`, `GetEntityCoords`) | none |
| Cost of an idle/AFK player | full, on every pass | zero |
| Data freshness | up to one interval stale (1.5s–5s) | immediate |

Each polled player is expensive because every one of those natives is a **Lua → C++
scripting-runtime boundary crossing**: argument marshalling, an entity-handle lookup, then the
read. Polling pays that per player, per native, on every pass, forever, even for players who
have not moved or changed anything. txAdmin even *lengthens* its interval as players grow
specifically to keep this scan from hitching, which is the trade-off the event model removes.

The key insight: **the server already decoded these exact values** while parsing the clone
packet in `ProcessClonePacket` health, vehicle and coords are read there either way.
Polling discards that work and re-fetches the same data through natives a few seconds later;
the event simply hands you the value the server already has. So the event-driven cache
replaces "scan N players × M natives on a timer" with "**one hash-map write when a field
genuinely changes**": no native call, no entity lookup, no `O(players)` loop, no staleness.
The cost is no longer a function of player count × poll frequency; it is a function of how
often state actually changes and an idle player costs nothing at all.

The only continuous piece is the live-map coordinate feed (`onPlayerCoordsUpdate`), which is
opt-in and rate-limited (fires only while a player moves, capped by
`sv_playerCoordsUpdateMinInterval`). Even that delivers coords the server already decoded,
instead of calling `GetEntityCoords` on every player on a timer.

Two new blocks are added in `ServerGameState::ProcessClonePacket`, right after the sync
tree is parsed:

1. **Player block** runs when `entity->type == NetObjEntityType::Player`.
2. **Vehicle block** runs for the vehicle entity types, gated by `sv_enableVehicleSyncEvents`.

Each event follows the same pattern:

- `parsingType == 1` (create) → seed the per-entity cache only, never fire (no false event
  on initial spawn).
- `parsingType == 2` (update) → compare the freshly parsed value against the cached previous
  value; if it changed, `QueueEvent2(...)`; then update the cache.
- A `…Valid` flag skips the first empty comparison; floats use an epsilon compare against
  sync jitter.
- High-frequency/continuous events (coords, velocity, camera, ping, steering, angular
  velocity, heli/plane control) are **threshold + rate-limited** (configurable minimum
  interval in ms) so they never flood the scheduler.

**Per-entity caches** are added to `struct SyncEntityState` (under `STATE_FIVE`) holding the
previous value of every tracked field. **21 convars** are registered in the existing
`InitFunction` to enable/tune the high-frequency events. All **continuous / high-frequency**
events (player coords/ping/velocity/camera and vehicle steering/angular-velocity/heli-plane
control) default to **OFF**: they can fire on every sync update while a player/vehicle is
moving (capped by their rate-limit) and would otherwise burn CPU serialising events even when
no resource listens; server owners opt in per group. Discrete events (the vast majority) fire
only on an actual change and are always on.

**`SyncTrees_Five.h`** change is minimal: a handful of weapon/death fields that the parser
already decoded but discarded are now stored into `data.*` so the event code can read them.
No new bit reads, no change to bit alignment. The `spectatorId` parse is left
**byte-identical** to upstream; a dedicated `bool isSpectating` field was added instead of
repurposing `spectatorId`, so no existing parsing behaviour changes. No upstream
`maxHealth`/`maxArmour` parsing or fields were removed.

#### Player events (28)

> Every event additionally provides the player as `source` / `[FromSource]`.

| Event | Parameters |
|---|---|
| `onPlayerModelUpdate` | `oldModel: number, newModel: number` |
| `onPlayerCoordsUpdate` | `oldX, oldY, oldZ, newX, newY, newZ: number, heading: number, msSinceLastUpdate: number` |
| `onPlayerAnimStart` | `scriptCommand: number, scriptTaskStage: number` |
| `onPlayerWeaponEquip` | `oldWeapon: number, newWeapon: number, tintIndex: number, componentCount: number, weaponState: number` |
| `onPlayerEnterVehicle` | `vehicle: number, seat: number` |
| `onPlayerExitVehicle` | `vehicle: number, seat: number` |
| `onPlayerSeatChange` | `vehicle: number, oldSeat: number, newSeat: number` |
| `onPlayerPingUpdate` | `oldPing: number, newPing: number, msSinceLastUpdate: number` |
| `onPlayerHealthChange` | `oldHealth: number, newHealth: number` |
| `onPlayerArmourChange` | `oldArmour: number, newArmour: number` |
| `onPlayerDeath` | `cause: number, source: number, deathState: number, killedByStealth: boolean, killedByTakedown: boolean` |
| `onPlayerFlashlightToggle` | `on: boolean` |
| `onPlayerActionModeToggle` | `on: boolean` |
| `onPlayerHandcuffChange` | `on: boolean` |
| `onPlayerStealthChange` | `on: boolean` |
| `onPlayerStrafeChange` | `on: boolean` |
| `onPlayerRagdollChange` | `on: boolean` |
| `onPlayerWantedLevelChange` | `oldLevel: number, newLevel: number` |
| `onPlayerWantedDetailChange` | `fakeWantedLevel: number, isWanted: number, isEvading: number` |
| `onPlayerTeamChange` | `oldTeam: number, newTeam: number` |
| `onPlayerFriendlyFireToggle` | `on: boolean` |
| `onPlayerInvincibleToggle` | `on: boolean` |
| `onPlayerSpectate` | `spectating: boolean` |
| `onPlayerSuperJumpToggle` | `on: boolean` |
| `onPlayerVoiceProximityChange` | `x: number, y: number, z: number` |
| `onPlayerModifierChange` | `modifier: string, oldValue: number, newValue: number` |
| `onPlayerVelocityUpdate` | `x: number, y: number, z: number` |
| `onPlayerCameraUpdate` | `camMode: number, relativeHeading: number, relativePitch: number` |

#### Vehicle events (27)

> `source` = the vehicle's current owner; the first argument `vehicle` is the vehicle network id.

| Event | Parameters |
|---|---|
| `onVehicleEngineToggle` | `vehicle: number, on: boolean, starting: boolean` |
| `onVehicleSirenToggle` | `vehicle: number, on: boolean` |
| `onVehicleHandbrakeToggle` | `vehicle: number, on: boolean` |
| `onVehicleLightsChange` | `vehicle: number, lightsOn: boolean, highbeamsOn: boolean, headlightsColour: number` |
| `onVehicleDoorsChange` | `vehicle: number, doorsOpen: number` |
| `onVehicleLockChange` | `vehicle: number, lockStatus: number` |
| `onVehicleRadioChange` | `vehicle: number, station: number` |
| `onVehicleStationaryChange` | `vehicle: number, on: boolean` |
| `onVehicleHealthChange` | `vehicle: number, engineHealth: number, petrolTankHealth: number, bodyHealth: number, health: number` |
| `onVehicleTyreBurst` | `vehicle: number, tyreIndex: number, status: number` |
| `onVehicleBulletDamage` | `vehicle: number, damaged: boolean` |
| `onVehicleWindowBreak` | `vehicle: number, windowIndex: number, broken: boolean` |
| `onVehicleColourChange` | `vehicle: number, primaryColour: number, secondaryColour: number` |
| `onVehicleModChange` | `vehicle: number` |
| `onVehiclePlateChange` | `vehicle: number, plate: string` |
| `onVehicleLiveryChange` | `vehicle: number, livery: number, roofLivery: number` |
| `onVehicleNeonChange` | `vehicle: number, red: number, green: number, blue: number, left: boolean, right: boolean, front: boolean, back: boolean` |
| `onVehicleDirtChange` | `vehicle: number, dirtLevel: number` |
| `onVehicleWindowTintChange` | `vehicle: number, tint: number` |
| `onVehicleSteeringChange` | `vehicle: number, angle: number` |
| `onVehicleAngVelocity` | `vehicle: number, x: number, y: number, z: number` |
| `onHeliHealthChange` | `vehicle: number, mainRotor: number, rearRotor: number, body: number, gasTank: number, engine: number, boomBroken: boolean` |
| `onHeliControlChange` | `vehicle: number, yaw: number, pitch: number, roll: number, throttle: number` |
| `onPlaneStateChange` | `vehicle: number, landingGearState: number, lockOnState: number` |
| `onPlaneControlChange` | `vehicle: number, nozzlePosition: number` |
| `onBoatStateChange` | `vehicle: number, wreckedAction: number, lockedToXY: boolean` |
| `onTrainStateChange` | `vehicle: number, trainState: number, trackId: number, direction: boolean` |

#### Convars (21)

> Continuous events are **off by default** (`enable` = `false`); discrete events are always on.
> Vehicle continuous events require both `sv_enableVehicleSyncEvents` **and** their own `enable`.

| Convar | Default | Purpose |
|---|---|---|
| `sv_enablePlayerCoordsUpdateEvent` | `false` | enable `onPlayerCoordsUpdate` |
| `sv_playerCoordsUpdateThreshold` | `1.0` | min squared-metres moved to fire |
| `sv_playerCoordsUpdateMinInterval` | `250` | min ms between emissions |
| `sv_enablePlayerPingUpdateEvent` | `false` | enable `onPlayerPingUpdate` |
| `sv_playerPingUpdateThreshold` | `10` | min ping delta (ms) to fire |
| `sv_playerPingUpdateMinInterval` | `1000` | min ms between emissions |
| `sv_enablePlayerVelocityUpdateEvent` | `false` | enable `onPlayerVelocityUpdate` |
| `sv_playerVelocityUpdateThreshold` | `1.0` | min squared (m/s) delta to fire |
| `sv_playerVelocityUpdateMinInterval` | `250` | min ms between emissions |
| `sv_enablePlayerCameraUpdateEvent` | `false` | enable `onPlayerCameraUpdate` |
| `sv_playerCameraUpdateThreshold` | `0.1` | min look delta (rad) to fire |
| `sv_playerCameraUpdateMinInterval` | `500` | min ms between emissions |
| `sv_enableVehicleSyncEvents` | `true` | master switch for **all** vehicle events |
| `sv_enableVehicleSteeringUpdateEvent` | `false` | enable `onVehicleSteeringChange` |
| `sv_vehicleSteeringUpdateThreshold` | `0.05` | min steering delta (rad) to fire |
| `sv_vehicleSteeringUpdateMinInterval` | `250` | min ms between emissions |
| `sv_enableVehicleAngVelocityUpdateEvent` | `false` | enable `onVehicleAngVelocity` |
| `sv_vehicleAngVelocityUpdateThreshold` | `0.5` | min squared (rad/s) delta to fire |
| `sv_vehicleAngVelocityUpdateMinInterval` | `250` | min ms between emissions |
| `sv_enableVehicleControlUpdateEvent` | `false` | enable `onHeliControlChange` / `onPlaneControlChange` |
| `sv_vehicleControlUpdateMinInterval` | `500` | min ms between heli/plane control emissions |

Server, Natives

### Successfully tested on

**Game builds:** all FiveM builds (the 3-bit `weaponState` weapon detail is gated to build >= 3258)

**Platforms:** Windows

### Checklist

### Notes for reviewers

- All events live under `STATE_FIVE`; the entire vehicle block can be disabled at runtime
  with `sv_enableVehicleSyncEvents false`.
- Continuous / high-frequency events default to **disabled** so there is zero CPU/scheduler
  cost out of the box; each group has an `enable`, `threshold` and `minInterval` convar and is
  opt-in. Discrete events (change-only) are always on and effectively free.
- `onPlayerAnimStart` only carries the scripted task command hash + stage the anim
  dict/clip strings are not replicated server-side.
- This is a large change; if the maintainers would prefer it split (player events first,
  vehicle events in a follow-up), I'm glad to break it up.

### DIFF
--- a/code/components/citizen-server-impl/include/state/ServerGameState.h
+++ b/code/components/citizen-server-impl/include/state/ServerGameState.h
@@ -325,12 +325,29 @@ struct CPedGameStateNodeData
 
 	int curWeapon;
 
+	// Weapon detail decoded alongside curWeapon (previously read and discarded in
+	// Parse). Surfaced through onPlayerWeaponEquip.
+	int weaponTintIndex;       // -1 = no tint synced this packet
+	int weaponComponentCount;  // number of attached weapon components
+	bool weaponHasAmmo;
+	bool weaponVisible;
+	uint8_t weaponState;       // reload/charge state, build >= 3258 only (else 0)
+
 	bool isHandcuffed;
 	bool actionModeEnabled;
 	bool isFlashlightOn;
 
+	// Death detail decoded in the ped game-state node (previously discarded).
+	// Surfaced through onPlayerDeath.
+	int deathState;            // 0 = alive, non-zero = dying/dead
+	bool killedByStealth;
+	bool killedByTakedown;
+
 	inline CPedGameStateNodeData()
-		: lastVehicle(-1), lastVehicleSeat(-1), lastVehiclePedWasIn(-1)
+		: lastVehicle(-1), lastVehicleSeat(-1), lastVehiclePedWasIn(-1),
+		  weaponTintIndex(-1), weaponComponentCount(0), weaponHasAmmo(false),
+		  weaponVisible(false), weaponState(0), deathState(0),
+		  killedByStealth(false), killedByTakedown(false)
 	{
 	}
 };
@@ -662,6 +679,7 @@ struct CPlayerGameStateNodeData
 
 	bool neverTarget;
 	int spectatorId;
+	bool isSpectating;
 
 	bool randomPedsFlee;
 	bool everybodyBackOff;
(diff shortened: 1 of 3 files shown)


dBozee:
giphy 1

MaskeZenY:
Wow

PiveteMenu:
homelander-the-boys-s4

ook3D:
This is a large change; if the maintainers would prefer it split (player events first,
vehicle events in a follow-up), I'm glad to break it up.

Mfw you don't even read the slop youre pasting

LeoPuntus:
the description was AI-generated, purely for convenience, that split note was a deliberate suggestion to the maintainers, nothing more.

ook3D:
If you say so. 
image

image

LeoPuntus:
those are just copy-pasted section headers, not rewritten each time if you look around you'll find the same comment in plenty of other places in the file, at this point you're just nitpicking.

ook3D:
> those are just copy-pasted section headers, not rewritten each time if you look around you'll find the same comment in plenty of other places in the file, at this point you're just nitpicking.

Youre not fooling anybody lmao

PiveteMenu:
youre life is based on blender update...

imLocutor:
so, what u think @radium-cfx @cerium-cfx @prikolium-cfx

gtasnail:
> youre life is based on blender update...

He's legit the only person i see actively contributing to the sollumz project, this is not an insult infact it's a compliment.

ook3D:
> youre life is based on blender update...

Try learning proper English before trying to insult someone

PiveteMenu:
Blender-keeps-crashing-on-PC

radium-cfx:
Thanks for your work on this change! Unfortunately, we’re currently in a feature freeze for FiveM, so we’re unable to merge any changes that modify or extend our APIs at this time.

However, improvements to the server-side sync logic are already planned.

PR #4050, opened 2026-06-27
feat(launcher): Wine/Proton compatibility for the FiveM client

Amzi-01:
- Add CfxIsProton() helper (detects Wine + Steam Proton env vars)
- Remove Wine-forced insecure mode from ComponentLoader; only ARM (xtajit64.dll) still forces sticky.dll
- Add WineCompat.Win32.cpp: installs a GetProcAddress hook before legitimacy.dll and adhesive.dll load, returning nullptr for wine_get_version/wine_get_build_id/wine_get_host_version so those components do not detect Wine and alter their behaviour
- DllGameComponent: graceful sticky.dll fallback if adhesive fails to load under Wine, instead of a hard crash
- EarlyMode: call WineCompat_InstallHooks() before any component loads
- Launcher stability fixes under Wine: skip SetProcessMitigationPolicy, EarlyLdrBlock (NVIDIA GameFilter), cross-process MiniDump operations, CLSID_TaskbarList, and false "outdated Windows" OS version warning
- GameSelect: auto-detect GTA V path via STEAM_COMPAT_INSTALL_PATH; fall back to GetOpenFileNameW when IFileDialog fails under Wine
- ViabilityChecks: skip MediaFeatureCheck under Wine (stub mfreadwrite)

### Goal of this PR
Allow the FiveM client launcher to start and run reliably under Wine/Proton on Linux. Currently several launcher and component code paths either assume a native Windows environment or deliberately change behaviour when Wine is detected, which leads to hard crashes, forced insecure mode, and a number of spurious failures/warnings. This PR makes the launcher Wine-aware so Linux users get a working, secure client without affecting native Windows behaviour.

The core of the change is a small compatibility layer that runs before any anti-cheat/legitimacy component loads:

- A `CfxIsProton()` helper centralises Wine/Proton detection (Wine env + Steam Proton env vars).
- `WineCompat.Win32.cpp` installs a `GetProcAddress` hook early (called from EarlyMode via `WineCompat_InstallHooks()`) that returns `nullptr` for `wine_get_version` / `wine_get_build_id` / `wine_get_host_version`, so `legitimacy.dll` no longer special-cases Wine.
- The previous Wine-forced insecure mode in `ComponentLoader` is removed — only ARM (`xtajit64.dll`) still forces `sticky.dll`. `DllGameComponent` now falls back to `sticky.dll` gracefully if `adhesive.dll` fails to load under Wine instead of crashing.
- Launcher-side stability fixes guard a set of Windows-only operations that fail or misbehave under Wine: `SetProcessMitigationPolicy`, `EarlyLdrBlock` (NVIDIA GameFilter), cross-process MiniDump operations, `CLSID_TaskbarList`, the false "outdated Windows" version warning, and `MediaFeatureCheck` in `ViabilityChecks` (stub `mfreadwrite`).
- `GameSelect` auto-detects the GTA V install path from `STEAM_COMPAT_INSTALL_PATH` and falls back to `GetOpenFileNameW` when `IFileDialog` fails under Wine.

All Wine-specific behaviour is gated behind the detection helper, so native Windows paths are unchanged.

FiveM, Launcher

### Successfully tested on
**Game builds:** 3258
**Platforms:** Windows, Linux

### Checklist

### AI disclosure
Claude (Anthropic) was used solely for code review purposes. All code was written and tested by me; no code in this PR was authored by an AI.

### DIFF
--- a/code/client/citicore/ComponentLoader.cpp
+++ b/code/client/citicore/ComponentLoader.cpp
@@ -172,27 +172,18 @@ void ComponentLoader::InitializeWithString(std::string_view cacheBuf)
 				continue;
 			}
 
-			if (CfxIsWine() || GetModuleHandleW(L"xtajit64.dll") != nullptr)
+			// Windows on ARM (xtajit64) cannot run the anti-cheat components —
+			// force insecure mode.  Under Wine we now attempt to load Adhesive
+			// normally; DllGameComponent::CreateComponent will fall back to
+			// sticky.dll if Adhesive fails to initialise at runtime.
+			if (GetModuleHandleW(L"xtajit64.dll") != nullptr)
 			{
 				auto cfxState = CfxState::Get();
 				if (cfxState->IsMasterProcess())
 				{
-					std::wstring environmentType;
-
-					if (CfxIsWine())
-					{
-						environmentType = L"Wine";
-					}
-					else if (GetModuleHandleW(L"xtajit64.dll") != nullptr)
-					{
-						environmentType = L"Windows on ARM";
-					}
-
-					MessageBoxW(NULL, 
-						va(
-							L"The game is running in insecure mode because %s is not supported by the anti-cheat components at this time.\n"
-							L"Most servers, as well as some authentication features will be unavailable.",
-							environmentType),
+					MessageBoxW(NULL,
+						L"The game is running in insecure mode because Windows on ARM is not supported by the anti-cheat components at this time.\n"
+						L"Most servers, as well as some authentication features will be unavailable.",
 						L"Cfx.re: Insecure mode", MB_OK | MB_ICONWARNING);
 				}
 
--- a/code/client/citicore/DllGameComponent.Win32.cpp
+++ b/code/client/citicore/DllGameComponent.Win32.cpp
@@ -20,6 +20,7 @@
 #include <Hooking.Aux.h>
 
 #include <MinHook.h>
+#include "LaunchMode.h"
 
 static NTSTATUS(NTAPI* g_origRtlRaiseException)(_In_ PEXCEPTION_RECORD ExceptionRecord);
 static BOOL(NTAPI* g_origZwQueryDebugFilterState)(_In_ ULONG ComponentId, _In_ ULONG Level);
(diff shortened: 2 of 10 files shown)


Gogsi:
A few months ago I implemented only the bare minimum changes to get FiveM to launch in insecure mode here https://github.com/Gogsi/fivem/commit/8ef2fccd7ddabf21f459ec906363f1c05447d876. Those changes were:
- Disabling the new Updater UI (TenUI), instead falling back to the simple Windows 8 update dialog. The lack of WindowsXamlManager made me not even bother with getting TenUI to work, when I noticed that the old code still exists.

- Disabling the RtlLookupFunctionTable patch in `SEHTableHandler.Win32.cpp`. This just enables a debugging feature and it's an undocumented, memory layout-dependent patch, so I got rid of it.
     
- Disabling the NtGlobalFlag patch in DllGameComponent.Win32.cpp. Another memory layout-dependent debug feature.
    
With those changes the game launched fine in insecure mode, tested on both latest `wine-staging` and Proton-GE 10.17 and Proton-CachyOS latest on multiple devices in May 2026.

Your PR has some nice improvements like disabling the Nvidia Game Filters (the check didn't exist when I wrote my code), improving the file selection dialog (it was kind of broken in my branch) and disabling more incompatible patches with the dump server. I'm not sure to the extent that you have tested it, but I don't think forgoing base Wine is a good idea, even if Proton is universally accepted to be "better" for games.

I recommend you also incorporate [this patch](https://github.com/Gogsi/fivem/commit/802559091ca1274b29bee15b07514b46388bd2da) which enables NUI Shared Resources/Texture on Wine, as later versions have support for it. This will greatly improve UI rendering and it removed the weird artifacting in the main menu I experienced in my tests.
 
What likely won't get merged any time soon is enabling adhesive under Wine/Proton. Even if it does load fine, which we can't know until it's tested, it is dangerous to enable, as anti-cheats often rely on subtle side effects to detect the presence of a cheat. Differences when running under Wine (such as different memory layout) can trigger detections even when you don't expect them to. It would take significant effort to test that no detection triggers unexpectedly and future detections would also need to be tested. We can wait for an official response from Cfx, but from the few private conversations I've had, it seem unlikely they would be willing to invest such effort, given the upcoming release of FiveM for GTA V Enhanced.

code/client/citicore/ComponentLoader.cpp
```diff
@@ -172,27 +172,18 @@ void ComponentLoader::InitializeWithString(std::string_view cacheBuf)
 				continue;
 			}
 
-			if (CfxIsWine() || GetModuleHandleW(L"xtajit64.dll") != nullptr)
```
radium-cfx, line 175: We can't enable the Anticheat on Linux, so this edit needs to be removed.

radium-cfx (changes requested):

code/client/citicore/ComponentLoader.cpp
Amzi-01: adhesive and legitimacy dlls from my research are compatible with the latest versions of wine/proton my changes are designed to not interfere with windows versions at all im gathering this is more of a policy issue rather than a technical limitation? as there is quite a lot of people that would like to play FiveM on linux

Amzi-01: tenor

NicolaiVdS: @radium-cfx 👀

TopinambourSansSel: Interest has not dropped at all! :)

aindrigo: Agreed, it's been almost 2 months with no word from CFX on this

Gogsi: The requested changes were not made, what word do you expect

tommy141x: The requested change goes against the purpose of the PR..

TopinambourSansSel: The requested change defeats the PR. And as Amzi-01 mentioned, the requested change does not come from a technical limitation but a policy.

Amzi-01: all CFX and R* is doing by denying this pr is shooting them selfs in the foot this PR dont effect Windows at all, and ill be honest the past few months this repo has been nothing but a glorified download page hardly any PRs have been murged and if they are its from those who are friends with the CFX team and also now it seems that they arent even working off this repo anymore by looking at the amount of pushes they merged from their internal repo, it feels like R* is going against everything CFX used to stand for

PR #4058, opened 2026-07-07
fix(gta-core-five): door auto-open framerate independent

DaniGP17:
### Goal of this PR
Make automatic sliding doors, gates and barriers actually open at high framerates. On high-FPS PCs they get stuck and you have to push them, this fix it and now they open on their own like they're supposed to.

https://github.com/user-attachments/assets/826ba35d-7a49-422c-af3a-e55208a828ef

Sliding doors divided their opening force by Max(0.03, timestep), so above ~33 FPS the force got scaled way down, with this I zero the 0.03 so it uses the real timestep. And every door type skipped ActivatePhysics() when the per-frame delta was under a fixed epsilon (which never happens at high FPS), so the collider never woke up.

FiveM

### Successfully tested on
**Game builds:** 3407

**Platforms:** Windows, Linux

### Checklist

### Fixes issues
https://forum.cfx.re/t/sliding-doors-and-gates-not-opening-unless-force-applied-to-them-first/

### DIFF
--- a/code/components/gta-core-five/src/PatchDoorFrameRate.cpp
+++ b/code/components/gta-core-five/src/PatchDoorFrameRate.cpp
@@ -1,0 +1,31 @@
+#include "StdInc.h"
+#include "Hooking.Patterns.h"
+struct PatternPair
+{
+	std::string_view pattern;
+	int offset;
+};
+
+static HookFunction hookFunction([]()
+{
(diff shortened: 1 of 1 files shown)


prikolium-cfx:
Good job as always, Daniel!

prikolium-cfx (approved):

PR #4061, opened 2026-07-10
feat(gta-core-five): raise the shop ped apparel item limit

xalva98:
### Goal of this PR

Raise the limit on how much addon clothing (`SHOP_PED_APPAREL_META_FILE` items) a server can use. Today, once a server reaches roughly 65,520 clothing items in total across all packs, players crash while joining — and removing any handful of items anywhere makes it work again, which is what shows it's a fixed limit and not a broken item.

That limit can't be raised with any convar or `increase_pool_size`, and it's hit well before the streaming pools run out — so in practice it's the real ceiling on how much clothing a server can add (around ~29.5k addon items on build 3095, since the base game already uses a large chunk of it). With it removed, raising the relevant pools finally lets servers add far more clothing. It should also help some setups that currently run into what looks like a "YMT limit," since this limit is reached sooner than that.

The change is opt-in and **off by default**.

The game keeps a single list of every clothing item, looked up by name. That list can only hold about 65,520 entries, and the release build doesn't guard against going past it — so once a server crosses that number the overflow corrupts memory and crashes the client. It happens on join because that's when every clothing pack loads and the list gets rebuilt.

This PR keeps its own, larger list on the side and points the game at it — both the code that rebuilds the list and the code that looks items up in it. When the fix is off (the default), everything behaves exactly as it does today.

Enable it server-side with:

```
setr game_enableExtendedApparelShopLimit true
```

Here's a clip of a large clothing pack (over 30k+ addon clothing) with the convar set to **false**
https://youtu.be/QoKaBW9ue0E
As you can see the game crashes as soon as the clothing packs load.

Here's a clip of that same clothing pack with the convar set to **true**:
https://youtu.be/CakPxO7zUjc
The game fully loads in and all 1400+ clothing pieces for jackets are available (and more is possible probably)

FiveM, Client

### Successfully tested on

**Game builds:** 3258

**Platforms:** Windows

### Checklist

### Fixes issues
https://discord.com/channels/779705925577080842/842554061734936596/1525121266275586211
[b1da67e0-faf7-471a-a8f7-5f5fbf2df96b.dmp](https://github.com/user-attachments/files/29907657/b1da67e0-faf7-471a-a8f7-5f5fbf2df96b.dmp)

PS: Even if my fix is not accepted I think everyone would highly appreciate this limit being increased to make full use of the pool size increases

### DIFF
--- a/code/components/gta-core-five/src/PatchShopPedApparelLookup.cpp
+++ b/code/components/gta-core-five/src/PatchShopPedApparelLookup.cpp
@@ -1,0 +1,197 @@
+#include <StdInc.h>
+#include <Hooking.h>
+#include <Hooking.FlexStruct.h>
+#include <Hooking.Stubs.h>
+
+#include <CoreConsole.h>
+#include <CrossBuildRuntime.h>
+#include <ICoreGameInit.h>
+
+#include <cassert>
+#include <cstdint>
+#include <iterator>
+#include <EASTL/hash_map.h>
+
+// Shop ped apparel items are resolved by name through one engine map whose backing
+// array counts entries in a u16; once the total <Item> count crosses ~65,520 it
+// overflows and corrupts the heap on join. This diverts the rebuild and both lookup
+// helpers to an unbounded side table (eastl::hash_map), leaving the engine map empty.
+
+static bool g_extendedApparelLimitEnabled = false;
+static bool g_builtApparelIndex = false;
+
+static int32_t g_mgrFilesPtrOffset;
+static int32_t g_mgrFilesCountOffset;
+static int32_t g_mgrMapSortedOffset;
+static int32_t g_mgrMapCountOffset;
+
+static constexpr int32_t kItemNameHashOffset = 0x20;
+
+// The three item sub-arrays in a ShopPedApparel file (component/prop/outfit).
+struct ItemArrayLayout
+{
+	int32_t ptrOffset;
+	int32_t countOffset;
+	size_t  stride;
+	uint8_t kind;
+};
+
+static constexpr ItemArrayLayout kItemArrays[3] = {
+	{ 0x28, 0x30, 144, 0 },
+	{ 0x38, 0x40, 136, 1 },
+	{ 0x48, 0x50, 160, 2 },
+};
+
+// Packed to match what the engine getters decode: kind[0:8) file[8:24) item[32:48).
+static inline uint64_t EncodeLocation(uint8_t kind, uint16_t fileIndex, uint16_t itemIndex)
+{
+	return static_cast<uint64_t>(kind)
+	     | (static_cast<uint64_t>(fileIndex) << 8)
+	     | (static_cast<uint64_t>(itemIndex) << 32);
+}
+
+static eastl::hash_map<uint32_t, uint64_t> g_apparelIndex;
+
+static void (*g_origRebuild)(hook::FlexStruct* manager);
+
+static void ApparelIndex_Rebuild(hook::FlexStruct* manager)
+{
+	g_builtApparelIndex = true;
+	g_apparelIndex.clear();
+
+	if (!g_extendedApparelLimitEnabled)
+	{
+		g_origRebuild(manager);
+		return;
+	}
+
+	uint16_t fileCount = manager->Get<uint16_t>(g_mgrFilesCountOffset);
+	auto files = manager->Get<hook::FlexStruct**>(g_mgrFilesPtrOffset);
+
+	for (uint16_t fileIndex = 0; fileIndex < fileCount; fileIndex++)
+	{
+		hook::FlexStruct* file = files[fileIndex];
+		if (!file)
+		{
+			continue;
+		}
+
+		for (const auto& arr : kItemArrays)
+		{
+			uint16_t count = file->Get<uint16_t>(arr.countOffset);
+			auto items = file->Get<uint8_t*>(arr.ptrOffset);
+			if (!items)
+			{
+				continue;
+			}
+
+			for (uint16_t itemIndex = 0; itemIndex < count; itemIndex++)
+			{
+				auto item = reinterpret_cast<hook::FlexStruct*>(items + static_cast<size_t>(itemIndex) * arr.stride);
+				uint32_t hash = item->Get<uint32_t>(kItemNameHashOffset);
+				// keep the first occurrence, matching the engine's lower-bound on ties
+				g_apparelIndex.emplace(hash, EncodeLocation(arr.kind, fileIndex, itemIndex));
+			}
+		}
+	}
+
+	// leave the engine map empty + sorted so nothing grows or reads it
+	manager->Set<uint16_t>(g_mgrMapCountOffset, 0);
+	manager->Set<uint8_t>(g_mgrMapSortedOffset, 1);
+}
+
+// The engine has two compiled copies of the lookup helper; divert both.
+static char* (*g_origLookupA)(hook::FlexStruct* map, uint32_t* nameHash);
+static char* (*g_origLookupB)(hook::FlexStruct* map, uint32_t* nameHash);
+
+static char* ResolveFromIndex(hook::FlexStruct* map, uint32_t nameHash)
+{
+	auto it = g_apparelIndex.find(nameHash);
+	if (it == g_apparelIndex.end())
+	{
+		return nullptr;
+	}
+
+	uint64_t packed = it->second;
+	uint8_t kind = static_cast<uint8_t>(packed & 0xFF);
+	uint16_t fileIdx = static_cast<uint16_t>((packed >> 8) & 0xFFFF);
+	uint16_t itemIdx = static_cast<uint16_t>((packed >> 32) & 0xFFFF);
+
+	// re-validate against the live manager so a stale entry can't point at an unloaded file
+	auto manager = &map->At<hook::FlexStruct>(-g_mgrMapSortedOffset);
+	if (fileIdx >= manager->Get<uint16_t>(g_mgrFilesCountOffset))
+	{
+		return nullptr;
+	}
+
+	hook::FlexStruct* file = manager->Get<hook::FlexStruct**>(g_mgrFilesPtrOffset)[fileIdx];
+	if (!file)
+	{
+		return nullptr;
+	}
+
+	// kind is always 0-2; the modulo keeps the index in range (silences C6385)
+	const ItemArrayLayout& arr = kItemArrays[kind % std::size(kItemArrays)];
+	if (itemIdx >= file->Get<uint16_t>(arr.countOffset))
+	{
+		return nullptr;
+	}
+
+	return reinterpret_cast<char*>(&it->second);
+}
+
+static char* ApparelIndex_LookupA(hook::FlexStruct* map, uint32_t* nameHash)
+{
+	return g_extendedApparelLimitEnabled ? ResolveFromIndex(map, *nameHash) : g_origLookupA(map, nameHash);
+}
+
+static char* ApparelIndex_LookupB(hook::FlexStruct* map, uint32_t* nameHash)
+{
+	return g_extendedApparelLimitEnabled ? ResolveFromIndex(map, *nameHash) : g_origLookupB(map, nameHash);
+}
+
+static void OnExtendedApparelLimitChange(internal::ConsoleVariableEntry<bool>* var)
+{
+	// Lock the value in once the index is built; changing it mid-session - or when
+	// the convar resets on disconnect - would desync the lookups from the index.
+	if (g_builtApparelIndex)
+	{
+		return;
+	}
+
+	g_extendedApparelLimitEnabled = var->GetRawValue();
+}
+
+static HookFunction hookFunction([]()
+{
+	static ConVar<bool> extendedApparelLimitVar("game_enableExtendedApparelShopLimit", ConVar_Replicated, false, &OnExtendedApparelLimitChange);
+
+	if (!xbr::IsGameBuildOrGreater<3258>())
+	{
+		return;
+	}
+
+	// Session over (and its apparel content gone) - unpin so the next one's value applies.
+	Instance<ICoreGameInit>::Get()->OnShutdownSession.Connect([]()
+	{
+		g_builtApparelIndex = false;
+		g_extendedApparelLimitEnabled = extendedApparelLimitVar.GetValue();
+	});
+
+	// Read the manager offsets from the displacements RebuildApparelLookup uses.
+	g_mgrMapSortedOffset = *hook::get_pattern<uint32_t>("33 D2 C6 81 ? ? ? ? 01 48 8B F9", 4);
+	g_mgrMapCountOffset = *hook::get_pattern<uint32_t>("48 8B F9 66 89 91 ? ? ? ? 8B F2", 6);
+	g_mgrFilesCountOffset = *hook::get_pattern<uint8_t>("0F B7 47 ? FF C6 89 75 ? 3B F0", 3);
+	g_mgrFilesPtrOffset = *hook::get_pattern<uint8_t>("48 8B 47 ? 8B CE 44 8B E2 4C 8B 34 C8", 3);
+
+#ifdef _DEBUG
+	// Catch a struct-layout change in a future game build.
+	assert(*hook::get_pattern<uint8_t>("48 8D 0C C0 49 8B 46 ? 45 0B EF", 7) == kItemArrays[0].ptrOffset);
+	assert(*hook::get_pattern<uint8_t>("4C 8B 34 C8 66 41 3B 56 ? 73", 8) == kItemArrays[0].countOffset);
+	assert(*hook::get_pattern<uint8_t>("44 89 6D D8 8B 5C C8 ? 8B 4D DC", 7) == kItemNameHashOffset);
+#endif
+
+	g_origRebuild = hook::trampoline(hook::get_pattern("40 55 53 56 57 41 54 41 55 41 56 41 57 48 8B EC 48 83 EC ? 33 D2"), ApparelIndex_Rebuild);
+	g_origLookupA = hook::trampoline(hook::get_pattern("48 89 5C 24 ? 57 48 83 EC ? 45 33 C9 48 8B FA 48 8B D9 66 44 39 49 ? 76 ? 8B 02 48 8B 49 ? 4C 89 4C 24 ? 89 44 24 ? 0F B7 43 ? 4C 8D 0D ? ? ? ? 4C 8D 04 40 4A 8D 14 81"), ApparelIndex_LookupA);
+	g_origLookupB = hook::trampoline(hook::get_pattern("4C 8B DC 49 89 5B ? 57 48 83 EC ? 45 33 C0 48 8B FA 48 8B D9 66 44 39 41 ? 76 ? ? ? 48 8B 49 ? 4D 89 43 ? 89 44 24"), ApparelIndex_LookupB);
+});


code/components/gta-core-five/src/PatchShopPedApparelLookup.cpp
```diff
@@ -0,0 +1,204 @@
...
+// The engine map's backing array grows its u16 capacity in steps of 16, so the
+// last size the grow math handles without wrapping is 65,520.
+static constexpr uint32_t kEngineMapMaxItems = 0xFFF0;
+
+// Offsets recovered for b3258/b3788. Manager `this`: apparel-file array ptr@+0x70,
+// u16 count@+0x78; lookup map sorted-flag@+0x80, u16 count@+0x90.
+static constexpr ptrdiff_t kMgrFilesPtr   = 0x70;
+static constexpr ptrdiff_t kMgrFilesCount = 0x78;
+static constexpr ptrdiff_t kMgrMapSorted  = 0x80;
+static constexpr ptrdiff_t kMgrMapCount   = 0x90;
```
chromium-cfx: It would be better if these offsets were patterns. or at least contain an assertion in debug if the struct layout changes in a future TU.

code/components/gta-core-five/src/PatchShopPedApparelLookup.cpp
```diff
@@ -0,0 +1,204 @@
...
+static inline T Peek(const void* base, ptrdiff_t offset)
+{
+	return *reinterpret_cast<const T*>(reinterpret_cast<const uint8_t*>(base) + offset);
+}
+
+
+static void (*g_origRebuild)(void* manager);
+
+// Raw <Item> count across all mounted files - what the engine map's u16 has to hold.
+static uint32_t CountTotalItems(void* manager)
```
chromium-cfx: Use FlexStruct and replace the 'Peek' usages.

code/components/gta-core-five/src/PatchShopPedApparelLookup.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+static void ApparelIndex_Rebuild(void* manager)
+{
+	// The convar is replicated, so it resets to its default on disconnect -
+	// BEFORE the apparel files unload - and the engine rebuilds the lookup one
+	// more time during teardown. Deciding by actual item count as well means
+	// that rebuild can never feed more items into the engine's 16-bit map than
+	// it can hold, regardless of convar state or teardown ordering.
+
+	g_sideTableActive = g_largeApparelPoolEnabled || CountTotalItems(manager) > kEngineMapMaxItems;
```
chromium-cfx: on disconnect, if the server was using this feature but didn't have more then 2^16 apparel entries this could lead to g_apparelIndex map not being cleared.

code/components/gta-core-five/src/PatchShopPedApparelLookup.cpp
```diff
@@ -0,0 +1,204 @@
...
+	g_apparelIndex.clear();
+
+	uint16_t fileCount = Peek<uint16_t>(manager, kMgrFilesCount);
+	void**   files     = Peek<void**>(manager, kMgrFilesPtr);
+
+	for (uint16_t fileIndex = 0; fileIndex < fileCount; fileIndex++)
+	{
+		void* file = files[fileIndex];
+		if (!file)
+			continue;
```
chromium-cfx, line 76: Formatting
```suggestion
		if (!file)
		{
			continue;
	    }
```

code/components/gta-core-five/src/PatchShopPedApparelLookup.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+		for (const auto& arr : kItemArrays)
+		{
+			uint16_t count = Peek<uint16_t>(file, arr.countOffset);
+			uint8_t* items = Peek<uint8_t*>(file, arr.ptrOffset);
+			if (!items)
+				continue;
```
chromium-cfx, line 85:
```suggestion
			if (!items)
			{
				continue;
			}
```

code/components/gta-core-five/src/PatchShopPedApparelLookup.cpp
```diff
@@ -0,0 +1,204 @@
...
+				uint32_t hash = Peek<uint32_t>(items + static_cast<size_t>(itemIndex) * arr.stride, kItemNameHash);
+				// keep first occurrence, matching the engine's lower-bound on ties
+				g_apparelIndex.emplace(hash, EncodeLocation(arr.kind, fileIndex, itemIndex));
+			}
+		}
+	}
+
+	// leave the engine map empty + sorted so nothing reads stale data or grows it
+	*reinterpret_cast<uint16_t*>(reinterpret_cast<uint8_t*>(manager) + kMgrMapCount)  = 0;
+	*reinterpret_cast<uint8_t*>(reinterpret_cast<uint8_t*>(manager) + kMgrMapSorted) = 1;
```
chromium-cfx: Use FlexStruct

code/components/gta-core-five/src/PatchShopPedApparelLookup.cpp
```diff
@@ -0,0 +1,204 @@
...
+// Consumers: both compiled copies of the map's lookup helper. The map arg is
+// manager + kMgrMapSorted, which we use to re-validate the cached location.
+static char* (*g_origLookupA)(void* map, uint32_t* nameHash);
+static char* (*g_origLookupB)(void* map, uint32_t* nameHash);
+
+static char* ResolveFromIndex(void* map, uint32_t nameHash)
+{
+	auto it = g_apparelIndex.find(nameHash);
+	if (it == g_apparelIndex.end())
+		return nullptr;
```
chromium-cfx, line 112:
```suggestion
	if (it == g_apparelIndex.end())
	{
		return nullptr;
    }
```

code/components/gta-core-five/src/PatchShopPedApparelLookup.cpp
```diff
@@ -0,0 +1,204 @@
...
+	uint8_t  kind    = static_cast<uint8_t>(packed & 0xFF);
+	uint16_t fileIdx = static_cast<uint16_t>((packed >> 8) & 0xFFFF);
+	uint16_t itemIdx = static_cast<uint16_t>((packed >> 32) & 0xFFFF);
+
+	uint8_t* manager   = reinterpret_cast<uint8_t*>(map) - kMgrMapSorted;
+	uint16_t fileCount = Peek<uint16_t>(manager, kMgrFilesCount);
+	void*    file      = (kind <= 2 && fileIdx < fileCount) ? Peek<void**>(manager, kMgrFilesPtr)[fileIdx] : nullptr;
+	uint16_t itemCount = file ? Peek<uint16_t>(file, kItemArrays[kind <= 2 ? kind : 0].countOffset) : 0;
+
+	if (kind > 2 || fileIdx >= fileCount || !file || itemIdx >= itemCount)
```
chromium-cfx: Could kind end up as a value greater then 2? kind should only be between 0-2 in ``ApparelIndex_Rebuild``, if the original game function doesn't include this check I don't think its needed for you to include it. 

and formatting.

code/components/gta-core-five/src/PatchShopPedApparelLookup.cpp
```diff
@@ -0,0 +1,204 @@
...
+			{
+				total += Peek<uint16_t>(file, arr.countOffset);
+			}
+		}
+	}
+
+	return total;
+}
+
+static void ApparelIndex_Rebuild(void* manager)
```
chromium-cfx:
```suggestion
static void ApparelIndex_Rebuild(hook::FlexStruct* manager)
```

code/components/gta-core-five/src/PatchShopPedApparelLookup.cpp
```diff
@@ -0,0 +1,204 @@
+// PatchShopPedApparelLookup.cpp
```
chromium-cfx: These comments can be removed or at least simplified and placed below the headers similar to other files. Ideally we don't need an overly detailed explanation of the entire file and its changes. What would be more beneficial for someone looking into this change is an explanation of the issue at hand, and how this patch aims to resolve it.

code/components/gta-core-five/src/PatchShopPedApparelLookup.cpp
```diff
@@ -0,0 +1,204 @@
...
+	ptrdiff_t countOffset;
+	size_t    stride;
+	uint8_t   kind;   // 0 component, 1 prop, 2 outfit
+};
+
+static constexpr ItemArrayLayout kItemArrays[3] = {
+	{ 0x28, 0x30, 144, 0 },
+	{ 0x38, 0x40, 136, 1 },
+	{ 0x48, 0x50, 160, 2 },
+};
```
chromium-cfx, line 43: Could these offsets be easily retrievable by a pattern? if not including an assertion if one or more of these differs between game builds would be great.

code/components/gta-core-five/src/PatchShopPedApparelLookup.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+static constexpr ptrdiff_t kItemNameHash = 0x20;
```
chromium-cfx: Same as the other two offsets comments.

code/components/gta-core-five/src/PatchShopPedApparelLookup.cpp
```diff
@@ -0,0 +1,204 @@
...
+		}
+	}
+
+	// leave the engine map empty + sorted so nothing reads stale data or grows it
+	*reinterpret_cast<uint16_t*>(reinterpret_cast<uint8_t*>(manager) + kMgrMapCount)  = 0;
+	*reinterpret_cast<uint8_t*>(reinterpret_cast<uint8_t*>(manager) + kMgrMapSorted) = 1;
+}
+
+// Consumers: both compiled copies of the map's lookup helper. The map arg is
+// manager + kMgrMapSorted, which we use to re-validate the cached location.
```
chromium-cfx: This comment isn't really needed.

code/components/gta-core-five/src/PatchShopPedApparelLookup.cpp
```diff
@@ -0,0 +1,204 @@
...
+}
+
+static char* ApparelIndex_LookupB(void* map, uint32_t* nameHash)
+{
+	return g_sideTableActive ? ResolveFromIndex(map, *nameHash) : g_origLookupB(map, nameHash);
+}
+
+static HookFunction hookFunction([]()
+{
+	static ConVar<bool> largeApparelPoolVar("game_enableLargeShopPedApparelPool", ConVar_Replicated, false, &g_largeApparelPoolEnabled);
```
chromium-cfx: This isn't a pool so the name is potentially misleading. A potentially better name would be `game_extendedApparelShopLimit` or `game_enableExtendedApparelShopLimit`

code/components/gta-core-five/src/PatchShopPedApparelLookup.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+	// patterns below are not found in 1604, but moving foward afaik the game build will be stable (3258 atm and then 37XX later?)
+	if (xbr::IsGameBuild<1604>())
```
chromium-cfx: It would better to change this check to not support any builds older then 3258.
```suggestion
	if (!xbr::IsGameBuildOrGreater<3258>())
```

code/components/gta-core-five/src/PatchShopPedApparelLookup.cpp
```diff
@@ -0,0 +1,204 @@
...
+#include <CrossBuildRuntime.h>
+
+#include <cstdint>
+#include <EASTL/hash_map.h>
+
+static bool g_largeApparelPoolEnabled = false;
+
+// Whether the side table is in use, decided by each rebuild (the lookups just
+// follow the last rebuild's choice, so both always agree on where the data is).
+static bool g_sideTableActive = false;
```
chromium-cfx: This naming could be improved. e.g. `g_usingExtendedApparelMap`

code/components/gta-core-five/src/PatchShopPedApparelLookup.cpp
```diff
@@ -0,0 +1,204 @@
...
+#include <Hooking.h>
+#include <Hooking.Stubs.h>
+
+#include <CoreConsole.h>
+#include <CrossBuildRuntime.h>
+
+#include <cstdint>
+#include <EASTL/hash_map.h>
+
+static bool g_largeApparelPoolEnabled = false;
```
chromium-cfx: this isn't a pool, the naming could be improved.

code/components/gta-core-five/src/PatchShopPedApparelLookup.cpp
```diff
@@ -0,0 +1,204 @@
...
+	     | (static_cast<uint64_t>(itemIndex) << 32);
+}
+
+static eastl::hash_map<uint32_t, uint64_t> g_apparelIndex;
+
+template<typename T>
+static inline T Peek(const void* base, ptrdiff_t offset)
+{
+	return *reinterpret_cast<const T*>(reinterpret_cast<const uint8_t*>(base) + offset);
+}
```
chromium-cfx: This can be replaced with using FlexStruct

code/components/gta-core-five/src/PatchShopPedApparelLookup.cpp
```diff
@@ -0,0 +1,204 @@
...
+	// lookup can never hand out an entry pointing at an unloaded file
+	uint64_t packed  = it->second;
+	uint8_t  kind    = static_cast<uint8_t>(packed & 0xFF);
+	uint16_t fileIdx = static_cast<uint16_t>((packed >> 8) & 0xFFFF);
+	uint16_t itemIdx = static_cast<uint16_t>((packed >> 32) & 0xFFFF);
+
+	uint8_t* manager   = reinterpret_cast<uint8_t*>(map) - kMgrMapSorted;
+	uint16_t fileCount = Peek<uint16_t>(manager, kMgrFilesCount);
+	void*    file      = (kind <= 2 && fileIdx < fileCount) ? Peek<void**>(manager, kMgrFilesPtr)[fileIdx] : nullptr;
+	uint16_t itemCount = file ? Peek<uint16_t>(file, kItemArrays[kind <= 2 ? kind : 0].countOffset) : 0;
```
chromium-cfx: It may be better to swap the ternary usage to exit early for the invalid cases.

chromium-cfx (changes requested):

code/components/gta-core-five/src/PatchShopPedApparelLookup.cpp
xalva98: done (although not easily retrievable by pattern) - so i went with the assertion way - i also checked that they exist in all builds above 3258, so it exists in 3788 and they are unique aswell

code/components/gta-core-five/src/PatchShopPedApparelLookup.cpp
```diff
@@ -0,0 +1,226 @@
...
+			uint16_t count = file->Get<uint16_t>(arr.countOffset);
+			auto     items = file->Get<uint8_t*>(arr.ptrOffset);
+			if (!items)
+			{
+				continue;
+			}
+
+			for (uint16_t itemIndex = 0; itemIndex < count; itemIndex++)
+			{
+				auto     item = reinterpret_cast<hook::FlexStruct*>(items + static_cast<size_t>(itemIndex) * arr.stride);
```
chromium-cfx: formatting.

```suggestion
				auto item = reinterpret_cast<hook::FlexStruct*>(items + static_cast<size_t>(itemIndex) * arr.stride);
```

code/components/gta-core-five/src/PatchShopPedApparelLookup.cpp
```diff
@@ -0,0 +1,226 @@
...
+		hook::FlexStruct* file = files[fileIndex];
+		if (!file)
+		{
+			continue;
+		}
+
+		for (const auto& arr : kItemArrays)
+		{
+			uint16_t count = file->Get<uint16_t>(arr.countOffset);
+			auto     items = file->Get<uint8_t*>(arr.ptrOffset);
```
chromium-cfx:
```suggestion
			auto items = file->Get<uint8_t*>(arr.ptrOffset);
```

code/components/gta-core-five/src/PatchShopPedApparelLookup.cpp
```diff
@@ -0,0 +1,226 @@
...
+	g_apparelIndex.clear();
+
+	if (!g_usingExtendedApparelMap)
+	{
+		g_origRebuild(manager);
+		return;
+	}
+
+	uint16_t fileCount = manager->Get<uint16_t>(g_mgrFilesCount);
+	auto     files     = manager->Get<hook::FlexStruct**>(g_mgrFilesPtr);
```
chromium-cfx:
```suggestion
	auto files = manager->Get<hook::FlexStruct**>(g_mgrFilesPtr);
```

code/components/gta-core-five/src/PatchShopPedApparelLookup.cpp
```diff
@@ -0,0 +1,226 @@
...
+}
+
+static char* ApparelIndex_LookupB(hook::FlexStruct* map, uint32_t* nameHash)
+{
+	return g_usingExtendedApparelMap ? ResolveFromIndex(map, *nameHash) : g_origLookupB(map, nameHash);
+}
+
+static HookFunction hookFunction([]()
+{
+	static ConVar<bool> extendedApparelLimitVar("game_enableExtendedApparelShopLimit", ConVar_Replicated, false, &g_extendedApparelLimitEnabled);
```
chromium-cfx: One concern. if the value of this convar gets changed on the server, any player currently connected will likely experience undefined behavior and more then likely crash. This should be changed to discard any changes once the client has started building the apparel index. This would also allow you to account for unloading.

a rough example.
```c++
static void OnExtendedApparelShopLimitChange(internal::ConsoleVariableEntry<bool>* var)
{
	if (hasBuiltList && Instance<ICoreGameInit>::Get()->HasVariable("networkInited"))
	{
		return;
	}

    g_extendedApparelLimitEnabled = var->GetRawValue();
}


// In HookFunction
static ConVar<bool> extendedApparelLimitVar("game_enableExtendedApparelShopLimit", ConVar_Replicated, false, &MinBlendDiffChanged);
```

code/components/gta-core-five/src/PatchShopPedApparelLookup.cpp
```diff
@@ -0,0 +1,226 @@
...
+
+static eastl::hash_map<uint32_t, uint64_t> g_apparelIndex;
+
+static void (*g_origRebuild)(hook::FlexStruct* manager);
+
+// Raw <Item> count across all mounted files - what the engine map's u16 must hold.
+static uint32_t CountTotalItems(hook::FlexStruct* manager)
+{
+	uint16_t fileCount = manager->Get<uint16_t>(g_mgrFilesCount);
+	auto     files     = manager->Get<hook::FlexStruct**>(g_mgrFilesPtr);
```
chromium-cfx:
```suggestion
	auto files = manager->Get<hook::FlexStruct**>(g_mgrFilesPtr);
```

code/components/gta-core-five/src/PatchShopPedApparelLookup.cpp
```diff
@@ -0,0 +1,226 @@
...
+	uint8_t kind; // 0 component, 1 prop, 2 outfit
+};
+
+static constexpr ItemArrayLayout kItemArrays[3] = {
+	{ 0x28, 0x30, 144, 0 },
+	{ 0x38, 0x40, 136, 1 },
+	{ 0x48, 0x50, 160, 2 },
+};
+
+static constexpr int32_t kItemNameHash = 0x20;
```
chromium-cfx: this should be renamed to make it clear that it is an offset and probably moved up to be with the rest of the offsets.
```suggestion
static constexpr int32_t kItemNameHashOffset = 0x20;
```

code/components/gta-core-five/src/PatchShopPedApparelLookup.cpp
```diff
@@ -0,0 +1,226 @@
...
+// size its grow math handles without wrapping is 65,520.
+static constexpr uint32_t kEngineMapMaxItems = 0xFFF0;
+
+// CExtraMetadataMgr member offsets: apparel-file array ptr / u16 count; lookup
+// map sorted-flag / u16 count. Resolved from RebuildApparelLookup at init so a
+// build that shifts these members adjusts on its own (see hookFunction).
+static int32_t g_mgrFilesPtr;
+static int32_t g_mgrFilesCount;
+static int32_t g_mgrMapSorted;
+static int32_t g_mgrMapCount;
```
chromium-cfx: These should be renamed to clearly indicate that they are offsets. e.g. g_mgrFiltersPtrOffset.

code/components/gta-core-five/src/PatchShopPedApparelLookup.cpp
```diff
@@ -0,0 +1,226 @@
...
+	g_mgrMapCount   = *hook::get_pattern<uint32_t>("48 8B F9 66 89 91 ? ? ? ? 8B F2", 6);
+	g_mgrFilesCount = *hook::get_pattern<uint8_t>("0F B7 47 ? FF C6 89 75 ? 3B F0", 3);
+	g_mgrFilesPtr   = *hook::get_pattern<uint8_t>("48 8B 47 ? 8B CE 44 8B E2 4C 8B 34 C8", 3);
+
+	// The three item sub-arrays share one loop body, so their offsets aren't
+	// individually patternable - but item 0's is (its index math is unique). In
+	// debug, assert the parser struct layout still matches what we hardcoded.
+	assert(*hook::get_pattern<uint8_t>("48 8D 0C C0 49 8B 46 ? 45 0B EF", 7) == kItemArrays[0].ptrOffset);
+	assert(*hook::get_pattern<uint8_t>("4C 8B 34 C8 66 41 3B 56 ? 73", 8) == kItemArrays[0].countOffset);
+	assert(*hook::get_pattern<uint8_t>("44 89 6D D8 8B 5C C8 ? 8B 4D DC", 7) == kItemNameHash);
```
chromium-cfx: These assertions will be compiled into release/retail builds. 
```suggestion
#ifdef _DEBUG
	assert(*hook::get_pattern<uint8_t>("48 8D 0C C0 49 8B 46 ? 45 0B EF", 7) == kItemArrays[0].ptrOffset);
	assert(*hook::get_pattern<uint8_t>("4C 8B 34 C8 66 41 3B 56 ? 73", 8) == kItemArrays[0].countOffset);
	assert(*hook::get_pattern<uint8_t>("44 89 6D D8 8B 5C C8 ? 8B 4D DC", 7) == kItemNameHash);
#endif
```

chromium-cfx (changes requested):

code/components/gta-core-five/src/PatchShopPedApparelLookup.cpp
xalva98: Fair point, I didn't even think about this I just assumed noone would change this in production. Maybe we don't even need the convar at all if everything is stable?

chromium-cfx: For now it would be better to keep it as an opt in for server's that need it. The main concern here is to prevent an accidental convar change in production (or a resource accidently writing a different value to the convar) which would lead to instability. It also allows you to do away with the ``g_usingExtendedApparelMap`` workaround which introduce to deal with disconnecting and reconnecting to a server with/without the convar set.

chromium-cfx (approved):

PR #4062, opened 2026-07-11
fix(server/state): don't insert into the world grid map under a shared lock

0x5267:
### Goal of this PR

While reading the relevance pass in `ServerGameState::Tick` I noticed it resolves the world grid differently from every other place that touches `m_worldGrids`:

```cpp
std::shared_lock _(m_worldGridsMutex);
const auto& grid = m_worldGrids[selfBucket];
```

Everything else uses `find()`. `operator[]` on a `std::map` inserts when the key is missing, and here that happens while holding only the reader side of the lock, racing any concurrent reader of the map. The key can actually be missing: `UpdateWorldGrid` creates grids at the start of the tick, but a routing bucket change from another thread (or a player entity appearing mid tick) leaves the pass looking at a bucket that has no grid yet.

The inserted entry is also a default constructed null `unique_ptr`, and since `UpdateWorldGrid` only creates a grid when `find()` fails, it keeps finding the null entry and bailing on its `if (!grid)` check. From then on nobody in that bucket gets world grid ownership until the bucket empties out.

Resolving the grid once per client with `find()` before the entity loop, like the other call sites do. This is safe because grids are only created and erased in `UpdateWorldGrid`, which runs earlier in the same tick on the same thread. A missing grid now just means no world grid relevance for that pass, which is what the next tick would produce anyway, instead of poisoning the bucket.

This also removes a `shared_mutex` acquisition and a map walk per entity per client from the pass.

Server, OneSync

### Successfully tested on

**Platforms:** Windows

### Checklist

### Fixes issues

### DIFF
--- a/code/components/citizen-server-impl/src/state/ServerGameState.cpp
+++ b/code/components/citizen-server-impl/src/state/ServerGameState.cpp
@@ -1048,6 +1048,21 @@ void ServerGameState::Tick(fx::ServerInstanceBase* instance)
 		auto slotId = client->GetSlotId();
 		auto netId = client->GetNetId();
 
+		// look up the world grid for the client's routing bucket once: grids are only created
+		// and erased in UpdateWorldGrid, which ran earlier in this tick on this thread, so the
+		// pointer stays valid for the whole pass and we don't need to lock per entity
+		const WorldGrid* worldGrid = nullptr;
+
+		if (playerEntity)
+		{
+			std::shared_lock _(m_worldGridsMutex);
+
+			if (auto gridIt = m_worldGrids.find(clientDataUnlocked->routingBucket); gridIt != m_worldGrids.end())
+			{
+				worldGrid = gridIt->second.get();
+			}
+		}
+
 		auto& currentSyncedEntities = clientDataUnlocked->syncedEntities;
 		decltype(clientDataUnlocked->syncedEntities) newSyncedEntities{};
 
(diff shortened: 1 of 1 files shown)


FabianTerhorst:
Hi, both emplace and erase of the world grid happens inside the tick. Please provide an scenario where emplace or erase of the world grid happens outside.

0x5267:
Hello! The emplace and erase in `UpdateWorldGrid` are fine, the issue is that `operator[]` in the relevance pass is itself an emplace when the bucket is missing, and readers of the map exist outside the sync thread.

`SET_PLAYER_ROUTING_BUCKET` calls `ClearClientFromWorldGrid` directly on the script runtime thread (unlike the drop path, which marshals `HandleClientDrop` onto the sync thread via `gscomms_execute_callback_on_sync_thread`). That does `m_worldGrids.find()` under a `shared_lock`, plus another in `SendWorldGrid`. The same native also sets `clientData->routingBucket` with no ordering against the tick, so the relevance pass can see a bucket with no grid and `operator[]` inserts into the tree under the reader lock while the script thread's `find()` walks it. Two shared locks do not exclude each other.

Even without the race, the inserted entry is a null `unique_ptr`. `UpdateWorldGrid` only emplaces when `find()` fails, so it keeps hitting its `if (!grid)` check and the bucket never gets a real grid for as long as it keeps clients.

PR #4068, opened 2026-07-14
fix(net): move routed packets instead of copying them

ShamZy9:
## Goal of this PR

Remove two redundant full-payload copies on the per-frame packet routing path in `NetLibrary`.

- `DequeueRoutedPacket`: `m_incomingPackets.front()` was copied by value (including the `std::string` payload) before being `memcpy`'d into the destination buffer. Now moved out of the queue instead.
- `RoutePacket`: the `RoutingPacket` (payload included) was copied again by `push`. Now moved into the queue.

Both functions run for every routed packet in/out, every frame, so this removes one full payload allocation+copy per packet in each direction.

Replace the by-value copy with `std::move` in both places. No behavior change: the moved-from element is immediately `pop()`'d / goes out of scope.

FiveM, RedM, Networking

## Successfully tested on

**Game builds:** n/a (behavior-neutral change)

**Platforms:** Windows

## Checklist

### DIFF
--- a/code/components/net/src/NetLibrary.cpp
+++ b/code/components/net/src/NetLibrary.cpp
@@ -226,7 +226,7 @@ bool NetLibrary::DequeueRoutedPacket(char* buffer, size_t* length, uint16_t* net
 			return false;
 		}
 
-		auto packet = m_incomingPackets.front();
+		auto packet = std::move(m_incomingPackets.front());
 		m_incomingPackets.pop();
 
 		memcpy(buffer, packet.payload.c_str(), packet.payload.size());
@@ -250,7 +250,7 @@ void NetLibrary::RoutePacket(const char* buffer, size_t length, uint16_t netID)
 	routePacket.netID = netID;
 	routePacket.payload = std::string(buffer, length);
 
-	m_outgoingPackets.push(routePacket);
+	m_outgoingPackets.push(std::move(routePacket));
 }
 
 #define	BIG_INFO_STRING		8192  // used for system info key only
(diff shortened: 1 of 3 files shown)


code/components/net/src/NetLibrary.cpp
```diff
@@ -226,7 +226,7 @@ bool NetLibrary::DequeueRoutedPacket(char* buffer, size_t* length, uint16_t* net
 			return false;
 		}
 
-		auto packet = m_incomingPackets.front();
```
AvarianKnight, line 229: This doesn't do anything, why didn't the AI just suggest taking a reference here and moving the `m_incomingPackets.pop` to after the usage of `timeval`? 🤔

code/components/net/src/NetLibrary.cpp
```diff
@@ -250,7 +250,7 @@ void NetLibrary::RoutePacket(const char* buffer, size_t length, uint16_t netID)
 	routePacket.netID = netID;
 	routePacket.payload = std::string(buffer, length);
 
-	m_outgoingPackets.push(routePacket);
+	m_outgoingPackets.push(std::move(routePacket));
```
AvarianKnight, line 254: ~~This takes a `const &` doing a std::move here is redundant as it already takes a reference.~~ 

~~Edit 1: tbb has a move constructor for this so moving here does make sense~~

Edit 2: This still doesn't make sense, this uses microsofts concurrent_queue not tbb so this still copies

AvarianKnight (commented):
If you're going to take time to do an AI PR atleast take the time to see if it makes any sense.

ShamZy9:
yep thanks, switched it to a const& on front() with the pop() after the timeval read.
on RoutePacket though, m_outgoingPackets is a concurrency::concurrent_queue and its push has a T&& overload so passing an rvalue move-constructs the RoutingPacket into the queue instead of copying the payload, noot a no-op, but I can revert if you'd rather keep it explicit

PR #4076, opened 2026-07-16
feat(gta-core-five): disable passive mode using convar

divocbn:
### Goal of this PR

Preventing abuse of the GTA:O friendly / passive mode sync which is being actively exploited and implemented in almost every cheat.

Patch out the TreatAsFriendlyForTargetingAndDamage check inside CPed::IsAllowedToDamageEntity when the _**game_disablePassiveMode**_ convar is enabled.

FiveM

### Successfully tested on
**Game builds:** 3258
**Platforms:** Windows

### Checklist

### Fixes issues
Cheats abusing passive mode: https://www.youtube.com/watch?v=IXXxhkDEt6A

### DIFF
--- a/code/components/gta-core-five/src/PatchPassiveMode.cpp
+++ b/code/components/gta-core-five/src/PatchPassiveMode.cpp
@@ -1,0 +1,53 @@
+#include <StdInc.h>
+
+#include <CoreConsole.h>
+
+#include <Hooking.h>
+#include <Hooking.Stubs.h>
+
+#include <jitasm.h>
+
+static bool g_DisablePassiveMode = false;
+
+static HookFunction hookFunction([]
+{
+	static ConVar<bool> enableVehicleHijackFix("game_disablePassiveMode", ConVar_Replicated, false, &g_DisablePassiveMode);
+
+	auto patchAllowedToDamagePassiveFlagLocation = hook::get_pattern("C1 E8 ? A8 ? 75 ? C1 E9 ? F6 C1 ? 75 ? 48 8B D3");
+	auto patchAllowedToDamageSkipLocation = hook::get_pattern("32 C0 E9 ? ? ? ? F7 83");
+
+	static struct : jitasm::Frontend
+	{
+		uintptr_t m_ContinueAddress;
+		uintptr_t m_SkipAddress;
+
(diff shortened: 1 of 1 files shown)


william-des:
I think it would be more flexible to expose a native similar to [OnesyncEnableRemoteAttachmentSanitization](https://docs.fivem.net/natives/?_0x30CE39D8)
For example, on my PvP server I use passive mode for teammates in certain game modes, but not all of them. I will not be able to use a global server set convar in this case.

It could also be useful to have a native that ignores passive mode for a specific player ped instead of applying it globally, if that's not too much complexity to add.

AngeloneISL:
> I think it would be more flexible to expose a native similar to [OnesyncEnableRemoteAttachmentSanitization](https://docs.fivem.net/natives/?_0x30CE39D8) For example, on my PvP server I use passive mode for teammates in certain game modes, but not all of them. I will not be able to use a global server set convar in this case.
> 
> It could also be useful to have a native that ignores passive mode for a specific player ped instead of applying it globally, if that's not too much complexity to add.

Agreed same here

divocbn:
> I think it would be more flexible to expose a native similar to [OnesyncEnableRemoteAttachmentSanitization](https://docs.fivem.net/natives/?_0x30CE39D8) For example, on my PvP server I use passive mode for teammates in certain game modes, but not all of them. I will not be able to use a global server set convar in this case.
> 
> It could also be useful to have a native that ignores passive mode for a specific player ped instead of applying it globally, if that's not too much complexity to add.

iirc when I asked about implementing that I got told to rather do it as a ConVar instead of a native, maybe @radium-cfxcan clarify whether to expose it as native or convar

Bossman02:
And couldn't both be implemented, with convar taking prevalence over the native method? That way, servers that want to block it entirely can do so using convar, and those that want to block it temporarily via a script can use the native method

xalva98:
> And couldn't both be implemented, with convar taking prevalence over the native method? That way, servers that want to block it entirely can do so using convar, and those that want to block it temporarily via a script can use the native method

New natives will not be added currently because it would mean double work for Gen9 (enhanced) if i understood it correctly.
My best guess is they want to have Enhanced "feature complete" before adding anything on top, making it 1 to 1 as legacy.

PR #4078, opened 2026-07-18
feat(extra-natives-five): add a global vehicle audio volume control

xalva98:
### Goal of this PR

Let players lower (or fully mute) **only** vehicle sounds - engine, exhaust,
tyres, horns, alarms, collisions - without affecting weapons, footsteps,
ambience or music.

https://github.com/user-attachments/assets/412f5500-076b-4371-bbcc-1403898cdb0e

Vehicle audio all descends from the RAGE `VEHICLES` audio category
(`0x74B31BE3`). The feature creates an `audCategoryControllerManager` controller
on that category and sets its linear volume - the same mechanism the game uses
for its own SFX/Music sliders. Setting the parent category scales every child
vehicle sound in one place.

1. The controller wrappers (`GetInstance` / `CreateController`) already exist in
   [`NuiAudioSink.cpp`](https://github.com/citizenfx/fivem/blob/25c937b1451be564e66afabab09b47a30d095385/code/components/extra-natives-five/src/NuiAudioSink.cpp#L2941-L2959)
   for the Mumble voice-volume feature; this reuses the same pattern on the
   `VEHICLES` category.
2. Exposed as the convar `profile_vehicleVolume` (archive convar)
3. The F8 menu previously rendered numeric convars as text
   inputs. I've added a slider for volumes which is easier to use in my opinion.

`SET_/GET_AUDIO_VEHICLE_VOLUME` native handlers are present but commented out
while gen8 is in feature freeze.

FiveM, Audio

### Successfully tested on

Game builds: **3258**
Platforms: **Windows**

### Checklist

### Fixes issues

Ears hurting from driving modded vehicles that have obnoxiously loud modded engine sounds

PS: I've tried implementing this functionality using existing natives, but I wasn't successful—hence the PR.

### DIFF
--- a/code/components/conhost-v2/src/DevGui.cpp
+++ b/code/components/conhost-v2/src/DevGui.cpp
@@ -69,6 +69,10 @@ struct DevGuiNode
 	std::string name;
 	std::string commandOrConVar;
 
+	float sliderMin = 0.0f;
+	float sliderMax = 0.0f;
+	float sliderStep = 0.0f;
+
 	std::list<std::unique_ptr<DevGuiNode>> children;
 
 	inline DevGuiNode* RegisterMenu(const std::string& name)
@@ -185,11 +189,59 @@ static void DevGui_Draw(const DevGuiNode* node)
 				auto value = intEntry->GetRawValue();
 				auto initialValue = value;
 
-				ImGui::InputScalar(node->name.c_str(), ImGuiDataType_S32, &value);
+				bool useSlider = node->sliderMax > node->sliderMin;
+
+				if (useSlider)
+				{
+					ImGui::SetNextItemWidth(160.0f);
+					ImGui::SliderInt(node->name.c_str(), &value, (int)node->sliderMin, (int)node->sliderMax);
+
+					if (node->sliderStep > 0.0f)
+					{
+						int step = (int)node->sliderStep;
+						value = (int)node->sliderMin + (int)std::round((value - (int)node->sliderMin) / (float)step) * step;
+					}
+				}
+				else
+				{
+					ImGui::InputScalar(node->name.c_str(), ImGuiDataType_S32, &value);
+				}
 
 				if (initialValue != value)
 				{
 					intEntry->SetRawValue(value);
+				}
+
+				return;
+			}
+
+			auto floatEntry = std::dynamic_pointer_cast<internal::ConsoleVariableEntry<float>>(entry);
+
+			if (floatEntry)
+			{
+				auto value = floatEntry->GetRawValue();
+				auto initialValue = value;
+
+				bool useSlider = node->sliderMax > node->sliderMin;
+
+				if (useSlider)
+				{
+					ImGui::SetNextItemWidth(160.0f);
+					ImGui::SliderFloat(node->name.c_str(), &value, node->sliderMin, node->sliderMax, "%.2f");
+
+					if (node->sliderStep > 0.0f)
+					{
+						value = node->sliderMin + std::round((value - node->sliderMin) / node->sliderStep) * node->sliderStep;
+					}
+				}
+				else
+				{
+					ImGui::InputScalar(node->name.c_str(), ImGuiDataType_Float, &value);
+				}
+
+				if (initialValue != value)
+				{
+					floatEntry->SetRawValue(value);
 				}
 
 				return;
@@ -246,6 +298,28 @@ static InitFunction initFunction([]()
 		node->commandOrConVar = convarName;
 	});
 
+	// devgui_convar with an explicit slider range: devgui_convar_slider "path" convar min max [step]
+	static ConsoleCommand devguiAddConVarSlider("devgui_convar_slider", [](const DevGuiPath& path, const std::string& convarName, float min, float max)
+	{
+		auto node = DevGui_InstantiatePath(path);
+
+		node->type = DevGuiNode::DevGuiNode_ConVar;
+		node->commandOrConVar = convarName;
+		node->sliderMin = min;
+		node->sliderMax = max;
+	});
+
+	static ConsoleCommand devguiAddConVarSliderStep("devgui_convar_slider", [](const DevGuiPath& path, const std::string& convarName, float min, float max, float step)
+	{
+		auto node = DevGui_InstantiatePath(path);
+
+		node->type = DevGuiNode::DevGuiNode_ConVar;
+		node->commandOrConVar = convarName;
+		node->sliderMin = min;
+		node->sliderMax = max;
+		node->sliderStep = step;
+	});
+
 #ifndef IS_FXSERVER
 	static ConVar<std::string> uiConnectHost("uiConnectHost", ConVar_Archive | ConVar_UserPref, "");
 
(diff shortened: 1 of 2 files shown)


lLukDreams:
Are you able to put it into the escape menu? this is for sure something needed for this annoying pay to win cars but i feel it would make more sense from the user experience to put it besides the escape menu settings

CeebDev:
What's the point of having this when you can literally change any volumes of SFX independently, including vehicle sounds, by script ?

I think it's server owners that should decide to either let people change the volume (even if it can be like cheating, e.g. cutting all sounds and only enabling vehicle sounds in a chase to know the position of a purchased vehicle), 

or let server owners create a "sound scene" for example, only lowering the weather volume so all players play with the same sounds (SFX volume lowered or raised for all scenes at once) and no player has an advantage.

xalva98:
> What's the point of having this when you can literally change any volumes of SFX independently, including vehicle sounds, by script ?
> 
> I think it's server owners that should decide to either let people change the volume (even if it can be like cheating, e.g. cutting all sounds and only enabling vehicle sounds in a chase to know the position of a purchased vehicle),
> 
> or let server owners create a "sound scene" for example, only lowering the weather volume so all players play with the same sounds (SFX volume lowered or raised for all scenes at once) and no player has an advantage.

The point is not needing to edit 100+ files

CeebDev:
> > What's the point of having this when you can literally change any volumes of SFX independently, including vehicle sounds, by script ?
> > I think it's server owners that should decide to either let people change the volume (even if it can be like cheating, e.g. cutting all sounds and only enabling vehicle sounds in a chase to know the position of a purchased vehicle),
> > or let server owners create a "sound scene" for example, only lowering the weather volume so all players play with the same sounds (SFX volume lowered or raised for all scenes at once) and no player has an advantage.
> 
> The point is not needing to edit 100+ files

What ? The script is like 100lines in a single file

xalva98:
> > > What's the point of having this when you can literally change any volumes of SFX independently, including vehicle sounds, by script ?
> > > I think it's server owners that should decide to either let people change the volume (even if it can be like cheating, e.g. cutting all sounds and only enabling vehicle sounds in a chase to know the position of a purchased vehicle),
> > > or let server owners create a "sound scene" for example, only lowering the weather volume so all players play with the same sounds (SFX volume lowered or raised for all scenes at once) and no player has an advantage.
> > 
> > 
> > The point is not needing to edit 100+ files
> 
> What ? The script is like 100lines in a single file

Enlighten me

CeebDev:
In fact the 100 lines is just config file but it's like 2 native calls StartAudioScene and SetAudioSceneVariable

xalva98:
> In fact the 100 lines is just config file but it's like 2 native calls StartAudioScene and SetAudioSceneVariable

I tried that without any success so if you can share a snippet on how to target vehicle/engine sounds only then this PR is of course pointless.

CeebDev:
```typescript
function getValueFromVolume(volume: number) {
	return 1 - (volume / 100);
}

const scene = "mutes_vehicles_scene";

StartAudioScene(scene)
SetAudioSceneVariable(scene, "apply", getValueFromVolume(45)) // 45% of max volume
```

Then if you need to update volume you just have to call this again without the start call

```typescript
SetAudioSceneVariable(scene, "apply", getValueFromVolume(70)) // 70% of max volume
```

xalva98:
> ```ts
> function getValueFromVolume(volume: number) {
> 	return 1 - (volume / 100);
> }
> 
> const scene = "mutes_vehicles_scene";
> 
> StartAudioScene(scene)
> SetAudioSceneVariable(scene, "apply", getValueFromVolume(45)) // 45% of max volume
> ```
> 
> Then if you need to update volume you just have to call this again without the start call
> 
> ```ts
> SetAudioSceneVariable(scene, "apply", getValueFromVolume(70)) // 70% of max volume
> ```

Works like a charm didn't even know that could be done 👍

CeebDev:
That's why I didn't understand the goal of this PR, since it can be done easily by script

n3xuuu:
It can be done via a script, but I still think it would be nice for players to be able to change it on their own through a command or via audio settings, since most servers likely don’t have the feature you're talking about

CeebDev:
But what's the usecase ? If vehicle are too loud it's server problem, GTA has made audio equilibrated

lLukDreams:
@CeebDev there are a lot of servers with extrem loud cars since people pay for that and are not interested to make them quiet, i can fully understand to give the users self the option since this is annoying af

CeebDev:
Yeah so players can just change server if it's badly developed

I mean if server owner decide to put shit stuff on their server, fivem shouldnt give a bandaid to players

If this is implemented we should at least allow server owner to have controls to avoid unfair advantage

AvarianKnight:
> Yeah so players can just change server if it's badly developed
> 
> I mean if server owner decide to put shit stuff on their server, fivem shouldnt give a bandaid to players
> 
> If this is implemented we should at least allow server owner to have controls to avoid unfair advantage

That doesn't exactly make much sense.

Some users will like louder cars, other users will like quieter cars.

I'm not sure how this would give an unfair advantage either, but giving configuration for users who like their cars quieter natively makes more sense to me.

CeebDev:
@AvarianKnight For vehicle sounds no, there is no "advantage" but if we do this for vehicle, why not for other scene ? Weather / Footsteps / Music / Radios etc

If this goes in this way, then players will just be able to max vehicle sounds and cut / lower other sounds

So in a pursuit, if a vehicle succeed to escape the cops, then a cop doing that can just know that the vehicle is on right / left etc since there is just vehicle sounds without even seeing it doing the turn

GTA audio is equilibrated, all scenes volumes are just rights. This only fix the issue of custom vehicle that are louder

If people like their car louder, either server owner let the possibility via script, or user adjust their sfx slider which lower or raise the entire vfx

Again, this can be done but at least let server owner to "force" percentages.

For example we lower a bit the weather volume on rain / thunder. But we want all players to have the same mix. Players can still lower or raise sfx but the mixed scene will be the same

Scullyy:
> @AvarianKnight For vehicle sounds no, there is no "advantage" but if we do this for vehicle, why not for other scene ? Weather / Footsteps / Music / Radios etc
> 
> If this goes in this way, then players will just be able to max vehicle sounds and cut / lower other sounds
> 
> So in a pursuit, if a vehicle succeed to escape the cops, then a cop doing that can just know that the vehicle is on right / left etc since there is just vehicle sounds without even seeing it doing the turn
> 
> GTA audio is equilibrated, all scenes volumes are just rights. This only fix the issue of custom vehicle that are louder
> 
> If people like their car louder, either server owner let the possibility via script, or user adjust their sfx slider which lower or raise the entire vfx
> 
> Again, this can be done but at least let server owner to "force" percentages.
> 
> For example we lower a bit the weather volume on rain / thunder. But we want all players to have the same mix. Players can still lower or raise sfx but the mixed scene will be the same

Couldn’t this easily be solved by setting a threshold on the min/max values so you can’t completely mute the audio and also can’t make it super loud? The default volume could be the max so you can’t make it any louder than it was prior to the change.

CeebDev:
This PR or my code dont overide the max value

it only allow to set a percentage of the original value

Limiting this will be useless

AvarianKnight:
> @AvarianKnight For vehicle sounds no, there is no "advantage" but if we do this for vehicle, why not for other scene ? Weather / Footsteps / Music / Radios etc

You're making a slippery slope argument.

There's no real reason for this to be applied to other things.

PR #4081, opened 2026-07-25
feat(natives): backport colshape natives to fivem + rdr3

xalva98:
### Goal of this PR

Make the enhanced colshape natives available on legacy GTA5 and RedM.

One shared manager (in `citizen-scripting-core`) implements the colshape
natives and geometry; a small per-side entity feed (five/rdr3 client, server)
ticks it with nearby entities and it fires `onColshapeEnter` / `onColshapeExit`.
A background-built BVH does broad-phase so detection stays cheap with many
shapes. Shapes are tied to the resource that created them and removed when it
stops.

Native names/hashes and event shape match the enhanced API, so scripts written
for enhanced work unchanged.

In plain terms, the detection loop is:

1. Every ~50 ms, on the game/tick thread, gather a snapshot of nearby entities
   (position + type). This is the only step that touches game state.
2. Hand that snapshot to a background worker thread and return immediately, so the
   frame/tick is never blocked by the actual detection.
3. The worker checks each entity against the shapes (a BVH narrows it down to a few
   candidates first) and compares the result to the previous tick to find who just
   entered or left.
4. The resulting enter/leave transitions are handed back and emitted as
   `onColshapeEnter` / `onColshapeExit` on the game/tick thread.

FiveM, RedM, Natives, FxServer

### Successfully tested on

Game builds: **3258 (FiveM)**, **1491 (RedM)**, **FxServer**
Platforms: **Windows**

Tested thoroughly on both games, client and server: full native/shape/getter
suite plus live enter/exit detection with moving peds, vehicles and objects and
per-entity-type gating. Create/query performance was compared against enhanced
and is roughly on par.

### Checklist

### DIFF
--- a/code/components/citizen-scripting-core/include/ColshapeManager.h
+++ b/code/components/citizen-scripting-core/include/ColshapeManager.h
@@ -1,0 +1,368 @@
+#pragma once
+
+#include <atomic>
+#include <cmath>
+#include <cstdint>
+#include <functional>
+#include <memory>
+#include <mutex>
+#include <shared_mutex>
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+#include "ComponentExport.h"
+
+namespace fx::colshape
+{
+enum class ColShapeType
+{
+	Circle,
+	Cuboid,
+	Cylinder,
+	Rectangle,
+	Sphere,
+	Polygon,
+};
+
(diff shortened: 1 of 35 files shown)


CENSOR37:
We need thisssss

PR #4088, opened 2026-07-29
fix(server): check `inVehicle` in `GET_VEHICLE_PED_IS_IN` for RedM

### LINKED ISSUE #4006: [Server] GetVehiclePedIsIn(ped, false) returns last vehicle when ped is not in any vehicle
### What happened?

Per the documented signature, the `lastVehicle` flag should switch the behavior:
- `false` - current vehicle
- `true` - last vehicle the ped occupied

In practice both calls return the same non-zero handle while the ped is on foot.

### Expected result

Server-side `GetVehiclePedIsIn(ped, false)` should return `0` when the ped is not currently sitting in any vehicle, matching the documented behavior and the client-side equivalent. Only `GetVehiclePedIsIn(ped, true)` should return the last-known vehicle handle in that case.

### Reproduction steps

1. Start a server and join with a client (player id `1`).
2. Enter any vehicle, then exit so the ped is on foot.
3. Confirm client-side that `IsPedInAnyVehicle(PlayerPedId(), false)` returns `false`.
4. From the **server console** run: `run GetVehiclePedIsIn(GetPlayerPed(1), false)`
5. Observe that the call returns the handle of the previously occupied vehicle instead of `0`.
6. Compare with `run GetVehiclePedIsIn(GetPlayerPed(1), true)` - same handle is returned.

Image

### Importancy

Unknown

### Area(s)

RedM, FXServer, OneSync, Natives, ScRT: Lua

### Specific version(s)

RedM/30408, FXServer-master SERVER v1.0.0.29199 win32

### Additional information

Server-side workaround currently used:
```lua
-- Server
local ped = GetPlayerPed(playerId)
local veh = IsPedInAnyVehicle(ped, false) and GetVehiclePedIsIn(ped, false) or 0

### PR DESCRIPTION
**Pull request by Github-Samuel**
### Goal of this PR
Make server-side `GetVehiclePedIsIn(ped, false)` return `0` on RedM when the ped is not sitting in a vehicle, instead of returning the last vehicle it was in.

Right now both `GetVehiclePedIsIn(ped, false)` and `GetVehiclePedIsIn(ped, true)` return the same handle once the ped is on foot, so there is no way to tell the two apart.

On RDR3 `CPedVehicleNodeData` splits this over two fields. `curVehicle` holds the vehicle id and keeps it after the ped gets out, it is only cleared when the vehicle slot stops being sent at all. `inVehicle` is the one that says whether the ped is seated right now.

This is already how the neighbouring natives in the same file work:

- `IS_PED_IN_ANY_VEHICLE` returns `pedVehicleData->inVehicle`
- `IS_PED_IN_VEHICLE` bails out on `if (!vehicleId || !inVehicle)`
- `GET_MOUNT` bails out on `if (!onHorse || !horseId)`

`GET_VEHICLE_PED_IS_IN` only checked `curVehicle == 0`, which is why it kept returning the stale handle. It now checks `inVehicle` as well for the current vehicle case:

```cpp
if ((lastVehicleArg == true && pedVehicleData->lastVehiclePedWasIn == 0) || (lastVehicleArg == false && (!pedVehicleData->inVehicle || pedVehicleData->curVehicle == 0)))
	 return (uint32_t)0;
```

The `lastVehicle == true` path is untouched, `lastVehiclePedWasIn` is kept up to date by `CPedVehicleDataNode::Serialize` and still works.

This is RedM only, the `#else` branch for FiveM already handles it because `CPedGameStateDataNode` writes `data.curVehicle = -1` when the ped leaves.

RedM, FXServer, OneSync, Natives

### Successfully tested on

**Game builds:** n/a, server-side only

**Platforms:** Windows, Linux

### Checklist

I could not do a full server build locally, so I left the first box unchecked. What I did check is that the changed condition compiles clean against the real `CPedVehicleNodeData`, and I stepped the real `CPedVehicleDataNode::Serialize` through enter vehicle -> exit -> re-enter -> swap vehicle to confirm `curVehicle` holds the old id while `inVehicle` is false, that the old condition returns the stale handle there, and that the new one returns `0` without changing any of the other cases. Happy to adjust if maintainers would rather this be shaped differently.

### Fixes issues
fixes #4006

### DIFF
--- a/code/components/citizen-server-impl/src/state/ServerGameState_Scripting.cpp
+++ b/code/components/citizen-server-impl/src/state/ServerGameState_Scripting.cpp
@@ -1280,7 +1280,7 @@ static void Init()
 		 if (!pedVehicleData)
 			 return (uint32_t)0;
 
-		 if ((lastVehicleArg == true && pedVehicleData->lastVehiclePedWasIn == 0) || (lastVehicleArg == false && pedVehicleData->curVehicle == 0))
+		 if ((lastVehicleArg == true && pedVehicleData->lastVehiclePedWasIn == 0) || (lastVehicleArg == false && (!pedVehicleData->inVehicle || pedVehicleData->curVehicle == 0)))
 			 return (uint32_t)0;
 
 		 lastVeh = pedVehicleData->lastVehiclePedWasIn;


outsider31000:
This is okay to be merged @prikolium-cfx , I missed that part of the `curVehicle` that keeps a holding on ids, I somehow also missed that issue report.

PR #4095, opened 2026-07-29
fix(launcher/citicore): fall back to Win32 UI and downlevel SEH handling under Wine

RatPrez:
### Goal of this PR
Get the legacy FiveM client to a point where it can actually start under Wine/Proton on Linux. Right now it doesn't get past the launcher, two separate things kill it before you even see a window.

Let me know if there is any genuine issues with this code being merged.

...

Two unrelated Wine-compat fixes, both gated behind the existing `CfxIsWine()` check in `LaunchMode.h` so native Windows behavior is untouched:

1. `client/launcher/UpdaterUI.cpp`

The update/patch splash screen is built with WinRT XAML Islands `DesktopWindowXamlSource`, which Wine doesn't implement. 

The code already has a working plain-Win32-controls fallback `g_uui.tenMode = false`, used for older Windows and safe mode. 

Added Wine to the existing `forceOff` checks in `UI_InitTen()` so it takes that path instead of trying to init XAML and dying.

4. `client/citicore/SEHTableHandler.Win32.cpp`

`CoreRT_SetupSEHHandler` disassembles `RtlLookupFunctionTable` in `ntdll.dll` to locate the internal `RtlpxLookupFunctionTable`. 

Wine's `ntdll` doesn't match real ntdll's binary layout, so the lookup never finds it, and since `IsWindows8OrGreater()` reports true under Wine, it was hitting a `FatalError()`. 

The code already has a fallback path for genuinely older/downlevel OS's that hooks `RtlLookupFunctionTable` directly instead, routed Wine into that same path instead of the fatal error.

...

FiveM, RedM

...

### Successfully tested on

**Platforms:** Windows, Linux

...

### Checklist

### DIFF
--- a/code/client/citicore/SEHTableHandler.Win32.cpp
+++ b/code/client/citicore/SEHTableHandler.Win32.cpp
@@ -14,6 +14,8 @@
 
 #include <udis86.h>
 
+#include <LaunchMode.h>
+
 static void* FindCallFromAddress(void* methodPtr, ud_mnemonic_code mnemonic = UD_Icall, bool breakOnFirst = false)
 {
 	// return value holder
@@ -210,7 +212,8 @@ extern "C" void DLL_EXPORT CoreRT_SetupSEHHandler(void* moduleBase, void* module
 			if (!internalAddress)
 			{
 				// and 2k3 to 7 don't even _have_ Rtlpx - so we directly hook the Rtl* function
-				if (IsWindows8OrGreater())
+				// (Wine's ntdll doesn't match real ntdll's binary layout, so it never finds RtlpxLookupFunctionTable there either. Treat it like a downlevel OS)
+				if (IsWindows8OrGreater() && !CfxIsWine())
 				{
 					FatalError("Could not find RtlpxLookupFunctionTable - hooking RtlLookupFunctionTable directly. This will break on a Win8+ system since RtlpxLookupFunctionTable is supposed to exist!\n");
 				}
--- a/code/client/launcher/UpdaterUI.cpp
+++ b/code/client/launcher/UpdaterUI.cpp
@@ -15,6 +15,8 @@
 
 #include "launcher.rc.h"
 
+#include <../citicore/LaunchMode.h>
+
 #include <ShellScalingApi.h>
 
 #include <winrt/Windows.Storage.Streams.h>
@@ -1330,6 +1332,12 @@ std::unique_ptr<TenUIBase> UI_InitTen()
 
 	if (getenv("CitizenFX_NoTenUI"))
 	{
+		forceOff = true;
+	}
+
+	if (CfxIsWine())
+	{
+		// WinRT XAML Islands not implemented in Wine/Proton, fall back to the plain Win32 controls UI.
 		forceOff = true;
 	}
 


RatPrez:
I did have a question for any maintainers on the `sticky` and `adhesive` differences.

I compared `sticky` and `adhesive` includes using `objdump` and noticed that sticky doesn't import `ws2_32` or any of the FiveM internal network dlls, which would explain why it never sends `cfxTicket2` on connection.

When you launch FiveM under `sticky` it does validate and assign you a R* license (visible in the main menu) and puts you into insecure mode because I am running linux, but it never sends the `cfxTicket2` ticket.

Is there any way to enable this on linux? Maybe even a local build workaround? Without it, it becomes difficult to develop using frameworks locally.

Obviously I understand the issues with validating a license when something is "insecure", but I thought it was still worth an ask. 

Thanks in advance :)

GekkeTovie:
Data point from testing this on a current build, in case it's useful: **the `UpdaterUI` half of this PR is confirmed necessary and sufficient, but the `SEHTableHandler` half does not appear to cover the failure I hit.**

### Environment

| | |
|---|---|
| Client | FiveM Legacy, `FiveM.exe` 2.0.0.**35843**, `CoreRT.dll` 1.0.0.35843, cache version 1256709 |
| Game | GTA V Legacy (Steam 271590), `GTA5.exe` **1.0.3889.0** |
| Proton | GE-Proton10-34 via umu-launcher 1.4.4, Steam Linux Runtime `sniper` 3.0.20260805.254768 |
| Host | Arch Linux, kernel 7.1.9, RX 7900 XT (RADV, Mesa 26.2.1) |
| Prefix | on a Linux filesystem (btrfs), not NTFS |

### `UI_InitTen` — confirmed

Without the change in this PR, the client dies before unpacking anything:

```
err:combase:RoGetActivationFactory Failed to find library for
    "Windows.UI.Xaml.Hosting.WindowsXamlManager"
err:seh:NtRaiseException Unhandled exception code c0000409
```

Setting `CitizenFX_NoTenUI=1` (the existing escape at `UpdaterUI.cpp:1331`) is a working
manual equivalent of this PR's `CfxIsWine()` branch. With it, the client installs, builds
its game cache for build 3889, and reaches the "Cfx.re: Insecure mode" notice. So that half
does exactly what it says. This also matches WineHQ
[bug 59987](https://bugs.winehq.org/show_bug.cgi?id=59987) (`RoGetActivationFactory` is a
semi-stub), filed against the CitizenFX client.

### The crash that follows

Immediately after dismissing the insecure-mode message box, every launch:

```
CoreRT.dll+1C471
Stack trace:
  CoreRT.dll+1C471
  CoreRT.dll+1C94F
  CoreRT.dll+144B1
  CoreRT.dll+1443B
  CoreRT.dll+20AA7
  FiveM.exe+A956B
  FiveM.exe+29079E
```

From the minidump's `MINIDUMP_EXCEPTION_STREAM`:

| field | value |
|---|---|
| `ExceptionCode` | `0xC0000005` (ACCESS_VIOLATION) |
| `ExceptionAddress` | `0x00006FFFFA44C471` |
| access type | **WRITE** (`ExceptionInformation[0] == 1`) |
| faulting address | `0x00006FFFFFEF1938` |

Module bases from the same dump: `CoreRT.dll` at `0x00006FFFFA430000`, `ntdll.dll` at
`0x00006FFFFFEF0000`. So this is a **write from `CoreRT.dll+0x1C471` into `ntdll+0x1938`**.

In GE-Proton10-34's `ntdll.dll`, RVA `0x1938` is inside `__wine_stub_RtlZeroHeap` — one of
Wine's stubs for unimplemented exports — which lives in `.text` (`R-X`):

```
.text    VA=0x001000  size=0x0731B0  R-X
.data    VA=0x075000  size=0x0048A0  RW-
.rdata   VA=0x081000  size=0x015760  R--
```

I don't have symbols for `CoreRT.dll`, so I'm deliberately **not** attributing this to a
specific function — just reporting where the write comes from and where it lands.

### Why I don't think the `SEHTableHandler` change covers this

This PR gates the `FatalError` inside the `if (!internalAddress)` branch. On this Wine build
that branch *is* reached, but it produces a different failure than what I see:

`FindCallFromAddress` returns `nullptr` whenever it finds **more than one** static call, and
Wine's `RtlLookupFunctionTable` has two:

```asm
000000017002d1a0 <RtlLookupFunctionTable>:
   17002d1b6:  call   170013bd0 <LdrFindEntryForAddress>
   17002d1e0:  call   170028e70 <RtlImageDirectoryEntryToData>
   17002d1ec:  ret
```

So `internalAddress` is `nullptr`, `IsWindows8Point1OrGreater()` is true (Wine reports Win10),
and without this PR it hits `FatalError`. With this PR it would instead fall through to
`internalAddress = baseAddress` and hook `RtlLookupFunctionTable` directly.

But the observed faulting write is at `ntdll+0x1938`, and `RtlLookupFunctionTable` is at
RVA `0x2D1A0` in this ntdll — so the crash I'm seeing is writing somewhere else entirely.
That makes me think this AV is a separate problem from the one the SEH hunk addresses, and
that landing this PR as-is would move the failure from `FatalError` to this crash rather
than past it.

Worth noting the same file already has a non-Windows path that would be safe here:

```cpp
else
{
    trace("Not running on Windows - no RtlLookupFunctionTable. Is this some fake OS?\n");
    MH_CreateHookApi(L"ntdll.dll", "RtlImageDirectoryEntryToData", ...);
}
```

It's never taken under Wine because the guard tests whether the *export exists*, and Wine's
ntdll does export `RtlLookupFunctionTable`. Gating the whole hooking path on `CfxIsWine()`
rather than only the `FatalError` might be the more robust shape — though that's a
suggestion, not something I've been able to test.

### Reproduction

1. Arch Linux, GE-Proton10-34 + umu-launcher, prefix on a Linux filesystem.
2. GTA V Legacy (271590) installed, build 1.0.3889.0.
3. `CitizenFX_NoTenUI=1 umu-run FiveM.exe`
4. Point it at the GTA V directory, accept the "needs to update the local game data" prompt.
5. Let the game data update finish, relaunch, dismiss the insecure-mode message box.
6. Crash. 100% reproducible across launches; the game cache persists, so it is not a
   partial-download artifact.

Happy to run a build with any change you want tested, or to pull more out of the minidump if
something specific would help.

**Scope note:** this is only about the client not crashing in open-source code — nothing here
is about the deliberate `CfxIsWine()` adhesive→sticky substitution in
`citicore/ComponentLoader.cpp`, which I'm treating as intended behaviour.

Issue #4083, opened 2026-07-26
cheater crash GTA5_b3258.exe+F2BC09 (vehicle)


mas7or123:
### What happened?

a cheater crashed every player inside a vehicle nearby

### Expected result

no crash

### Reproduction steps

no idea

### Importancy

Crash

### Area(s)

FiveM

### Specific version(s)

FiveM/31592

### Additional information

[cheater video (proof)](https://streamable.com/p60dr2)

[CfxCrashDump_2026_07_26_00_46_17.zip](https://github.com/user-attachments/files/30381942/CfxCrashDump_2026_07_26_00_46_17.zip)

PR #4100, opened 2026-08-02
fix(net): guard invalid vehicle archetype pointers

rafaadeveloper:
### Goal of this PR

Prevent a client crash caused by dereferencing an invalid vehicle archetype pointer during remote vehicle updates.

This change adds a defensive validation before the affected game-side path is reached, preventing malformed or corrupted vehicle archetype pointers from being passed to GTA's internal code.

- Adds a guard that validates vehicle archetype pointers before they are dereferenced.
- Rejects updates that contain an invalid or non-dereferenceable archetype pointer.
- Preserves the normal update path for valid vehicle entities.
- Prevents malformed network state from propagating into the game code where the reported crash occurs.

- FiveM
- Networking
- Vehicle synchronization

### Successfully tested on

**Game builds:** b3258

**Platforms:** Windows

> Note: This change is a defensive mitigation. The original corruption source has not yet been conclusively identified, but the guard prevents invalid archetype pointers from reaching the reported crash location.

### Checklist

### Fixes issues

Related to #4083

### DIFF
--- a/code/components/gta-net-five/src/CrashFixes.NetObjVehicleArchetype.cpp
+++ b/code/components/gta-net-five/src/CrashFixes.NetObjVehicleArchetype.cpp
@@ -1,0 +1,109 @@
+#include <StdInc.h>
+
+#include <Hooking.h>
+#include <Hooking.Patterns.h>
+#include <Hooking.Stubs.h>
+
+#include <EntitySystem.h>
+#include <netObject.h>
+
+//
+// GH-4083: a malicious client is able to leave a cloned vehicle with a fwEntity::m_archetype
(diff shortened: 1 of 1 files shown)


DaniGP17:
This is not a good solution, the fix should address the root cause.

rafaadeveloper:
> This is not a good solution, the fix should address the root cause.

Thanks for the feedback. I agree that the root cause should be fixed.

This was intended as a defensive mitigation since I couldn’t identify where the corruption originates from the information available in the issue. If there’s a reproducible case or additional context, I’d be happy to investigate the underlying cause.

mas7or123:
any idea how to fix it in my server for now

Issue #3389, opened 2025-05-09
Rockstar Editor crash everytime


Getobz:
### What happened?

Hello,

I'm currently experiencing a consistent crash when using the Rockstar Editor on my FiveM server.

When I record a clip and then go into the Rockstar Editor to view or edit it, the clip appears to load normally. However, right after the loading completes, FiveM crashes with the following error message:

> An exception occurred (c0000005 at 0x7ff753ed2f03) during loading of griz_cayo_restaurant in data file mounter 00007ffd46d11af8. The game will be terminated.

I've tested this issue on both the stable (Release) version of FiveM and the Canary (Latest/Unstable) version. Unfortunately, the crash happens in both cases.

Please let me know if there's any additional information I can provide to help troubleshoot this issue.

Thank you!

[CitizenFX_log_2025-05-09T155814.log](https://github.com/user-attachments/files/20125345/CitizenFX_log_2025-05-09T155814.log)

[36149485-6a42-4384-8681-c3ef59ebf88a.dmp](https://github.com/user-attachments/files/20125346/36149485-6a42-4384-8681-c3ef59ebf88a.dmp)

### Expected result

Start my clip for edit him

### Reproduction steps

1. Recording with Rockstar Editor in my server
2. Close FiveM
3. Launch FiveM and go to Rockstar Editor in the menu
4. Load the clip

### Importancy

Crash

### Area(s)

FiveM

### Specific version(s)

FiveM 14964 / Release version

Mathu-lmn:
Hey, I think it's related to escrowed assets, do you have any ? (also looks kinda similar to #3118 )

Gogsi:
[This is the resource](https://forum.cfx.re/t/mlo-cayo-perico-restaurant/5240814) in the error, I think it's not escrowed? Since it says code is accessible and it doesn't have the `cfx-` preset.

Getobz:
I'm not sure to be honest — I'm not the owner or developer of the server where I recorded the clip using Rockstar Editor, so I don't know if any escrowed assets are used.
Is there a way for me to check if an asset is escrowed, even without access to the server files?

manzarek123:
Rockstar editor will crash if you have a DLC_ITYP_REQUEST with an YTYP that contains an MLO instance.
Remove the incorrect DLC_ITYP_REQUEST line from the fxmanifest.lua and it should fix your issue
Also fix https://github.com/citizenfx/fivem/issues/3118 and some other issues about this problem
[...]

PR #4111, opened 2026-08-06
fix(gta/streaming): don't resolve game packfile handles when unregistering streamed files

TheMadGriffin:
### Goal of this PR

Fix a reliably reproducible client crash when entering the Rockstar Editor while connected to
any server that streams a file overriding a base-game asset (occlusion ymaps, `v_int_*` ytyps
and similar — very common on roleplay servers).

Crash signature: `EXCEPTION_ACCESS_VIOLATION` reading `0x10` at `GTA5_b3570.exe+1357D4F`,
called from `gta-streaming-five.dll+440AD`, legacy hash `summer-football-princess`.

This is very likely the mechanism behind the long-standing scattered reports of "the Rockstar
Editor always crashes on server X but works on server Y" — whether a server happens to stream
a base-asset override decides it.

When a streamed file overrides an asset that is already registered, `LoadStreamingFile.cpp`
pushes the **original game packfile handle** (collection index >= 2) onto `g_handleStack`
before overwriting the entry, then pushes the new cfx raw handle on top:

```cpp
// LoadStreamingFile.cpp:1934-1948 — registration, override branch
// if no old handle was saved, save the old handle
auto& hs = g_handleStack[strId + strModule->baseIdx];

if (hs.empty())
{
    hs.push_front(entry.handle);          // ORIGINAL game handle
}

entry.handle = (collectionId << 16) | idx;
g_handlesToTag[entry.handle] = tag;

// save the new handle
hs.push_front(entry.handle);              // cfx raw handle
```

The handle stack therefore legitimately contains handles the cfx raw streamers do not own.
The registration path already accounts for this and guards raw streamer access:

```cpp
// LoadStreamingFile.cpp:1965
// only for 'real' rawStreamer (mod variant likely won't reregister)
if (streaming::IsRawHandle(entry.handle))
```

The matching unregistration path in `CfxCollection_RemoveStreamingTag` has no such guard —
it walks the whole stack and unconditionally does:

```cpp
// LoadStreamingFile.cpp:2622-2634
auto rawStreamer = streaming::GetRawStreamerByIndex(streaming::GetCollectionIndex(*it));
auto entryName = rawStreamer->GetEntryName(streaming::GetEntryIndex(*it));
```

Rockstar Editor activation forces a disconnect whose session shutdown destroys the base
packfile collections **before** resources unmount. By the time the unregister loop runs,
resolving a saved game handle dereferences freed collection state.

This PR applies the same ownership check the registration side already uses, and null-checks
the returned streamer. No behaviour is lost by skipping non-raw handles: `file` is a resource
cache path, so it can never string-compare equal to a game packfile entry name — those
compares were dead in healthy runs and fatal during teardown.

The loop rewrite also fixes a pre-existing iterator bug, independent of the crash: the
original `it = handleData.erase(it)` inside a `for (...; ++it)` header skipped the element
following each erasure, and incremented past `end()` when the erased element was the last one.
The RDR3 branch is behaviourally unchanged.

<details>
<summary>Crash evidence (4 minidumps, one client, one server)</summary>

- 3 dumps: RIP `GTA5_b3570.exe+1357D4F`, called from `gta-streaming-five.dll+440AD`; the
  crashing frame's locals contain `cs6_occl_00` (an occlusion ymap override).
- 1 dump: RIP `GTA5_b3570.exe+1357D38` (same function), locals contain `v_int_1.ytyp`
  (a base interior ytyp override) — different session, different first override in unmount
  order.
- Exception context in all four: `RAX = 0`, AV read at `0x10`.
- R* gamelog at crash time: 0 vehicles, 0 objects, 1 ped, streaming idle. All local entities
  had been deleted before activation as a mitigation attempt and the crash reproduced
  identically, so world entity state is not a factor.
- Stopping the same resources *mid-session* does not crash, consistent with the collections
  still being alive at that point.

Dumps available on request.

</details>

**Repro:** connect to a server streaming any file whose name collides with a base-game asset
(a `*_occl_*.ymap` from any free MLO is the easiest source), then trigger
`ACTIVATE_ROCKSTAR_EDITOR`. The client crashes during the resource unmount following
`HS_FORCE_DISCONNECT`, before the editor appears.

FiveM

### Successfully tested on

**Game builds:** 3570

**Platforms:** Windows

**Testing status — please read.** This compiles clean with no new warnings, and the root cause
is established from four minidumps plus source analysis. I have **not** been able to complete
a runtime before/after run: a locally built `gta-streaming-five.dll` is rejected when dropped
into a retail client — the component loader deletes `content_index.xml` on the failed load and
the launcher's updater then restores the stock DLL on next start — so I could not get a
self-built component to load in a stock install to do the comparison.

If a maintainer with a dev-signed client can run it, the repro above is deterministic and
takes under a minute. Stock crashes every time; the patched build should enter the editor
cleanly. Happy to provide the dumps, the server-side repro setup, or to make any changes
requested.

### Checklist

- [ ] Code compiles and has been tested successfully. *(compiles clean; runtime testing blocked — see above)*

### DIFF
--- a/code/components/gta-streaming-five/src/LoadStreamingFile.cpp
+++ b/code/components/gta-streaming-five/src/LoadStreamingFile.cpp
@@ -2619,18 +2619,29 @@ void DLL_EXPORT CfxCollection_RemoveStreamingTag(const std::string& tag)
 				// erase existing stack entry
 				auto& handleData = g_handleStack[strId + strModule->baseIdx];
 
-				for (auto it = handleData.begin(); it != handleData.end(); ++it)
+				for (auto it = handleData.begin(); it != handleData.end(); )
 				{
+					bool matches = false;
+
 #ifdef GTA_FIVE
-					auto rawStreamer = streaming::GetRawStreamerByIndex(streaming::GetCollectionIndex(*it));
-					auto entryName = rawStreamer->GetEntryName(streaming::GetEntryIndex(*it));
-					if (entryName && strcmp(file.c_str(), entryName) == 0)
-#elif IS_RDR3
-					if (*it == idx)
-#endif
+					// Only handles owned by a cfx raw streamer can refer to this resource
+					// file: `file` is a resource cache path, which game packfile entry
+					// names never match. More importantly, saved handles for an asset we
+					// overrode point at game packfile collections, and during session
+					// shutdown (e.g. Rockstar Editor activation forcing a disconnect)
+					// those collections are already destroyed - resolving their entry
+					// names dereferences freed collection state and crashes.
+					if (streaming::IsRawHandle(*it))
 					{
-						it = handleData.erase(it);
+						auto rawStreamer = streaming::GetRawStreamerByIndex(streaming::GetCollectionIndex(*it));
+						auto entryName = rawStreamer ? rawStreamer->GetEntryName(streaming::GetEntryIndex(*it)) : nullptr;
+						matches = entryName && strcmp(file.c_str(), entryName) == 0;
 					}
+#elif IS_RDR3
+					matches = (*it == idx);
+#endif
+
+					it = matches ? handleData.erase(it) : std::next(it);
 				}
 
 				// if not empty, set the handle to the current stack entry


menordoodio1333333aa-spec:
Screenshot_20260807-125041

menordoodio1333333aa-spec:
https://github.com/citizenfx/fivem.git

menordoodio1333333aa-spec:
Muito bom fivem te amooooooooooooooo muitoooooooooooooioooooooo fivem

valerisn:
Hey! Quick question, and absolutely no offense intended. Is there a linked issue or existing report for this PR that I could take a look at?

Also, I may be completely wrong here, but the PR description reads a little AI-assisted to me. Was AI used anywhere while putting this together? Not asking as a criticism at all, just curious about the context and how the PR was prepared.

TheMadGriffin:
Hey Valerisn!

No offense at all! I opened this PR to resolve the long standing issue of Rockstar editor crashing when launched from in game. There have been quite a few issues opened for it. Here is one: https://github.com/citizenfx/fivem/issues/3389

I did use Claude Fable 5 to aid in putting this together and manually reviewed the description and code. I apologize if that is not permissible.

valerisn:
No no its alright, I get Writing Documentation with AI, was just curious thats all

TheMadGriffin:
> No no its alright, I get Writing Documentation with AI, was just curious thats all

It's crazy how good some of these tools are getting. I appreciate you taking the time to look at my PR.

PR #4123, opened 2026-08-12
fix(nui): make <input type="file"> file upload work

### LINKED ISSUE #3091: NUI issue with input files
### What happened?

Repro script available here:
https://forum.cfx.re/t/nui-issue-with-input-files/4929477

### Expected result

The expected behavior is to open the file browser again, you can try it in the browser, and it works perfectly.

### Reproduction steps

1. Type focus in F8
2. Click on the input to open the file browser and select any file
3. Click on the input again to change the file. It won’t open the file browser again, even if you reset the input value.

### Importancy

Slight inconvenience

### Area(s)

FiveM, RedM

### Specific version(s)

FiveM/RedM

### Additional information

_No response_

frostfire575:
### Goal of this PR

Make `<input type="file">` work in NUI. Today the OS file picker opens once and then never again until the game restarts (citizenfx/fivem#3091).

This matters because native file upload is a real building block for FiveM resources: MDT / CAD scripts let officers attach mugshots and evidence photos, character creators and identity systems let players upload an avatar, and admin / reporting tools attach screenshots. All of these rely on `<input type="file">`, and today it is unreliable — it works once and then silently stops, so authors resort to awkward workarounds (paste-a-URL fields, external uploaders) instead of a plain file picker. Fixing it makes the standard web pattern authors already know just work inside NUI.

NUIClient never implemented CefDialogHandler, so `<input type="file">` fell through to CEF's internal default dialog. The NUI browser is windowless with a null parent HWND, so that dialog has no owner window and strands CEF's single pending-file-chooser slot; once stuck, every later click resolves instantly as cancelled and the picker stops opening until restart.

An in-process IFileDialog does not work here: in the hooked, hand-mapped game process, shell-extension creation (`AssocCreateForClasses` -> `SHExtCoCreateInstanceString`) parks the shell's own worker-pool STAs in non-pumping waits, so `IFileDialog::Show()` blocks forever inside shell init without ever creating a window, and the stuck workers accumulate across attempts.

So the dialog is shown in a separate process: a clean copy of the main launcher exe (spawned via `MakeCfxSubProcess` as `FiveM_FileDialog.exe`) receives the request through an inherited file mapping, shows IFileDialog outside all of the game's hooks, and returns the result through the mapping plus an inherited completion event. The game waits on a pumping STA thread; a helper that never creates a window within the timeout is killed, so a wedge can no longer take the picker down. An RAII guard resolves the CEF callback exactly once on every path, so the pending-chooser slot cannot leak.

The helper defeats the foreground lock so the dialog reliably lands in front of the fullscreen game instead of opening behind it.

Because this CEF is Chromium 103 (predating the `<input type="file">` `cancel` event from Chrome 113), a `nuiFileDialogResult` message is pushed to the originating frame on every resolution (pick, cancel, busy-reject) so pages can detect a cancelled picker. It and the input's own `change` event ride different queues, so pages must tolerate either order.

Covers all four FileDialogMode values, translates `accept_filters` into `COMDLG_FILTERSPEC`, and sets `FOS_NOCHANGEDIR` so the dialog can't mutate the process-wide CWD.

Note on design: this spawns a helper process (a copy of the launcher exe via the existing `MakeCfxSubProcess`/DumpServer pattern) from the game process, using an inherit-handles `CreateProcess`. Failure modes are graceful — a `CreateProcess` failure or a helper that makes no window within the timeout both resolve as cancel, and the IPC block is fixed-layout with all reads bounds-checked. The one gap is that a fault inside the game-side wait loop escapes to breakpad rather than the C++ guard (SEH, not `catch(...)`), as noted in-code. The helper being a signed-launcher copy is the item most likely to need a maintainer's input on the release/signing pipeline.

FiveM, NUI

### Successfully tested on

**Game builds:** b3258

**Platforms:** Windows

Example + reproduction resource: https://github.com/frostfire575/fivem-nui-file-upload
A small resource that both demonstrates `<input type="file">` usage in NUI and reproduces #3091. Repeatedly opened single / multi-select / filtered pickers and cancels; picker now opens every time, cancels are detected via `nuiFileDialogResult`, no restart needed.

### Checklist

### Fixes issues

fixes #3091

### DIFF
--- a/code/client/launcher/GameSelect.cpp
+++ b/code/client/launcher/GameSelect.cpp
@@ -447,4 +447,405 @@ std::optional<int> EnsureGamePath()
 
 	return {};
 }
+
+// Out-of-process file dialog server for the game process.
+//
+// The game process cannot show IFileDialog itself: shell extension creation deadlocks on
+// the shell's own worker pool inside the hooked/hand-mapped game process (see
+// nui-core/src/NUIFileDialog.cpp). nui-core spawns a clean copy of this exe with
+// "-filedialog:<mappingHandle>:<eventHandle>" (inherited-handle-as-integer convention,
+// like -dumpserver/-switchcl) and reads the result back from the mapping. A helper that
+// hangs is simply killed by the game - that containment is the point of the design.
+//
+// Lives in GameSelect.cpp because this file is already main-personality-only and already
+// hosts the launcher's other IFileDialog use.
+
+#include <FileDialogIPC.h>
+
+namespace
+{
+class FileDialogServerEvents : public IFileDialogEvents
+{
+public:
+	explicit FileDialogServerEvents(fdipc::Response* response)
+		: m_response(response)
+	{
+	}
+
+	// IUnknown
+	IFACEMETHODIMP QueryInterface(REFIID riid, void** ppv) override
+	{
+		if (!ppv)
+		{
+			return E_POINTER;
+		}
+
+		if (riid == IID_IUnknown || riid == IID_IFileDialogEvents)
+		{
+			*ppv = static_cast<IFileDialogEvents*>(this);
+			AddRef();
+			return S_OK;
+		}
+
+		*ppv = nullptr;
+		return E_NOINTERFACE;
+	}
+
+	IFACEMETHODIMP_(ULONG) AddRef() override
+	{
+		return InterlockedIncrement(&m_refCount);
+	}
+
+	IFACEMETHODIMP_(ULONG) Release() override
+	{
+		auto refCount = InterlockedDecrement(&m_refCount);
+
+		if (!refCount)
+		{
+			delete this;
+		}
+
+		return refCount;
+	}
+
+	// IFileDialogEvents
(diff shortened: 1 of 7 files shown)


yorick2002:
hello claude please fix input type = file in the fivem make no mistakes thank you claude

frostfire575:
> hello claude please fix input type = file in the fivem make no mistakes thank you claude

Indeed , hey claude lets get heli and plane blades speed 😻

yorick2002:
> > hello claude please fix input type = file in the fivem make no mistakes thank you claude
> 
> Indeed , hey claude lets get heli and plane blades speed 😻

at least i put effort into that, you just prompted a bot to do something for you, + its bad

frostfire575:
> > > hello claude please fix input type = file in the fivem make no mistakes thank you claude
> > 
> > 
> > Indeed , hey claude lets get heli and plane blades speed 😻
> 
> at least i put effort into that, you just prompted a bot to do something for you, + its bad

Thanks for your effort for the community, keep it up though! i dont really want to argue over pointless stuff anyway. Have a good day ❤️

PR #4141, opened 2026-08-25
fix(rdr3): allocate third AMV buffer

Abardenn:
### Goal of this PR

Fix a RedM crash when disconnecting from a server and reconnecting again without restarting the client.

**AMVPresentBuffer** can access a 3rd **BitsetNumBytes**-sized buffer, but the replacement AMV bitset only allocated space for two.

This could overwrite future allocations in the shared stub memory area, including hook stubs used during reload.

Allocate space for all three AMV buffers to prevent the out-of-bounds write.

RedM

### Successfully tested on

**Game builds:**
1491

**Platforms:**
Windows

### Checklist

### Fixes issues

Tested with around 15 disconnect/reconnects without restarting RedM. The crash reproduced consistently before this change and did not reproduce after it.

Follow-up to #3145. This allocates space for the third AMV buffer without changing the original bitset sizing.


Abardenn:
Closed in favor of #4151

PR #4142, opened 2026-08-26
tweak(gta-streaming-five): Cut connect time in half on asset heavy servers

xalva98:
### Goal of this PR

Cut the time it takes to connect to an asset-heavy server roughly in half:
**67s -> 34s (1.98x)**, measured from the first init function to spawning in on a
95,480 file / 36.9 GB streaming set.

The asset load itself, which is the part this PR touches, goes from **42.5s to
16.0s (2.66x)**. The game's own init work inside that window measures ~10s on my machine and is untouched

- `ResourceCacheDeviceV2::GetEntryForFileName` now remembers the last file it
  resolved. The load path asks for the same file two or three times in a row -
  existence, resource version, open - and each resolution walked the resource
  manager and two maps. Invalidated on resource start/stop.
- `ResourceCacheDeviceV2` gains a `GetAttributes` override, which
  `ResourceCacheDevice` already had. Without it the base implementation opens and
  closes the file, allocating a stream object and a handle per call.
- `GetBaseName` probed the packfile index for every streaming file to check for a
  `platform:/textures/` override, on both the registration and the load path.
  That directory is now enumerated once (134us -> 0.2us per call). Nested names
  still fall back to a probe, but only when the directory has subdirectories -
  `subdir_file_mapping` rewrites `^` to `/`, which most asset-pack files contain.
- The existence check and the size fixup resolved the same path twice; the device
  pointer is reused.
- `SearchTypeName` is cached per mounter; the name depends only on the vtable.
- The per-file `loading`/`done`/`can't register` traces become `DPrintf`, since
  print listeners run inline on the calling thread. The skipped-file diagnostic
  becomes one summary line per pass naming the resources and counts.
- The `LoadDataFiles` frame pump ran once per data file and drives a Scaleform
  frame advance plus a streaming progress scan. `game_loadDataFileFrameInterval`
  defaults to 33ms; `0` restores per-file.

Which assets get registered is unchanged.

FiveM, RedM

### Successfully tested on

Game builds: **3258 (FiveM)**
Platforms: **Windows**

95,480 streaming files / 36.9 GB, both builds alternating which one runs first:

| | run 1 | run 2 | run 3 | median |
|---|---|---|---|---|
| before | 103.6s | 64.8s | 67.0s | **67.0s** |
| after | 42.8s | 33.8s | 33.7s | **33.8s** |

Note: the 103.6s is a cold start / empty cache, so is 42.8s

Faster in every round. Verified after each step with a test resource that
requests vehicle models, the freemode ped and addon clothing and reports
pass/fail - all load, registration unchanged. 

### Checklist

Note: I did NOT test this in RedM as I don't really have enough streaming assets to test with

### DIFF
--- a/code/components/citizen-resources-client/include/ResourceCacheDeviceV2.h
+++ b/code/components/citizen-resources-client/include/ResourceCacheDeviceV2.h
@@ -167,6 +167,8 @@ public:
 
 	virtual std::shared_ptr<RcdBulkStream> OpenBulkStream(const std::string& fileName, uint64_t* ptr) override;
 
+	virtual uint32_t GetAttributes(const std::string& fileName) override;
+
 	virtual bool ExtensionCtl(int controlIdx, void* controlData, size_t controlSize) override;
 
 	virtual size_t GetLength(const std::string& fileName) override;
--- a/code/components/citizen-resources-client/src/ResourceCacheDeviceV2.cpp
+++ b/code/components/citizen-resources-client/src/ResourceCacheDeviceV2.cpp
@@ -300,6 +300,11 @@ size_t ResourceCacheDeviceV2::GetRealSize(const ResourceCacheEntryList::Entry& e
 	return realSize;
 }
 
+uint32_t ResourceCacheDeviceV2::GetAttributes(const std::string& fileName)
+{
+	return GetEntryForFileName(fileName) ? 0 : -1;
+}
+
 bool ResourceCacheDeviceV2::ExistsOnDisk(const std::string& fileName)
 {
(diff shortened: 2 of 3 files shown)


Legacy-TacticalGamingInteractive:
Hopefully this gets implemented!

Legacy-TacticalGamingInteractive:
are you still working on this

xalva98:
> are you still working on this

No, I just fixed a bug for RedM.
If you want to check it out, it is on a feature branch `specific/feature/faster-asset-load-on-connect`

PR #4145, opened 2026-08-26
fix(gta-streaming-five): save_gta_cache crashing since b2802

xalva98:
### Goal of this PR

Make `save_gta_cache` work again. It currently crashes the client, or writes a
258 byte file with three empty module sections when something catches the
exception.

b2802 added an extra parameter to the `strCacheLoader` module save callback.
Calling it without one left whatever happened to be in `rcx` to be used as the
stream, which faulted in the flush before the cache append was ever reached.
Declare the parameter and pass `nullptr`; builds below b2802 ignore it.

FiveM

### Successfully tested on

Game builds: **3258 (FiveM)**
Platforms: **Windows**

Checked every build from 1604 to
3751 for the signature change: 1604-2699 take no argument, 2802-3751 do.

### Checklist

### DIFF
--- a/code/components/gta-streaming-five/src/CacheLoader.cpp
+++ b/code/components/gta-streaming-five/src/CacheLoader.cpp
@@ -59,10 +59,12 @@ struct
 {
 	struct
 	{
-		void(*loadCache)();
-		void(*saveCache)();
-		char name[48];
-		int size;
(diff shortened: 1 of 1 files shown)


Nobelium-cfx (approved):
LGTM

PR #4152, opened 2026-08-29
feat(gta-streaming-five): add ped metadata data file mounters

### LINKED ISSUE #2188: [REQUEST] Streaming facial_overlays.meta
Would it be possible to add data_file to mount  facial_overlays.meta <CPedFacialOverlays>?

gurkansahinn:
### Goal of this PR

`facial_overlays.meta` and `ped_skin_tones.meta` are loaded once from fixed paths during startup and are not present in the game's data file type table. Because of this, resources cannot add custom makeup, facial hair, eyebrows, or skin tones.

Requested in #2188.

Adds two Cfx-side data file types, resolved before the game enum lookup:

```lua
data_file 'CFX_PED_FACIAL_OVERLAYS' 'facial_overlays.meta'
data_file 'CFX_PED_SKIN_TONES'      'ped_skin_tones.meta'
```

Both mounters use the game's parser to load the file into a game-allocated instance and append its entries to the live store.

Duplicate entries are skipped, so resources can include either the full file or only their additions. On unload, only the entries appended by that resource are removed.

Also exposes `rage::LoadFromStructure`, which is used by both mounters.

Tested with a resource containing 141 overlay textures and a `ped_skin_tones.meta` containing the 12 default entries plus 4 additional entries:

| Component | Before | After |
|---|---:|---:|
| Facial hair | 29 | 87 |
| Eyebrows | 34 | 68 |
| Skin tones per component | 12 | 16 |

`GetPedHeadOverlayNum` returns the new counts, and the added overlays render correctly on freemode peds. This was also tested in multiplayer with the overlays synchronized between players.

Test: https://www.youtube.com/watch?v=NW949uSiLDY&feature=youtu.be
A client without these changes logs both data file types as invalid, as expected.

`CPedSkinTones` also contains `males`, `females`, `uniqueMales`, and `uniqueFemales`. These are indices based on the ordering of the loaded file and cannot be appended without remapping them against the component offsets. They are ignored for now, and the ignored entry count is logged.

FiveM

### Successfully tested on

**Game builds:** 3258, 3751

**Platforms:** Windows

**Multiplayer:** Tested successfully with multiple clients

### Checklist

### Fixes issues

Closes #2188

### DIFF
--- a/code/components/gta-streaming-five/src/LoadStreamingFile.cpp
+++ b/code/components/gta-streaming-five/src/LoadStreamingFile.cpp
@@ -1559,6 +1559,148 @@ public:
 
 static CfxPedFacialOverlaysMounter g_pedFacialOverlaysMounter;
 
+// layout per the game's parser definition: three component arrays at 0, then four byte
+// arrays at 48/64/80/96, 112 bytes in total
+struct CPedSkinTonesStore
+{
+	atArray<uint32_t> comps[3];
+
+	// males/females/uniqueMales/uniqueFemales - indices into the file's own ordering, so
+	// appending them would need remapping; only read to report that they were ignored
+	atArray<uint8_t> genderIndices[4];
+};
+
+// these stores load from a run of near-identical call sites and the path operand cannot
+// anchor them, so each match is validated against the shape CPedSkinTones has to have
+static CPedSkinTonesStore* GetPedSkinTonesStore()
+{
+	static CPedSkinTonesStore* store = ([]() -> CPedSkinTonesStore*
+	{
+		auto matches = hook::pattern("48 8B 0D ? ? ? ? 48 8D 05 ? ? ? ? 48 8D 15 ? ? ? ? 4C 8B C3");
+		matches.count_hint(8);
+
+		if (matches.empty())
+		{
+			trace("CFX_PED_SKIN_TONES: no candidate call sites found\n");
+			return nullptr;
+		}
+
+		for (size_t i = 0; i < matches.size(); i++)
+		{
+			// the store is the operand of the second instruction in the run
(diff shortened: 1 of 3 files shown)


gurkansahinn:
Small update: wolfram from CFX said an RFC would still be good for tracking this for Enhanced too.

I will open it this week with the technical details and everything I found while reversing/testing this. This PR is still the tested implementation for 3258/3751 and multiplayer, but the RFC should make the approach easier to review/discuss.

code/components/gta-streaming-five/src/LoadStreamingFile.cpp
```diff
@@ -1398,6 +1398,314 @@ struct CInteriorProxy
...
+static uint16_t AppendUniqueHashes(atArray<uint32_t>& dst, atArray<uint32_t>& src)
+{
+	uint16_t srcCount = src.GetCount();
+	if (srcCount == 0)
+	{
+		return 0;
+	}
+
+	// reserve for the worst case, where none of the entries are already present
+	if (dst.GetCount() + srcCount > dst.GetSize())
```
divocbn, line 1450: iirc this causes or can cause an OOB write since the addition in the if statement is being promoted to an int, but Expand down below takes a uint16_t, so basically everything above uint16_t maximum gets truncated (atArray capacity shrinks / or fails to expand -> OOB)

gurkansahinn: fixed in 70a85db.

i now clamp srcCount before the reserve based on how much room is left in the array. this keeps the total within uint16_t, so nothing gets truncated when passed to Expand and m_count can't wrap.

1DEMENT1:
@gurkansahinn 
How can I get in touch with you on Discord?

PR #4166, opened 2026-09-02
feat(server): support a list of players in TriggerClientEvent

spacevx:
### Goal of this PR
Allow `TriggerClientEvent` on the server to send an event to a list of players in one call, instead of looping and calling it once per player.

Based on the PR of @PichotM https://github.com/citizenfx/fivem/pull/1210

I added a new native, `TRIGGER_MULTICAST_CLIENT_EVENT_INTERNAL` (i didn't want to edit TriggerClientEventInternal, so i made a new one) that takes the same event name and payload as `TRIGGER_CLIENT_EVENT_INTERNAL` plus the targets as a msgpack array of net ids. The packet is built once, the list is deduplicated so a player never receives the event twice, and the packet is sent to every connected client of the list. The packet building of the existing `TriggerClientEvent` was moved to a shared function so both paths produce the exact same bytes, the scalar path itself is unchanged.

```lua
local players = { 4, 5, 6 }
TriggerClientEvent("sendmsg", players, "Hello everyone")

-- You can also do
local players = { "4", "5", "6" }
TriggerClientEvent("sendmsg", players, "Hello everyone")
```

Server, Natives, ScRT: Lua, ScRT: JS, ScRT: C#

### Successfully tested on

**Platforms:** Windows

### Checklist

### Fixes issues
None


AvarianKnight:
This is a great change, but I don't think we should add this onto existing paths, we should just make new wrapper functions for the multicast variant.

spacevx:
> This is a great change, but I don't think we should add this onto existing paths, we should just make new wrapper functions for the multicast variant.

I did it like this, because you said: "From a general perspective adding more functions might be more confusing than just overloading TriggerClientEvent, especially when they copy the same behavior"

And for me too it's better to not recreate another native, but if everyone prefer a new native, that's okay for me x)

spacevx:
@FabianTerhorst What's your opinion on this?

manups4e:
~Maybe i missed that part in the code.. but what if i send the event to a list with server id repeated such as `{1, 1, 2, 2, 3, 3, 3, 3}`?~
Edit: Found it! 😄  seems nice to me.. especially in the .net part

PR #4169, opened 2026-09-04
feat(extra-natives-five): add curb boost control native

MidgetHD:
### Goal of this PR

Add `SET_VEHICLE_CURB_BOOST_ENABLED`, allowing resources to disable curb-induced speed boosting on individual vehicles without enabling the unrelated suspension, downforce, and drag behavior associated with `CF_USE_DOWNFORCE_BIAS`.

The new client native applies a Cfx-only marker to the vehicle's current wheel objects.

An allocation free hook in the longitudinal tyre-slip update checks this marker and skips the relevant smoothing operation when curb boosting is disabled. The game's original suspension-raise condition remains unchanged.

A native declaration and usage documentation are included.

FiveM, Natives

### Successfully tested on

**Game builds:** 3095, 3570

**Platforms:** Windows

Testing included:

- Debug x64 compilation of `extra-natives-five`
- Disabling curb boosting with `SET_VEHICLE_CURB_BOOST_ENABLED`
- Driving repeatedly over a large bump
- Confirming the vehicle no longer gained speed from the bump
- Confirming no crash occurred during the test

### Checklist

### Fixes issues

None.

### DIFF
--- a/code/components/extra-natives-five/src/VehicleExtraNatives.cpp
+++ b/code/components/extra-natives-five/src/VehicleExtraNatives.cpp
@@ -322,6 +322,39 @@ static int WheelPowerOffset;
 static int WheelHealthOffset; // = 0x1E8; // 75 24 F3 0F 10 81 ? ? ? ? F3 0F
 static int WheelSurfaceMaterialOffset;
 static int WheelFlagsOffset;
+// GTA V uses bits 0-25 of this 32-bit field. Reserve bit 30 for a Cfx-only
+// override so the physics hook can remain allocation-free.
+static constexpr uint32_t CfxCurbBoostDisabledWheelFlag = 0x40000000;
+
+static void SetVehicleCurbBoostEnabledForWheels(fwEntity* vehicle, bool enabled)
+{
+	const unsigned char numWheels = readValue<unsigned char>(vehicle, NumWheelsOffset);
+	const uintptr_t wheelsAddress = readValue<uintptr_t>(vehicle, WheelsPtrOffset);
+
+	if (!wheelsAddress)
+	{
+		return;
+	}
+
+	for (unsigned char wheelIndex = 0; wheelIndex < numWheels; ++wheelIndex)
+	{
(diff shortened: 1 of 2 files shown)


code/components/extra-natives-five/src/VehicleExtraNatives.cpp
```diff
@@ -761,6 +794,54 @@ static HookFunction initFunction([]()
 		WheelFlagsOffset = *(uint32_t*)(location + 7);
 	}
 
+	{
+		auto curbBoostLocation = hook::get_pattern<char>("0F 2E BB 4C 01 00 00 75 ? 44 38 2D ? ? ? ? F3 0F 10 05 ? ? ? ? 74 ? 48 8B 83 20 01 00 00 0F 2F 40 14");
```
ook3D: Pattern is not valid <2802, needs to use xbr

mickeygeecom:
Love this! Whats the difference between using the Open Wheel flag (on each handling.meta) and this?

Gogsi:
> Love this! Whats the difference between using the Open Wheel flag (on each handling.meta) and this?

The Open Wheel flag (CF_USE_DOWNFORCE_BIAS) expects the vehicle to have a rear wing which changes the downforce calculations. When applied to vehicles that don't have that, the result isn't great :D

This patch enables the part of CF_USE_DOWNFORCE_BIAS which prevents curbboosting, without the extra downforce calculations

mickeygeecom:
Would be lovely to have this as a global setting for all vehicles instead of having to use the native for each vehicle spawn. So basicly a replicated client variable?

alberttheprince:
> Would be lovely to have this as a global setting for all vehicles instead of having to use the native for each vehicle spawn. So basicly a replicated client variable?

Yeah, I'm not sure when a server would want to selectively apply this to vehicles rather than all vehicles.

ook3D:
> > Would be lovely to have this as a global setting for all vehicles instead of having to use the native for each vehicle spawn. So basicly a replicated client variable?
> 
> Yeah, I'm not sure when a server would want to selectively apply this to vehicles rather than all vehicles.

minigames servers for example, modes with and without

PR #4172, opened 2026-09-04
feat(gta-core-rdr3): support resource audio wavepacks

Mosquito-Man:
### Goal of this PR

Port FiveM's resource `AUDIO_WAVEPACK` support to RedM.

This allows RedM resources to stream custom AWC wavepacks using the same fxmanifest layout as FiveM, 

Uses a similar approach to FiveM's PatchAudioWavePackOverlay for RedM.

Resource wavepacks are checked when the game builds the path for an audio bank, while keeping the original game path as fallback.

RedM resources can mount after the audio engine has initialized, so already loaded banks affected by a new or removed wavepack are requested again.

The RPF2 VFS metadata is also populated so RAGE can read these files properly from resources.

Replacement wavepacks have been tested. I wasn't able to test completely new wavepacks, as exposing new sounds and events generally also requires matching audio metadata, which is outside the scope of this PR.

RedM

### Successfully tested on

**Game builds:** 1491

**Platforms:** Windows

Tested 33 AWC banks across 7 wavepacks, including resource restarts while affected bank requests were in flight, mounting before and after audio engine initialization, and local/distant weapon sounds.

### Checklist

### Fixes issues

### DIFF
--- a/code/components/gta-core-rdr3/src/PatchAudioWavePackOverlay.cpp
+++ b/code/components/gta-core-rdr3/src/PatchAudioWavePackOverlay.cpp
@@ -1,0 +1,421 @@
+/*
+ * This file is part of the Cfx project - https://cfx.re/
+ *
+ * See LICENSE in the root of the source tree for information
+ * regarding licensing.
+ */
+
+#include <StdInc.h>
+#include <Error.h>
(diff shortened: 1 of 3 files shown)


HalconBM:
Thanks!

PR #4175, opened 2026-09-07
fix(mono): keep RELAX_DELEGATE, drop RELAX_REFLECTION

vertexitde:
### Goal of this PR
A player who joins a server should not have their local disk queried by that server's C# client resource. On current master they do. This started as a duplicate of #3987. That PR dropped RELAX_REFLECTION and was closed by pointing at #3138. Closing it did not change the code. The option is still set. The hole is still there.

This matters because the trust boundary is backwards. The server is untrusted. The filesystem belongs to the player. CoreCLR is turned on specifically so resource scripts stay on the transparent side of that line. No prompt. No permission. No way for the player to refuse it without leaving the server.

A remote resource only needs a path it already knows. Common install locations, launchers, tools, known cheat filenames. The answer comes back as an exception type. That is a software inventory of the player's machine, taken from anyone who clicks Join. You do not need to list a directory and you do not need to brute-force names. A wordlist of well-known paths is enough, and it can be repeated on every connect.

People have already treated this as "it only tells you exist / not exist, so it is fine." That is the wrong bar. Existence of a path is still data the player did not hand over.

I also checked the mechanism before sending this. RELAX_REFLECTION is not what makes Assembly.LoadFrom callable. In the shipped client mscorlib that method is SecuritySafeCritical. Transparent code can call it with the flag on or off. I embedded the repo's client Mono (4.9.1). Present non-assembly path: BadImageFormatException. Missing path: FileNotFoundException. Same result for flags 3 (master), 2 (delegate only), and 0. File.ReadAllText stays SecurityException. So #3987 named the right problem and locked the wrong door. Leaving RELAX_REFLECTION on is still extra surface for other Critical APIs. It is not the LoadFrom channel.

This has already shown up in real client resources. Filing it once and walking away did not make it unused.

Two changes, both client Mono.

1. Keep RELAX_DELEGATE, drop RELAX_REFLECTION.

BaseScript wires [Tick] and events with CreateDelegate. Newtonsoft.Json does the same when it reads private members. That is what #3138 actually was. Taking RELAX_DELEGATE out broke live servers and got reverted in cc97bd6. This PR does not touch that flag.

RELAX_REFLECTION is the other switch. It relaxes the CoreCLR reflection access check. #3987 already proposed this split and was closed as if it would repeat #3138. Please read those stacks before doing that again. They are all CreateDelegate. That is RELAX_DELEGATE.

2. Stop path-based assembly opens from leaving the FiveM app directory.

That is the LoadFrom channel. Resource scripts are loaded from bytes (OpenHostFile + Assembly.Load), so they do not need arbitrary filesystem LoadFrom. This hooks the Mono open-from-filename APIs and allows only paths under GetAbsoluteCitPath() (Citizen and cache). Anything else returns MONO_IMAGE_ERROR_ERRNO / ENOENT. The target file is not opened. Present and missing look the same.

#3987 is the duplicate of the flag half. This PR is that half plus the path gate that actually closes the disk probe.

Related / duplicates:
- #3987 (flag-only, closed, nothing landed)
- #3138 (different flag; that regression is why the flag half keeps getting waved off)

ScRT: C#, FiveM (client)

### Successfully tested on
Not a full live client. The shipped client Mono was run in a separate process with the same CoreCLR platform callback. Critical APIs from transparent code are refused. LoadFrom still reaches disk unless the path gate is applied.

The path rule was checked against the repo client tree: platform mscorlib is allowed, including a .. form that still resolves inside the tree. Audit fixtures and C:\Windows are denied. Present and missing get the same deny. Resource assemblies loaded from bytes are not on this path.

### Checklist

### DIFF
--- a/code/components/citizen-scripting-mono/component.lua
+++ b/code/components/citizen-scripting-mono/component.lua
@@ -6,6 +6,8 @@ return function()
 	add_dependencies { 'vendor:eastl' }
 
 	if os.istarget('windows') then
+		add_dependencies { 'vendor:minhook' }
+
 		filter 'architecture:x64'
 			links { "mono-2.0-sgen" }
 	
--- a/code/components/citizen-scripting-mono/src/MonoComponentHostShared.cpp
+++ b/code/components/citizen-scripting-mono/src/MonoComponentHostShared.cpp
@@ -9,11 +9,17 @@
 #include <mono/metadata/exception.h>
 #include <mono/metadata/mono-debug.h>
 #include <mono/metadata/mono-gc.h>
+#include <mono/metadata/image.h>
 
 #ifndef IS_FXSERVER
 extern "C" {
 #include <mono/metadata/security-core-clr.h>
 }
+#endif
+
+#if !defined(IS_FXSERVER) && defined(_WIN32)
+#include <cerrno>
+#include <MinHook.h>
 #endif
 
 #ifdef _WIN32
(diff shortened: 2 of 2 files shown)


vertexitde:
### Why this matters outside the code

Adding context, since the flag half of this was waved off once already.

FiveM ships to players in the EU, and there is a separate privacy layer to what a joined server can learn from a player's local machine.

Under ePrivacy Art. 5(3), implemented in Germany by § 25 TDDDG, gaining access to information already stored on a user's terminal equipment generally requires prior consent unless the access is strictly necessary to provide a service explicitly requested by the user. Importantly, this protection is not limited to personal data.

A resource script probing whether a specific local path exists obtains information about the state of the player's terminal equipment. Joining a server does not by itself amount to consent to arbitrary filesystem probing, and checking unrelated software or install locations is difficult to characterize as strictly necessary for executing the server's resources.

So "it only reveals existence, therefore it is fine" is not the relevant threshold.

There is a GDPR dimension as well. A FiveM server already processes identifiers associated with a player, such as license identifiers and IP addresses, and may also associate Steam or Discord identifiers with that session. A collection of local path-existence observations associated with those identifiers can become information relating to an identifiable player and may contribute to a device or software-environment profile. Device fingerprinting has specifically been addressed in Article 29 Working Party Opinion 9/2014 in the context of ePrivacy.

The more important architectural issue is that the server operator may be the party responsible for processing performed by its resources, while the platform exposes a capability that the operator may not reasonably expect a resource to possess. That makes it unnecessarily difficult for operators to understand and control what information their server can obtain from players' devices.

This is also why privacy by design matters here. Recital 78 GDPR expressly calls on producers of products, services and applications to take data-protection principles into account so that controllers are able to fulfil their obligations.

That is exactly the direction of this PR.

Resource assemblies are already loaded from bytes. Arbitrary path-based `Assembly.LoadFrom` outside the Citizen tree is therefore not a capability ordinary resource scripts need. Removing that capability from the untrusted side prevents the information from being exposed in the first place rather than expecting every server operator to somehow police its use.

There is also a forward-looking cybersecurity angle. Regulation (EU) 2024/2847 (Cyber Resilience Act) requires products with digital elements, where applicable, to be designed around principles including secure-by-default configuration, limitation of attack surfaces, and processing only data _- personal or otherwise -_ that is adequate, relevant and limited to what is necessary for the product's intended purpose (Annex I, Part I). Article 14 applies from 11 September 2026; the Regulation generally applies from 11 December 2027.

None of this replaces the technical argument above, and I would prefer this PR to be judged on that technical argument.

But the "it only leaks existence" framing does not make the privacy issue disappear. A remote, untrusted resource being able to distinguish whether arbitrary known paths exist on a player's local machine is itself the capability that should not be exposed.

dark0x6:
I'm the author of #3987. Nice work, and your analysis holds.

What worries me more is the delay. My PR was opened four months ago, closed by pointing at #3138, and nothing has moved since: the option is still set on master and the probe still works. The problem was never disputed on the merits, it just never got handled.

Hopefully this one gets a real review : )

Issue #3456, opened 2025-06-17
Server begins hitching intermittently


Damen57:
### What happened?

Hi all,

Just over a week ago we migrated our server from Linux to Windows and also changed hosts. Since that change, we've been running into some serious and consistent issues with the server tick rate. At seemingly random intervals, the tick rate starts spiking rapidly, and then the sync thread begins hitching repeatedly until the server either partially hangs or becomes unstable enough that we have to restart it manually, or txAdmin restarts it for us.

At first, we thought it might be caused by excessive MySQL queries flooding the connection pool and overwhelming oxmysql. In response, we put a lot of work into cutting down the number of queries. Most of them now run in batches, and we've also reworked our logging so that file writes are done in batches rather than on every single log event. These changes have improved overall performance, but the main issue is still happening. The tick rate will suddenly spike and continue getting worse until a restart is necessary.

We’ve tried capturing procdumps, but since the server doesn’t actually crash and instead just becomes unusable, those haven’t been useful. We also ran server profiles, but the data is extremely noisy. It mostly reflects the downstream effects of the issue, like excessive thread hitching, without pointing to a specific root cause. We haven’t been able to narrow it down to one resource or pattern that consistently triggers it.

At this point, we’re almost out of ideas. Reverting back to Linux is an option (we never had this problem on Linux), but we’d prefer to stick with Windows if possible, since it gives us better access to tools for diagnosing problems. (It has been a huge help in the optimisations we have made so far)

If anyone has seen something like this before, or has suggestions for what we can try next, we’d really appreciate it.

https://i.gyazo.com/d93b354f1d5fa59b39f22959e22d5222.png
https://i.gyazo.com/ad9e7e90d4a8efa3a391e81783f724b2.png
https://i.gyazo.com/400b46ad2ff950e1564ddd688ab67ab8.png

### Expected result

The server continue to operate

### Reproduction steps

Start the server on a Windows-based host using the same configuration that previously worked on Linux

Allow players to connect and play normally (issue does not appear immediately)

Monitor tick rate using resmon or profiler

At a seemingly random time (often during regular load, but sometimes low pop), tick rate spikes significantly

Sync thread begins hitching, often escalating quickly

Server either partially hangs or requires restart due to unusability

Sometimes the issue happens very quickly after a restart and requires a second restart shortly after

### Importancy

Crash

### Area(s)

FiveM, FXServer, OneSync

### Specific version(s)

Fivem fx: b15414/Windows

[...]

PR #4181, opened 2026-09-08
perf(state/five): pool Windows ped sync-tree allocations

Damen57:
### Goal of this PR

Use a bounded raw-storage pool for `std::allocate_shared<CPedSyncTree>` on Windows/FiveM. This targets allocator overhead and contention during frequent ped sync-tree creation/destruction. Other entity factories, Linux and RDR3 are unchanged.

This is a **draft for review**, extracted from a locally deployed artifact-35245 build and applied cleanly to upstream `master` at `0d8a2a6f78a9922445d8930305af82a7b1826980`. It is not a claim that all OneSync or server-main-thread hitching is fixed. A full build of the extracted patch against this upstream base remains outstanding.

Related investigation: [#3456 — Server begins hitching intermittently](https://github.com/citizenfx/fivem/issues/3456). This patch addresses the ped sync-tree allocation bottleneck identified during that investigation, not every cause of the broader hitching report.

Our diagnostic investigation found expensive ped sync-tree construction during degraded server operation. A real `CPedSyncTree` microbenchmark also showed a substantial difference between repeated stock shared allocation and pooled storage under contention. We deployed the targeted allocator change separately from a later Node event-loop change.

Subsequent production captures reached 300+ connected clients while average sync-tree construction remained in the low microseconds. This is observational evidence, not a controlled production A/B result. Some sync hitches, pending-create accumulation and missing-entity ownership traffic remain. Separate svMain traces still show general heap contention; those traces do not establish that this pool fixes every allocator bottleneck.

- Replace only the Windows/FiveM `CPedSyncTree` factory's `make_shared` with `allocate_shared` using a dedicated allocator.
- Pool **raw storage**, not live tree instances. Standard shared-pointer machinery still constructs and destroys the tree; the slot is not returned until the shared allocation's final weak reference releases it.
- Use Windows interlocked SLists for free-slot push/pop. Only slab growth takes a separate SRW lock.
- Allocate lazy 1 MiB slabs using `VirtualAlloc`, bypassing the shared CRT heap for normal pooled allocations. Alignment is checked at compile time.
- Retain slabs until process exit. The tested limit is **512 MiB per rebound allocation type**; there is one expected production shared-allocation type here. At the tested MSVC slot size of 16,688 bytes, the limit holds 31,744 slots. This is compiler/layout-dependent.
- If the cache is full or slab allocation fails, use aligned CRT allocation/free. The cache limit is not a ped/entity limit.
- Preserve default allocator behaviour for array allocations. No changes to ownership, state bags, clone acceptance, timeouts, saved frames or pending-create cleanup.

The two-file diff retains inexpensive allocator counters used by the standalone regression check. Private HTTP/Prometheus exporters and the wider diagnostic instrumentation are deliberately not included.

### Verification

- Original 512 MiB state DLL: Release build succeeded; isolated artifact-35245 startup and endpoint smoke checks passed.
- Standalone allocator regression: freshly rebuilt and passed before submission, with 818,002 test-object constructions checking fresh initialization, virtual destruction, `shared_from_this`, last-weak-reference lifetime, constructor exceptions, eight-thread churn, cross-thread release, alignment, cache exhaustion and fallback. The standalone harness and MSVC runner are available in the local review bundle, not committed in this two-file PR; upstream test integration remains a review item.
- Historical real-tree harness: fresh acknowledgement state matched the stock tree across reuse; 20,000 retained trees exceeded the old 256 MiB limit with zero fallbacks; cross-thread destruction and weak-reference retention passed.
- Historical synthetic benchmark: eight threads / 400,000 real trees took approximately 309 ms stock versus 42 ms pooled on the test machine. This is not a production latency prediction.
- The two-file patch passed application and whitespace checks against the upstream base above. The submitted allocator's Git blob matches the tested bundle. Do not equate the original full diagnostic-DLL build with a fresh full build of this clean diff against upstream HEAD.

#### Review points / limitations

1. Agree the upstream cache policy: retain the tested 512 MiB cap, lower it, or add a startup configuration option in a separately tested revision. This draft deliberately does not invent an untested default/convar.
2. Review allocator rebind, alignment and shared/weak lifetime behaviour across supported MSVC/runtime versions. Only the tested x64 toolchain has been exercised.
3. Process-lifetime slab retention is intentional. Review component unload/lifecycle assumptions and whether upstream wants a different storage ownership design.
4. Decide where the standalone regression belongs in the upstream test suite and whether the small counters should remain or be adapted to existing upstream instrumentation.
5. Run the full Windows build and relevant platform CI before marking ready for merge, and refresh the base if needed. The non-Windows/RDR3 branches are textually unchanged, not newly CI-validated here.
6. Aggregate pool counters do not prove absence of memory corruption. No exhaustive production-load or memory-safety proof is claimed.

#### Scope deliberately excluded

Node `DrainTasks`/foreground-pump changes, TCP/TLS buffer changes, resource scripts, private diagnostic endpoints, server dashboard changes, old-runtime build compatibility workarounds and generated binaries.

FiveM, FXServer, OneSync.

### Successfully tested on

**Game builds:** FXServer artifact 35245 with the local diagnostic build and allocator change described above. A specific GTA V client game-build number is not recorded in this review bundle.

**Platforms:** Windows x64. The standalone regression was freshly rebuilt with MSVC v14.44. The deployed diagnostic DLL passed the original Release build and isolated startup checks; production observations are described above. The extracted clean patch has not yet had a full build against current upstream. Linux and RedM/RDR3 have not been tested for this submission.

### Checklist

### Fixes issues

Refs #3456. Deliberately not an automatic closing reference: remaining hitching paths need separate investigation.


wolfram-cfx:
Hi, can you add benchmarks between creation time with pooling and creation time without pooling?

Gogsi:
I think it would be worth checking if boost::fast_pool_allocator will achieve the same result as it is:
1) Tried and tested
2) Platform agnostic
3) boost is already in the codebase, although [boost pool](https://github.com/boostorg/pool) will have to be added as a submodule

Damen57:
> Hi, can you add benchmarks between creation time with pooling and creation time without pooling?

Hey, I've got some before/after data saved from the server. Codex has been helping me investigate this and put the patch together, so I'm still getting my head around the allocator side of it.

The diagnostics measure time spent in MakeSyncTree. From the saved captures:

- Before pooling: 238.32 µs average across 12.60 million constructions
- First run with pooling: 2.42 µs across 8.46 million
- Latest capture: 2.48 µs across 28.92 million

These are from production, not a controlled benchmark. They include all entity types, and the before capture includes the periods where the server was struggling, so I don't want to claim that whole difference is purely down to pooling.

There's also a standalone test using CPedSyncTree that compares pooled and non-pooled creation/destruction. Would results from that be closer to what you're after?
| Period | Mean construction time | Constructions |
| --- | ---: | ---: |
| Before pooling | 238.32 µs | 12.60 million |
| Initial pooled deployment | 2.42 µs | 8.46 million |
| Recent pooled deployment | 2.48 µs | 28.92 million |

Damen57:
For some context on how we collected those numbers, I added this timing scope around the sync-tree factory in our diagnostic DLL:

```cpp
{
    clone_diag::Stage diagnosticTree(clone_diag::SyncTreeConstruct);
    entity->syncTree = MakeSyncTree(objectType);
}
```

`Stage` uses `std::chrono::steady_clock` to measure elapsed time until the scope ends. It adds that duration and one call to the packet's counters, which are then accumulated into these exported metrics:

```text
fxserver_onesync_clone_phase_seconds_sum{phase="sync_tree_construct"}
fxserver_onesync_clone_phase_seconds_count{phase="sync_tree_construct"}
```

Our panel saves the changes in those counters between scrapes. For the figures above, we calculated:

```text
average µs = total recorded seconds / total recorded calls × 1,000,000
```

So this measures the factory call, including allocation and construction—not the whole entity creation flow or its later destruction. It's elapsed wall time, so waiting/contention and time spent descheduled are included too.

It covers all entity types going through that instrumented path, rather than peds alone.

wolfram-cfx:
Can you tell me the sizes of each sync tree? I measured the same in enhanced and the result was basically the same with and without pooled sync trees. And benchmarks using boost::pool would be nice too.

Gogsi:
> Can you tell me the sizes of each sync tree?

CPedSyncTree is 16648 bytes, CAutomobileSyncTree is 14056 bytes

> And benchmarks using boost::pool would be nice too.

Alternatively I just saw the citizen-server-state-* components already have a CPool/CDynamicPool implementation that's used for script handles. It should also work for this purpose, the only concern being that when the pool is full it grows by 50% and copies all entries - will probably cause a single sync thread hitch when it happens.

Damen57:
One thing I didn't explain well in my earlier comment: the problem wasn't a consistently high creation cost from startup. We were seeing construction times increase as the server cycle progressed, alongside the sync hitching. Restarting would improve things again. The overall averages I posted hide that behaviour.

My earlier report, Issue #3456, has some background on the recurring hitching after we moved to Windows. I can't be certain every incident since then had the same cause, but that's what started this investigation.

We do have the saved time series data from before and after pooling, so I can provide graphs showing that progression rather than just the averages. Those measurements cover all entity types and different production workloads, though, so they aren't a controlled comparison.

PR #3681, opened 2025-10-18
tweak(data/five): Increase MaxExtraWeaponModelInfos to 4096

Katos1:
### Goal of this PR

This PR increases MaxExtraWeaponModelInfos to allow more weapon archetypes to be registered. This is useful for servers with a lot of weapons and weapon skins.

### How is this PR achieving the goal

Increase MaxExtraWeaponModelInfos in `gameconfig.xml`.

### This PR applies to the following area(s)

FiveM

### Successfully tested on

**Game builds:** 3095

**Platforms:** Windows

### Checklist

- [x] Code compiles and has been tested successfully.
- [x] Code explains itself well and/or is documented.
- [x] My commit message explains what the changes do and what they are for.
- [x] No extra compilation warnings are added by these changes.

PR #4185, opened 2026-09-09
Fix/weapon archetype limit

### LINKED ISSUE #2992: Request increase pool size [CWeaponComponentInfo / CWeaponInfoBlob]
### What happened?

Hi ! I would like to kindly request an increase in the limits for the following poolsize:

`
CWeaponComponentInfo: 2048 → 4096
`
`
CWeaponInfoBlob: 1024 → 2048
`

image

Thank you for your future reply. 

### Expected result

to be able to increase this poolsize with increase_pool_size CWeaponComponentInfo / CWeaponInfoBlob

### Reproduction steps

Start resources with enough weapons to overcome current limits.

### Importancy

Crash

### Area(s)

FiveM

### Specific version(s)

FiveM - 3258 (FXServer-master SERVER v1.0.0.11775 win32)

### Additional information

_No response_

### LINKED ISSUE #3624: Weapon limit crash's
### What happened?

Hello,

I’ve been running some tests regarding client crashes when loading into a FiveM server with custom melee weapon packs. I’m the creator behind Polars Armoury, and I’ve released 10 weapon packs so far.

Here’s what I’ve observed:

When I run all 10 of my weapon packs on a server, clients crash upon loading in.

If I remove some of the weapons, the crashing stops and players can join normally.

At first, I assumed the crashes were caused by my weapons specifically. However, after testing with weapons from other creators, I noticed the same issue — the game crashes when too many weapon packs are active at once.

The crash reports in those cases reference weapons from other creators, not mine, which leads me to believe the issue is not tied to a specific creator’s assets.

Based on these tests, it seems like there may be a limit to how many custom weapons can be loaded into a server before it starts causing client crashes.

Can you confirm if there is a known limit to the number of weapon packs or custom weapons that can be added to a server? If so, is there any chance this limit could be increased in the future to allow more weapons without causing instability? Also, is there any recommended workaround in the meantime?

Thank you for your time and support.

— Polars Armoury

ps. i have attached a few crash reports referencing my own weapons and other creators weapons

### Expected result

the game should have loaded up without crashing 

### Reproduction steps

1. you require 10 of my weapon packs (happy to provide)
2. place them in your test server
3. it will crash
4. remove 2 weapon packs (for example halloween pack and throwables pack as they have a decent amount of weapons inside them)
5. load up server and see the crash doesnt happen
6. to remove suspicion its due to my weapons add a these free weapon packs (https://github.com/NoobySloth/Custom-Weapons) , (https://github.com/Branqueador/GGC-Weapons), (https://forum.cfx.re/t/free-addon-gang-weapons/5244495)
7. the same will crash with the similiar crash report you had from my weapons but now with one of those other creator weapons 

### Importancy

Crash

### Area(s)

FiveM

### Specific version(s)

FiveM Artifacts 17000, Gamebuild 3258

### Additional information

Hello,

I’ve been running some tests regarding client crashes when loading into a FiveM server with custom melee weapon packs. I’m the creator behind Polars Armoury, and I’ve released 10 weapon packs so far.

Here’s what I’ve observed:

When I run all 10 of my weapon packs on a server, clients crash upon loading in.

If I remove some of the weapons, the crashing stops and players can join normally.

At first, I assumed the crashes were caused by my weapons specifically. However, after testing with weapons from other creators, I noticed the same issue — the game crashes when too many weapon packs are active at once.

The crash reports in those cases reference weapons from other creators, not mine, which leads me to believe the issue is not tied to a specific creator’s assets.

Based on these tests, it seems like there may be a limit to how many custom weapons can be loaded into a server before it starts causing client crashes.

Can you confirm if there is a known limit to the number of weapon packs or custom weapons that can be added to a server? If so, is there any chance this limit could be increased in the future to allow more weapons without causing instability? Also, is there any recommended workaround in the meantime?

Thank you for your time and support.

— Polars Armoury

ps. i have attached a few crash reports referencing my own weapons and other creators weapons

Image

Image

[CfxCrashDump_2025_09_07_22_43_08 (2).zip](https://github.com/user-attachments/files/22199060/CfxCrashDump_2025_09_07_22_43_08.2.zip)

[CfxCrashDump_2025_09_07_22_37_22.1.zip](https://github.com/user-attachments/files/22199061/CfxCrashDump_2025_09_07_22_37_22.1.zip)

usershhh:
### Goal of this PR

Let servers load more add-on weapons without the client crashing with `c0000005` while mounting a `weaponarchetypes.meta`.

The crash comes from the limit on weapon archetypes.

Every entry from `weaponarchetypes.meta` is added to the same `atArray<CWeaponModelInfo>`. The size of this array is set when the server starts, based on `MaxExtraWeaponModelInfos`. The problem is that the allocator doesn't check if the array is already full. It keeps increasing `m_count`, calculates the next address with `m_offset + idx * 0x190`, and then calls the object at that address.
Once we go past the allocated space, we're basically reading random data from the heap as if it was a valid object. The vtable is no longer valid, which eventually causes the crash.
On b3258, the crash dump shows the count reaching `0x800` (2048), which is exactly the current value of `MaxExtraWeaponModelInfos`.
Add-on weapons also use more than one archetype. A weapon usually needs one for the weapon itself and one for each of its component models, around 6 entries per weapon on average. This is why the limit is reached at roughly 250 weapons. The specific weapon file reported in the crash isn't necessarily the actual cause; depending on the order things are loaded, any weapon can end up being the one that triggers it.
`MaxExtraWeaponModelInfos` was already increased back in 2021 to address the same issue, so this is essentially the same limitation coming back with larger weapon packs.
This PR increases the limit from 2048 to 4096. It also increases `CWeaponInfoBlob` from 512 to 1024 and adjusts `CWeaponComponentInfo` as well (including its pool and relocated array), so we don't just hit the exact same problem somewhere else after increasing the weapon limit.
A bounds check is also added so that if the limit is reached, we get an actual error explaining which limit was hit instead of continuing into invalid memory and corrupting the heap.

Finally, this also fixes an existing out-of-bounds read in `ShiftWeaponInfoBlobsDown` that can happen when the array is completely full.

FiveM

### Successfully tested on

**Game builds:** 

Game builds: 3258
Platforms: Windows

**Platforms:** Windows, Linux

### Checklist

### Fixes issues

fixes #3624 fixes #2992

### DIFF
--- a/code/components/gta-core-five/src/CrashFixes.cpp
+++ b/code/components/gta-core-five/src/CrashFixes.cpp
@@ -489,10 +489,15 @@ void ShiftWeaponInfoBlobsDown(CWeaponInfoBlob* pArray[], uint16_t startIndex)
 {
 	uint16_t lastItemIndex = *g_weaponInfoArrayCount - 1;
 
+	uint16_t arraySize = g_weaponInfoArrayCount[1];
+
 	g_origShiftWeaponInfoBlobsDown(pArray, startIndex);
 
-	// Only clear if a shift occured
-	if (startIndex < lastItemIndex)
+	// Only clear if a shift occured, and only while the entry past the last one is still part of the
+	// allocation: a completely full array has no spare constructed blob to copy from, and reading one
+	// would be an out-of-bounds read of sizeof(CWeaponInfoBlob) bytes. This also covers an empty array,
+	// where lastItemIndex underflows to 0xFFFF.
+	if (startIndex < lastItemIndex && (lastItemIndex + 1) < arraySize)
 	{
 		// Copy the next empty weapon info into this unused one to reset it
 		(*pArray)[lastItemIndex] = (*pArray)[lastItemIndex + 1];
--- a/code/components/gta-streaming-five/include/Local.h
+++ b/code/components/gta-streaming-five/include/Local.h
@@ -1,3 +1,3 @@
 #pragma once
 
-constexpr int kNumWeaponInfoBlobs = 512;
+constexpr int kNumWeaponInfoBlobs = 1024;
--- a/code/components/gta-streaming-five/src/PatchWeaponArchetypeLimit.cpp
+++ b/code/components/gta-streaming-five/src/PatchWeaponArchetypeLimit.cpp
@@ -1,0 +1,62 @@
+#include "StdInc.h"
+
+#include <Hooking.h>
+
+#include <Error.h>
+#include <MinHook.h>
+
+struct atArrayHeader
+{
+	void* m_offset;
+	uint16_t m_count;
+	uint16_t m_size;
+};
+
+static void* g_weaponArchetypeStore;
+static atArrayHeader* (*g_getWeaponArchetypeArray)(void* store, uint32_t storeKey);
+static void* (*g_origAllocateWeaponArchetype)(void* definition, bool registerModel, uint32_t storeKey);
+
+static void* AllocateWeaponArchetypeStub(void* definition, bool registerModel, uint32_t storeKey)
+{
+	if (auto array = g_getWeaponArchetypeArray(g_weaponArchetypeStore, storeKey))
+	{
+		// a zero m_size means this isn't an array we understand - stay out of the way
+		if (array->m_size != 0 && array->m_count >= array->m_size)
+		{
+			AddCrashometry("weapon_archetype_limit", "%d", (int)array->m_size);
+
+			FatalError(
+				"Ran out of weapon archetype slots: all %d of them (MaxExtraWeaponModelInfos) are in use.\n"
+				"\n"
+				"Every add-on weapon registers one CWeaponModelInfo for each model it ships - the weapon itself "
+				"plus every one of its component models - so this limit is hit long before the number of weapons "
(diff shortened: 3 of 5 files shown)


SolotyFiveM:
I have quite a few clients encountering this issue because they are using more and more weapon add-ons, and newer versions of FiveM game builds contain more weapons than before; it would therefore be good to increase this limit

pachxl:
See https://github.com/citizenfx/fivem/pull/3681#issuecomment-3421225367

usershhh:
> See [#3681 (comment)](https://github.com/citizenfx/fivem/pull/3681#issuecomment-3421225367)

I just took a look, but some elements are missing from these changes. Furthermore, it's now more important than ever to raise this limit with the new game builds that have added weapons, and servers are now using more add-on weapons.

VeqtaDev:
Great idea i think it could work with the engine of the game but it need to be tested more deeply ! Nice PR

usershhh:
> Great idea i think it could work with the engine of the game but it need to be tested more deeply ! Nice PR

For my part, I tested it on a server with gamebuild 3258 and everything works; I also tested the error message in case a server exceeds the new limit. 

image

sanglitchercorp:
We need this

faythegoat:
All server FiveM need this

PR #4192, opened 2026-09-11
feat(gta-net-five): apply net velocity to peds during blending (Strafe & Speedboost Fix)

divocbn:
### Goal of this PR
Fix the well-known remote ped desync during rapid direction changes (WASD strafe spam & speed boosting) where a remote player drifts away from their real position and syncs back. (btw player movement is now overall more accurate)

Demo: https://www.youtube.com/watch?v=k1EZ22Jsrpo

Hooks ProcessPrePhysics, then overwrites the remote velocity with the received net velocity (plus a bounded position correction) so it stays at the players real position instead of drifting.

The behaviour is disabled by default and can be enabled with the replicated ConVar:
```
setr game_usePedNetworkVelocity true
```

FiveM

### Successfully tested on

**Game builds:** b1604, b3258, b3570, b3788

**Platforms:** Windows

### Checklist


yannbcf:
Thanks YAMP

PR #4196, opened 2026-09-11
feat(extra-natives-five): add GET_BLIP_DISPLAY_NAME

andrejmaricc:
### Goal of this PR

Allow client resources to read a blip's current display name, including a name assigned by another resource. A custom map currently cannot recover those names through a getter and otherwise needs cooperation from the resources that create the blips.

This draft proposes `GET_BLIP_DISPLAY_NAME(Blip)`. The latest update moves the resolver into the implementation file, uses `hook::pattern`, and removes the build-3570 restriction after additional runtime testing.

Resolve the game's blip pool and existing blip-name function using signatures. Validate the pool index and generation, then call `DOES_BLIP_EXIST` before reading the name. The existence check matters: runtime testing showed that a deleted blip can retain its pool slot and generation, and the name function can otherwise fall back to the player blip.

Read directly from current game state without intercepting setters or caching names. Return the custom name or localized default supplied by GTA; preserve formatting codes. Invalid/deleted handles and unavailable implementations return null. Category headings and grouped counts are outside the contract.

The draft also adds `IS_BLIP_DISPLAY_NAME_AVAILABLE()` so callers can detect failed signature initialization. Both handlers remain registered if initialization fails; the capability returns false and the getter returns null. Feedback on whether to retain this capability native is welcome.

The name function is already used by `gta-core-five/src/PatchBlipCategories.cpp`. The blip-pool layout and signatures were cross-checked against ScriptHookVDotNet's `NativeMemory` implementation.

FiveM, Natives (client, GTA V Legacy)

### Successfully tested on

**Game builds:** GTA V Legacy executables 3258, 3323, 3407, 3570, 3751, 3788, and 3889. All passed the listed in-game assertions. The 3889 executable used 3788 content (`sv_defaultGameBuild 3889`, `sv_enforceGameBuild 3788`); full 3889 content remains untested because the installation lacks `mp2026_01/dlc.rpf`. Initialization logs confirmed the executable build for each run.

**Platforms:** Windows x64.

Built from upstream commit `0d8a2a6f78a9922445d8930305af82a7b1826980` using Visual Studio 2022/v143 and Windows SDK 10.0.26100.0. Native bindings and return-type metadata were generated with prebuild. The final incremental build completed with 0 errors and 3 linker/postbuild warnings; no compiler diagnostic was emitted for the new source. A clean baseline warning comparison has not been performed.

Two separate local client resources were used: a producer using ordinary GTA blip natives and a reader invoking the new getter. The in-game test passed:

- A custom name created in the other resource.
- UTF-8 (`Čevapi – Žuta garaža`).
- Preserved formatting (`~g~NOS garage~s~`).
- A nonempty default label for sprite 60 (`Police Station` in this session).
- Null results for handles 0 and -1.
- A live rename visible immediately through the getter.
- Null after removing the blip.

Runtime output:

```text
[blip-names] Experimental name reader initialized for build 3570.
[blip-names] Compare this default with GTA map: Police Station
[blip-names] PASS: cross-resource, UTF-8, formatting, default, invalid, rename, removal.
```

A standalone C++ resolver test passed 13 assertions covering bounds, missing slots, invalid handles, generation mismatches, and slot reuse. The in-game test covers liveness of a retained deleted slot separately. The default label was not visually compared against the GTA map. Executables older than 3258, Enhanced, full 3889 content, and a full production resource set have not been tested.

#### Minimal cross-resource reproduction

Create two Lua client resources with `fx_version 'cerulean'`, `game 'gta5'`, and `client_script 'client.lua'` in their manifests. Start `blip_name_producer` before the reader.

Producer (`blip_name_producer/client.lua`):

```lua
local blip = AddBlipForCoord(0.0, 0.0, 72.0)
BeginTextCommandSetBlipName('STRING')
AddTextComponentSubstringPlayerName('Cross-resource test')
EndTextCommandSetBlipName(blip)
exports('handle', function() return blip end)
exports('remove', function() RemoveBlip(blip) end)
AddEventHandler('onResourceStop', function(resource)
    if resource == GetCurrentResourceName() then RemoveBlip(blip) end
end)
```

Reader (`client.lua`):

```lua
RegisterCommand('testblipgetter', function()
    local availableHash = joaat('IS_BLIP_DISPLAY_NAME_AVAILABLE') & 0xFFFFFFFF
    assert(Citizen.InvokeNative(availableHash, Citizen.ResultAsInteger()) == 1)
    local getHash = joaat('GET_BLIP_DISPLAY_NAME') & 0xFFFFFFFF
    local function read(handle)
        return Citizen.InvokeNative(getHash, handle, Citizen.ResultAsString())
    end
    local handle = exports.blip_name_producer:handle()
    assert(read(handle) == 'Cross-resource test')
    assert(read(0) == nil and read(-1) == nil)
    exports.blip_name_producer:remove()
    assert(read(handle) == nil)
    print('PASS')
end, false)
```

Run `testblipgetter` in the client console. Restart the producer before repeating.

### Checklist

The warning comparison remains open as described above. Before marking ready for review, build coverage, behavior on unsupported builds, and the public API shape need further validation.

### Fixes issues

No linked issue.


code/components/extra-natives-five/include/BlipHandleResolver.h
ook3D, line 1: Not sure why this needs a new header file

code/components/extra-natives-five/src/BlipNameNatives.cpp
ook3D, line 1: not sure why you're using module_pattern here, i didnt even know this existed, just use normal hooking stuff

andrejmaricc: Will change it to the usual hook::pattern. I'll include the changes in the next update.

code/components/extra-natives-five/include/BlipHandleResolver.h
andrejmaricc: I separated it for standalone testing, but it's only used in this file. I'll mmove the resolver into the .cpp file and remove the header

xalva98:
In my opinion, a smarter approach would be to intercept the natives used to set blip names and cache the result externally.

```cpp
// pseudocode
std::unordered_map<Blip, std::string> blipNames;

wrap(END_TEXT_COMMAND_SET_BLIP_NAME, [](ctx) {
    Blip blip = ctx.GetArgument<Blip>(0);

    original(ctx);

    blipNames[blip] = currentTextCommand;
});

GET_BLIP_DISPLAY_NAME(blip) {
    return blipNames[blip];
}
```

Example on how it is already done inside of the project (this is just "hooking a native call so to say", or "intercept", whatever makes more sense in your mind)
https://github.com/citizenfx/fivem/blob/master/code/components/gta-core-five/src/SetCursorLocation.cpp
or this
https://github.com/citizenfx/fivem/blob/0d8a2a6f78a9922445d8930305af82a7b1826980/code/components/gta-net-five/src/NetHook.cpp#L1965

Then it can be done in one file with < 60 lines or so

andrejmaricc:
Thanks for the examples. I’ll look into wrapping the naming natives and caching the resolved text, including cleanup when blips are removed. I’ll also check the default-label case, since the current getter returns those as well.

yannbcf:
Its not smarter to cache the name no, you then manage the lifetime of that thing, just point a getter to the blip name based on its internal id

code/components/extra-natives-five/src/BlipNameNatives.cpp
```diff
@@ -0,0 +1,95 @@
...
+	fx::ScriptEngine::RegisterNativeHandler("GET_BLIP_DISPLAY_NAME", [](fx::ScriptContext& context)
+	{
+		const char* result = nullptr;
+		if (IsAvailable() && context.GetArgumentCount() > 0)
+		{
+			auto handle = context.GetArgument<uint32_t>(0);
+			auto blip = ResolveBlip(g_blipPool, *g_blipCount, handle);
+			// Removed blips can retain their slot and generation. Ask GTA for
+			// liveness before its name function can fall back to the player blip.
+			auto exists = fx::ScriptEngine::GetNativeHandler(0xA6DB27D19ECBB7DA);
```
AvarianKnight, line 86: We shouldn't be calling `GetNativeHandler` every time this function is called, you should move this outside the call and just pass it into the lambda

PR #4197, opened 2026-09-13
feat(extra-natives-rdr3): add screenshot natives

abdulkadiraktas:
### Goal of this PR

RedM has no equivalent of [screenshot-basic](https://github.com/citizenfx/screenshot-basic): that resource is built on the FiveM-only backbuffer capture path in `rage-graphics-five`, which has no RDR3 counterpart. This PR adds a small set of client natives so RedM resources can capture the game window, hand the result to a script callback, upload it to an HTTP endpoint, or write it to disk.

Four natives are added in `extra-natives-rdr3`:

```c
BOOL IS_SCREENSHOT_AVAILABLE();
void REQUEST_SCREENSHOT(int quality, int maxDimension, func callback);
void REQUEST_SCREENSHOT_UPLOAD(char* url, char* field, char* fileName, int quality, int maxDimension, char* extraFieldsJson, func callback);
BOOL SAVE_SCREENSHOT_TO_FILE(char* fileName, int quality);
```

- Capture, bilinear downscaling, JPEG encoding (`vendor:toojpeg`) and the WinHTTP upload all run on a single persistent worker thread, so the script thread is never blocked.
- Results are delivered through `fx::FunctionRef` and invoked from `ResourceManager::OnTick`. No shared result state is exposed to scripts, several requests can be in flight at once, and a callback is dropped if its resource stops while the worker is busy.
- `vendor:toojpeg` and `vendor:botan` are added as dependencies, mirroring `citizen-scripting-core`, which already uses the same pair to JPEG-encode and base64 the profiler screenshot. Both already ship with the client, so nothing new is added to the distribution.

Known limitation, called out up front: the game window is captured through GDI, which reads the composited desktop. That means:

- **Exclusive fullscreen is not supported** - it yields a black frame. Borderless or windowed is required.
- Anything drawn on top of a focused game window (overlays, notifications, other applications) appears in the capture.

A backbuffer capture path in `rage-graphics-rdr3`, mirroring `OnRequestInternalScreenshot`/`OnInternalScreenshot` in `rage-graphics-five`, would avoid both and let `screenshot-basic` work unchanged. That is a considerably larger change; I am happy to take that direction instead if maintainers prefer it.

Two deliberate restrictions on what a resource can do with these:

- Capture is refused unless the game window owns the foreground, so a resource cannot capture unrelated applications while the player is alt-tabbed.
- `SAVE_SCREENSHOT_TO_FILE` resolves its path inside `data/screenshots` and rejects drive letters, UNC prefixes, rooted paths and `..` segments, so it cannot write to arbitrary locations on the client's disk.

Script-supplied multipart field and file names are sanitized before they reach the request headers, only `http`/`https` URLs are accepted, WinHTTP timeouts are set, and the response body surfaced to scripts is capped at 64 KiB.

RedM, Natives

### Successfully tested on

**Game builds:** 1491

**Platforms:** Windows

Tested with a RedM port of the `screenshot-basic` API: capture to a base64 data URL, and a direct multipart upload to a Discord webhook (HTTP 200, image received intact).

### Checklist

### Fixes issues

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ook3D:
at least read the slop youre committing.... "I am happy to take that direction instead if maintainers prefer it."

divocbn:
Even if we were to review this fairly right now, this code introduces so many different security issues, I am really in favor of banning people who just submit their untested sloppy AI PRs

abdulkadiraktas:
Fair points. GDI capture was the wrong foundation for this - proper backbuffer capture at the graphics pipeline level is where this belongs. Closing.

PR #4199, opened 2026-09-14
fix(gta-core-five): guard vehicle mod-stat queries without a draw handler

Damen57:
### Goal of this PR

Prevent a client access violation when vehicle physics queries a performance modifier while the vehicle has no draw handler. Three crash reports on b3258 share `GTA5_b3258.exe+0x626158` (`ohio-rugby-spring`); the full dump identifies a BF400 during an internal gearbox-modifier query.

The caller passes `drawHandler + 0x40`, so an absent handler becomes `RCX = 0x40`. The mod-kit getter then reads `[RCX + 0x3B4]`, faulting at `0x3F4`.

Guard the shared integer mod-stat lookup before its first memory access. Return zero for the missing-handler sentinel, matching the function's existing no-mod-kit result. Valid pointers replay the displaced five-byte prologue and resume the original function.

The hook covers calls into this shared lookup, rather than a particular model or gearbox call site. It has no game-build restriction: the signature matches the lookup's prologue and existing no-mod-kit return path. If the signature is missing or ambiguous, installation is skipped with a diagnostic. The guard handles the `drawHandler + 0x40` calling convention; other layouts and lookup functions are outside this patch's scope.

This prevents the demonstrated invalid access. Why the draw handler was absent, and whether that state can cause additional failures later in the vehicle update, remain open.

FiveM native client (`gta-core-five`), game builds matching the lookup signature. Only b3258 has been verified against a crash dump.

### Successfully tested on

- Windows x64, MSVC 14.51: complete new source file passes a syntax check with CitizenFX headers.
- Standalone optimized JIT harness: 151 checks passed against the actual production stub for the absent-handler sentinel, valid pointer/slot forwarding, prologue replay, zero values and a full-width result.
- Production installer exercised with a substituted hook API: missing and duplicate matches skip installation; a unique match installs the executable guard at the expected address.
- Full-dump signature scan: exactly one match at game image `+0x627088`.
- Patch application and Git whitespace checks passed against upstream `0d8a2a6f78a9922445d8930305af82a7b1826980`.

**Draft:** the full client build/link, hook installation in FiveM and in-game stock/upgraded BF400 stream-in/out and crash-reproduction testing remain outstanding, as does verification on other game builds. Raw dumps and the local standalone harness are not included in this PR.

### Checklist

### Fixes issues

No existing issue linked.


code/components/gta-core-five/src/CrashFixes.VehicleModStat.cpp
```diff
@@ -0,0 +1,57 @@
...
+		L("noModifier");
+		xor(eax, eax);
+		ret();
+	}
+};
+
+static HookFunction hookFunction([]()
+{
+	// The function layout and missing-handler value are verified on b3258.
+	if (!xbr::IsGameBuild<3258>())
```
divocbn: I dont see any logical reason to restrict the whole thing to just b3258

